PR #13, opened 2018-06-13
fix entry listing endpoint

ltalirz:
* fix description of 'properties' key
* fix example

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -581,7 +581,6 @@ MUST include the following information in the "data" field:
   type, where each key is a property ID.
   Each value is a dictionary, with the required key 'description' and optional
 key 'unit'.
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -578,7 +578,9 @@ MUST include the following information in the "data" field:
 
 * **description**: Description of the entry.
 * **properties**: A dictionary describing queryable properties for this entry
-  type, where each key is a property ID, each value is a description of the
+  type, where each key is a property ID.
+  Each value is a dictionary, with the required key 'description' and optional
+key 'unit'.
   property, along with the units.
```
merkys: @ltalirz, I guess this line should be removed.

PR #14, opened 2018-06-13
fix #7

### LINKED ISSUE #7: Missing sentence separator
A sentence in [4.1.2 JSON API response schema](https://github.com/Materials-Consortia/OPTiMaDe/blob/master/optimade.md#412-json-api-response-schema) subsection reads:

> **immutable_id**: an OPTIONAL field containing the entry's immutable ID db specific properties need to be prefixed by the db specific prefix

From the source of .md file I assume that the part "db specific properties need to be prefixed by the db specific prefix" should constitute a separate paragraph (not connected to the description of ``immutable_id``).

ltalirz:


merkys:
Fix seems to be trivial, therefore I accept it. Thanks!

Issue #17, opened 2018-06-14
Handle unset optional properties


rartino:
We should adopt policies around unset optional properties.

PR #47, opened 2018-07-12
Handle unset optional properties

ctoher:
Dear all,

I've inserted a first draft of the section on how to handle unset optional properties (i.e. properties with "null" values - #17 on the list of issues), as per the discussion during the meeting in Lausanne. 

I've added this as a new subsection 3.3.4, but let me know if you think some of the information would be better suited in a different section, e.g. if the "<property> IS KNOWN" filter option should be in the filter section instead. Please let me know if you have any other comments, questions or suggestions on this topic.

Best regards

Cormac.


optimade.md
```diff
@@ -298,6 +299,30 @@ An example of a full response:
...
+Any comparisons involving unset properties should return a value of false,
+i.e. by definition the value of 'null' is outside of any defined search range.
+
+If a property is explicitly requested in a search query without value range filters,
+then all entries otherwise satisfying the query should be returned, including those with null values for this property.
+The value of these properties should be set to the JSON 'null' value in the response.
+
+Note that this only applies to properties that are described by a single value,
+and not to properties with multiple attributes such as the structure.
+In this case, attributes that have null value are simply omitted from the returned structure.
```
rartino: In an OPTIMaDe meeting discussion, we note that we have some trouble understanding exactly what this means. What we think is needed is to say that the text in this section only describes how the api handles properties that themselves are null, and does not regulate which values inside the data structure used by a properties can be null. That is rather to be described in relation to the definition of those data structures.

So:
```suggestion
The text in this section describes how the api handles properties that are null. It does not regulate the handling of values inside property data structures that can be null. That use of null values is described in relation to the definition of those data structures.
```
(Note: I cannot get github to make a suggestion over several lines, but this is meant to replace the whole paragraph.)

optimade.md
```diff
@@ -298,6 +299,30 @@ An example of a full response:
 | 418  | Asked for a non-existent keyword                                      |
 | 422  | Database returned (200) but the translator failed                     |
 
+### <a name="h.3.3.4">3.3.4. Unset optional properties</a>
+
+Unset optional properties in a database are properties that exist and have a specific value within a database for some materials entries,
+but are undefined for other entries, i.e. have the value 'null' within a JSON file or SQL database.
+
+Unset properties are not returned in the response unless explicitly requested in the search query.
```
```suggestion
Unset properties MUST NOT be returned in the response unless explicitly requested in the search query.
```

optimade.md
```diff
@@ -298,6 +299,30 @@ An example of a full response:
 | 418  | Asked for a non-existent keyword                                      |
 | 422  | Database returned (200) but the translator failed                     |
 
+### <a name="h.3.3.4">3.3.4. Unset optional properties</a>
+
+Unset optional properties in a database are properties that exist and have a specific value within a database for some materials entries,
+but are undefined for other entries, i.e. have the value 'null' within a JSON file or SQL database.
+
+Unset properties are not returned in the response unless explicitly requested in the search query.
+Any comparisons involving unset properties should return a value of false,
```
```suggestion
Any comparisons involving unset properties MUST be evaluated as false,
```

optimade.md
```diff
@@ -298,6 +299,30 @@ An example of a full response:
...
+Unset optional properties in a database are properties that exist and have a specific value within a database for some materials entries,
+but are undefined for other entries, i.e. have the value 'null' within a JSON file or SQL database.
+
+Unset properties are not returned in the response unless explicitly requested in the search query.
+Any comparisons involving unset properties should return a value of false,
+i.e. by definition the value of 'null' is outside of any defined search range.
+
+If a property is explicitly requested in a search query without value range filters,
+then all entries otherwise satisfying the query should be returned, including those with null values for this property.
+The value of these properties should be set to the JSON 'null' value in the response.
```
```suggestion
These properties MUST be set to the JSON 'null' value in the response.
```

optimade.md
```diff
@@ -298,6 +299,30 @@ An example of a full response:
...
+
+Unset optional properties in a database are properties that exist and have a specific value within a database for some materials entries,
+but are undefined for other entries, i.e. have the value 'null' within a JSON file or SQL database.
+
+Unset properties are not returned in the response unless explicitly requested in the search query.
+Any comparisons involving unset properties should return a value of false,
+i.e. by definition the value of 'null' is outside of any defined search range.
+
+If a property is explicitly requested in a search query without value range filters,
+then all entries otherwise satisfying the query should be returned, including those with null values for this property.
```
```suggestion
then all entries otherwise satisfying the query SHOULD be returned, including those with null values for this property.
```

ctoher:
Dear all,

I've updated the draft of the section on how to handle unset optional properties (i.e. properties with "null" values - #17 on the list of issues), to incorporate Rickard's suggested changes. 

I've slightly modified his suggestion for the last sentence of the paragraph about null values inside property data structures to try to make it clearer - let me know if this makes sense, or if you think something else would work better. 

Best regards

Cormac.

giovannipizzi:
After discussion in the skype call, what is missing in this PR is an update of the grammar. This can be done at the same time (or right after having discussed) issue #58

ctoher:
Dear all,

I've updated the specification for the handling of "null" properties. In particular, I updated the "IS KNOWN" and "IS UNKNOWN" discussion, including a cross-reference to section 5.2 (the filter specification where IS KNOWN and IS UNKNOWN are described). Please let me know if you have any comments or suggestions.

Best regards

Cormac.

optimade.md
```diff
@@ -377,6 +378,30 @@ the next course of action SHOULD be to fetch the resource objects under the
 to the corresponding database ID that was originally queried, using the object's
 `base_url` value.
 
+### <a name="h.3.3.4">3.3.4. Unset optional properties</a>
+
+Unset optional properties in a database are properties that exist and have a specific value within a database for some materials entries, but are undefined for other entries, i.e. have the value 'null' within a JSON file or SQL database.
```
giovannipizzi: To be consistent, I would always write null or 'null' as `null` instead, with backticks - both here and in a number of places below.

giovannipizzi: I don't know if we want to mention SQL here. Maybe just say "i.e. have the value `null` or are missing in the data source (that can be a JSON file, a database, ...)" or something like this? (assuming I'm understanding what you want to say here)

optimade.md
```diff
@@ -377,6 +378,30 @@ the next course of action SHOULD be to fetch the resource objects under the
 to the corresponding database ID that was originally queried, using the object's
 `base_url` value.
 
+### <a name="h.3.3.4">3.3.4. Unset optional properties</a>
+
+Unset optional properties in a database are properties that exist and have a specific value within a database for some materials entries, but are undefined for other entries, i.e. have the value 'null' within a JSON file or SQL database.
+
+Unset properties MUST NOT be returned in the response, unless explicitly requested in the search query. 
+
+Any comparisons involving unset properties MUST be evaluated as false,
```
giovannipizzi: Also here, I would use `false` instead of false?

ctoher:
I fixed the `null` and `false` formatting as per Giovanni's comment. Let me know if there are any other formatting changes that should be made.

I removed the mention of SQL; this part now reads: "e.g. have the value null within a JSON file." The idea behind this part is just to give a concrete example of the types of values we're talking about; let me know if you think we should write this in a different way.

Issue #23, opened 2018-06-14
Structure entry standardization


rartino:
We need a standardization of the structure entry type that allows communicating "where the atoms are", composition, etc.

davidwaroquiers:
Points related to structure :
- How to deal with occupancies (also variable occupancies)
- How to deal with correlation (in a similar way of what "assembly disorder groups" are in CIF files)
- How to extend that to trajectories ? (keep in mind the variable occupancies above)

giovannipizzi:
Here is the suggested format that resulted from the group discussion:
```json
{
  "dimension_types": [ 
    1, 
    1, 
    1
  ], 
  "lattice_vectors": [ 
    [  
      4.0, 
      0.0, 
      0.0
    ], 
    [
      0.0, 
      4.0, 
      0.0
    ], 
    [
      0.0, 
      0.0, 
      4.1
    ]
  ], 
  "cartesian_site_positions": [ 
    [ 
      0.0, 
      0.0, 
      0.0
    ], 
    [
      2.0, 
      2.0, 
      1.95
    ], 
    [
      2.0, 
      2.0, 
      2.15
    ],     
    [
      2.0, 
      2.0, 
      0.0
    ], 
    [
      2.0, 
      0.0, 
      2.0
    ], 
    [
      0.0, 
      2.0, 
      2.0
    ]
  ],
  "species_at_sites": [  
    "BaCa", 
    "Ti", 
    "Ti", 
    "O", 
    "O", 
    "O"
  ], 
  "species": { 
    "BaCa": {
      "chemical_symbols": ["Ba", "Ca"], 
      "concentration": [0.4, 0.5], 
      "mass": 88.5 
    },
    "O": {
      "chemical_symbols": ["O"],
      "concentration": [1.0],
      "mass": 15.9994
    },
    "Ti": {
      "chemical_symbols": ["Ti"],
      "concentration": [1.0],
      "mass": 47.867
    }
  },
  "assemblies": [
    { 
      "sites_in_groups": [ [1], [2] ], 
      "group_probabilities": [ 0.5, 0.5 ] 
    }
  ], 
}
```

And some comments:
- `dimension_types`: Compulsory; follows definitions of https://esl.cecam.org/ESCDF_-_System but we allow only values of 0 (non-periodic direction specified by the lattice vector) or 1 (periodic)
[...]

PR #49, opened 2018-10-12
Adding specifications for specifying crystal structure atomic positions

giovannipizzi:
This PR extends the specifications defining the specifications for a structure endpoint, fixing #23.
I hope I correctly interpreted all comments, feel free to comment.

Regarding @sauliusg 's comment in #23:
> Actually, there seems to be the third way to represent atoms with unknown positions: specify no site for them (so no 'null' coordinates), and have a species that is on no site.

I think this is not a problem - the specification of this PR allow to define this case (there `null`s, and I don't think we need to define multiple ways of defining unknown atoms.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -39,6 +39,13 @@
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.2. nelements](#h.6.2.2)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.3. chemical\_formula](#h.6.2.3)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.4. formula\_prototype](#h.6.2.4)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.5. dimension\_types](#h.6.2.5)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.6. lattice\_vectors](#h.6.2.6)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.7. cartesian\_site\_positions](#h.6.2.7)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.8. all\_coordinates\_known](#h.6.2.8)
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.9. species\_at\_sites](#h.6.2.9)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.10. species](#h.6.2.10)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.11. assemblies](#h.6.2.11)  
 
 &nbsp;&nbsp;&nbsp;&nbsp;[6.3. 'Calculation' entries](#h.6.3)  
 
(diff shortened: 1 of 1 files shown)


rartino:
Thanks for writing this up for the specification, it looks great. The assembly construct really is a great design.

A few points:
1. Do you recall the motivation for standardizing on cartesian coordinates over fractional coordinates? I seem to recall that the working group had a good motivation, which it may be nice to preserve here in the discussion to not have the issue resurface. (My gut reaction is to prefer fractional coordinates since then internal positions can often be expressed exactly instead of as chopped off infinite decimal sequences, thus arguably being more precise and also cutting down the string representations...)
2. For `lattice_vectors` I agree with specifying that the convention is a non-zero determinant. However, I think it also should be added that for the `lattice_vectors` matrix to be valid, each lattice vector must be strictly confined either to the periodic or the non-periodic subspace defined by `dimension_types`. One MUST NOT have lattice vectors that mix periodic and non-periodic dimensions.
3. Again for `lattice_vectors`, I'm not so sure about pushing for that non-periodic dimensions should have lengths that enclose the structure. A nice use for `lattice_vectors` in non-periodic dimensions is to suggest a sensible basis for reduced coordinates. E.g., a slab can have the non-periodic lattice vector be the layer distance, giving nice reduced z coordinates of 0, 1, 2, 3, ...
4. The text is now very clear on that `species` names that match standard chemical names are not to be interpreted as having special meaning. An alternative way that arguably can avoid a lot of user confusion down the line is to say that standard chemical symbols are reserved names that must mean precisely that element at concentration 1, and are then not to be included in the `species` dict. That would also cut down the representation a lot for normal standard structures. 
5. The text now implements the 'null' approach to unknown coordinates. I really think we should think this through before committing this to the standard. Can we first commit a version without 'nulls' for now?
6. There are a few places where the text says 'convention',  'not required', or use lowercase 'must' where I think it may be more clear to explicitly type out SHOULD or MUST. I'll try to use github's line comments to mark those.

rartino:
>  Although I don't remember the rationale for the conventions for representing the unit cell (Cartesian vectors over the cell lengths and angles) and site positions (Cartesian over fractional) thus they seem a little odd to me,

Is it clear how one would combine something like `dimension_types` with a cell defined as cell lengths and angles? I.e. how is the periodic and non-periodic directions aligned relative to the cell vectors?

merkys:
> Is it clear how one would combine something like `dimension_types` with a cell defined as cell lengths and angles? I.e. how is the periodic and non-periodic directions aligned relative to the cell vectors?

Ah yes, now I remember. Sorry for the noise!

giovannipizzi:
Dear all,
thanks for the comments! I reply here to @rartino 's comments (also discussing with @merkys who's here in Lausanne these days):

> Do you recall the motivation for standardizing on cartesian coordinates over fractional coordinates? 

For partially non-periodic systems (e.g. 2D) the non-periodic dimension would still need to be Cartesian, and mixing Cartesian and fractional in the same structure is confusing. 

> For lattice_vectors I agree with specifying that the convention is a non-zero determinant. However, I think it also should be added that for the lattice_vectors matrix to be valid, each lattice vector must be strictly confined either to the periodic or the non-periodic subspace defined by dimension_types. One MUST NOT have lattice vectors that mix periodic and non-periodic dimensions.

I'm not sure if I understand this - maybe there is a misunderstanding on the `dimension_types`.
`dimension_types = [1,0,1]` means that the system is periodic *on the plane spanned by the first and third lattice vector* (and not `x` and `z` Cartesian directions). The second one (if the determinant is not null) will be outside of the periodic plane. So there is no risk that a lattice vector mixes periodic and non-periodic dimensions, in my understanding. If you refer to having the second vector orthogonal to the first and third, I'm not sure if we want to enforce it.

> Again for lattice_vectors, I'm not so sure about pushing for that non-periodic dimensions should have lengths that enclose the structure. A nice use for lattice_vectors in non-periodic dimensions is to suggest a sensible basis for reduced coordinates. E.g., a slab can have the non-periodic lattice vector be the layer distance, giving nice reduced z coordinates of 0, 1, 2, 3, ...

Ok. I will remove that from the specs.

> The text is now very clear on that species names that match standard chemical names are not to be interpreted as having special meaning. An alternative way that arguably can avoid a lot of user confusion down the line is to say that standard chemical symbols are reserved names that must mean precisely that element at concentration 1, and are then not to be included in the species dict. That would also cut down the representation a lot for normal standard structures.

I see your point. I think however that even if this makes things a bit harder for clients, it reduces a bit the risk of 'dumb' clients being implemented, that do not check properly all the data types. Also, it makes it a bit harder to comply to the specs if a server application internally allows for any name; or e.g. for Lennard-Jones I would imagine species "A", "B", "C", ..., and already "B" and "C" would not be valid (they could be represented with a "X" species in Optimade). Then the server should convert these names to something else, and when importing back know how to map the names back to the ones used internally.

So, personally I would keep the specs as it is, but happy to hear comments also from other people before deciding what to do (or move the discussion in another issue, for the future).

> The text now implements the 'null' approach to unknown coordinates. I really think we should think this through before committing this to the standard. Can we first commit a version without 'nulls' for now?

Very happy to - I added because I had the understanding that is was needed from the COD side, but as @merkys is happy to postpone, there is no problem for me. I'll move to a new issue.

> There are a few places where the text says 'convention', 'not required', or use lowercase 'must' where I think it may be more clear to explicitly type out SHOULD or MUST. I'll try to use github's line comments to mark those.

Thanks, very useful! I'll now work on the text.


Finally, a few more points:
- about the suggestion of enforcing a positive determinant: I'm not sure we want to do it, if internally the lattice vectors are defined using a left-handed set, the API would need to swap some axes changing the structure, which is confusing I believe.

- about vacancies: we could explicitly specify vacancies. In that case, I'd prefer `""` (empty string) and leave "X" for a non-chemical element (e.g. for Lennard-Jones). However, only doubt here (about rounding) is the following: by specifying that the sum must be 1, should we also specify a minimum required threshold for a valid return value? (and for errors below that, the API client can e.g. decide to renormalise the values so that the sum is 1). The other option instead is to leave it as it is and the client decides what to do - it depends on who we want to put the burden. E.g., I imagine the following problematic case, e.g. for COD, when exposing CIF files that they didn't generate: if they have `0.333 0.333 0.333` as occupations, should they assume that it's actually 1/3 1/3 1/3, or specify 0.001 as a vacancy? If they do the latter, they might actually doing something wrong: I would leave this responsibility to the client that needs to use the structures and anyway will have some algorithms to deal with these problems, if it can deal with partial occupancies. At the moment I would still prefer to leave vacancies unspecified, but again, happy to hear opinions.

optimade.md
```diff
@@ -908,6 +915,165 @@ Multiple Entry Types", as well as the following properties:
...
+    lattice vector: `[1, 0, 1]`
+  * For a bulk 3D system: `[1, 1, 1]`
+
+### <a name="h.6.2.6">6.2.6. lattice\_vectors</a>
+
+* Description: The three lattice vectors in Cartesian coordinates, in angstrom
+* Requirements/convention: 
+  * This property is required, except when `dimension_types` is equal to `[0, 0, 0]` (in this case it is optional).
+  * It must be a 3 times 3 list of lists, each element being a float.
+  * The fastest index runs over the x, y, z Cartesian coordinates, while
```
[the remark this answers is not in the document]
giovannipizzi: Ok thanks. It's common when speaking of matrices in memory, but indeed it might not be clear in this context (even if since we have a list of lists, it should be clear). We have examples so I'm merging this.

giovannipizzi:
I've done my changes, happy to hear more feedback or to get approval and move open questions to new issues. 
Still potentially open questions (unless I forget something):
- [ ] should we have a simplified species/sites definition, where species are optional if they their name is a chemical symbol
- [ ] is it ok NOT to enforce a positive determinant for the lattice_vectors (only >0) 
- [ ] should we enforce that the sum of concentrations is 1 (and have a symbol for vacancies), or allow for sum < 1 and what is left is a vacancy (and the client will implement checks for rounding errors)

rartino:
>> Do you recall the motivation for standardizing on cartesian coordinates over fractional coordinates?

> For partially non-periodic systems (e.g. 2D) the non-periodic dimension would still need to be Cartesian, and mixing Cartesian and fractional in the same structure is confusing.

Ok, but this isn't the only way to do it (or even the most common.) It would be more consistent to have fractional coordinates for a non-periodic direction be in fractions of the non-periodic lattice vector. I.e., as in the example above with a lattice vector orthogonal to a surface of length = distance between places, one would have factional coordinate 1.0, 2.0, 3.0, ...

So, to me, this variant of fractional coordinates is more elegant. But, I don't feel very strongly about it. Hence, unless others speak up preferring fractional coordinates, I'll approve the merge as is.

> maybe there is a misunderstanding on the dimension_types.
dimension_types = [1,0,1] means that the system is periodic on the plane spanned by the first and third lattice vector (and not x and z Cartesian directions).

Sorry, my bad. I was indeed confused. (Which partly invalidates my comment to @merkys; but with angles and lengths I still guess it isn't self-evident which is the 1st, 2nd and 3td lattice vector.)

>> An alternative way that arguably can avoid a lot of user confusion down the line is to say that standard chemical symbols are reserved names that must mean precisely that element at concentration 1, and are then not to be included in the species dict. 

> I think however that even if this makes things a bit harder for clients, it reduces a bit the risk of 'dumb' clients being implemented, that do not check properly all the data types. Also, it makes it a bit harder to comply to the specs if a server application internally allows for any name; or e.g. for Lennard-Jones I would imagine species "A", "B", "C", ..., and already "B" and "C" would not be valid

I agree especially with the second point. In addition, the space of "all chemical symbols" isn't technically well defined, as we are still occasionally adding entries to the end of the periodic table. But, it still seems as a mistake to over-complicate the format for what I think is the vast, *vast*, majority of use cases. 

So, if we go for this, we should separate the namespaces. E.g., by a prefix that isn't allowed otherwise, e.g. ":C" or, e.g., using the atomic number "6" (though, for the sake of datatypes I'd argue for using the *string* "6".) 

Nevertheless, with a separate namespace, this works as something we can add later as an addition to the spec. So, if no one else chimes in wanting to pushing this through now, I'll approve the merge without this.

> is it ok NOT to enforce a positive determinant for the lattice_vectors (only >0)

Note that my suggestion was to put this as a SHOULD. I.e., you are not strictly in violation of the spec with a negative determinant. But isn't it det > 0 generally preferred essentially everywhere? It isn't a very important point though, so, I'm ok with just "SHOULD have non-zero determinant".

> about vacancies: we could explicitly specify vacancies. In that case, I'd prefer "" (empty string) and leave "X" for a non-chemical element (e.g. for Lennard-Jones). However, only doubt here (about rounding) is the following: by specifying that the sum must be 1, should we also specify a minimum required threshold for a valid return value? (and for errors below that, the API client can e.g. decide to renormalise the values so that the sum is 1). The other option instead is to leave it as it is and the client decides what to do - it depends on who we want to put the burden. E.g., I imagine the following problematic case, e.g. for COD, when exposing CIF files that they didn't generate: if they have 0.333 0.333 0.333 as occupations, should they assume that it's actually 1/3 1/3 1/3, or specify 0.001 as a vacancy? If they do the latter, they might actually doing something wrong: I would leave this responsibility to the client that needs to use the structures and anyway will have some algorithms to deal with these problems, if it can deal with partial occupancies. At the moment I would still prefer to leave vacancies unspecified, but again, happy to hear opinions.

Here my opinion is a bit stronger than for the other points above. As a general stance, I prefer the OPTIMaDe format to have as little ambiguity as possible, rather than to be designed to allow passing on ambiguity to users. I rather have the user deal with some single extra instances of faulty data due to a conversion gone wrong (which one always have to be ready for anyway) than have every user invent assumptions of their own when interpreting retrieved data. If we take the decision to pass on this ambiguity, the same argument will apply throughout the work with optimade anywhere a popular source format has an ambiguity, and we'll end up with a lot of ambiguity.

So, for this point two of us (Fawzi and I) want to make vacancies explicit. It'd great to hear from others.

optimade.md
```diff
@@ -908,6 +914,158 @@ Multiple Entry Types", as well as the following properties:
...
+* Description: Name of the species at each site (where values 
+  for sites are specified with the same order of the 
+  `cartesian_site_positions` property). The properties of 
+  the species are found in the `species` property. 
+* Requirements: 
+  * This property is REQUIRED.
+  * It MUST be a list of strings. It MUST have 
+    length equal to the number of sites in the structure (first dimension of the `cartesian_site_positions` list). 
+  * Each string MUST be a 
+    valid key of the dictionary specified by the `species` property.
```
rartino: In line with my longer comment: to make it possible for future extensions of the spec to give reserved names in a separate namespace special meaning (e.g., ":C" or "6" for standard Carbon). Also, it may be nice to not force clients to handle overly weird species names, e.g., with non-printable unicode in them. 
```suggestion
    valid key of the dictionary specified by the `species` property. The requirements on this string are the same as for property names, i.e., it can be of any length, may use upper and lower case letters, the underscore, and the digits 0-9, but cannot begin with a digit.    
```

optimade.md
```diff
@@ -908,6 +914,158 @@ Multiple Entry Types", as well as the following properties:
   occurrence number in the **reduced** chemical formula and replace them with
   subsequent alphabet letters A, B, C and so on.
 
+### <a name="h.6.2.5">6.2.5. dimension\_types</a>
+
+* Description: a list of three integers. For each of the three 
+  directions indicated by the three lattice vectors (see property `lattice_vectors`) this list indicates if that direction is periodic (value 1) or non-periodic (value 0). Note: each periodicity direction is *not* one of the Cartesian x, y, z directions but the directions specified by the `lattice_vectors`.
```
rartino: Some very minor language editing, because, technically, the directions *could*, of course, be the Cartesian x,y,z directions, if that is where the lattice vectors point. 
```suggestion
  directions indicated by the three lattice vectors (see property `lattice_vectors`) this list indicates if that direction is periodic (value 1) or non-periodic (value 0). Note: the elements in this list each refer to the direction of the corresponding entry in `lattice_vectors` and *not* the Cartesian x, y, z directions.
```

giovannipizzi:
Hi @rartino, thanks for pinging!
I was waiting to see if there was more feedback, and on top of this I was busy and couldn't work on it yet.

I agree to focus only on explicit vacancies, and thinking a bit better I'm ok to explicitly put them. 
(A client could still complain if it gets something weird, and instead if it get something close to 1 like 0.999999 it can always renormalise it or do something similar).

I don't think we should use `X` though. I'm not sure using an empty string is a good idea (e.g. for querying, assuming we want at some point to allow this?) but I don't know what to suggest here. Maybe a different key (`{"chemical_symbols": ["Ti"], "concentration": [0.9], "vacancy": 0.1}`) to avoid to collide with possible names, but on the other hand it's maybe ugly... And thinking better to it we just need not to collide with chemical symbols, not with any string used as a species, so another suggestion could be to use a reserved word, e.g.: `{"chemical_symbols": ["Ti", "vacancy"], "concentration": [0.9, 0.1]}`.

I like also the suggestion to namespace simple chemical species (e.g. `:C`) but I would do it later (feel free to open another issue!)

If you want to suggest the changes (this GitHub feature is very nice!) please go ahead an I will comment/merge in this PR. Otherwise if it's too complicated to do in the browser either make a PR against my branch, or let me know. Thanks!

giovannipizzi:
After discussion with @rartino @gmrigna @sauliusg @fawzi @gjc29 there's been an agreement for the option `{"chemical_symbols": ["Ti", "vacancy"], "concentration": [0.9, 0.1]}` - I will now update the specs with all details discussed.

optimade.md
giovannipizzi: The only comment I have here is if the DB internally uses "unallowed" characters, then we are forcing it to rename all sites. 

**Suggestion**: Maybe a safe enough option is to just limit our namespace, e.g. simply forbid names to start with a character (e.g. `:`)? 

In the unlikely case that a DB is already doing it, it could automatically do some sort of escaping (also to make sure that when reading the name back, it can convert it back to the internal representation). 

E.g. (this is just an example, it's client specific and will not go in the specs, just for the discussion here):
- use an escape character, say `~`
- prepend `~` if the first character is `:` or the first character is `~`

so e.g.
- "A" -> "A"
- ":B" -> "~:B"
- "~C" -> "~~C"
that guarantees that we don't make it impossible to store names with different characters, nor to make round trips (DB->Optimade->DB)

rartino: But don't we still want to avoid forcing clients to handle names with 'very unexpected characters', e.g., non-printable unicode? (Say, the right-to-left mark...) Species names are a kind of 'variable names', and for such it is common to limit what is allowed.

I think it actually would be safe to let implementations invent any scheme of their own to 'downconvert' identifiers into the allowed character set. But, sure, if we really want roundtrips to be unicode-safe, what about saying that any character outside [a-zA-Z_][a-zA-Z_0-9]* needs to be escaped using, e.g., javascript notation; e.g., \xA9 is the copyright symbol, \x5C is a backslash, and a right-to-left mark would be \u200F. A client that want to worry about these things will thus interpret "Ti\xA9" as Ti©. But, a client that does not care will do fine by just regarding "Ti\xA9" as an abstract species identifier.

rartino: Perhaps like so:
```suggestion
    valid key of the dictionary specified by the `species` property. This string can be of any length, may use upper and lower case letters, the underscore, the hyphen -, the colon :, the period ., and the digits 0-9. While it is recommended to keep species identifiers within these limitations, characters outside the allowed set may be encoded using the backslash symbol \. A sequence \uABCD where ABCD is a four digit hexadecimal number refers to the unicode character with this id. To encode an explicit backslash, use \u005C. At this point, no other backslash constructs are allowed beyond those on the form \uABCD. 
```
If you feel strongly that the second sentence should be expanded with more characters, feel free to suggest that. I am really stretching my imagination to see what this is meant to be used for. Note: if we adopt this, a notation for "special species" should stay outside what is allowed.

PR #10, opened 2018-06-12
Grammar documentation checks

sauliusg:
I have transferred the filter grammar into the main OPTiMaDe document, and the current code now produces grammar files from this documentation, and runs regressions tests against them with 'make tests'.

PR #53, opened 2018-12-19
[WIP] Adding travis file to check grammar

### LINKED ISSUE #52: Add travis tests for grammar checks
Also, add protection on the GitHub branch

giovannipizzi:
Note this also includes a script by Saulius, downloaded
from http://saulius-grazulis.lt/~saulius/scripts/
with the only modification of changing `perl5` -> `perl`

Note: Grammar tests were added in #10

This fixes #52

Please follow [the guidelines here](https://github.com/Materials-Consortia/OPTiMaDe/wiki/How-to-contribute#submit-proposed-changes) to create a new pull request.
If your pull request requires discussion, make sure the OPTIMaDe developer list is alerted (for more info see the link above.)


giovannipizzi:
@sauliusg The Makefile needs some fixes and I need your help here.
The .travis.yml file I added make the test run and "pass", but actually the tests are failing.

Here the issues I found.

High priority:
1. now tests are failing, still we get a zero error status - this should be fixed (both when the files are not found at the beginning:
   ```
   make tests
   tests/makefiles/Makelocal-grammars:27: .grammars.d: No such file or directory
   tests/makefiles/Makelocal-tests:41: tests/cases/.ebnf_numbers_001.d: No such file or directory
   tests/makefiles/Makelocal-tests:41: tests/cases/.ebnf_numbers_002.d: No such file or directory
   [...]
   ```
   and when the later tests fail. (one way, if done in bash, is to use `set -e` at the top of the script (see e.g. [this explanation](https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script) and to make sure that, if you pipe in case of errors, still the error code of each line is the correct one: e.g. this is wrong
   ```
   ls filename || echo "not found"
   ```
   while this is ok
   ```
   ls filename || ( echo "not found" ; exit 1 )
   ```
 1. I added a script that was missing as indicated in the commit message. Still, for some reason some files are not found, I don't know why

Lower priority:
1. is there any guarantee that the tools that we are downloading are always going to be there (grammatica, grammatiker, ...). Maybe it's better if we fork that project inside Materials-Consortia and we download the release from our fork, so we guarantee that the tests don't stop passing because the author removed his repo?

@sauliusg can you fix these issues and make another PR (or take my commit and work on it, as you wish)? Thanks!

sauliusg:
> make tests
> tests/makefiles/Makelocal-grammars:27: .grammars.d: No such file or directory
> tests/makefiles/Makelocal-tests:41: tests/cases/.ebnf_numbers_001.d: No such file or directory
> tests/makefiles/Makelocal-tests:41: tests/cases/.ebnf_numbers_002.d: No such file or directory
> [...]

Actually, these lines are *not* test failures, these are regular Make messages; and later Make wuild build the missing files himself. :)

giovannipizzi:
Ok thanks for the clarification. Maybe it would be good do do a `>/dev/null 2>&1` then?
In any case, the important point is to make sure that if the final state is not fully correct, the exit status is not zero.

giovannipizzi:
This is superseded by #62

PR #54, opened 2018-12-19
Fixing mkcomdepend and grammatica dependencies

sauliusg:
Correcting the missing or external dependencies:
* adding 'mkcomdepend' to tests/tools and setting PATH in the GNUmakefile to find it;
* downloading Grammatica from the Materials-Consortia fork; checking and deleting the downloaded file if the download fails.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -30,6 +30,10 @@ ifneq ("${MAKECONF_FILES}","")
 include ${MAKECONF_FILES}
 endif
 
+TOOLS_PATH ?= tests/tools
+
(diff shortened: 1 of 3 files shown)


giovannipizzi:
Could you do this on top of my PR (or add the .travis.yml I prepared in your PR) so we check if the tests actually pass correctly?
Also, I think you worked on the same branch of the other PR which was squashed, so now we see a lot of commits - maybe you want to do a git rebase or just work off a new branch from develop and apply the few relevant changes?

Issue #22, opened 2018-06-14
Discovery mechanism


rartino:
We have discussed the many benefits to add a discovery mechanism to the OPTIMaDe API.

This requires establishing relations to other databases as `siblings`, `parents`, or `children`. The resonable place for this seems to be a section in the output of the info endpoint.

Some notes:
* All endpoints need to have a string identifier of "which database" is served at a given endpoint so clients can avoid duplication.
* There should be a standard for the name of the "topmost" database (to distinguish it from children), e.g., database_specific_prefix + ".main".

* Children should be named database_prefix + "." + child_name
* A testing database, if provided, should be named: database_prefix + ".test"

rartino:
Here is a concrete suggestion: Extend the /info endpoint data to return information about these relationships:
```
{
  ... <other response items> ...
  "data": [
    {
      "type": "info",
      "id": "/",
      "attributes": {
        "api_version": "v0.9",
        "available_api_versions": {
          "0.9.5": "http://db.example.com/optimade/v0.9/",
          "0.9.2": "http://db.example.com/optimade/v0.9.2/",
          "1.0.2": "http://db.example.com/optimade/v1.0/"
      },
      "formats": [
        "json",
        "xml"
      ],
      "entry_types_by_format": {
        "json": [
          "structure",
          "calculation"
        ],
        "xml": [
          "structure"
        ]
      },
      "available_endpoints": [
        "entry",
        "all",
        "info"
      ]
      "relationships": {
        "siblings": [
          {
            "database_id": "_exmpl2_.main",
            "uri": "http://two.example.com/optimade/v0.9",
            "description": "A database of machine learning generated perovskites."
          },{
            "database_id": "_exmpl3_.main",
            "uri": "http://three.example.com/optimade/v0.9"
          }
        ],
        "parent": {
          "database_id": "_exmpl_.main"
          "uri": "http://example.com/optimade/v0.9"
        },
        "children": [
          {
            "database_id": "_exmpl_.sub_1_1",
            "uri": "http://example.com/1_1/optimade/v0.9"
          },{
            "database_id" = "_exmpl1_.sub_1_2",
            "uri" = "http://example.com/1_2/optimade/v0.9"
          }
       ]
     }        
  ]
}
```

rartino:
Specific member names needs to be discussed to conform to standards. And at this point only siblings. The reason we need database_id and uri/url is that more than one base url can serve data from the same database and the client want to be able to do deduplication.

[...]

PR #57, opened 2019-02-25
simple discovery

fawzi:
Add simple discovery of Optimade APIs

- an api can consist only of the info endpoint
- info endpoint  moves further toward discovery (already had discovery
of other versions). This looks like a better use of info, description
of endpoints is better done via swagger (standard, widely used)
- minimal change: aknowledged_apis attribute
- bootstrap directory in git is a valid api that can provide the first
link to the implementations (via gitlab raw access)
- discovery of sub-databases or other implementations with the same
mechanism

Notes
- names open for discussion
- aknowledged_apis could be extended to be a full info object and contain other aknowledged_apis (to return full trees with fewer requests)
- sorry branch name should have started with issue #22

### DIFF
--- a/bootstrap/info
+++ b/bootstrap/info
@@ -12,6 +12,9 @@
             "type": "api",
             "api_base_uri": "https://raw.githubusercontent.com/Materials-Consortia/OPTiMaDe/simple-discovery/bootstrap/",
             "api_description":["Top level entry point to find optimade compatible apis"]
+            "provider_name": "Optimade",
+            "provider_description": "Root optimade provider",
+            "api_id": "root_db"
           }
         ]
         "formats": [
--- a/bootstrap/info
+++ b/bootstrap/info
@@ -1,0 +1,32 @@
+{
+  "data": [
+    {
+      "type": "info",
+      "id": "/",
+      "attributes": {
+        "api_version": "v0.9.8",
+        "available_api_versions": {
+        },
+        "acknowledged_apis":[
+          {
+            "type": "api",
+            "api_base_uri": "https://raw.githubusercontent.com/Materials-Consortia/OPTiMaDe/simple-discovery/bootstrap/",
+            "api_description":["Top level entry point to find optimade compatible apis"]
+          }
+        ]
+        "formats": [
+          "json"
+        ],
+        "entry_types_by_format": {
+          "json": []
+        },
+        "available_endpoints": [
+          "info"
+        ],
+        "info_description":[
+          "Top level entry point to find optimade compatible apis"
+        ]
+      }
+    }
+  ]
+}
--- a/optimade.md
+++ b/optimade.md
@@ -542,6 +542,11 @@ The response dictionary MUST include the following fields
 	        * **api\_base\_uri**: a string containing an url.
 	            The provided url MUST adhere to the rules of [section '3.1. Base URL'](#h.3.1).
 	        * **api\_description**: a list of strings providing a human readable description in markdown format of the content available through the API
+	        * **provider_name**: a short name for the database provider.
+	        * **provider_description**: a longer description of the database provider.
+	        * **provider_prefix**: an OPTIONAL field giving the database-provider-specific prefix as found in
+	            [appendix 1: Database-provider-specific namespace prefixes](#h.app1).
+	        * **api_id**: an OPTIONAL field containing the provider's local ID.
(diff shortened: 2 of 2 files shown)


sauliusg:
Could you please provide more details on what API you suggest?
IMHO, the API should be as implementation-neutral as possible. I do not know if Swagger has API specification for discovery; if yes please quote it, and if it is mature enough we can opt for using it. As far as implementations go, I thing we should have all options open (so Swagger Codegen should be regarded one of multiple possibilities, and even then generating code strikes me as a bad choice, actually).

I do not quite understand what is a suggested github role in the API. The Github is a possible repo for hosting implementation code, true; but the API specification should be independent of github, I think.

rartino:
A summary of some arguments when we discussed this on the last meeting.

- I don't think these proposals are so different. But, the alternative in #56 seems more future-proof. If we introduce `/databases/siblings`, this opens up for later adding things like `/databases/sub`, `/databases/super`, `databases/providers`, etc. that can be developed into a complete system for discovery of large amounts of other base URLs, including filtering, etc. 
- While putting everything in /info as suggest here in #57 is somewhat simpler, the difference isn't huge. You can still achieve the same thing by just hosting three files instead, e.g., /info/index.html + /databases/index.html + /databases/siblings/index.html.
- After reading over the actual PR text of #57 again, I'm a bit confused over the discussion we had about #57 supporting "a tree" vs. #56 only supporting "two levels, provider + databases". 

  The present description of the field `acknowledged_apis` is: `[APIs] somehow connected to the implementor of this API`, which I read a synonymous with "OPTiMaDe Base URLs connected to the same database provider" (correct me if I'm misunderstanding this). That would make this field the exact same thing as "/databases/siblings" in #56. 

  But perhaps the idea is to change that description to really allow for a tree? I.e., that Base URLs pointed out in `acknowledged_apis` are not supposed to point back 'upwards'? If so, please do make that change to the PR. However, I think I prefer waiting with this form of reference between OPTiMaDe Base URLs, and for now just strictly allowing discovery of siblings = all Base URLs of one and the same database provider.

fawzi:
@sauliusg the swagger thing was more a comment it should be moved to another place. But yes swagger open api json description 
https://swagger.io/docs/specification/about/
Should be used, and then you can use many nice tools, that is what we did (and many frameworks help you in defining the open api json document.

wrt github, I wanted to show that it can be used as server for a set of static files, and if the API can be reduced to few static files then GitHub can be an API that is always up to date wrt git.
Indeed an “owned” domain (and web server) is cleaner.

fawzi:
@rartino 
Indeed two proposal now are close. If database/siblings is interpreted as base uris. I plan to rewrite it with that to minimimize the differences.
The aknowledged_uris builds a tree because it might refer to both top level apis empty (index) dB and leaves DBs. At the moment you get only one level (children) with one request, one could return also the whole tree in a single request.
56 differentiates top level, index and leaves, I would not, that is the only remaining diff (if as I plan this we I rewrite it with databases/siblings

rartino:
@fawzi 
> The aknowledged_uris builds a tree because it might refer to both top level apis empty (index) dB and leaves DBs. 

But, does this interpretation really match your definition of the field `acknowledged_apis` in the PR? Or do you intend to update that definition? 

To clarify:
- In the PR the definition of `acknowledged_apis` is given as: *"List of Optimade APIs that are aknowledged by the implementor of this API [and which must be] connected to the implementor of this API"*. The way I read that, it means a provider should just put all their base URLs into `acknowledged_apis` everywhere---i.e., all the base URLs that this provider manages will contain the exact same list of siblings in /info. The geometry of that is a 'complete graph' (i.e., all nodes point at all other nodes), not a tree. I don't see how the present description prevents (or even discourages) two nodes from pointing at each other, which would be needed for `acknowledged_apis` to describe a tree. 

- Also the way I read that definition, it does not allow setting up a bootstrap/info where `acknowledged_apis` reference all different provider's base URLs---since they would not all be *"connected to the implementor of this API"*.

But, perhaps this is a side discussion. Whichever geometry we prefer could just as well be accommodated under, e.g., `/databases/`. Maybe it is more important to first agree on whether these links (regardless if they describe a tree or a list of siblings) belong under `/info` or under its own endpoint prefix, e.g. `/databases`.

gmrigna:
This is going to be handled in #63

PR #62, opened 2019-03-22
Travis CI setup

sauliusg:
This PR sets up the Travis CI system to check grammars in the OPTiMaDe documentation. Everything seems to work :)

CAVIAT: It currently depends in svn://saulius.grazulis.lt/grammatiker; I know you will not like it; you would like it even less if you knew what hardware it runs in :P, so I will not tell you (unless you ask) :). But for now it works, and we can change it later, gradually (the 'grammatiker' that it hosts there is also pushed to Github.

Please review and merge if OK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,14 @@ addons:
         packages:
             - unzip
             - default-jdk
+            - subversion
 
 # Skip cpanm installation
 install: []
 
 script:
     - make --version
+    - make -n
     - make -n tests
+    - make
     - make tests 2>&1 | tee make.log && if grep FAILED make.log; then exit 123; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,5 @@ install: []
 script:
     - make --version
     - make tools
-    - make tests 2>&1 | tee make.log && if grep :\ FAILED make.log; then exit 123; fi
+    - make tests
+    - exit $(make listdiff | head | wc -l)
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ install: []
 
 script:
     - make --version
-    - make 2>&1 | tee make.log && if grep :\ FAILED make.log; then exit 123; fi
+    - make tests 2>&1 | tee make.log && if grep :\ FAILED make.log; then exit 123; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,6 @@ addons:
 install: []
 
 script:
+    - make --version
+    - make -n tests
     - make tests 2>&1 | tee make.log && if grep FAILED make.log; then exit 123; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,5 @@ install: []
 
 script:
     - make --version
-    - make -n
-    - make -n tests
     - make
     - make tests 2>&1 | tee make.log && if grep FAILED make.log; then exit 123; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ addons:
 install: []
 
 script:
-    - make tests 2>&1 | tee make.log && if ! grep FAILED make.log; then exit 123; fi
+    - make tests 2>&1 | tee make.log && if grep FAILED make.log; then exit 123; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,5 @@ script:
     - make tools
     - make grammars
     - make tests
+    - make listdiff
     - exit $(make listdiff | head | wc -l)
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,4 @@ install: []
 
 script:
     - make --version
-    - make
-    - make tests 2>&1 | tee make.log && if grep FAILED make.log; then exit 123; fi
+    - make 2>&1 | tee make.log && if grep FAILED make.log; then exit 123; fi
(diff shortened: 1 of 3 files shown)


sauliusg:
On 2019-03-22 18:31, Rickard Armiento wrote:
> But, to comment more generally on all the things under tests/. I am
> very grateful for the hard work put into testing part of the filter
> grammar. But, I am a bit skeptical that the specification repository
> really is the right place for all this technical implementation of a
> testing framework itself.

The tests are needed to check whether:

a) the formally specified grammar is unambiguous;

b) text strings we deem as correct are indeed derivable in the grammar
(and flagged as correctly parsed by a parser derived from the grammar);

c) text strings that we agree are incorrect are not derived from the
grammar (and flagged as errors by a parser derived from the grammar).

I have seen projects where grammar is written in "documentation", and
not checked by a computer; such grammars more often than not contain
mistakes. Automated testing of the grammar against known texts does not
of course prove that the grammar is correct, but at least guards against
some mistakes. Thus, I took opportunity to implement this here so that
we can have automated grammar checks. In my experience, they really help
to catch errors in the *grammar* itself (we ar not talking about any
implementation yet).

Imagine, for example, that we forget to include braced expressions into
the grammar, like in the change:

Term = Comparison |
-       '(', [Spaces], Expression, [Spaces], ')'  |
        NOT, [Spaces], Term
        ;

do that, and you will see immediately that multiple tests are failing.
This is a clear indication that something is wrong with your grammar.

You need some framework to test the grammar in you *documentation*;
these tests are not supposed to check your implementation(s), but your
*documentation* (which can be buggy as well ;). If you do not have any
test tools (I would not call them framework, they are rather a set of
tools), you can not test.

> At the very least I think there is some documentation needed to
> explain what the purpose of all the interconnected scripts are.

The point is well taken, but given our limited resources allocated for
the Optimade project, we need to take priorities. The installed test
directory is rather simple, very general, and at least several people in
Optimade community now them very well. We use similar testing method for
most of our projects, and it works very well. So while I'm happy to
provide more docs on the subject, I can not do that right now. I suggest
we discuss during the upcomming CECAM meeting; I can give a short
explanation on the test system in use.

> For testing my own implementation, what I see a clear use for is
> files from where one can extract long lists of 'filter input string'
> -> 'parse tree' with the correct parse trees represented using some
> nice application-independent abstract form with the symbols in the
> grammar. We have tests/inputs and tests/outputs, but the output
> format to check against is too application-specific; I'd have to work
> almost as hard as I did to write the parser to parse those parse
> trees into something I can compare against :-)

The purpose of the tests was not to fix parse trees, but only to make
sure that certain strings are accepted or not accepted by the grammar
(i.e. by a parser automatically built from the grammar).

If you think that specifying default parse trees would be useful, we can
talk and pick a standard tree representation format so that any
implementation can check against them (maybe Lisp lists?).

On the other hand, I am not sure if parse trees *must* be the same for
every implementation; maybe AST's can be, but not the detailed
derivations. There can be several equivalent parsed expression
representations, application can make equivalent transformations, and
even for identical trees we can have different interpretations (i.e.
different semantics attached to the tree). Defining semantics formally
and testing it is currently beyond the scope of the suggested tests.

Hope this clarifies the situation.

Regards,
Saulius

-- 
Dr. Saulius Gražulis
Vilnius University Institute of Biotechnology, Saulėtekio al. 7
LT-10257 Vilnius, Lietuva (Lithuania)
fax: (+370-5)-2234367 / phone (office): (+370-5)-2234353
mobile: (+370-684)-49802, (+370-614)-36366

rartino:
>>  At the very least I think there is some documentation needed to
 explain what the purpose of all the interconnected scripts are.

> The point is well taken, but given our limited resources allocated for
the Optimade project, we need to take priorities. 

Thanks for taking the time to explain. Since you did put in this work to explain to me, perhaps you don't mind if I just take most of what you've replied with here and put in a README.md in the test/ directory? While that does not provide 'complete instructions' for the test tools, even just such a README would be quite useful for someone who checks out the repository and wonders what all these things are.

> On the other hand, I am not sure if parse trees *must* be the same for
every implementation; maybe AST's can be, but not the detailed
derivations. 

I may be using the terminology incorrectly here, but what I mean by a parse tree in this context (in contrast to an AST) is a structure that exactly represents the grammar reductions made when parsing. E.g., let `a,b,c,d` be terminals and `A, B, C` be non-terminals. With the grammar rules:
```
A -> B, D, d;
B -> a;
D -> b, c;
```
The parse tree for the input `a b c d` would in LISP-like notation be:
```
  (A, ( 
      (B, (a))
      (D, (b, c))
   d) )
```
These parse trees should be unique if the grammar is non-ambiguous, right? Or is that what you are saying isn't the case? Can you give an example? I assume we want our grammar to be non-ambiguous?(Because otherwise there would be more than one valid 'filter' to run for a given OPTIMaDe filter string, which seems a bad thing.)

> You need some framework to test the grammar in you *documentation*; these tests are not supposed to check your implementation(s), but your *documentation* (which can be buggy as well ;). If you do not have any test tools (I would not call them framework, they are rather a set of tools), you can not test.

I completely agree with you that these tests are a great idea to do before including a grammar change in the specification. I'm just not 100% convinced the 'test tools' (rather than just the 'test cases') belong in the repository of the specification. But it isn't something I feel very strongly about for the moment, so lets indeed postpone that discussion to the meeting.

Issue #46, opened 2018-06-19
Support sub-databases within each database


giovannipizzi:
This issue suggests a possible implementation of the concept of sub-databases within a database provider registered in the OPTiMaDe specification.

This is useful for database providers (like e.g. [Materials Cloud](http://www.materialscloud.org)) where there isn't necessarily a "main" database, but many databases, all however provided by the same database provided (Materials Cloud, in this case), exposing the same API, but managed independently by different researchers (as an analogue, imagine GitHub repositories, where each user can manage his own GitHub repository, but the API is provided for all DBs by GitHub itself).

While this could be achieved by implementing the current API for each sub-DB, this would require to register each of them with the current specification (that, at the bottom, lists DB providers and their URL) but this is not a viable approach for cases where the sub-DBs within one registered DB provider can be added and removed by users at any time.

This proposal has the following aims:
- provide a minimal extension to the current API, supporting a minimal but useful usecase (having multiple DBs), and defer to the future further interesting extensions (see e.g. #22, that focuses more on discovering other DBs)
- allow for the concept of 'default' DB for those who only have only one "default" sub-DB
- it should require only minimal effort to implement (especially for those who only have only one "default" sub-DB) - the current addition can be implemented as a static JSON file by DBs that only have one (or a fixed number of) sub-DBs.

As side-effect:
- this allows to easily define a "test" DB (that was discussed at the June 2018 Optimade meeting), see #44.

## Proposed implementation

1. Implement the same API (the one already defined by OPTiMaDe) for each sub-DB, where each sub-DB has
   a different `base_url`, for instance:
   - http://example.com/mainDB/optimade/v0.9/
   - http://example.com/testDB/optimade/v0.9/
   - http://example.com/AThirdDB/optimade/v0.9/
   
   (versions might be different for each DB - no requirement here). It is up to the implementation to decide the best  approach to reuse the API code.

1. In the OPTiMaDe API specification Appendix, where now we 
   list only a prefix with a generic name of the DB, e.g.:
   ```
   _exmpl_: used for examples, not to be assigned to a real database
   ```
   replace this with a common, well defined API endpoint, e.g. `http://www.example.com/optimade-discovery/`, like
   ```
   _exmpl_: http://www.example.com/optimade-discovery/ (used for examples, not to be assigned to a real database)
   ```
   whose format we describe below.
   Ideally, I think we should also add a machine-readable list for automated discovery.

[...]

Issue #44, opened 2018-06-19
Each DB provider should have a test sub-DB


giovannipizzi:
This would allow for
- centralized testing of the API (unit tests written only once and tested on all known/registered DB providers)
- easy checking  of the currently implemented features

Moreover, if we agree on the content of the test DB (say ~200 structures) each DB can store them in the internal format of the DB, and it is easy to check additionally that the way structures are returned by the API (e.g. in corner cases like partial occupancies, low-dimenasional systems, assemblies, ... - see #23) is consistent and correct across implementations.

rartino:
I like this, but, specifically using the word SHOULD, i.e., an API implementation is still conformant without it.

merkys:
At COD we already have a test API endpoint: http://www.crystallography.net/cod-test/optimade/info. Feel free to test it! Keep in mind that we also use that sub-database as a testbed, from time to time adding crafted crystal structures or removing them altogether.

ml-evs:
Doing a bit of issue triage, I think this can be closed after the eventual approach we took with index meta-databases and the validation dashboard.

PR #63, opened 2019-03-22
Introducing 'links' endpoint and 'index' meta-database.

CasperWA:
# Overview
New PR as the result from consortium discussions and meetings.
Related to PRs #56 and #57.
Related to issues #22, #44, and #46.

This will bump the version up from v0.9.7-develop to v0.9.8-develop.

## `links` endpoint
This endpoint can contain 'child' objects and up to a single 'parent' object.
These objects are related OPTiMaDe databases/implementations with respect to the current.

The new `links` endpoint thereby provides a way for one OPTiMaDe API database/implementation to point to a number of related OPTiMaDe API databases/implementations  by the *same* provider.

The structure of the endpoint can be seen in the diff, where an example is also provided.

## Index meta-database
A non-mandatory minimal index API that only serves the endpoints `info` and `links`.

The main purpose of the index is to allow for automatic discoverability of all databases of a given provider. Thus, it acts as a meta-database.

If a provider wants to be part of the "List of providers" (i.e. the new `providers.json` file, see below), it is mandatory for the provider to supply the index meta-database.

The index is defined such that it can be setup using static JSON files, i.e. the endpoints need not be queryable.

There are minor requirements for a provider to implement when also implementing the index meta-database. An example of this is the choice to include a [JSON API relationships object](https://jsonapi.org/format/#document-resource-object-relationships) in the `info` endpoint *only* for the index meta-database. This is to have the possibility of defining a provider default database.
See the details in the diff.

## `providers.json`
The list of database providers from the API specification has been moved into a separate JSON file called 'providers.json'.
The file serves as a valid OPTiMaDe API response, with the exceptions of the members `api_version` and `time_stamp` in the top-level `meta` field.
This is so, because the file should be considered a separate entity from the evolution of the specifications. It is therefore also suggested to at a later stage completely remove this file from the current repository. See issue #64.

## Minor changes
Updated base URL `info`-endpoint to include `available_endpoints`.
This item was included in the JSON example, but not in the official description.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1,4 +1,4 @@
-# OPTIMADE API specification v0.9.7-develop
+# OPTiMaDe API specification v0.9.8-develop
 
 [1. Introduction](#h.1)
 
@@ -11,6 +11,7 @@
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[3.3.1. Response format](#h.3.3.1)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[3.3.2. JSON API response schema: common fields](#h.3.3.2)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[3.3.3. HTTP response status codes](#h.3.3.3)  
+&nbsp;&nbsp;&nbsp;&nbsp;[3.4. Index](#h.3.4)  
 
 [4. API endpoints](#h.4)  
 &nbsp;&nbsp;&nbsp;&nbsp;[4.1. Entry listing endpoints](#h.4.1.)  
@@ -25,7 +26,12 @@
 &nbsp;&nbsp;&nbsp;&nbsp;[4.4. Info endpoints](#h.4.4)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[4.4.1. Base URL info endpoint](#h.4.4.1)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[4.4.2. Entry listing info endpoints](#h.4.4.2)  
-&nbsp;&nbsp;&nbsp;&nbsp;[4.5. Custom extension endpoints](#h.4.5)  
+&nbsp;&nbsp;&nbsp;&nbsp;[4.5. Links endpoint](#h.4.5)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[4.5.1. Response schema](#h.4.5.1)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[4.5.2. 'parent' and 'child' objects](#h.4.5.3)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[4.5.3. Index links endpoint](#h.4.5.3)  
+&nbsp;&nbsp;&nbsp;&nbsp;[4.6. Custom extension endpoints](#h.4.6)  
+
 [5. API Filtering Format Specification](#h.5)  
 &nbsp;&nbsp;&nbsp;&nbsp;[5.1. Lexical tokens](#h.5.1)  
 &nbsp;&nbsp;&nbsp;&nbsp;[5.2. The filter language syntax](#h.5.2)  
@@ -204,47 +210,83 @@ Every response MUST contain the following fields:
           "_example.com_db_version": "3.2.1"
         }, 
         ``` 
-* **meta**:
-    * **query**: information on the query that was requested.
-        * It MUST be a dictionary with these items:
-            * **representation**: a string with the part of the URL that follows the base
-            URL.
-    * **api\_version**: a string containing the version of the API .
-    * **time\_stamp**: a string containing the date and time at which the query
-      was executed, in [ISO 8601](https://www.iso.org/standard/40874.html)
-      format.  Times MUST be timezone-aware (i.e. MUST NOT be local times),
-      in one of the formats allowed by [ISO 8601](https://www.iso.org/standard/40874.html)
-      (i.e. either be in UTC, and then end with a Z, or indicate explicitly
-      the offset).
-    * **data\_returned**: an integer containing the number of data returned for the query.
-    * **more\_data\_available**: 'False' if all data for this query has been
-      returned, and 'True' if not.
-    * The dictionary MAY also include these OPTIONAL items:
-        * **data\_available**: an integer containing the total number of data
-          available in the database. 
-        * **last\_id**: a string containing the last ID returned.        
-        * **response\_message**: OPTIONAL response string from the server.
-        * Other OPTIONAL additional information _global to the query_ that is not specified
-          in this document, MUST start with a database-provider-specific prefix as defined in [Appendix 1](#h.app1).
-    * Example:  
-        For a request made to http://example.com/optimade/v0.9/structures/?filter=a=1 AND b=2
+* **meta**: a [JSON API meta object](https://jsonapi.org/format/#document-meta)
+  that contains non-standard meta-information.
+  
+  It MUST be a dictionary with these members:
 
-        ```
-        {
-          "meta": {
-            "query": {
-              "representation": "/structures/?filter=a=1 AND b=2",
-            },
-            "api_version": "v0.9",
-            "time_stamp": "2007-04-05T14:30Z",
-            "data_returned": 10,
-            "data_available": 10,
-            "more_data_available": false
+  * **query**: information on the query that was requested.
+
+    It MUST be a dictionary with these items:
+    * **representation**: a string with the part of the URL that follows the base URL.
+  * **api\_version**: a string containing the version of the API .
+  * **time\_stamp**: a string containing the date and time at which the query
+    was executed, in [ISO 8601](https://www.iso.org/standard/40874.html)
+    format.  Times MUST be timezone-aware (i.e. MUST NOT be local times),
+    in one of the formats allowed by [ISO 8601](https://www.iso.org/standard/40874.html)
+    (i.e. either be in UTC, and then end with a Z, or indicate explicitly
+    the offset).
+  * **data\_returned**: an integer containing the number of data returned for the query.
+  * **more\_data\_available**: 'false' if all data for this query has been
+    returned, and 'true' if not.
+  * **provider**: a dictionary describing the provider of this OPTiMaDe implementation.
+  It MUST contain the following members:
+    * **name**: a short name for the database provider.
+    * **description**: a longer description of the database provider.
+    * **prefix**: database-provider-specific prefix as found in
+    [appendix 1: Database-provider-specific namespace prefixes](#h.app1).
+
+    **provider** MAY also include these OPTIONAL members:
+
+    * **homepage**: a [JSON API Links object](http://jsonapi.org/format/#document-links),
+    pointing to the homepage of the provider,  either directly as a string, or as a link object which
+    can contain the following members:
+      * `href`: a string containing the homepage URL.
+      * `meta`: a meta object containing non-standard meta-information about the provider's homepage.
+    * **index_base_url**: a [JSON API Links object](http://jsonapi.org/format/#document-links)
+    pointing to the `base_url` for the `index` databae of the provider as specified in [appendix 1](#h.app1), either
+    directly as a string, or as a link object which can contain the following members:
+      * `href`: a string containing the `index_base_url` for the provider.
+      * `meta`: a meta object containing non-standard meta-information about this link.
+
+    If the index meta-database (see section [3.4 Index](#h.3.4)) is implemented by the provider, the `index_base_url` member MUST be included.
+
+  **meta** MAY also include these OPTIONAL items:
+
+  * **data\_available**: an integer containing the total number of data
+    available in the database.
+  * **last\_id**: a string containing the last ID returned.
+  * **response\_message**: OPTIONAL response string from the server.
+  * Other OPTIONAL additional information _global to the query_ that is not specified
+  in this document, MUST start with a database-provider-specific prefix as defined in
+  [Appendix 1](#h.app1).
+
+  * Example:  
+      For a request made to http://example.com/optimade/v0.9/structures/?filter=a=1 AND b=2
+
+      ```json
+      {
+        "meta": {
+          "query": {
+            "representation": "/structures/?filter=a=1 AND b=2",
+          },
+          "api_version": "v0.9",
+          "time_stamp": "2007-04-05T14:30Z",
+          "data_returned": 10,
+          "data_available": 10,
+          "more_data_available": false,
+          "provider": {
+            "name": "Example provider",
+            "description": "Provider used for examples, not to be assigned to a real database",
+            "prefix": "exmpl",
+            "homepage": "http://example.com"
           }
-          ...
-          <additional response items>
         }
-        ```
+        ...
+        <additional response items>
+      }
+      ```
+
 * **data**: The schema of this value varies by endpoint, it can be either a [JSON API Resource Object](http://jsonapi.org/format/#document-resource-objects) or a list or JSON API Resource Objects.
   Every resource object need the `type` and `id` fields, and its attributes (described in [section '4. API endpoints'](#h.4) of this document) need to be in a dictionary corresponding to the **attributes** field.
 
@@ -258,7 +300,7 @@ If there were errors in producing the response all other fields can be skipped, 
 
 An example of a full response:
 
-```
+```json
 {
   "links": {
     "next": null,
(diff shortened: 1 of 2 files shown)


CasperWA:
> Could you please clarify this point? With this clarification in place, I feel like approving this PR.

I'll try my best :)

I have always thought of the `providers.json` to be handled in the following way, feel free to pinpoint where I may have gone wrong in these presumptions. I will try to do so myself afterwards:

- The different implementations, i.e. the providers, do *not* know of `providers.json`.
- Only those clients who choose to implement a "discovery" feature will know of the list of providers.
- Since they expect any response to come in the standard OPTiMaDe API form, I tried to design the file as such.
  When doing this, the `time_stamp`, which has a dynamic value, makes no sense, since we want `providers.json` to be a static file.
- Simultaneously I thought of the separation of the organizational and technical aspects, and the idea that `providers.json` shouldn't necessarily follow the versioning of the OPTiMaDe API specs.

However, on this last point, I can see it doesn't make much sense, since if the technical specs are changed in a way that will make the response of `providers.json` invalid. So the member `api_version` should be kept in.


Now. I see my error - I think.
I believe instead what should be done is the following:

- Make `providers.json` only focus on the top-level `data` member/field. Either by making
  - a response that only returns `data` or
  - making it contain directly the array/list of provider objects, i.e. the value of `data`.

The latter makes it possible to pipe it directly through in a full response.

This will, at the moment, create a special clause for the clients on how to deal with the file. They will either have to create the full response themselves, or simply treat the file as the `data` field, forgetting about `meta` and `links`.

This will also make it easier for the future to (as an example) implement a `providers` endpoint specially for the "index" meta-database that serves this file as the value for `data` and thus the separate provider implementations will start to know about each other. Cross-implementational auto-discovery is/will be one step closer :)

This is a bit long, but I hope I have cleared it up a bit.
What do you think?

rartino:
Can you clarify why the providers.json at all need a meta and links section? Here is the only thing about this I find in the jsonapi specification:
```
A document MUST contain at least one of the following top-level members:

- data: the document’s “primary data”
- errors: an array of error objects
- meta: a meta object that contains non-standard meta-information.
The members data and errors MUST NOT coexist in the same document.

A document MAY contain any of these top-level members:

- jsonapi: an object describing the server’s implementation
- links: a links object related to the primary data.
- included: an array of resource objects that are related to the primary data and/or each other (“included resources”).
```
So, the way I read that, is it not perfectly fine to just provide a static file with only a `data` field? Am I missing something?

I thought we were going in the direction (in this PR or a later one) to setup the 'providers.json' file in a valid OPTIMaDe base-url-mimic:ing structure already in the repository; I.e.,
```
   OPTiMaDe/providers/info
   OPTiMaDe/providers/links
   OPTIMaDe/providers/v0.9/info
   OPTIMaDe/providers/v0.9/links
```
Where 'info' is a static file with some basic info about this being an OPTIMaDe index pointing to providers, and 'links' is a static file with the contents of providers.json. But, now as I type this out, I see it requires duplicating the files and looks a bit inelegant in a repository. So perhaps the separate 'providers.json' file is OK, and we just host it like above on `www.optimade.org/providers`? 

If this is the direction we are going, shouldn't we rather say that the format of providers.json mimics a valid response to the 'links' section?

(Finally, rather than explicitly referencing 'github' explicitly, I'd just replace that with 'the repository' or 'the main repository'. It'd be nice if an event that forces us to move the repository to another host doesn't invalidate the specification.)

CasperWA:
> So, the way I read that, is it not perfectly fine to just provide a static file with only a data field? Am I missing something?

Concerning just the JSON API specs, this is correct. And I will happily do this, as I also mentioned in my previous comment.

My understanding was simply related to this comment of yours:

> I thought we were going in the direction (in this PR or a later one) to setup the 'providers.json' file in a valid OPTIMaDe base-url-mimic:ing structure already in the repository; I.e.,

To relate this to the above, I thought the file should also replicate an OPTiMaDe API specification response, and not the "vanilla" JSON API specification. In OPTiMaDe, the fields `links` and `meta` (as well as `data`) are (as far as I can see) mandatory fields in a response. The only way to legally omit them is to include the `error` field.

However, that being said, I much prefer having the `providers.json` file only return the `data` field, or simply the value of that field (i.e. the array/list of providers). The latter way will make it easy to pipe it through to another server entity or similar.


Concerning what you really ment in your comment above when writing: 

> I thought we were going in the direction (in this PR or a later one) to setup the 'providers.json' file in a valid OPTIMaDe base-url-mimic:ing structure already in the repository; I.e.,

I think you answered that yourself nicely :)

> ... But, now as I type this out, I see it requires duplicating the files and looks a bit inelegant in a repository. So perhaps the separate 'providers.json' file is OK, and we just host it like above on www.optimade.org/providers?

I like this idea as well.

Keeping the complete `data` field in `providers.json` or just the value (as described above), will make future additions easy and the serving of the file can be easily separated from the file itself by implementing a stripped down OPTiMaDe implementation at www.optimade.org/providers.

> If this is the direction we are going, shouldn't we rather say that the format of providers.json mimics a valid response to the 'links' section?

As `links` is defined as being relative to the current implementation, and its contained resource objects are essentially internal implementations or databases, I do not think this is the way to go. Instead, keeping it as a separate special minimal implementation at www.optimade.org/providers will make it a simple step to introduce a similar `providers` endpoint (or something else) in the future to enable cross-provider auto-discovery of other OPTiMaDe providers and their implementations. The special and minimal implementation may act as a "prototype" of this future `providers` endpoint (or similar).
Granted, the endpoint may look similar to `links`, but will be on a different hierarchical level.

> (Finally, rather than explicitly referencing 'github' explicitly, I'd just replace that with 'the repository' or 'the main repository'. It'd be nice if an event that forces us to move the repository to another host doesn't invalidate the specification.)

This makes sense. I will change it accordingly, thanks :+1:

rartino:
> To relate this to the above, I thought the file should also replicate an OPTiMaDe API specification response, and not the "vanilla" JSON API specification. In OPTiMaDe, the fields links and meta (as well as data) are (as far as I can see) mandatory fields in a response. The only way to legally omit them is to include the error field.

Oh, that is what I had missed completely (somehow I thought those were optional also in our specification). Then we have a few options, either do as you did and say that time_stamp doesn't need to be updated, or in this PR also loosen the criteria that these parts must be included to SHOULD. Because I really think it is iffy to require these "dynamic-ish" response parts when hosting static files. It is also not really anything in these that *must* be there for the client, so I'm not sure why were mandating them. Maybe someone else can comment on that.

> As links is defined as being relative to the current implementation, and its contained resource objects are essentially internal implementations or databases, I do not think this is the way to go. Instead, keeping it as a separate special minimal implementation at www.optimade.org/providers will make it a simple step to introduce a similar providers endpoint (or something else) in the future to enable cross-provider auto-discovery of other OPTiMaDe providers and their implementations. The special and minimal implementation may act as a "prototype" of this future providers endpoint (or similar).
Granted, the endpoint may look similar to links, but will be on a different hierarchical level.

Wait, so how do you envision the URL structure on www.optimade.org that hosts the contents of providers.json?

I thought the idea with all the parent/child/providers/auto-discovery stuff was that it would go into `links/`.  Hence, this is the structure I was envisioning:
```
  http://www.optimade.org/providers/info   
  http://www.optimade.org/providers/providers/links <- this is providers.json
  http://www.optimade.org/providers/providers/v0.9/info 
  http://www.optimade.org/providers/providers/v0.9/links  <- this is also providers.json
```
But then providers.json needs to be "links"-like. I'd be fine if it points out providers as `type="child"` (which is the model @fawzi was pushing quite heavily for.) I'd be equally fine if it uses `type="provider"`. 

Using this model keeps down on the amount of different types of responses a client needs to parse. All responses that represent "links" to other OPTIMaDe base_urls will then be in the format of a response to `links/`

CasperWA:
> But then providers.json needs to be "links"-like. I'd be fine if it points out providers as `type="child"` (which is the model @fawzi was pushing quite heavily for.) I'd be equally fine if it uses `type="provider"`.
> 
> Using this model keeps down on the amount of different types of responses a client needs to parse. All responses that represent "links" to other OPTIMaDe base_urls will then be in the format of a response to `links/`

This is fine for me. Although, it - in my opinion - will mix different resource objects (since one is provider, i.e. whole other possible tree of OPTiMaDe databases) and another is databases that are part of the current tree).
However, they are of course all common in the sense that they all reference an OPTiMaDe base URL.
And because this is the case, it is fine for me, as long as the "provider"-objects are differentiated with the type `"provider"`.
This will also make it immediately possible to auto-discover all other OPTiMaDe providers - which I do not think was the initial goal of this PR - but I do not mind it as an added bonus in this way, if the loudest majority thinks it a good idea ;)

As a partly philosophical note, I also do not see this as a mix of organizational and technical details, but only as a technical detail (i.e. the text concerning `"provider"` resource objects can stay in the specifications and this repository). Since it will NOT be a MUST to retrieve `providers.json` and serve it (but perhaps it is RECOMMENDED?). In either way, it should be stated that `links` is the only endpoint under which it may be served and the resource objects MUST have `"type"="provider"`.

> Wait, so how do you envision the URL structure on www.optimade.org that hosts the contents of providers.json?

As an OPTiMaDe implementation I didn't envision anything else than an amputated implementation of a single endpoint at www.optimade.org/providers. But I see why this is maybe not ideal :)
Especially, since the official specification does not mention a `providers` endpoint at all.

CasperWA:
@sauliusg and @rartino.

I have now updated both the specifications and `providers.json` in accordance with our discussion in this PR.
Please, update your reviews/comments, thank you :)

@rartino: I will make another PR for the rule of SHOULD for the top-level response members/fields, since it is essentially a different subject.

rartino:
After thinking a bit more about the `meta` and `links` sections, an alternative way of handling this compared to #66, would be to append the relaxation of that requirement specifically only for Index endpoints right under the new section 3.4. Index, I.e., right below this:
*These endpoints do not need to be queryable, i.e. they MAY be provided as static JSON files.
However, they MUST return the correct and updated information on all currently provided databases.*

Maybe add this:

To simplify serving Index endpoints with static hosting, a few of the otherwise mandatory requirements of the OPTIMaDe specification are relaxed specifically for Index endpoints as follows:
- It is allowed (and for static hosting recommended) that the response contains *only* the `data` field.
- The jsonapi requirements for content negotiation are not mandatory, i.e., the server may respond to requests for any content-type, and is not mandated to respond with Content-Type: application/vnd.api+json, as long as the contents of the response is encoded in UTF-8 and parseable as a json dictionary.

Note, however, that the requirement that the client MUST be allowed to access the endpoints also under an url where the OPTIMaDe API version number is appended is mandatory also for Index endpoints.

What do you think? The segment about dropping the Content-Type requirement is as far as I can see needed here in any case.

rartino:
> However, they are of course all common in the sense that they all reference an OPTiMaDe base URL.
And because this is the case, it is fine for me, as long as the "provider"-objects are differentiated with the type "provider".

> As a partly philosophical note, I also do not see this as a mix of organizational and technical details, but only as a technical detail (i.e. the text concerning "provider" resource objects can stay in the specifications and this repository).

 Well, arguably, any reference to a 'provider' can be seen as an organizational thing that does not belong in the specification. An alternative way of doing things would be to put in the specification ONLY that there are OPTIMaDe base_url:s, they can have children, and they can have one or zero parents. And they can host actual data or be empty (i.e., they can be an Index). And that is all that goes into the specification. I *think* this was more or less what @fawzi was arguing for (feel free to agree or correct me.) 

Then, we'd draft a separate document about the "organizational rules of the OPTIMaDe working group" (or what we call ourselves :) ). That document would say something like this: 

---
The OPTIMaDe working group provides an optional service to keep track of hosted OPTIMaDe databases. To join you have to fulfill the following criteria:
- You need to provide materials data of somewhat sufficient quantity and quality.
- You need to host what the OPTIMaDe specification refers to as an Index. The children of that Index must be all the OPTIMaDe base_urls that host your data. (It must be an Index, i.e., this first level cannot provide any data.) 
- Your children may not have further children.

If you fulfill all these criteria, you will be assigned a provider-specific prefix and your Index URL will be added as a child to our Index OPTIMaDe base_url.

---
This way all organizational matters would be separated from the specification, and we allow for other organizations to pop up with different rules.

**With that said**, while I think it may be a good idea to eventually go for this complete separation of technical and organizational parts,  I do not see this as having any urgency. So, for now, I'm happy with wherever we land with this `providers.json` file. (And I'm honestly starting to think that it may have been a mistake to combine the 'parent' + 'child' + 'Index' PR with edits of the providers.json...)

optimade.md
```diff
@@ -618,7 +762,141 @@ Example:
...
+current.
+
+> **Note**: The RECOMMENDED number of layers is two: The top layer *only* contains the 'root' OPTiMaDe
+> implementation. This is considered the 'parent' of all implementations in the second
+> layer. The second layer contains *all other* OPTiMaDe implementations. These have no
+> 'child' objects in their `links` endpoint and exactly one `parent` object pointing to 'root'.
+
+### <a name="h.4.5.3">4.5.3. 'provider' objects</a>
+
+JSON API Resource objects that MAY be present under the `links` endpoint.
```
rartino: I find section 4.5.3 difficult to understand. This I don't think is a complete sentence: *"JSON API Resource objects that MAY be present under the `links` endpoint."* I also question the need to mandate that exactly the `providers.json` file is served. Even so, I also don't understand the intent with `It is RECOMMENDED to serve these objects`. For whom?

Can't this section just say that ~ "Providers objects are meant to indicate links to Indexes hosted by database-providers." (which are terms already defined in the specification). Does all the strict mandating of who may include which objects have a purpose?

optimade.md
```diff
@@ -618,7 +762,141 @@ Example:
...
+### <a name="h.4.5.4">4.5.4. Index links endpoint</a>
+
+If the provider implements an index meta-database (see section [3.4 Index](#h.3.4)),
+the index meta-database MUST be the 'root' OPTiMaDe implementation.
+
+All other OPTiMaDe databases by the provider MUST be `child` objects under the `links` endpoint
+of the index meta-database.
+
+It is understood that the provider's OPTiMaDe databases MUST include the index meta-database as their
+single `parent` object under their respective `links` endpoints.
```
rartino: I thought we decided *not* to enforce only two layers as a technical requirement? @sauliusg argument was that we cannot really add that we allow that later without breaking backwards compatibility. 

Also, it is not very clear what the requirement "MUST be the 'root' OPTiMaDe implementation" exactly means. I think 'MUST'-type requirements should be things very clear how to validate.

optimade.md
```diff
@@ -530,16 +603,44 @@ The response dictionary MUST include the following fields
...
+
+If this is an index meta-database base URL (see [3.4. Index](#h.3.4)), then the response dictionary
+MUST also include the member:
+
+* **relationships**: a dictionary that MAY contain a single
+[JSON API relationships object](https://jsonapi.org/format/#document-resource-object-relationships):
+
+  * **default**: a reference to the `child` object under the `links` endpoint that the provider
+  has chosen as their "default" OPTiMaDe API database. A client SHOULD present this database as the
+  first choice when an end-user chooses this provider. A client MAY choose to only serve the default
```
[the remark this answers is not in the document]
CasperWA: Agreed. I will remove the sentence.

optimade.md
CasperWA: > Does all the strict mandating of who may include which objects have a purpose?

I will cut it down. I wanted only to make sure all providers in the list are included in a response if one chooses to return the `provider` objects. But maybe this does not belong in the specification, but more for the organizational aspects?

CasperWA: It will be changed to direct to appendix 1 instead for a list of providers.

optimade.md
CasperWA: Yeah, this part should really be part of the new repository that holds all information on organizational decisions.
But since it is still not present and other organizational things are also still within the current specifications I decided to include it here.
It should however definitely be removed from a final technical specification. Maybe it can also be removed here, since the recommendation is already in place previously?

rartino: Yes, I would for omitting that requirement from the specification, and live with that it is an "undocumented policy" to be included in the main OPTIMaDe registry for the time being. But really, since this PR discussion has dragged us in that direction, one of us should just draft up that policy document more or less directly after this PR is approved.

CasperWA: Completely agree with your last sentiment. If you wish to start making a draft, you are more than welcome :) I am quite willing to help you, maybe in a Google Doc? Or as a new PR.

CasperWA: > Also, it is not very clear what the requirement "MUST be the 'root' OPTiMaDe implementation" exactly means. I think 'MUST'-type requirements should be things very clear how to validate.

I am changing the whole section to the following:

If the provider implements an index meta-database (see section [3.4 Index](#h.3.4)),
it is RECOMMENDED to adopt a structure, where the index meta-database is the `parent`
implementation of the provider's other OPTiMaDe databases.

CasperWA:
I have added changes again. @rartino, please check section "4.5.4. Index links endpoint".
The addition of the excellent notes by @rartino concerning 404 errors have also been added and updated to reflect the current changes.

CasperWA:
> This now looks good to me.

Thanks @rartino :)

> It also would be very nice to get these changes approved finally, as I have a number of other edits I'd like to propose but which I have been holding off to submit to avoid conflicts with this rather major edit.

I agree, I also want to go through the document and correct spelling mistakes, do minor clarifications and unify terms as best as I can, and already have a branch working on this.

rartino:
I agree with @giovannipizzi and @gmrigna that it is time to go ahead with this, as there has not been further input except from me for weeks. The concerns in @sauliusg original review seems thoroughly discussed and addressed, so I will now click 'dismiss' on the outstanding "request for changes" that blocks the merge. But, as @giovannipizzi suggests, it will be fine to further adjust the multiple database feature with additional PRs before we release the present devel version into a non-devel one.

giovannipizzi:
Considering the comments above, I am going to merge this - further discussion can go in a different PR.

PR #66, opened 2019-03-25
Change top-level response fields to be more JSON API compliant

CasperWA:
### Overview
As suggested [here](https://github.com/Materials-Consortia/OPTiMaDe/pull/63#issuecomment-476124565), the common fields in an OPTiMaDe API response should be more JSON API-like, see [here](https://jsonapi.org/format/#document-top-level).

#### Changes
At least one of the fields MUST be included:
- `meta`
- `links`
- `data`

All of those fields SHOULD be included (instead of MUST).

If any errors: ONLY the field `errors` MUST be included.

#### Personal note
I think it is useful to *not* only return the `errors` field if errors occur, but this is the way JSON API specifies to do it.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -190,7 +190,7 @@ Not only in response format, but also in, e.g., how content negotiation is imple
 
 ### <a name="h.3.3.2">3.3.2. JSON API Response Schema: Common Fields</a>
 
-Every response MUST contain the following fields:
+Every response SHOULD contain the following fields, and MUST contain at least one:
 
 * **meta**: a [JSON API meta member](https://jsonapi.org/format/1.0/#document-meta)
   that contains JSON API meta objects of non-standard meta-information.  
@@ -328,7 +328,7 @@ related to the primary data contained in `data`.
 A response with related resources under `included` are in the JSON API known as
 [compound documents](https://jsonapi.org/format/1.0/#document-compound-documents).
 
-If there were errors in producing the response all other fields MAY be skipped, and the following field MUST be present
+If there were errors in producing the response all other fields MAY be present, but the top-level `data` field MUST be skipped, and the following field MUST be present:
 
 * **errors**: a list of [JSON API error objects](http://jsonapi.org/format/1.0/#error-objects).
 


rartino:
Requesting review of Fawzi, because we think you were the one originally adopting the specification to jsonapi, and thus likely are the one who put in the extra requirement of having these fields. Fawzi, is there a good reason not to relax this requirement back to what is required by jsonapi?

CasperWA:
Concerning my personal note:

> #### Personal note
> I think it is useful to _not_ only return the `errors` field if errors occur, but this is the way JSON API specifies to do it.

This PR does not reflect this, but I think it should. If you agree I will update the PR with this sentiment instead, unless we want to follow the JSON API very strictly in this regard?

merkys:
> This PR does not reflect this, but I think it should. If you agree I will update the PR with this sentiment instead, unless we want to follow the JSON API very strictly in this regard?

@CasperWA what do you mean? I assume that there is no data to return in case of errors. Otherwise if there is data to return, the situation is most likely not a critical error.

CasperWA:
> (...) Otherwise if there is data to return, the situation is most likely not a critical error.

This is maybe hitting the nail on the head - I believe there are cases where we would like to return an error, but where it is not a critical error, and so it would be nice to have what results back that can be returned, in order to trace the cause of the error?
On the other hand, we can simply decide that if an error occurs, it is _always_ critical. Therefore we do not expect (or want) any results, we simply want the error returned, which should be enough to deduce how to fix it.

merkys:
> I believe there are cases where we would like to return an error, but where it is not a critical error, and so it would be nice to have what results back that can be returned, in order to trace the cause of the error?

Can we think of a such recoverable error? Server-side errors are most likely fatal, as are ones related to improper user-supplied queries. We might have deprecation warnings in future, but I feel that such messages belong in `meta`.

CasperWA:
In accordance with the discussion here and offline, the issue #80 has been put forth.
In short, we suggest to distinguish between critical and non-critical errors, keeping only critical errors in the `errors` field, returning ONLY this if an error occurs, in accordance with JSON API, and having non-critical errors in a separate field `warnings` under `meta`.

This leaves this PR in the "correct" state as is.

rartino:
The distinction between critical errors and non-critical errors is that the former return an HTTP status code that is not 200, right?

But you now suggest placing non-critical errors under the warnings key. IF we have anything that truly falls in the category of non-critical error, that seems somewhat jsonapi-non-conformant. It may mean that if one uses a jsonapi-helper-library it wouldn't pick them up as errors. 

But I'm indeed also struggling to come up with something that is a non-critical error... Maybe: "Only partial data returned, because I lost connection to the backend after getting some results..."

CasperWA:
> The distinction between critical errors and non-critical errors is that the former return an HTTP status code that is not 200, right?

Sure. However, why couldn't you have a "successful" request, where the return code is not 200? E.g., the 1xx or other 2xx status codes?
In any case, this particular discussion has moved to issue #80 and PR #85.

> But you now suggest placing non-critical errors under the warnings key. IF we have anything that truly falls in the category of non-critical error, that seems somewhat jsonapi-non-conformant. It may mean that if one uses a jsonapi-helper-library it wouldn't pick them up as errors.
> 
> But I'm indeed also struggling to come up with something that is a non-critical error... Maybe: "Only partial data returned, because I lost connection to the backend after getting some results..."

Find use cases in issue #80, and please refer all discussion there (and PR #85, if necessary).

Issue #42, opened 2018-06-15
Extend query (filter) language to include substring comparison operators


sauliusg:
Let's add the following operators to the filter language:

- string_property LIKE "value"  # as in SQL 'select string_property from data where string_property like "%value%"'

- string_properties STARTS WITH "value" # as in SQL 'select string_property from data where string_property like "value%"'

- string_properties ENDS WITH "value" # as in SQL 'select string_property from data where string_property like "%value"'

- string_properties UNLIKE "value" # as in SQL 'select string_property from data where string_property NOT like "%value%"'

dwinston:
I agree to extend the query filter language to include the capability for substring comparison. I propose an alternative to the above (or perhaps including the above as syntactic sugar): we can add a single operator, REGEX, to the filter language. The above four cases would map to the following equivalents:

* string_property REGEX "value"
* string_property REGEX "^value"
* string_property REGEX "value$"
* NOT string_property REGEX "value"

Furthermore, filters more powerful than substring comparison are enabled by REGEX. I propose the value for this operator be interpreted by the server as a Perl compatible regular expression (i.e. “PCRE” ) version 8.39 with UTF-8 support.

dwinston:
I suggest an optional attribute "operator_notes" returned by the base URL info endpoint. This attribute is a dictionary, with keys being operators and values being short notes of interest to a client. For the REGEX operator, if an OPTIMaDe implementation uses PCRE 8.39 with UTF-8 support, they MAY provide the field. However, if an implementation interprets a regex differently, it MUST include "operator_notes.regex"  and a corresponding value understandable to a human.

I don't see another way around having a default spec for what a regex is (in this case, that supported by MongoDB 3.2+), and, if we insist on implementors being able to deviate from that and still claim to support the operator of the same name, to provide metadata in /info to inform a client implementation.

rartino:
What you suggest seems problematic from a user perspective. Someone who wants to send a regex-type query to many databases will now need to manually deal with these differences in support. They would have to go through all target databases to check compatibility, and possibly manually translate between different regex formats. A user that isn't careful will easily end up with a mix of data resulting from different interpretations of their regex.

Individual databases can already support an extended filtering language on a query parameter like `_exmpl_filter=....`. Maybe it is reasonable to stay with the simpler substring operators in the standard filter language (which hopefully can be supported as specified in any reasonable backend), and defer full support for REGEX to database-specific extended filtering?

[...]

PR #69, opened 2019-05-13
Updates to filtering language

### LINKED ISSUE #16: Filter language extensions
It is arguably desirable to be able to write filter implementations that can translate between an optimade filter string and most backend languages without having to know the specifics for each and every property we define. This requires lexically, or at least syntaxtically, distinguishable datatypes so the filter implementation can deduce sufficient information from the filter string itself.

In the OPTIMaDe v0.9 syntax this has the implicit meaning of returning only structures with both "Si" AND "Ge" and other elements:
```
filter=elements="Si,Ge"
```
A filter language implementation that does not know how we defined the elements property cannot tell this apart from a string equality. It won't know what to do with this. 

Also, we'd like to investigate extensions for searching for correlated 'list-form' data, e.g., element_ratios that tie elements with ratios.

### LINKED ISSUE #17: Handle unset optional properties

### LINKED ISSUE #19: Adopt a length predicate to the filtering language
The suggestion is to adopt a length predicate to the filtering language
```
LENGTH set_property < 5
```

### LINKED ISSUE #58: Some suggestions for the filter grammar
First, a big thank you to (I think primarily) @sauliusg for the work with the formal grammar for the filtering language. Now, that I have gone through actually implementing these filters based on the grammar, I have a few thoughts.

**I realize it may at first seem a quite major thing to propose changes to the filtering grammar at this point. However, I stress that none of the changes proposed below actually changes any behavior of presently working API implementations. These aren't changes of how API implementations should interprete filtering strings, they are changes in how to best define the present behavior in a way that is as unambiguous, consistent and as useful for implentors as possible.**

### The 'filter=' keyword
I propose 'filter=' should not formally be part of the filtering language grammar at all. To me 'filter=' it is the delivery mechanism of the filter in the url query which is outside the filtering language itself. My two primary motivations:

- I now want to refer to 'OPTIMaDe filter strings' in various contexts, not just as delivered in the query API. It seems awkward to keep 'filter=' as a prefix in those contexts when it has no function there, and it seems equally awkward to talk about, e.g., "the standard OPTIMaDe grammar but starting from the <Expression> node". 

- It introduces an obstructive keyword that may be in the way of relevant queries. It is fairly easy to understand that you cannot name your Identifiers AND, OR, or NOT, but 'filter=' will down the line surely get in the way for someone's attempts to query on a field named 'filter' with potentially weird error messages (since if one in an Expression has "filter='test" it will be tokenized as (Keyword:'filter=', Identifier:'test') instead of ('Identifier:'filter', Operator:'=', Identifier:'test') as expected.)
  
Hence, my proposal is to change the top of the grammar tree to:
```
OPTIMaDeFilter = Expression ; 
```
and completely remove `<Filter>` and `<Keyword>` from the grammar.

### Spaces

Is this way of handling whitespace with [Spaces] a good idea? Is any other 'standard' published EBNF using anything resembling this? It seems rare when looking around. The wikipedia article says "Whitespaces and comments are typically ignored in EBNF grammars". 

Isn't rather the common way to deal with this to defer whitespace to be handled by the tokenizer? (unless whitespace really is an integral part of the syntax of a language). I believe we could just say something in the specification along the line of: `Except for strings, tokens do not span whitespace. All other whitespace (space, tab and newline) should be discarded during tokenizing.` As the specification presently stands, in my implementation that uses a lexer that handles whitespace "in the normal way", I'm very tempted to fetch the official grammar and then just do: `grammar = grammar.replace("[Spaces]", "")`. Is that what we want implementors to do?

### The non-standard definition of UnicodeHighChar

I think it is not such a good idea to insert a Grammatica-specific syntax in the middle of the otherwise standard EBNF. IF we could create a completely resolved standard EBNF I would be all for including that in the specification, but I do not see how it can be done with the choice of allowing arbitrary unicode in strings. 

That means we need to go to some "non-standard-EBNF" of defining the `<String>` token anyway. And since that must be done, I suggest splitting the present EBNF into two machine-separable parts. One as the formal standard EBNF grammar of non-terminals. The other would define all the suggested tokens to use in the lexer in a format useful for implementors; I propose POSIX-Extended Regular Expressions. 

Below follows the POSIX-Extended Regular Expressions token definitions I presently use in my implementation, which I believe are equivalent with the present specification. I suggest we incorporate in the specification:
```
  Identifier: [a-zA-Z_][a-zA-Z_0-9]*
  String: "([^\"]|\\.)*"
  Number: (\+|-)?([0-9]+(\.[0-9]*)?|\.[0-9]+)((e|E)(\+|-)?[0-9]+)?
```
(Note that due to differences in escaping backslash in regex flavors the `String` one is edited from what I use in Python, the one above should be right for Posix ERE.) These definitions then technically obsoletes all of `<UnicodeHighChar>`, `<EscapedChar>`, `<UnescapedChar>`, `<Punctuator>`, `<Exponent>`, `<Sign>`, `<Digits>`, `<Digit>`, which would be removed from the formal standard EBNF grammar part of the specfication. 

But I'm certainly not opposed to ALSO include a 'grammatica' definition of the tokens, which would be the present EBNF-like version of those definitions with the grammatica extension. 

**EDIT 2018-03-22: (To keep everything up here, I've appended another issue)**

### Allowing `value=value`, `value=identifier`, and `identifier=identifier`

Arguably, the most commonly expected construct in, what is meant to be, a somewhat straightforward filtering language is on the form `identifier <operator> value`. But, the grammar explicitly allows also the following constructs: `value <operator> value`, `value <operator> identifier`, and `identifier <operator> identifier`. As I am trying to implement the handling of these, I get into some difficulties because OPTIMaDe doesn't (yet) properly define types for its fields. (I brought that up on the last CECAM meeting, but I couldn't find an issue filed for it; I need to look more or and if it is not there, also file it an issue.)

* `value=identifier`: from the technical standpoint, this one is trivial, I've included it here only because one can question if there is a need to allow it, or if the querying language would be simpler by disallowing it. 

* `identifier=identifier`: as the specification presently stands, what is the formally correct way of handling such a comparison if the identifiers are not the same? E.g., `chemical_formula=prototype_formula` or `nelements > _exmp_other_numerical_field`. Note that presently the OPTIMaDe type model essentially makes every property is its own type, which defines its own semantics for comparisons. E.g., elements are equal regardless of order and equal even if they contain subsets of elements (which really seems an abuse of the equal operator when there exists an `>=`...). 

  However, I suspect that the correct handling here is to simply reject any comparison of two different identifiers unless it is clear in the specification that they have the same semantics (e.g., for integers with no comments about non-standard comparisons...) But, we are not so clear on that in the spec presently. 

But, the one that truly baffles me how to implement correctly is this one:

* `value <operator> value`: Since we don't have a type model where we can unambiguously detect the type from the expression of the value, I don't see how I can derive the semantics for this comparison. If I see "Al, Ga" = "Ga": is that an "element"-type comparison? Or a string comparison? How can I know?

So: in summary: going forward we **absolutely** need to think about the typing system for OPTIMaDe. In my opinion we need a type system where types (including the semantics for comparisons) are clearly derivable from the value expression. Then one can confidently either carry out a comparison, or throw a type error if the types do not match. With that, `value <operator> value`, `identifier <operator> identifier` becomes well-defined. This means that if we want to keep the particular comparison semantics for, e.g., the elements property, we need to define that as a "set"-type and give it a form of expression that is recognizable preferably down on the token level.

Until we have sorted that out, would it be better to disallow all other forms than `identifier <operator> value`?

rartino:
This is an attempt to address multiple updates to the filter language discussed across numerous issues and PRs. Since all filter language changes affect more or less the same part of the text, I did not see how to do separate commits and still retain a coherent flow. But, we should discuss the points being addressed individually and we can surely back-track the text on specific features.

Outline of changes:
1. Remove the required start of filter with 'filter=' from the grammar, fixes #58
2. Keywords and constructs to handle unset properties; IS KNOWN, IS UNKNOWN fixes #17, also see #47 
3. Keywords and constructs for fuzzy string filtering, CONTAINS, STARTS, ENDS partially addresses #42
  - Note: In contrast to the suggestion in #42 LIKE was changed to CONTAINS to avoid reusing the SQL keyword with a similar but different definition. HAS cannot be used as there would be an actual conflict in interpretation with set operators. 
  - Also the additional WITH keyword was left out as it was deemed unnecessary both in terms of linguistics, and to avoid unnecessarily complicate the grammar.
  - ~~We might want to add LIKE one day with wildcard support like the SQL version.~~
4. Handle set-type filtering over multi-valued properties with HAS, HAS ALL, HAS EXACTLY, HAS ANY, HAS ONLY, and LENGTH. fixes #16, fixes #19.
  - ~~Note: in contrast to the original suggestions, HAS and HAS ALL are completely synonymous in this PR (both allow lists, originally HAS only allowed a single value). This is to avoid a complication in the grammar otherwise.~~ **Edit:** now HAS only allows a single value. 
  - As per our recent discussion: the definitions of `elements`, `chemical_formula` and `prototype_formula` have been updated accordingly for `elements` to be a multi-valued string, and for `chemical_formula` and `prototype_formula` to be normal strings with normal string semantics so we don't have to keep track of separate semantics for all properties we define.  
5. Handle set-type zip-filtering over combinations of multi-valued properties as an optional feature #16 
6. Weaken the formulation about how to handle type mismatches in comparisons.  (The current formulation is very strict and appears to be a major obstacle for 'dumb' translator-type implementations.) 
7. ~~Force all comparisons to be of type 'identifier operator value', rather than 'value operator value', etc.~~ **Edit:** now force left hand side to always be identifier, implementations are allowed to OPTIONALLY allow the right hand side to be identifier. #58
8. Remove ability to filter on ordering of strings, which would be very problematic to implement if string rewrites take place in the OPTIMaDe layer. 
  - Note: this removes the ability to filter on 'id > ....' which may interfere with people's plans for pagination. However, pagination is now done with the 'links->next' member, with no specifics on what provider-specific syntax can go into that URL.
9. More standard definition of UnicodeHighChar #58
10. Minor renaming of some levels of Expression to more clearly mark them as such (ExpressionClause, ExpressionPhrase), rather than having an AndExpression that may not contain any AND, and using Term for something that isn't a term in the math sense (we might need that later...).
11. Mini-version of grammar no longer included in the text (all is in the appendix), since even the relevant part is so much longer now, and we now talk a lot about the features anyway.

I've updated and run the main repository tests, and run my own tests, and find no issues (e.g., reduce/reduce conflicts) with this grammar. ~~But, I don't quite know how the checked-in test system works so it would be nice if someone can see if this grammar update can be validated with that system as well.~~

**Edit: I had forgotten to add this to the list of changes:**
12. The handling of Spaces in the grammar has been removed and replaced with a sentence saying that: "All whitespace (space, tab, and newline) between tokens should be discarded."


rartino:
Maybe someone with insight into the grammar testing framework can help to look into why the automated travis test fails? After I made a few adjustments (e.g., apparently grammatica does not want comments inside rules) there are no more error messages, but the tests seem to stall on:
```
$ make tests
tests/cases/Filter_001.inp: 
```

rartino:
Ok, I figured out how most of the grammar test framework works and updated the grammar to be compatible with (as far as I understand) limitations in the LL-type parsing supported by grammatica. This required *major* changes to the grammar. I've also updated the tests to match the output. The travis check seems to run correctly now. 

However, I should go back and re-test this grammar with my own tools now, which I have not completed yet.

giovannipizzi:
My comments:
1. ok for me
2. ok for me
3. I would leave `WITH`: `name STARTS "Si"` sounds weird to me, `name STARTS WITH "Si"` is better. I don't have any problem to call the operator `STARTSWITH` without spaces if this simplifies the grammare
4. ok
5. I need to check how the syntax looks like now
6. what is the current behaviour? Is `A = "16"` returning false if `A` is an integer? (I would be ok with this)
7. ok
8. ok to drop the possibility to sort strings (or better, keep it optional) but I would make this mandatory instead for ID, at least
9. this is a bit technical for me, I need a bit more time to understand - I'm ok to do it if you think it's the right thing to do
10. ok
11. ok

Thanks a lot for the great work!

rartino:
Some notes from the Skype meeting:

* We had a big discussion in general about whether to allow optional features at all in the grammar. But, I think towards the end of the meeting we came to a consensus that the way it is done in the PR now, with a core set of features + optional, but thoroughly standardised, extensions is the best we can do. (Anyone who disagree, please comment.)

* Regarding 8: We absolutely want to require string inequalities for IDs. We also concluded that a database backend that 1) doesn't store strings in the way they are standardised in OPTIMaDe AND 2) doesn't have a rewrite function that can make inequality tests against a rewritten string CANNOT correctly support even ENDS, BEGINS and CONTAINS. So, to not loose too much power from the core set of features, we are ok if the specification leads to one of those being a de facto requirement. And, in that case, one can use the same features to support inequality operators for strings. So, maybe let us make those mandatory. HOWEVER, we are opening a can of worms there with different ways of ordering strings, so, I don't know where we stand on that. **No change required to the grammar, I will update the PR with a text change describing all string comparison support as mandatory.**

* There is no equality operator for lists since HAS EXACTLY tests for set equality, i.e., the elements may be in another order as long as they are all there. We could support '<identifier> = val1, val2', or, e.g., '<identifier> EQUALS val1, val2'. **Unless someone objects, I'll update the grammar to support EQUALS for this.** (Not using = is a simplification in the grammar.)

* There is some disagreement whether we really should disallow `value <op> value`. If combined with keeping the requirement/recommendation of type conversions in point 6 I'm still unsure how to evaluate "16" < 3. Is it: "16" < "3" = False, or 16 < 3 = True? @sauliusg suggested introducing different operators for string comparisons and number comparisons. E.g. "EQUALS". Should we do that? **I'll make no change for now, until we've concluded something.**

* About 3: As @fawzi comments: the best idea so far seems to let the grammar allow both STARTS and STARTS WITH, ENDS and ENDS WITH. This can be done in the grammar, it is not a big issue. **I'll update the grammar for this.**

* About 6: Right now *BEFORE this PR* the specification reads:
```
When comparisons are performed, comparison operators '<', '<=', '=', '!=' and so on are interpreted either as numeric value comparisons or as string comparisons, depending on the type of the search parameter. The literal string value MUST be converted to the type of the search parameter (e.g., 'x > "0.0"' where x is a coordinate MUST treat this expression as numeric filter 'x > 0', and 's = 0' search against text parameter 's' MUST perform string comparison as in 's = "0"'). Strings are converted to numbers using the token syntax specified in section '5.1. Lexical tokens', p. "Numeric values"; numbers SHOULD be converted to strings using the libc '%g' format. In all cases the application MUST return a warning when conversion was performed and specify the actual search values that were used.
```
This means that if A is specified in OPTIMaDE to be a property of type integer, then A = "16" MUST match entries where A = 16. The change I've made here is essentially to say that this is the RECOMMENDED behaviour, but implementations are allowed to throw an error.

rartino:
I've pushed a number of further changes to PR #69 based on our discussions last Friday:

* optional WITH keyword for fuzzy string comparisons 
* require inequalities with strings to be supported for all fields
* allow left hand side to also be a value, and allow value = value as an OPTIONAL feature
* allow inequalities in set-type filters
* add text about floating point comparisons
* update type conversion text to say what I think we agreed it should say:
  - Implementations MUST either throw an error or non-silently convert values. 
  - The type semantics used in comparisons of type `id <op> value` follows the type of `id`.
  - `Value = Value` comparisons MUST throw an error if there is an apparent type-mismatch.
* Add in text that OPTIONAL features must either work as intended or throw a 501 error.

There were some discussion about adding new features in new PRs instead of updating this one. However, editing the grammar (at least under the requirement that it remains LL-parsable by Grammatica) isn't a trivial thing, and one may easily need to rework large parts of the grammar to add what might even seem as minor feature on top of other features. 

These changes also alter the parse trees completely, so one need to rebuild the test case output, and more importantly, re-write parse code (e.g., as I continuously do for these changes in python-optimade-candidate-reference-implementation to test the proposed grammars.) Hence, it really makes sense to try to get every grammar change we think we need for OPTIMaDe 1.0 into one go, i.e., this PR. This is also why I try to be somewhat feature-complete with providing all OPTIONAL features we've discussed, because that means that we are sure we don't need to rework the grammar again to accommodate them.

Please take a look at this PR as soon as possible and let me know if there is anything you disagree with (and otherwise approve it). Regardless of approvals I'll wait until next Monday to merge it, in case someone comes back with feedback. It would great if we can get these changes through before the workshop so we have a good base for standardizing new fields.

Issue #12, opened 2018-06-13
conflicting definitions of content of "entries"


ltalirz:
There are currently two, conflicting definitions of the basic content of an "entry"
 * [Section 4.1.2](https://github.com/Materials-Consortia/OPTiMaDe/blob/v0.9.5/optimade.md#412-json-api-response-schema) describing the case of multiple entries
 * [Section 6.1](https://github.com/Materials-Consortia/OPTiMaDe/blob/v0.9.5/optimade.md#61-properties-used-by-multiple-entry-types) describing properties used by multiple entry types

First of all, having two definitions seems unnecessary, since [4.2.2](https://github.com/Materials-Consortia/OPTiMaDe/blob/v0.9.5/optimade.md#422-response-schema) states: 
> The response for a 'single entry' endpoint is the same as for 'entry listing' endpoint responses, except that the value of the "data" key MUST have only one entry.

Second, one needs to decide for one of the two, in particular
 * `local_id` vs `id`
 * ~~`last_modified` vs `modification_date`~~
 * `immutable_id` vs nothing

I'm not sure whether this decision has already been made at some point and just hasn't found its way into the spec yet, could someone please comment on that?

~~Let me just reiterate that `modification_date` is a standard field in JSON schema as already pointed out in #1~~
**Edit**: Issue #1 has been fixed in PR #77, keeping only `last_modified`.

rartino:
I'm sure the writing/organization of things can be improved and clarified, but I don't see a bug other than that `last_modified` in Sec 4.1.2 should be the jsonapi mandated `modification_date` as was reported in #1. 

Note that Section 6.1.1 doesn't say that the fields it lists are the ones (and the only ones) that should appear in any specific output. 6.1.1 is really just intended mostly as an 'appendix' that lists field names and gives more precise details on how they are defined, wherever they appear in the output. The requirements for what MUST and MAY be included in a response is meant to be in section 4. So, if you are wondering if you should be reading section 4 or 6 to know which fields needs to be in some specific output, the answer is section 4.

The three ID fields were an intentional result from the 2016 workshop, with the following intent:
- `id` is the jsonapi mandated id that sits at top level. This is an id that OPTIMaDe puts some requirements on in 6.1.1. Hence, it is possible that for some databases this cannot be their own preferred id.
- `local_id`, on the other hand, has no requirements whatsoever. Here databases get to communicate their own ids for each record. 
[...]

Issue #73, opened 2019-06-11
Uniqueness of ids


merkys:
We have to decide whether `id`s are unique in:
1. a namespace (structure with `id` of 1 and calculation with `id` of 1 coexist) or
2. a database (no two entries with the same `id` can coexist in a database, just like issues and PRs on GitHub).

I personally prefer option 1. For example, JSON API uses tuple of `type` and `id` to uniquely identify an entry in the database.

merkys:
Issue was brought up in a discussion of #21.

dwinston:
@merkys I believe [https://jsonapi.org/format/#document-resource-objects](https://jsonapi.org/format/#document-resource-objects) gives some guidance:

> Within a given API, each resource object’s type and id pair MUST identify a single, unique resource. (The set of URIs controlled by a server, or multiple servers acting as one, constitute an API.)

Thus (type, id) is unique in the database, but an id may appear in multiple entry types.

merkys:
Citing [4.1.2. JSON API Response Schema](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.md#412-json-api-response-schema):

> **id**: a string which together with the type uniquely identifies the object, this can be the local database ID

Thus I consider this issue resolved. However, it would be beneficial to describe this property in the term definition and refer to the term definition elsewhere in the text.

CasperWA:
Closed in PR #104.

PR #82, opened 2019-06-11
updates to id, local_id, and immutable_id definitions

coreyoses:
references #12

Please follow [the guidelines here](https://github.com/Materials-Consortia/OPTiMaDe/wiki/How-to-contribute#submit-proposed-changes) to create a new pull request.
If your pull request requires discussion, make sure the OPTIMaDe developer list is alerted (for more info see the link above.)

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -508,7 +508,7 @@ represent individual entries. In the JSON API format every dictionary
 needs the following fields:
 
 * **type**: field containing the type of the entry
-* **id**: a string that, together with the type, uniquely identifies the object and
+* **id**: a string which together with the type uniquely identifies the object and
 strictly follows the requirements/conventions as specified by [id](#h.6.1.1).
 This can be the local database ID.
 * **attributes**: a dictionary, containing key-value pairs representing the
--- a/optimade.md
+++ b/optimade.md
@@ -508,21 +508,25 @@ represent individual entries. In the JSON API format every dictionary
 needs the following fields:
 
 * **type**: field containing the type of the entry
-* **id**: a string which together with the type uniquely identifies the object,
-this can be the local database ID
+* **id**: a string that, together with the type, uniquely identifies the object and
+strictly follows the requirements/conventions as specified by [id](#h.6.1.1).
+This can be the local database ID.
 * **attributes**: a dictionary, containing key-value pairs representing the
   entry's properties and the following fields:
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -508,21 +508,25 @@ represent individual entries. In the JSON API format every dictionary
 needs the following fields:
 
 * **type**: field containing the type of the entry
-* **id**: a string which together with the type uniquely identifies the object,
-this can be the local database ID
+* **id**: a string that, together with the type, uniquely identifies the object and
```
CasperWA: The original definition here is lifted from (and is identical to) the JSON API definition of `id`.
However, I do like the reference to `id` from the list of entry properties.
Just to keep closer to the JSON API, I would prefer to keep the original sentence, adding a separate sentence afterwards that links to `id` in Entry List.

CasperWA: Especially considering issue #73 and [this](https://github.com/Materials-Consortia/OPTiMaDe/issues/73#issuecomment-501161075) comment.

coreyoses: Ok I reverted back, thanks for checking.

merkys: I suggest describing the requirements for `id` in 'Term Definitions' section and reference it everywhere else. This will maintain the Single Point Of Truth and prevent the specs bloat.

CasperWA: @merkys personally, I am a bit confused about `id` and ID. I believe ID is (and should be) the term definition, and `id`, being a field, is different. However, the value of `id` is an ID, hehe.
In the end, maybe it is better to overwrite the term defintion with `id`, but I don't know if it is the correct approach.
ID is used in other places as well during the spec.

merkys: But the content of `id` is the ID, or am I wrong here? So we describe ID in the 'Term Definition', and specify that `id`s value is ID elsewhere in the document.

CasperWA: Yup, exactly.
My guess is that the Term Definitions is for defining a nomenclature, i.e., more abstract things and giving them a context, while `id` is a specific thing, defined from the Term Definitions as a "field". Its value or content is then ID, another defined term.

merkys: I agree, but still the requirements for `id` arising from ID should stay with ID :)

Nevertheless, this PR can be accepted for now.

CasperWA: Agree :) With both statements.

PR #84, opened 2019-06-12
Clarifying requirements for the entry resource object

merkys:
1. Defining `id`, `type` and `attributes` as MUST-have.
2. Describing `type` as property used by multiple entry types.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1304,6 +1304,7 @@ This section defines standard entry types and their properties.
 ### <a name="h.6.1.2">6.1.2. type</a>
 
 * **Description**: the type of an entry.
+* **Requirements/Conventions**: MUST be an existing entry type.
 * **Example**: `structure`
 
 ### <a name="h.6.1.3">6.1.3. last\_modified</a>
--- a/optimade.md
+++ b/optimade.md
@@ -1306,7 +1306,12 @@ This section defines standard entry types and their properties.
 * **Example**: `"2007-04-05T14:30Z"`
 * **Querying**: Date-time queries are permitted ([RFC 3339](http://tools.ietf.org/html/rfc3339)).
 
-### <a name="h.6.1.3">6.1.3. database-provider-specific properties</a>
+### <a name="h.6.1.3">6.1.3. type</a>
+
+* **Description**: the type of an entry.
+* **Example**: `structure`
+
+### <a name="h.6.1.4">6.1.4. database-provider-specific properties</a>
 
 * **Description**: Database providers are allowed to insert database-provider-specific entries
   in the output of both standard entry types and database-provider-specific entry types.
--- a/optimade.md
+++ b/optimade.md
@@ -505,7 +505,7 @@ Examples:
 key. The value of this key MUST be a list containing dictionaries that
 represent individual entries. In the JSON API format every dictionary
 ([resource object](http://jsonapi.org/format/#document-resource-objects))
-needs the following fields:
+MUST have the following fields:
 
 * **type**: field containing the type of the entry
 * **id**: a string which together with the type uniquely identifies the object,
--- a/optimade.md
+++ b/optimade.md
@@ -40,8 +40,8 @@
 [6. Entry List](#h.6)  
 &nbsp;&nbsp;&nbsp;&nbsp;[6.1. Properties Used by Multiple Entry Types](#h.6.1)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.1.1. id](#h.6.1.1)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.1.2. last\_modified](#h.6.1.2)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.1.3. type](#h.6.1.3)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.1.2. type](#h.6.1.2)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.1.3. last\_modified](#h.6.1.3)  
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -1306,7 +1308,12 @@ This section defines standard entry types and their properties.
 * **Example**: `"2007-04-05T14:30Z"`
 * **Querying**: Date-time queries are permitted ([RFC 3339](http://tools.ietf.org/html/rfc3339)).
 
-### <a name="h.6.1.3">6.1.3. database-provider-specific properties</a>
+### <a name="h.6.1.3">6.1.3. type</a>
```
CasperWA: Since `type`, together with `id`, are MUSTs for the JSON API, I think this should be number 2 instead of 3 - a small thing, but it keeps the numbering equal to the examples in the spec.

optimade.md
```diff
@@ -1306,7 +1308,12 @@ This section defines standard entry types and their properties.
 * **Example**: `"2007-04-05T14:30Z"`
 * **Querying**: Date-time queries are permitted ([RFC 3339](http://tools.ietf.org/html/rfc3339)).
 
-### <a name="h.6.1.3">6.1.3. database-provider-specific properties</a>
+### <a name="h.6.1.3">6.1.3. type</a>
+
+* **Description**: the type of an entry.
+* **Example**: `structure`
```
CasperWA: There should be a "Requirements/Conventions" stating something like: "MUST be a valid object type". They will then have to scour the spec to find them all, but I think it's fine - for now. Later we can always add a non-binding list in the wiki or the README or as an appendix?

optimade.md
```diff
@@ -505,7 +507,7 @@ Examples:
 key. The value of this key MUST be a list containing dictionaries that
 represent individual entries. In the JSON API format every dictionary
 ([resource object](http://jsonapi.org/format/#document-resource-objects))
-needs the following fields:
+MUST have the following fields:
```
CasperWA: Very good. I have no problem that we extend the JSON API and make `attributes` a MUST for objects as well - it seems natural. However, I would like a third opinion on this (@rartino, @giovannipizzi, @ltalirz, @dwinston).

optimade.md
merkys: I am a bit reserved about renumbering the sections in the spec as they might be referred somewhere outside the text. But it's no big harm, will fix it.

optimade.md
merkys: Section 6. Entry List gives the list of supported entries, we may refer to it.

optimade.md
merkys: I will check what specs say about the `attributes` for other entrypoints.

optimade.md
CasperWA: I understand - but it should be a quick search and replace, since the sections numbers should be consistently written together with titles.

optimade.md
CasperWA: No completely - we have introduced other objects, `"child"`, `"parent"` and more, if I remember correctly.
I understand that these are not considered data entries, however, they are still essentially valid object entry types

optimade.md
CasperWA: Note: If updated, this should (possibly) be added to the term definition of a Resource Object. Otherwise, we may define an Entry Resource Object as something that needs `attributes`, and (at this point in time) can _only_ be of type `"structure"` or `"calculation"`, right?

optimade.md
CasperWA: Came up with another requirement from issue #73 and your [comment](https://github.com/Materials-Consortia/OPTiMaDe/issues/73#issuecomment-501161075): Both for `id` and `type`, a requirement should be that `id` and `type` should be unique together, i.e., this should be added to the requirements of `id` as well under 6.1.1.

merkys: Do we really want to repeat it once more? I'd suggest describe the uniqueness requirement in the term definition and remove it from everywhere else, just as in any legal document: requirements stated in terms are not repeated in order to avoid consistency problems and specs bloat.

optimade.md
merkys: As I understand, this section applies only to the dictionaries representing entries, not all Resource Objects. Otherwise this definition should be placed in Term Definitions or like.

optimade.md
CasperWA: Repetitions can surely be a problem, bloating the specs. But I see the entry list, and the different fields here, as a look-up for implementers. Therefore it seems natural to have the requirements and conventions repeated here for the specific fields - even though it is also stated in the definitions above.
Indeed, I would say it the other way around: If it is explicitly written in the entry list, then it can be a bit more "fluffy" elsewhere in the spec, since one may always reference the entry list as the ultimate truth for each field.

optimade.md
```diff
@@ -1301,18 +1301,19 @@ This section defines standard entry types and their properties.
   * `"nomad/L1234567890"`
   * `"42"`
 
-### <a name="h.6.1.2">6.1.2. last\_modified</a>
+### <a name="h.6.1.2">6.1.2. type</a>
+
+* **Description**: the type of an entry.
+* **Requirements/Conventions**: MUST be an existing entry type.
+* **Example**: `structure`
```
CasperWA: Probably add "s here to be consistent with the other examples.

optimade.md
CasperWA: True. Let's keep it as you have it here. In my opinion we ALWAYS want `attributes` to be present for entry resource objects - so no problem.

optimade.md
merkys: I'd like to eliminate the "s in `s altogether, but OK, this could be done later globally.

optimade.md
merkys: Well, I'd still prefer moving general stuff to a single place. Maybe let's wait for more opinions here, OK? :)

giovannipizzi:
Does this reference a specific issue? I'm working to sort the current work in [this project](https://github.com/Materials-Consortia/OPTiMaDe/projects/1)

CasperWA:
> Does this reference a specific issue? I'm working to sort the current work in [this project](https://github.com/Materials-Consortia/OPTiMaDe/projects/1)

I think perhaps issue #2? But I am not completely sure.

giovannipizzi:
@merkys could you fix the conflict? then we can merge

optimade.md
CasperWA: This is to be discussed and solved through issue #106.

giovannipizzi:
About uniqueness of IDs: see issue #73

PR #85, opened 2019-06-12
Introducing 'warnings' list in 'meta'

### LINKED ISSUE #80: Introduce "warning" messages under meta
Related to the discussion in PR #66 between @merkys and @CasperWA.

### Suggestions

- Top-level `errors` should be for _critical_ errors only. In accordance with JSON API, if an error occurs, ONLY the top-level `errors` should be returned with a [JSON Error Object](https://jsonapi.org/format/#error-objects) and the appropriate HTTP error code.
- The keyword `warnings` should be reserved under the top-level `meta`, to include errors or rather warning messages that are _non-critical_.

### Possible use cases

An example of a use case may be that a query keyword was misspelt from the client side (something that the client developer would definitely want to know and correct). This will according to the current spec result in the keyword to be ignored, and an otherwise normal response.
A solution to this would be for a warning message and that the client always checks for the presence of the `warnings` key under `meta` and relay these messages in a development environment (or even in production if needed).

Another example is for deprecation warnings to be expressed under this keyword.

merkys:
Fixes #80.


merkys:
> I am willing to try and write a PR / update this PR, with a suggestion of what I have in mind, if you don't want to do it :)

Sure, please do so! :) I like the Warning Resource Object idea.

CasperWA:
Hereby done. Please review :)

optimade.md
```diff
@@ -240,6 +240,22 @@ Every response MUST contain the following fields:
...
+  * **warnings**: a list of warning resource objects representing non-critical errors or warnings.  
+    A warning resource object is defined similarly to a [JSON API error object](http://jsonapi.org/format/#error-objects), but MUST instead be of `type` `"warning"`.
+    The field `detail` SHOULD contain a non-critical message, e.g., reporting unrecognised search attributes or deprecated features.
+
+    Example:  
+    For a deprecation warning
+
+    ```json
+    {
+      "id": "dep_chemical_formula",
```
merkys: It'd be probably good to put a note here that `id`s MUST NOT be trusted to identify the **exceptional situations** (i.e. they are not error codes). They can only be trusted to be unique in the `warnings` list.

CasperWA: True. It is essentially stated elsewhere, but it may be good to point out again, when using this example.

CasperWA: Done

optimade.md
```diff
@@ -240,6 +240,25 @@ Every response MUST contain the following fields:
...
+    ```json
+    {
+      "id": "dep_chemical_formula_01",
+      "type": "warning",
+      "title": "Deprecation Warning",
+      "detail": "chemical_formula is deprecated, use instead chemical_formula_hill"
+    }
+    ```
+
+    Note: `id`s MUST NOT be trusted to identify the exceptional situations (i.e., they are not error codes, use instead the field `code` for this).
```
ltalirz: So, just to understand: warnings should be able to have codes as well or not?
If yes, I would suggest to add a code to the example as well.

CasperWA: Will do.
As stated, warning resource objects are defined similarly as [JSON error objects](https://jsonapi.org/format/#error-objects), so they MAY contain all the fields that are listed (including `code` and even `status`).
We could consider making some of the fields mandatory for the warning resource objects, e.g., `detail`?

merkys: Let's make `detail` mandatory for both Error and Warning objects. JSON API specs seem to be too liberal here.

merkys: Concerning the `code`, is it provider-specific, or global? I feel that the former option is fine for now, as the latter one will open a can of worms...

CasperWA: I guess the `code` is simply the HTTP response code. As this is set by the server developer/provider in most cases, the former option is the "correct" option.

However, we try to conform to a global interpretation of what a specific code means (from section [3.3.3. HTTP Response Status Codes](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.md#h.3.3.3)).

merkys: > I guess the `code` is simply the HTTP response code. As this is set by the server developer/provider in most cases, the former option is the "correct" option.

Indeed, but this makes little sense for warnings. Can we omit the `code` for warnings then?

CasperWA: Ah no, sorry, `status` is the HTTP response code, not `code`. Sorry!
I think, yes, we can state that `status` is not a valid field for a warning resource object, but in the end, I think it is fine, since you can still have a valid response, which is mildly disturbed (the 2xx and 1xx responses). If you're a truly detailed-minded developer, you may even use these codes in your warnings, and/or setting them as the HTTP Response Code.

rartino: I think the intent (although not the letter) of jsonapi is that code is a number encoded as a string, just like status is. See examples here: https://jsonapi.org/examples/#error-objects-error-codes . Hence, I think our examples should stick to that. 

In similarity to how HTTP status work, I think one is meant to assign these warnings numerical codes that are easy to process by a client, and the title is supposed to always be the same when the same numerical code is used (but possibly localized, but we have no facility for that). Only the `details` field may vary between warnings that have the same warning code.

Do you agree with clarifying this in the text?

optimade.md
```diff
@@ -312,7 +334,9 @@ A response with related resources under `included` are in the JSON API known as
 
 If there were errors in producing the response all other fields MAY be skipped, and the following field MUST be present
 
-* **errors**: a list of [JSON API error objects](http://jsonapi.org/format/#error-objects).
+* **errors**: a list of [JSON API error objects](http://jsonapi.org/format/#error-objects).  
+An additional field `type` MUST be present (along with `id` and `detail`).
+The value of `type` MUST be `"error"`.
```
rartino: When errors and warnings appear under different keys, why mandate a `type` field? I don't mind it so much for warnings, but it does seem a bit odd to mandate it for what otherwise is standard jsonapi error objects.

optimade.md
merkys: I feel that OPTiMaDe-specific error/warning codes should be unified across all the implementations, and I don't think that we should get into defining these codes. Therefore I'd use HTTP error codes for Error objects, and do not use this field in Warnings at all.

optimade.md
CasperWA: Yeah, you're right. Let's keep error resource objects as is specified in the JSON API v1.0.

CasperWA: But let's still mandate `detail`, I think?

giovannipizzi: I agree

optimade.md
giovannipizzi: My suggestion: let's leave space here for Optimade-defined warning codes. What about requiring that any non-standard warning MUST start with the provider-specific prefix? (and for now we have zero standard warning codes, and we should have a section about "list of optimade warnings, saying that the list is empty).
So e.g. `"code": "_exmpl_dep_chemical_formula"`

optimade.md
CasperWA: Done

optimade.md
CasperWA: I like this.
While the example above clearly shows integer codes, I see no reason why it should have to be.
In the end, a provider may have internal error codes they want to use, so prepending the database-provider-specific prefix seems natural for these codes.
This leaves a natural space for OPTiMaDe to have general warning codes for its own purposes for different warnings as well. Ones which do not have to be defined at this moment, for this PR.

giovannipizzi: The only thing missing is to explicitly specify that the code MUST be starting with a provider-specific prefix, or be an OPTIMADE-defined warning; point to a section; and in that section say that there are for the time being no OPTIMADE-defined warnings.

giovannipizzi:
Can be merged after final comment is taken care of

optimade.md
merkys: > The only thing missing is to explicitly specify that the code MUST be starting with a provider-specific prefix, or be an OPTIMADE-defined warning; point to a section; and in that section say that there are for the time being no OPTIMADE-defined warnings.

Done.

rartino:
Conflicts need to be resolved.

PR #89, opened 2019-06-12
Describing 'meta' field 'implementation'.

### LINKED ISSUE #8: Include implementation details into meta
I propose including implementation details in top-level ``meta`` subobject. An example:
```JSON
{
   "meta":{
      "implementation":{
         "maintainer":{
            "email":"cod-bugs@ibt.lt"
         },
         "source_url":"svn://www.crystallography.net/cod/trunk/cod/cgi-bin/optimade.pl@205080",
         "title":"Crystallography Open Database",
         "version":"v0.9.5-alpha.1"
      }
}
```
Description:
* **maintainer**: JSON object. Attributes:
  * **email**: e-mail address to contact the webmaster of the site
* **source_url**: URL of the source of the program, which has handled the request
* **title**: title of the database
* **version**: version of the program, which has handled the request

All attributes may be defined as optional.

merkys:
Closes #8.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -245,7 +245,7 @@ Every response MUST contain the following fields:
     * **name**: name of the implementation.
     * **version**: version string of the current implementation.
     * **source_url**: URL of the implementation source, either downloadable archive
-      or version contol system.
+      or version control system.
     * **maintainer**: a dictionary providing details about the maintainer of the
       implementation, MUST contain a single field **email** with maintainer's email
       address.
--- a/optimade.md
+++ b/optimade.md
@@ -247,8 +247,8 @@ Every response MUST contain the following fields:
     * **source_url**: URL of the implementation source, either downloadable archive
       or version control system.
     * **maintainer**: a dictionary providing details about the maintainer of the
-      implementation, MUST contain a single field **email** with maintainer's email
-      address.
+      implementation, MUST contain the single field **email** with the maintainer's
+      email address.
   * Other OPTIONAL additional information _global to the query_ that is not specified
   in this document, MUST start with a database-provider-specific prefix as defined in
   [Appendix 1](#h.app1).
--- a/optimade.md
+++ b/optimade.md
@@ -241,7 +241,7 @@ Every response MUST contain the following fields:
   * **last\_id**: a string containing the last ID returned.
   * **response\_message**: response string from the server.
   * **implementation**: a dictionary describing the server implementation, containing
-    OPTIONAL fields:
+    the OPTIONAL fields:
     * **name**: name of the implementation.
     * **version**: version string of the current implementation.
     * **source_url**: URL of the implementation source, either downloadable archive
--- a/optimade.md
+++ b/optimade.md
@@ -240,6 +240,15 @@ Every response MUST contain the following fields:
     available in the database.
   * **last\_id**: a string containing the last ID returned.
   * **response\_message**: response string from the server.
+  * **implementation**: a dictionary describing the server implementation, containing
+    OPTIONAL fields:
+    * **name**: name of the implementation.
+    * **version**: version string of the current implementation.
+    * **source_url**: URL of the implementation source, either downloadable archive
+      or version contol system.
+    * **maintainer**: a dictionary providing details about the maintainer of the
+      implementation, MUST contain a single field **email** with maintainer's email
+      address.
   * Other OPTIONAL additional information _global to the query_ that is not specified
   in this document, MUST start with a database-provider-specific prefix as defined in
   [Appendix 1](#h.app1).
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -240,6 +240,15 @@ Every response MUST contain the following fields:
     available in the database.
   * **last\_id**: a string containing the last ID returned.
   * **response\_message**: response string from the server.
+  * **implementation**: a dictionary describing the server implementation, containing
+    OPTIONAL fields:
+    * **name**: name of the implementation.
+    * **version**: version string of the current implementation.
+    * **source_url**: URL of the implementation source, either downloadable archive
+      or version contol system.
```
CasperWA: control

optimade.md
```diff
@@ -240,6 +240,15 @@ Every response MUST contain the following fields:
     available in the database.
   * **last\_id**: a string containing the last ID returned.
   * **response\_message**: response string from the server.
+  * **implementation**: a dictionary describing the server implementation, containing
+    OPTIONAL fields:
```
CasperWA: (... containing) the (OPTIONAL fields ...)

optimade.md
```diff
@@ -240,6 +240,15 @@ Every response MUST contain the following fields:
...
   * **last\_id**: a string containing the last ID returned.
   * **response\_message**: response string from the server.
+  * **implementation**: a dictionary describing the server implementation, containing
+    OPTIONAL fields:
+    * **name**: name of the implementation.
+    * **version**: version string of the current implementation.
+    * **source_url**: URL of the implementation source, either downloadable archive
+      or version contol system.
+    * **maintainer**: a dictionary providing details about the maintainer of the
+      implementation, MUST contain a single field **email** with maintainer's email
```
CasperWA: (... MUST contain) at least the (single field ...)

CasperWA: Maybe write what other fields it can contain? `name`?

CasperWA: Otherwise, it should simply be "(... MUST contain) the (single field ...)", which is also fine. It leaves room for future additions, if wanted.

merkys: Fixed the wording as suggested. Let's leave the other properties of the maintainer for the future.

merkys:
> Thinking out loud: If by "implementation" the OPTiMaDe implementation is understood, which has nothing to do with the data repository that is served, then it makes sense to keep it under `meta`, since the base URL `info` endpoint should contain information about the data repository, and be potentially agnostic to how the OPTiMaDe implementation is done.

Well, `/info` endpoint is kind of *meta* by itself. And I believe the software serving the OPTiMaDe queries has to be quite strongly related to the data repository. So I am not sure what is the best way to organize the metadata here.

CasperWA:
> (...) So I am not sure what is the best way to organize the metadata here.

Yeah - I would like some more comments on this, @rartino, @sauliusg, @giovannipizzi, @fawzi, @ltalirz, etc.

merkys:
> The only thing I'm somewhat questioning here is why maintainer absolutely MUST contain an email, rather than SHOULD, when the whole field is optional. Maybe the maintainer really don't want to give out their email. But I suppose then you can just leave out maintainer, so... it's not a very important point.

It's just the way to say that if we have `maintainer` field, then it MUST contain `email`. Otherwise there is no need to include `maintainer` in the first place.

rartino:
> It's just the way to say that if we have `maintainer` field, then it MUST contain `email`. Otherwise there is no need to include `maintainer` in the first place.

This really is a minor point, but, the idea with having a json Object here, instead of just calling it `maintainer_email` must be that other things may go in the Object (right?) Then, why do we have to mandate that email is always present? E.g., right now I cannot do:
```
maintainer: {
  'name':'Rickard Armiento',
}
```
If that is what I want, I'll just have to not include the maintainer field at all.

merkys:
> This really is a minor point, but, the idea with having a json Object here, instead of just calling it `maintainer_email` must be that other things may go in the Object (right?) Then, why do we have to mandate that email is always present?

I get the point. The idea behind this is to have a standardized way to include the email address of the maintainer. We may change the wording to allow other (non-standard) properties in `maintainer`, is this OK?

CasperWA:
Maybe we should simply add the OPTIONAL field `name` now, then?

merkys:
> Maybe we should simply add the OPTIONAL field `name` now, then?

Sure, if we can't think of anything else. I just don't like compound names like `maintainer_email` in cases when it's clear that we will want to introduce more `maintainer_*` fields. On the other hand I have a voice in my head whispering "[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)" from time to time. So I prefer to leave `maintainer` with `email` only, as I feel that it might be expanded, but currently I need only `email`.

PR #90, opened 2019-06-12
Update HTTP Response Status Codes according to JSON API

### LINKED ISSUE #71: Consider rewriting/updating the HTTP Response Status Codes
Lifted from PR #68.

> (...) I also don't like those comments in this table anyway. People may think those are supposed to be included as part of the response message, and that really shouldn't be the case. (These messages are not meant to vary with code other than localization). Should we remove them and instead focus on being more clear in the specification on for what conditions which errors are supposed to be returned? Furthermore, I also think this table misses some error codes specified by the jsonapi specification...

_Originally posted by @rartino in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjM2NjMzNDQw/pull_request_reviews/more_threads_

CasperWA:
Fixes #71

Removed code 418 (which is an unassaigned code, used (in)famously for [I am a teapot](https://www.google.com/teapot)).
Removed all parentheses in the description of codes.

Referencing [IANA](http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) and [RFC7231](https://tools.ietf.org/html/rfc7231) for codes and their definitions.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -361,21 +361,27 @@ An example of a full response:
 }
 ```
 
-### 3.3.4. HTTP Response Status Codes
+### <a name="h.3.3.3">3.3.3. HTTP Response Status Codes</a>
 
-| Code | Message                                                               |
-|:----:|:--------------------------------------------------------------------- |
-| 200  | OK (Successful operation)                                             |
-| 400  | Bad request (e.g., mistyped URL)                                       |
-| 401  | User does not have permission                                         |
-| 403  | Forbidden (the request was understood but not authorized)             |
-| 404  | Not found (e.g., database not found)                                   |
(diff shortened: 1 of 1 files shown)


merkys:
A drastic suggestion: can't we just remove the table? These are vanilla HTTP codes, aren't they?

CasperWA:
> A drastic suggestion: can't we just remove the table? These are vanilla HTTP codes, aren't they?

Yes, they are.
With the added links, I believe it should be fine as well.
Mainly, this will also highlight the "note" better (hopefully), since that will be the main content then of this section. And the note is essentially important, since it is our description of what should happen if a certain response code is received.

merkys:
Great, thanks @CasperWA!

Issue #51, opened 2018-11-22
Property to indicate which optional structure features are used


rartino:
Across discussions on various aspects of the structure format, i.e., #23 resolved by PR #49 it has been discussed that by introducing optional features in separate properties, we open for the risk of naive clients that miss the use of some optional features of the format. 

In the format we ended up accepting in PR #49, the most clear example is the `assemblies` property. However, arguably also the use of disorder in structures can give similar issues. In #50 we have started discuss a separate optional property to indicate what is 'unknown' in the structural information (e.g., unknown coordinates), etc. One could make the argument here that a client is already able to filter for no optional properties by, e.g., querying for `assemblies IS NOT KNOWN` (a suggested syntax in #17). But if OPTIMaDe in later versions adds further optional properties, e.g., `magnetic_moments`, code that rely on this will break, because it will miss to filter out structures that uses this feature. 

Hence, if we continue along this direction it would be beneficial to have a property that specifies precisely what optional structure features are present for a structure. This would make it easier for clients to only query for and operate on structures they are prepared to handle (e.g., no assemblies, no unknown coordinates.)

Some options include a format like this:
```
structure_features: {
  'simple': False
  'disordered': True,
  'assembly': False,
  'unknown_positions': True
} 
```
Or this:
```
structure_features: ["disordered", "unknown_positions"]
```
Personally, I prefer the latter. Together with the syntax in #16 it will be easy to formulate sensible queries to only operate on structures with the features that one wants. E.g.,
```
filter=LENGTH structure_features=0 
```
Means to only return structures using no optional features.
```
filter=structure_features HAS ONLY "disordered" 
```
for simple structures and structures with disorder, but no other optional features.

giovannipizzi:
I think it's a very good idea.
Another comment, though (is this already in the specs?):
all keys NOT starting with an underscore should be understood (while the custom ones with underscore, if not understood, can be ignored), and the client should "stop" or complain if it finds something it does not understand. This could be already a good method to have a safe client (and maybe this client behaviour should be specified in the specs?)

Anyway, if you think this is not enough, I like the approach `structure_features: ["disordered", "unknown_positions"]` (better than the dict, not only because of queries, but also because a client can easily have a list of "understood" features, and tell if there are new that it cannot understand.  Instead with the dict there could be a new feature but set to False and the client might not be 100% sure if "False" means that it's safe to ignore it).

rartino:
[...]

PR #93, opened 2019-06-12
Adding support for unknown coordinates

### LINKED ISSUE #50: Allowing for unknown/null coordinates
As discussed in #49, we split part of the discussion of #23 to this new issue.

From the comment of @rartino in #23:

> There are differing opinions on how to handle systems with atoms at unknown coordinates. How to resolve this is deferred into a separate issue. Two alternatives is to allow null in the coordinate lists, OR put an extra optional property that alongside the known coordinates describe the information that exists about the atoms at unknown positions.
> - The separate x,y,z coordinates are allowed to be given as 'null' if unknown. E.g., for disordered systems where coordinates are not fixed, and certain situations where light atoms can be near heavier atoms without knowing exactly where. However, we note that the latter situation can alternatively be represented by a species ~ "CH4", which can also be used in an assembly with "C" to indicate that it is not clear which one of the C atoms has the four H. (If it is known that all 4 always sits on on C...)
> - There is an extra required property `all_coordinates_known` that is true or false depending on whether any coordinate is a null value. (So clients can easily use filter to get only structures where all coordinates are known.)

And from @sauliusg :
> Actually, there seems to be the third way to represent atoms with unknown positions: specify no site for them (so no 'null' coordinates), and have a species that is on *no* site.

giovannipizzi:
This fixes #50 and #93 

This also adds the `all_coordinates_known` property as discussed
in #50.

I did also some section renumbering because I felt that
it is better to keep `all_coordinates_known` close to the
`cartesian_site_positions` section.

I also added a note for implementers for `NaN`s vs `null`s.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1467,8 +1467,8 @@ an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual cry
     e.g., the [6.2.10.`assemblies`](#h.6.2.10) property).
   * If a component of the position is unknown, the `null` value should be provided instead. 
     Otherwise, it should be a float value, expressed in angstrom. Note that if at least one
-    of the coordinates is unknown, the correct flag in the 
-    [`structure_features`](#h.6.2.11) list MUST be set.
+    of the coordinates is unknown, the correct flag MUST be set
+    in the list `structure_features` (see section [6.2.11 `structure_features`](#h.6.2.11)).
 * **Notes**: (for implementers) While this is unrelated to this OPTiMaDe specification:
   if you decide to store internally the `cartesian_site_positions` as a float array,
   you might want to replace `null` values with `NaN` values, the latter being valid float numbers
@@ -1509,14 +1509,15 @@ by multiple chemical elements.
   * This property is REQUIRED.
   * It MUST be a dictionary, where keys represent the species' name, and values are themselves
   dictionaries with the following keys:
-    * **chemical\_symbols**: REQUIRED; MUST be a list of strings of all chemical elements composing this species. 
-      *  It MUST be one of the following:
+    * **chemical\_symbols**: REQUIRED; MUST be a list of strings of all chemical elements composing this species.
+      * It MUST be one of the following:
         * a valid chemical-element name, or
         * the special value `"X"` to represent a non-chemical element, or
         * the special value `"vacancy"` to represent that this site has a non-zero probability of having
         a vacancy (the respective probability is indicated in the `concentration` list, see below).
       * If any one entry in the `species` list has a `chemical_symbols` list that 
-        is longer than 1 element, the correct flag in the [`structure_features`](#h.6.2.11) list MUST be set.
+        is longer than 1 element, the correct flag MUST be set
+        in the list `structure_features` (see section [6.2.11 `structure_features`](#h.6.2.11)).
   
 
     * **concentration**: REQUIRED; MUST be a list of floats, with same length as `chemical_symbols`.
@@ -1567,8 +1568,8 @@ by multiple chemical elements.
 * **Description**: A description of groups of sites that are statistically correlated.
 * **Requirements/Conventions**:
   * This key is OPTIONAL (it is absent if there are no partial occupancies).
-  * If present, the correct flag in the 
-    [`structure_features`](#h.6.2.11) list MUST be set.
+  * If present, the correct flag MUST be set
+    in the list `structure_features` (see section [6.2.11 `structure_features`](#h.6.2.11)).
   * Client implementations MUST check its presence (as its presence changes the
     interpretation of the structure).
   * If present, it MUST be a list of dictionaries, each of which represents an assembly and MUST have
@@ -1676,7 +1677,7 @@ by multiple chemical elements.
 * **Requirements/Conventions**: 
   * This property is REQUIRED.
   * This property MUST be returned as an empty list if no special features are used.
-  * This list MUST be sorted.  
+  * This list MUST be sorted alphabetically.  
   * If a special feature listed below is used, the corresponding string MUST be set.
   * If a special feature listed below is not used, the corresponding string MUST NOT be set. 
 * **List of special structure features**:
--- a/optimade.md
+++ b/optimade.md
@@ -49,9 +49,10 @@
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.5. dimension\_types](#h.6.2.5)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.6. lattice\_vectors](#h.6.2.6)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.7. cartesian\_site\_positions](#h.6.2.7)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.8. species\_at\_sites](#h.6.2.8)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.9. species](#h.6.2.9)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.10. assemblies](#h.6.2.10)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.8. all\_coordinates\_known](#h.6.2.8)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.9. species\_at\_sites](#h.6.2.9)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.10. species](#h.6.2.10)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.11. assemblies](#h.6.2.11)  
 &nbsp;&nbsp;&nbsp;&nbsp;[6.3. Calculation Entries](#h.6.3)  
 
 [Appendix 1: Database-Provider-Specific Namespace Prefixes](#h.app1)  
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -48,10 +48,11 @@
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.4. formula\_prototype](#h.6.2.4)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.5. dimension\_types](#h.6.2.5)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.6. lattice\_vectors](#h.6.2.6)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.7. cartesian\_site\_positions](#h.6.2.7)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.8. species\_at\_sites](#h.6.2.8)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.9. species](#h.6.2.9)  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.10. assemblies](#h.6.2.10)  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.7. cartesian\_site\_positions](#h.6.2.7)
```
CasperWA: This line shouldn't have changed - check that you are keeping the end-of-line spaces.

optimade.md
```diff
@@ -1414,21 +1415,34 @@ an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual cry
   * This property is REQUIRED.
   * It MUST be a list of length N times 3, where N is the number of sites in the structure.
   * An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see
-  e.g., the [6.2.10.`assemblies`](#h.6.2.10) property).
+    e.g., the [6.2.11.`assemblies`](#h.6.2.11) property).
+  * If a component of the position is unknown, the `null` value should be provided instead. 
+    Otherwise, it should be a float value, expressed in angstrom. (See also the property `all_coordinates_known`.)
```
CasperWA: I think at some point I found in the official list of SI units (where Å is not actually an SI unit) they wrote it as "Ångström" :) But I guess it is fine.

optimade.md
```diff
@@ -1414,21 +1415,34 @@ an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual cry
   * This property is REQUIRED.
   * It MUST be a list of length N times 3, where N is the number of sites in the structure.
   * An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see
-  e.g., the [6.2.10.`assemblies`](#h.6.2.10) property).
+    e.g., the [6.2.11.`assemblies`](#h.6.2.11) property).
+  * If a component of the position is unknown, the `null` value should be provided instead. 
+    Otherwise, it should be a float value, expressed in angstrom. (See also the property `all_coordinates_known`.)
+* **Notes**: (for implementers) While this is unrelated to this OPTiMaDe specification:
+  if you decide to store internally the `cartesian_site_positions` as a float array,
+  you might want to replace `null` values with `NaN` values, that are valid float numbers
```
CasperWA: no comma before "that"

optimade.md
```diff
@@ -1414,21 +1415,34 @@ an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual cry
...
   * It MUST be a list of length N times 3, where N is the number of sites in the structure.
   * An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see
-  e.g., the [6.2.10.`assemblies`](#h.6.2.10) property).
+    e.g., the [6.2.11.`assemblies`](#h.6.2.11) property).
+  * If a component of the position is unknown, the `null` value should be provided instead. 
+    Otherwise, it should be a float value, expressed in angstrom. (See also the property `all_coordinates_known`.)
+* **Notes**: (for implementers) While this is unrelated to this OPTiMaDe specification:
+  if you decide to store internally the `cartesian_site_positions` as a float array,
+  you might want to replace `null` values with `NaN` values, that are valid float numbers
+  in the IEEE 754 standard (possibly with an appropriate payload).
```
CasperWA: Do you have a link you could add here? This would be consistent with the rest of the spec, where we always have direct links when mentioning references.

optimade.md
giovannipizzi: It's up to discussion - but often units are written without special characters and lower-case, so I'd keep it as is

CasperWA: No problem - just because noone has the proper keyboard... :)

optimade.md
giovannipizzi: I didn't know about the fact that two trailing spaces have a meaning in markdown!!

optimade.md
giovannipizzi: Rephrased

optimade.md
giovannipizzi: Done - I reference both the 1985 standard (as probably most hardware implemented that) and the most recente 2008 standard (that anyway didn't change the part about `NaN`s).

giovannipizzi:
@CasperWA fixed your suggestions, thanks for the review

giovannipizzi:
WARNING: Note that both this and #100 change the numbering of some sections so there will be conflicts

optimade.md
```diff
@@ -1414,21 +1415,35 @@ an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual cry
...
+  if you decide to store internally the `cartesian_site_positions` as a float array,
+  you might want to replace `null` values with `NaN` values, the latter being valid float numbers
+  in the IEEE 754 standard in [IEEE 754-1985](https://doi.org/10.1109/IEEESTD.1985.82928) and in the updated
+  version [IEEE 754-2008](https://doi.org/10.1109/IEEESTD.2008.4610935).
 * **Examples**:
   * `[[0,0,0],[0,0,2]]` indicates a structure with two sites, one sitting at the origin and one along
   the (positive) `z` axis, 2 Å away from the origin.
 
-### <a name="h.6.2.8">6.2.8. species\_at\_sites</a>
+### <a name="h.6.2.8">6.2.8. all\_coordinates\_known</a>
```
rartino: So, are we going to do this with a boolean field? Because in issue #51 we had the proposal of rather doing this with:
  structure_features: ["disordered", "unknown_positions"]
I don't feel very strongly about it. Maybe separate booleans is the way to go. Though, I'm inconveniently realizing that the filtering language right now cannot do booleans...

giovannipizzi: Ups... I missed the discussion in #51. Maybe good to discuss this tomorrow morning?
For the filtering language and booleans, maybe this should be added then... We have a few booleans apart from this I think.

rartino:
So, as per discussion, is it probably better to go with:
```
  structure_features: ["disorder", "unknown positions"]
```
There is good discussion about this in #51, but in short:- 
- To filter on NO extraordinary features:
```
LENGTH structure_features = 0
```
- To keep structure features limited to the set you are prepared to handle:
```
structure_features HAS ONLY "unknown_positions", "assemblies"
```
And, this you always SHOULD do. Because if we add things to the spec in the future that you aren't prepared to handle, then you do not want those structures.

As for what we right now standardize on as being part of structure_features, we should have:
- `"disorder"` (any one species has a chemical_symbols list that is longer than 1 element OR assemblies is present (?))
- `"unknown positions"` (there is null in any one site position in `cartesian_site_positions`)
- `"assemblies"` (`assemblies` exist and is a non-empty list)
- `"unperiodic"` (`dimension_types` indicate one or more non-periodic dimension)

I see how we might get into deep water into having `"unperiodic"` and not `"perodic`" as the feature flag, though.

giovannipizzi:
I'm not sure about `unperiodic` - not much for the name, but because it's not really a big feature and it's easy (?) to query for this in other ways?

giovannipizzi:
Agreed to change what is written in the comment and then do a final revision

optimade.md
giovannipizzi: I'm going to fix this now

optimade.md
```diff
@@ -1499,12 +1509,15 @@ by multiple chemical elements.
...
-      * the special value `"vacancy"` to represent that this site has a non-zero probability of having
-      a vacancy (the respective probability is indicated in the `concentration` list, see below).
+    * **chemical\_symbols**: REQUIRED; MUST be a list of strings of all chemical elements composing this species. 
+      *  It MUST be one of the following:
+        * a valid chemical-element name, or
+        * the special value `"X"` to represent a non-chemical element, or
+        * the special value `"vacancy"` to represent that this site has a non-zero probability of having
+        a vacancy (the respective probability is indicated in the `concentration` list, see below).
+      * If any one entry in the `species` list has a `chemical_symbols` list that 
+        is longer than 1 element, the correct flag in the [`structure_features`](#h.6.2.11) list MUST be set.
```
CasperWA: To keep conventions, this line should probably be changed to:

```markdown
(...), the correct flag MUST be set in the list `structure_features` (see section [6.2.11 `structure_features`](#h.6.2.11)).
```

optimade.md
```diff
@@ -1499,12 +1509,15 @@ by multiple chemical elements.
...
   * It MUST be a dictionary, where keys represent the species' name, and values are themselves
   dictionaries with the following keys:
-    * **chemical\_symbols**: REQUIRED; MUST be a list of strings of all chemical elements composing
-    this species. It MUST be one of the following:
-      * a valid chemical-element name, or
-      * the special value `"X"` to represent a non-chemical element, or
-      * the special value `"vacancy"` to represent that this site has a non-zero probability of having
-      a vacancy (the respective probability is indicated in the `concentration` list, see below).
+    * **chemical\_symbols**: REQUIRED; MUST be a list of strings of all chemical elements composing this species. 
+      *  It MUST be one of the following:
```
CasperWA: Perhaps loose the "point" for this line. You can add two trailing spaces to the previous line to have it on its own instead, if wanted. It is just that as it is now, the markdown, for some reason, doesn't recognise the indentation and sets this line at the same level of the following "points".

optimade.md
```diff
@@ -1554,8 +1567,10 @@ by multiple chemical elements.
 * **Description**: A description of groups of sites that are statistically correlated.
 * **Requirements/Conventions**:
   * This key is OPTIONAL (it is absent if there are no partial occupancies).
-  * Client implementations MUST check its presence (as its presence changes the interpretation of the
-  structure).
+  * If present, the correct flag in the 
+    [`structure_features`](#h.6.2.11) list MUST be set.
```
CasperWA: same as the comment above.

optimade.md
```diff
@@ -1655,6 +1670,29 @@ by multiple chemical elements.
     probability; the pair (0, 3) with 0.2\*0.7 = 14 % probability; the pair (1, 2) with
     0.8\*0.3 = 24 % probability; and the pair (1, 3) with 0.8\*0.7 = 56 % probability).
 
+### <a name="h.6.2.11">6.2.11. structure\_features</a>
+* **Description**: A list of strings, flagging which special features are used by
+  the structure.
+* **Requirements/Conventions**: 
+  * This property is REQUIRED.
+  * This property MUST be returned as an empty list if no special features are used.
+  * This list MUST be sorted.
```
CasperWA: How must it be sorted?

optimade.md
```diff
@@ -1655,6 +1670,29 @@ by multiple chemical elements.
...
+  * This property MUST be returned as an empty list if no special features are used.
+  * This list MUST be sorted.  
+  * If a special feature listed below is used, the corresponding string MUST be set.
+  * If a special feature listed below is not used, the corresponding string MUST NOT be set. 
+* **List of special structure features**:
+  * `disorder`: This flag MUST be present if any one entry in the `species` list has a 
+    `chemical_symbols` list that is longer than 1 element.
+  * `unknown_positions`: This flag MUST be present if at least one component of the
+    `cartesian_site_positions` list of lists has value `null`.
+  * `assemblies`: This flag MUST be present if the [`assemblies`](#h.6.2.10)
```
CasperWA: same as two previous comments

optimade.md
giovannipizzi: It's because I mistakenly put two spaces in front of `It MUST`, I'm fixing it now.

PR #95, opened 2019-06-12
Introducing relationships for references and provenance

### LINKED ISSUE #27: Provenance / Publication references
OPTIMaDe needs to provide some facility to communicate data provenance.

Specifically, some data provenance property needs to facilitate the aggregation of OPTIMaDa IDs (i.e., database prefix + local id) and other ids.

Dates?

Publications?

merkys:
@rartino, @muhrin, @ml-evs, others (sorry, I forgot the GitHub handles, feel free to add them in) and I have put together a proposal to describe the provenance and publication references (closes #27). The proposal relies on [JSON API Relationships](https://jsonapi.org/format/#document-resource-object-relationships) and introduces a new entry type `reference` with fields based on <s>[BibJSON specification](http://okfnlabs.org/bibjson/)<s>BibTeX.

<s>To implement filtering by references, the reference list of any entry MUST be duplicated in `references` property, as we currently have no way to filter by relationships (#96).</s>


optimade.md
```diff
@@ -1611,6 +1625,65 @@ by multiple chemical elements.
...
 
+## <a name="h.6.4">6.4. Reference Entries</a>
+
+`"reference"` entries describe bibliographic references. To do so, they MUST
+use properties defined by [BibJSON specification](http://okfnlabs.org/bibjson/).
+At least one property is REQUIRED.
+
+## <a name="h.6.5">6.5. Relationships Used by Multiple Entry Types</a>
+
+[JSON API Relationships](https://jsonapi.org/format/#document-resource-object-relationships)
```
ml-evs: Could you change this link to pin to v1.0 of the JSON API?

e.g. https://jsonapi.org/format/1.0/#document-resource-object-relationships

merkys: Fixed.

optimade.md
```diff
@@ -1611,6 +1625,83 @@ by multiple chemical elements.
...
+        }
+      ]
+    }
+  },
+  "included": [
+    {
+      "type": "reference",
+      "id": "1234",
+      "attributes": {
+        "identifier": [{"type": "doi","id": "10.1234/1234"}]
```
giovannipizzi: ~If we want to do it this way, we should define an `identifier` type, explain what `type=doi` is, ... - I don't think this done either here or in the JSON API specs. Otherwise, why don't we just have `identifier` be a URL? So using the standard `links` of JSON API?~
I see now that this is the bibJSON standard mentioned elsewhere - see then my other comments - remove `included` and leave only `relationships`, and add an example in the `reference` entry

optimade.md
```diff
@@ -1611,6 +1625,83 @@ by multiple chemical elements.
...
+          "type": "reference",
+          "id": "5678",
+          "meta": {
+            "description": "This article has been retracted"
+          }
+        }
+      ]
+    }
+  },
+  "included": [
```
giovannipizzi: Do we really want to define both a new resource for references (with `relationships`), and also have the possibility to include them "in-line" with `included`? It seems too much to me

optimade.md
```diff
@@ -1323,6 +1328,15 @@ This section defines standard entry types and their properties.
...
   * \_exmpl\_workflow\_id
 
+### <a name="h.6.1.4">6.1.4. references</a>
+
+* **Description**: to simplify querying entries by their references, IDs of references,
+  if provided as JSON API relationships as described in [6.5.1. references](#h.6.5.1),
+  MUST be also provided as a list of IDs.
+* **Requirements/Conventions**: MUST be a list of strings enumerating references
+  existing in the database.
+* **Example**: `[ "Dijkstra1968", "Babbage1864" ]`
```
giovannipizzi: Not sure I understand how this is currently used (and how this helps)

optimade.md
```diff
@@ -1611,6 +1625,83 @@ by multiple chemical elements.
 `"calculation"` entries have the properties described above in section
 [6.1. Properties Used by Multiple Entry Types](#h.6.1).
 
+## <a name="h.6.4">6.4. Reference Entries</a>
+
+`"reference"` entries describe bibliographic references. To do so, they MUST
+use properties defined by [BibJSON specification](http://okfnlabs.org/bibjson/).
+At least one property is REQUIRED.
```
giovannipizzi: Can we add a short example here instead of inside "included" in the example below?

optimade.md
```diff
@@ -1611,6 +1625,83 @@ by multiple chemical elements.
...
+
+### <a name="h.6.5.1">6.5.1. References</a>
+
+`"references"` relationship is used to provide bibliographic references for any
+of the entry types. It relates an entry with any number of `"reference"` entries.
+`"reference"` entries, mentioned in the returned JSON document, SHOULD be included
+in the `"included"` field as per
+[JSON API 1.0 specification](https://jsonapi.org/format/1.0/#fetching-includes).
+
+Example:
```
CasperWA: This example seems wrong, according to the JSON API, see the example in [this](https://jsonapi.org/format/#document-resource-object-linkage) section.

In short, `relationships` should be a part of the returned resource object, i.e., on the same level as `attributes`. This makes sense, because the references or "related" objects, belong to the individual entry object.

optimade.md
```diff
@@ -1611,6 +1625,83 @@ by multiple chemical elements.
...
+    "attributes": {
+      "formula": "Es2",
+      "local_id": "example.db:structs:1234",
+      "url": "http://example.db/structs/1234",
+      "immutable_id": "http://example.db/structs/1234@123",
+      "last_modified": "2007-04-07T12:02Z"
+    },
+  },
+  "relationships": {
+    "references": {
```
CasperWA: According to the link in the previous comment, `relationships` should either:
- Return `null` or a single resource object (i.e., it shouldn't be `"references"` in plural, but singular), when one knows that maximum a single relationship can occur.
- Return a list of linkage resource objects, i.e., `relationships` becomes a list (if I understood correctly).

CasperWA: I may have gotten it slightly wrong here. A good example I found is [here](https://jsonapi.org/format/#document-compound-documents).

optimade.md
```diff
@@ -1611,6 +1625,83 @@ by multiple chemical elements.
...
+      }
+    }
+  ]
+}
+```
+
+### <a name="h.6.5.2">6.5.2. Calculations</a>
+
+Relationships with calculations MAY be used to indicate provenance where a structure
+is either an input to or an output of a calculations.  In future versions of this API
```
CasperWA: One space too much (and a missing comma before "where" in the previous line.

merkys: I agree with the space, but don't feel strongly about the need of a comma before "where".

optimade.md
merkys: Added.

optimade.md
merkys: You are right, this was a mistake. Fixed now.

optimade.md
merkys: > I may have gotten it slightly wrong here. A good example I found is [here](https://jsonapi.org/format/#document-compound-documents).

Could you please check the fixed example and let me know which parts seem departed from the JSON API spec?

optimade.md
merkys: Defining `references` endpoints will allow both querying and selective inclusion of bibliographic data into the response JSONs. `included` is mandated by the JSON API specs. Inclusion or exclusion of `included` in JSON API is a user-requested thing. We haven't decided yet what would be the default behaviour here.

merkys: `included` field is described as well in [3.3.2. JSON API Response Schema: Common Fields](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.md#332-json-api-response-schema-common-fields) as a container for referenced entries.

optimade.md
merkys: This is clearly redundant, but it's needed to filter the entries by their references while we find a way to solve #96.

optimade.md
merkys: Please see my responses to your comments below.

optimade.md
```diff
@@ -532,7 +537,7 @@ contain the field
 * **meta**: a [JSON API meta object](https://jsonapi.org/format/#document-meta) that contains
 non-standard meta-information about the object
 * **relationships**: a dictionary containing references to other resource objects as defined in
-[JSON API relationships object](http://jsonapi.org/format/#document-resource-object-relationships)
+[6.5. Relationships Used by Multiple Entry Types](#.6.5)
```
CasperWA: missing an "h" in the link

optimade.md
```diff
@@ -1323,6 +1328,15 @@ This section defines standard entry types and their properties.
   * \_exmpl\_trajectory
   * \_exmpl\_workflow\_id
 
+### <a name="h.6.1.4">6.1.4. references</a>
+
+* **Description**: to simplify querying entries by their references, IDs of references,
```
CasperWA: This sentence is a bit confusing to me. Could you try and rewrite it? Or maybe completely remove it from here, since references will be separate entries in the `references` endpoint, and I don't think we have structures here, for example?
The definition of references would then be under 6.4. Reference Entries.
But maybe I have misunderstood something?

optimade.md
CasperWA: Maybe this is related with my comment as well.
I still think it is not needed to have this, i.e. section 6.1.4. references. Since we also don't have this for the other entry types.
There should, however, definitely be a separate section about how to query attributes and relationships in general. But this should be under a filter section or similar.

optimade.md
```diff
@@ -1611,6 +1625,101 @@ by multiple chemical elements.
...
+    }
+  }
+}
+```
+
+## <a name="h.6.5">6.5. Relationships Used by Multiple Entry Types</a>
+
+[JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships)
+MAY be used to describe the relations between entries. Human-readable description
+of a relationship MAY be provided using `"description"` field inside `"meta"`
```
CasperWA: the (`"description"` field inside) the (`"meta"` dictionary ...)

optimade.md
```diff
@@ -1611,6 +1625,101 @@ by multiple chemical elements.
...
+dictionary of a relationship.
+
+### <a name="h.6.5.1">6.5.1. References</a>
+
+`"references"` relationship is used to provide bibliographic references for any
```
CasperWA: The (`"references"` relationship ...)

optimade.md
```diff
@@ -1611,6 +1625,101 @@ by multiple chemical elements.
...
+of the entry types. It relates an entry with any number of `"reference"` entries.
+`"reference"` entries, mentioned in the returned JSON document, SHOULD be included
+in the `"included"` field as per
```
CasperWA: (in the) top-level (`"included"` field as per) the ([JSON ....)

optimade.md
```diff
@@ -1611,6 +1625,101 @@ by multiple chemical elements.
...
+      "id": "1234",
+      "attributes": {
+        "identifier": [{"type": "doi","id": "10.1234/1234"}]
+      }
+    },
+    {
+      "type": "reference",
+      "id": "5678",
+      "attributes": {
+        "identifier": [{"type": "doi","id": "10.5678/5678"}]
```
CasperWA: Just for "fun", could you add (or exchange one of the existing, perhaps the first vanilla reference) with the example of a reference resource object that you have in the previous section (`"id": "Dijkstra1968"`) :)
Because in this case, I would think that in the `"included"` field, you would need to show the full resource, right?

optimade.md
```diff
@@ -1611,6 +1625,101 @@ by multiple chemical elements.
...
+      }
+    }
+  ]
+}
+```
+
+### <a name="h.6.5.2">6.5.2. Calculations</a>
+
+Relationships with calculations MAY be used to indicate provenance where a structure
+is either an input to or an output of a calculations. In future versions of this API
```
CasperWA: A rewriting of the second sentence:

"""
We intend to implement in a future version a standardized way (or mechanism, you can choose yourself here :) ) of differentiating between these two cases, thus allowing databases a common way of exposing the full provenance tree with inputs and outputs between structures and calculations.

At the moment we suggest database providers to extend their API the way the choose, always using their database-provider-specific prefix in non-standardized fields.
"""

optimade.md
merkys: @fawzi and @rartino agreed to discontinue this feature in favour of a good solution to #96.

optimade.md
merkys: Well spotted! Fixed.

optimade.md
merkys: "6.1.4 references" has been removed altogether.

optimade.md
merkys: Fixed.

optimade.md
merkys: Fixed.

optimade.md
merkys: Fixed.

optimade.md
merkys: Done.

optimade.md
merkys: Done :)

> Because in this case, I would think that in the `"included"` field, you would need to show the full resource, right?

True.

optimade.md
CasperWA: image

optimade.md
```diff
@@ -53,6 +53,10 @@
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.9. species](#h.6.2.9)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.10. assemblies](#h.6.2.10)  
 &nbsp;&nbsp;&nbsp;&nbsp;[6.3. Calculation Entries](#h.6.3)  
+&nbsp;&nbsp;&nbsp;&nbsp;[6.4. Reference Entries](#h.6.4)  
+&nbsp;&nbsp;&nbsp;&nbsp;[6.5. Relationships Used by Multiple Entry Types](#.6.5)
```
CasperWA: missing "h"

optimade.md
```diff
@@ -1611,6 +1615,107 @@ by multiple chemical elements.
...
+    }
+  }
+}
+```
+
+## <a name="h.6.5">6.5. Relationships Used by Multiple Entry Types</a>
+
+[JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships)
+MAY be used to describe the relations between entries. Human-readable description
+of a relationship MAY be provided using `"description"` field inside the `"meta"`
```
CasperWA: (... provided using) the (`"description"` field ...)

optimade.md
CasperWA: @giovannipizzi, consider `"relationships"` to be a list of foreign keys, and the OPTIONAL list under `"included"` to be the actually retrieved data.
This makes sense - especially because the content in `"included"` is optional.

optimade.md
```diff
@@ -1611,6 +1615,107 @@ by multiple chemical elements.
...
+      }
+    }
+  ]
+}
+```
+
+### <a name="h.6.5.2">6.5.2. Calculations</a>
+
+Relationships with calculations MAY be used to indicate provenance where a structure
+is either an input to or an output of a calculations. We intend to implement in the
```
CasperWA: (... implement in) a (future version ...)

optimade.md
```diff
@@ -1611,6 +1615,107 @@ by multiple chemical elements.
...
+future version of this API a standardized mechanism to differentiate these two cases,
+thus allowing databases a common way of exposing the full provenance tree with inputs
+and outputs between structures and calculations. At the moment the database providers
```
CasperWA: This last sentence should eventually be removed, so I think it should be separated from the main text here, i.e., put a couple of EOL before it :)

optimade.md
```diff
@@ -53,7 +53,11 @@
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.9. species](#h.6.2.9)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[6.2.10. assemblies](#h.6.2.10)  
 &nbsp;&nbsp;&nbsp;&nbsp;[6.3. Calculation Entries](#h.6.3)  
-&nbsp;&nbsp;&nbsp;&nbsp;[6.4. Database-Provider-Specific Entry Types](#h.6.4)  
+&nbsp;&nbsp;&nbsp;&nbsp;[6.4. Reference Entries](#h.6.4)  
+&nbsp;&nbsp;&nbsp;&nbsp;[6.5. Database-Provider-Specific Entry Types](#h.6.5)  
+&nbsp;&nbsp;&nbsp;&nbsp;[6.6. Relationships Used by Multiple Entry Types](#.6.6)
```
CasperWA: missing "h" in the link

merkys: Done.

optimade.md
merkys: Done.

optimade.md
```diff
@@ -1620,13 +1624,116 @@ by multiple chemical elements.
...
+      }
+    }
+  ]
+}
+```
+
+### <a name="h.6.6.2">6.6.2. Calculations</a>
+
+Relationships with calculations MAY be used to indicate provenance where a structure
+is either an input to or an output of a calculations. We intend to implement in a
```
CasperWA: no "a" here before calculations - or go singular instead - but I think it should support multiples?
```suggestion
is either an input to or an output of calculations. We intend to implement in a
```

merkys: A structure may be used as input for many calculations, so let's leave the plural form.

CasperWA: Agree, i.e., the "a" should be removed.

merkys: Done!

merkys:
> Some beautiful stuff here, thanks @merkys! sunglasses

Thanks @CasperWA for approval! :)

optimade.md
```diff
@@ -1620,13 +1624,116 @@ by multiple chemical elements.
 `"calculation"` entries have the properties described above in section
 [6.1. Properties Used by Multiple Entry Types](#h.6.1).
 
-## <a name="h.6.4">6.4. Database-Provider-Specific Entry Types</a>
+## <a name="h.6.4">6.4. Reference Entries</a>
+
+`"reference"` entries describe bibliographic references. To do so, they MUST
+use properties defined by [BibJSON specification](http://okfnlabs.org/bibjson/).
+At least one property is REQUIRED.
```
giovannipizzi: Replace with something like "At least one field from the bibJSON specification MUST be present."

optimade.md
```diff
@@ -1620,13 +1624,116 @@ by multiple chemical elements.
 `"calculation"` entries have the properties described above in section
 [6.1. Properties Used by Multiple Entry Types](#h.6.1).
 
-## <a name="h.6.4">6.4. Database-Provider-Specific Entry Types</a>
+## <a name="h.6.4">6.4. Reference Entries</a>
+
+`"reference"` entries describe bibliographic references. To do so, they MUST
+use properties defined by [BibJSON specification](http://okfnlabs.org/bibjson/).
```
giovannipizzi: Verify that we know exactly what bibJSON requires, and make sure we are not allowing completely free-form

Also maybe specify that for additional keys, they cannot just start with `_` as bibJSON says, but with the identifier of the provider (e.g. `_exmpl_`).

optimade.md
```diff
@@ -1620,13 +1624,116 @@ by multiple chemical elements.
...
+[JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships)
+MAY be used to describe the relations between entries. Human-readable description
+of a relationship MAY be provided using the `"description"` field inside the
+`"meta"` dictionary of a relationship.
+
+### <a name="h.6.6.1">6.6.1. References</a>
+
+The `"references"` relationship is used to provide bibliographic references for any
+of the entry types. It relates an entry with any number of `"reference"` entries.
+`"reference"` entries, mentioned in the returned JSON document, SHOULD be included
```
merkys: Reformulate sentence in order to not start it with `"reference"`.

giovannipizzi:
Can be merged after comments have been addressed

optimade.md
merkys: Fixed.

optimade.md
merkys: I still have to sort this thing out. BibJSON seems to inherit properties from BibTeX, but the only list of BibTeX properties I was able to find on the spot is this: http://bibtexml.sourceforge.net/btxdoc.pdf (year 1988, so no DOIs and stuff). Will come back to this later.

CasperWA:
Current status is expressed in a review comment by @merkys, linked [here](https://github.com/Materials-Consortia/OPTiMaDe/pull/95#discussion_r293488185), rewritten here:

> I still have to sort this thing out. BibJSON seems to inherit properties from BibTeX, but the only list of BibTeX properties I was able to find on the spot is this: http://bibtexml.sourceforge.net/btxdoc.pdf (year 1988, so no DOIs and stuff). Will come back to this later.

merkys:
Hmm, BibJSON does not seem to have a proper specification:
> BibJSON is not a complete standard, the aim is to demonstrate the simplicity with which we can represent this data in a JSON object, using namespaces to extend keys as necessary.

(from [Jones et al., 2011](https://static-content.springer.com/esm/art%3A10.1186%2F1758-2946-3-47/MediaObjects/13321_2011_228_MOESM1_ESM.HTML))

Although BibJSON uses BibTeX fields, [BibTeX v0.99b spec](http://bibtexml.sourceforge.net/btxdoc.pdf) does not contain fields `url` and `doi`, which we urgently need. I see two options how we can continue:
1. define the BibJSON as we understand it, including `url`, `doi` and other fields (essentially a fork of the BibJSON spec);
2. choose another standard (suggestions?)

Pinging provenance group: @fawzi, @rartino, @ml-evs.

merkys:
Ditched BibJSON as it does not seem to have a formal specification. I've placed a note that we support the BibTeX specification + `doi` and `url`.

optimade.md
```diff
@@ -1667,13 +1671,137 @@ by multiple chemical elements.
...
+* **address**, **annote**, **booktitle**, **chapter**, **crossref**,
+  **edition**, **howpublished**, **institution**, **journal**, **key**,
+  **month**, **note**, **number**, **organization**, **pages**, **publisher**,
+  **school**, **series**, **title**, **type**, **volume**, **year**: meanings
+  of these properties match the
+  [BibTeX specification](http://bibtexml.sourceforge.net/btxdoc.pdf), values
+  are strings;
+* **authors** and **editors**: lists of *person objects* which are dictionaries
+  with the following fields:
+  * **name**: name of the person, REQUIRED.
```
CasperWA: Perhaps specify that this is the full name, since you are specifying first and last name in the next line.

optimade.md
```diff
@@ -1667,13 +1671,137 @@ by multiple chemical elements.
...
 
 Names of database-provider-specific entry types MUST start with
 database-provider-specific namespace prefix as given in [Appendix 1](#h.app1).
 Database-provider-specific entry types MUST have all properties described above
 in section [6.1. Properties Used by Multiple Entry Types](#h.6.1).
 
+## <a name="h.6.6">6.6. Relationships Used by Multiple Entry Types</a>
+
+[JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships)
+MAY be used to describe the relations between entries. Human-readable description
```
CasperWA: Either:
"A human-readable description ..."
or
"Human-readable descriptions ..."

optimade.md
```diff
@@ -1667,13 +1671,137 @@ by multiple chemical elements.
...
+  **month**, **note**, **number**, **organization**, **pages**, **publisher**,
+  **school**, **series**, **title**, **type**, **volume**, **year**: meanings
+  of these properties match the
+  [BibTeX specification](http://bibtexml.sourceforge.net/btxdoc.pdf), values
+  are strings;
+* **authors** and **editors**: lists of *person objects* which are dictionaries
+  with the following fields:
+  * **name**: name of the person, REQUIRED.
+  * **firstname**, **lastname**: parts of the person's name, OPTIONAL.
+* **doi** and **url**, strings.
```
CasperWA: I feel like this whole listing could be done more rigidly (with descriptions etc. of all these fields), but at the same time, I'm loving the conciseness and how short it is. Furthermore, one can simply go to the linked resource to get the descriptions (hopefully).
The only thing is perhaps this line. Maybe just make it a bit more human-readable and fit your other points above. Perhaps something like:

```markdown
* **doi** and **url**: values are strings.
```

optimade.md
```diff
@@ -1667,13 +1671,137 @@ by multiple chemical elements.
...
+      }
+    },
+  },
+  "included": [
+    {
+      "type": "reference",
+      "id": "Dijkstra1968",
+      "attributes": {
+        "authors": [
+            { "name": "Edsger Dijkstra" }
```
CasperWA: Update this to match the shorter example above.

optimade.md
```diff
@@ -1667,13 +1671,137 @@ by multiple chemical elements.
...
+      }
+    }
+  ]
+}
+```
+
+### <a name="h.6.6.2">6.6.2. Calculations</a>
+
+Relationships with calculations MAY be used to indicate provenance where a structure
+is either an input to or an output of calculations. We intend to implement in a
```
CasperWA: After some thought, we should probably consider putting everything after the first sentence here in a markdown "note". What do you think?

giovannipizzi:
Before merging, could you replace (*ONLY* in the *new* examples that you added) 
```
 ```json
```
with
```
 ```jsonc
```
?

(see also #118 and PR #121)

optimade.md
merkys: Fixed.

optimade.md
merkys: Fixed.

optimade.md
merkys: @CasperWA why? The more examples, the better, isn't this true?

optimade.md
merkys: I've left BibTeX field name list unchanged. I don't think I'd describe these BibTeX fields better than they are given in the original publication.

> The only thing is perhaps this line. Maybe just make it a bit more human-readable and fit your other points above. Perhaps something like:

Fixed.

optimade.md
merkys: Sure, done as suggested.

optimade.md
CasperWA: Hehe, true. It was simply because it is otherwise completely similar to the previous example (that includes `firstname` and `lastname`) and this may leave a reader in doubt whether resources returned under `include` should not return ALL of the properties/fields of a resource object.

merkys: I see the point. Made the examples match.

Issue #20, opened 2018-06-14
Allow sorting and/or paging of output


rartino:
Allow a sort query parameter, e.g., something like:
```
    ?filter=element_ratios HAS element = "Si"&order=chemical_formula
```
Ascending, descending? Multiple fields?

rartino:
So, how about the syntax:
```
 ?filter=element_ratios HAS element = "Si"&order=chemical_formula
 ?filter=element_ratios HAS element = "Si"&order=chemical_formula:asc
 ?filter=element_ratios HAS element = "Si"&order=chemical_formula:desc

 ?filter=element_ratios HAS element = "Si"&order=chemical_formula:asc,elements:dsec
```
Where the first defaults to desc = descending.

rartino:
Approved as optional

merkys:
Suggestion for discovery mechanism: at the entry introspection endpoints viz. ``/structures/info`` we have the following data structure:
```
{
  "properties" :
    {
      "some_property" : { "description" : "free text description", "unit" : "unit description" }
    }
}
```
We may include additional attribute ``sortable`` alongside ``description`` and ``unit``. Value of this attribute would be boolean, and ``False`` by default. Therefore, databases that choose to implement sorting would need to explicitly indicate so for all sortable properties.

dwinston:
I suggest using the `page` key for pagination concerns as per [https://jsonapi.org/format/1.0/#fetching-paginationI](https://jsonapi.org/format/1.0/#fetching-pagination). Specifically, I suggest replacing the `response_limit` key in our spec with `page[limit]` and using its counterpart `page[offset]` in order to fully specify pagination as an offset-based strategy. The alternative suggested by JSON API 1.0 is `page[number]` and `page[size]` for a page-based strategy. My preference is for an offset-based strategy. Thus, while paging is OPTIONAL, I would us to specify the paging strategy and keys used if one implements paging.

dwinston:
### Decisions on sorting
- Support for the `sort` query parameter for an entry listing endpoint is optional. If an implementation supports sortable queries, it MUST use the `sort` query parameter with format as specified by the [JSON API 1.0 spec](https://jsonapi.org/format/1.0/#fetching-sorting).
- It is not required that an implementation supports multiple sort fields for a single query. However, if it does, it again must conform to the [JSON API 1.0 spec](https://jsonapi.org/format/1.0/#fetching-sorting).
[...]

PR #97, opened 2019-06-12
sorting and pagination updates

dwinston:


dwinston:
Closes #20.

dwinston:
Reproducing last comment of #20 (conclusions of working group) here:

### Decisions on sorting
- Support for the `sort` query parameter for an entry listing endpoint is optional. If an implementation supports sortable queries, it MUST use the `sort` query parameter with format as specified by the [JSON API 1.0 spec](https://jsonapi.org/format/1.0/#fetching-sorting).
- It is not required that an implementation supports multiple sort fields for a single query. However, if it does, it again must conform to the [JSON API 1.0 spec](https://jsonapi.org/format/1.0/#fetching-sorting).
- If an implementation supports sorting for an entry listing endpoint, then the` /<entries>/info` endpoint MUST include, for each field name `<field>` in its `data.properties.<field>` response value, the key "sortable" with value `true`. This is in addition to the property description (and optional unit).
### Decisions on pagination
- Pagination is MANDATORY. Implementations MUST conform to the [JSON API 1.0 spec](https://jsonapi.org/format/1.0/#fetching-pagination). That is, when a server does not return all data matching a query (which it must be capable of, because a user may request a small `page[limit]` -- see below), it must provide a `links` object in the response that includes `next`, `prev`, `first`, and `last` values -- in particular, we require each of these keys to have either a valid URL or `null` as their value. However, there are only two situations where the value of a key is (and MUST be) `null`: in the first-page response, the value of `prev` is `null`, and in the last-page response, the value of `next` is `null`. In all other cases, the value of each of these keys must be a valid URL.
- Following recommendations of the JSON API 1.0 spec and its reservation of the `page` query argument for pagination, implementers MUST use an offset-based pagination strategy using the `page[offset]` and `page[limit]` query arguments. It is trivial for an implementation to map this strategy to a page-based strategy using arithmetic on the server side. **NOTE: this means the `response_limit` query parameter will be removed.**
- If a server sets a maximum `page[limit]` and a user specifies a larger limit, the server MUST return an error. For example, if a server `page[limit]` max is 100, and a user specifies 200, the server cannot return 100 entries with a warning. The server MUST return an error.
- Thus, gien the above specification on pagination, each URL value of a `links` field will include the appropriate `page[offset]` and `page[limit]` values while reproducing all other query parameters of the original request.

optimade.md
```diff
@@ -275,6 +275,16 @@ or a _list_ of JSON API resource objects. Every resource object needs the `type`
 and its attributes (described in section [4. API Endpoints](#h.4))
 need to be in a dictionary corresponding to the `attributes` field.
 
+<<<<<<< HEAD
```
CasperWA: Remove - It seems there is an incomplete merge / merge conflicts that haven't been resolved?

CasperWA: It is difficult for me to understand which of these lines I should review - will wait :)

giovannipizzi: There is a merge conflict that has remained into the final commit

optimade.md
```diff
@@ -285,7 +295,14 @@ The response MAY OPTIONALLY also return resources related to the primary data in
     data, or if there is no way to fetch additional data. The link object can contain
     the following members:
     * **href**: a string containing the link’s URL.
+>>>>>>> upstream/develop
     * **meta**: a meta object containing non-standard meta-information about the link.
+  * **prev**: is `null` only when the current response is the first page of data.
+  * **last**: the last page of data.
+  * **first**: the first  page of data.
+  
+    NOTE: The links `next`, `prev`, `last`, and `first` are MANDATORY when **more_data_available** is `true`, that is, for
```
CasperWA: I think you can leave out the last part and just end on "`true`".

optimade.md
```diff
@@ -446,6 +463,18 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
 
 * **filter**: a filter string, in the format described below in section
   [5. API Filtering Format Specification](#h.5).
+* **page[limit]**: sets a numerical limit on the number of entries returned. See [https://jsonapi.org/format/1.0/#fetching-pagination](https://jsonapi.org/format/1.0/#fetching-pagination). The API
```
CasperWA: Have instead "`(...) See the [JSON API v1.0](<link>) specification on pagination. (...)`"

optimade.md
```diff
@@ -446,6 +463,18 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
 * **filter**: a filter string, in the format described below in section
   [5. API Filtering Format Specification](#h.5).
+* **page[limit]**: sets a numerical limit on the number of entries returned. See [https://jsonapi.org/format/1.0/#fetching-pagination](https://jsonapi.org/format/1.0/#fetching-pagination). The API
+  implementation MUST return no more than the number specified. It MAY return
+  less. The database MAY have a maximum limit and not accept larger numbers (in
+  which case an error code MUST be returned). The default limit value is up
+  to the API implementation to decide.  
+  
+  Example: <http://example.com/optimade/v0.9/structures?page[limit]=100>
+* **page[offset]**: Implements, along with **page[limit]**, an offset-based strategy for
```
CasperWA: Bold, when "title" in the listing, otherwise "code", i.e.:

```markdown
* **page[offset]**: Implements, along with `page[limit]`, an offset-based ...
```

optimade.md
```diff
@@ -446,6 +463,18 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+  pagination. See [https://jsonapi.org/format/1.0/#fetching-pagination](https://jsonapi.org/format/1.0/#fetching-pagination).
```
CasperWA: See previous comment.

optimade.md
```diff
@@ -446,6 +463,18 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+  
+  Example (equivalent to second "page" of 100 entries): <http://example.com/optimade/v0.9/structures?page[offset]=100&page[limit]=100>
+* **sort**: If supporting sortable queries, an implementation MUST use the **sort** query parameter with format as specified by [https://jsonapi.org/format/1.0/#fetching-sorting](https://jsonapi.org/format/1.0/#fetching-sorting). It is not required that an implementation supports multiple sort fields for a single query. However, if it does, it again must conform to the JSON API 1.0 spec. If an implementation supports sorting for an [entry listing endpoint](#h.4.4.2), then the `/<entries>/info` endpoint MUST include, for each field name `<fieldname>` in its "data.properties.`<fieldname>`" response value, the key "sortable" with value true. This is in addition to each property description (and optional unit).
```
CasperWA: See previous comment (`sort` vs. **sort**)

CasperWA: See previous comment (link `[JSON API v1.0](<link>)`)

CasperWA: Maybe give each sentence its own line in the markdown? It makes it a bit easier to read when coded.

CasperWA: true should be in code (last word of the second to last sentence).

rartino:
The sort part of this is great! I'd be happy to approve that separately.

About paging: we had quite some discussion about it today; and just to note down my personal opinions on this matter:
* In my opinion pagination should only say that the `next` link is MANDATORY, and that is only for unsorted data (using any mechanism supported within jsonapi) and then the phrasing should be that "API implementations SHOULD also provide prev, last and first links unless that is overly difficult for the underlying database."
* For sorted data the API implementation "SHOULD provide next, prev, last, and first links unless that is overly difficult for the underlying database."

For users who are concerned with getting all the data, this really should be sufficient: they can page through the unsorted data and download it all. If they are downloading it all, it cannot be super-important to get it pre-sorted, the client can do the sort itself once it has everything. 

On the API implementer side, this should be OK even if they don't want to keep cursors and even if [offsets are expensive](https://www.eversql.com/faster-pagination-in-mysql-why-order-by-with-limit-and-offset-is-slow/). They can output the data sorted on any unique field (id, insert number, timestamp) and then do pagination by essentially append to the filter, e.g., `id > last value`. (However, I suggest standardizing optional use of `page_above=last_value` for this just so the server can avoid rewriting the filter string.)

For sorted data, it is great if pagination can be supported. But, I really see a possible problem scenario for an API implementation when:
* The underlying database has trouble handling sorting and, in particularly, expressing inequality search for a combination field (e.g. spacegroup + id). 
* But it would still be useful for users to sort on spacegroup even though possibly millions of entries has the same spacegroup, so 'spacegroup > 42'-type paging wouldn't be possible.

dwinston:
Notes from morning discussion:

- Client can use `page_limit` to limit number of entries per response. An implementation must respect this -- however, a server may set a maximum value and in this case, if the client sets a higher value, must return an error.
- It is mandatory to implement pagination in the case of no user-specified sorting.
- At a minimum, pagination means providing a non-`null` `next` link as long as there is more data available for a query to return.
- If an implementation chooses to implement
  - _offset-based pagination_: using `page_offset` and `page_limit` is recommended.
  - _cursor-based pagination_: using `page_cursor` and `page_limit` is recommended.
  - _page-based pagination_: using `page_number` and `page_limit` is recommended.
  - _value-based pagination_: using `page_above`/`page_below` and `page_limit` is recommended.

optimade.md
merkys: @CasperWA all the JSON API says about **page[limit]** and the like are _not_ the specification, but an example.

optimade.md
dwinston: Fixed

optimade.md
dwinston: I refactored this section

optimade.md
```diff
@@ -275,17 +275,28 @@ or a _list_ of JSON API resource objects. Every resource object needs the `type`
...
-    * **meta**: a meta object containing non-standard meta-information about the link.
+The response MAY OPTIONALLY also return resources related to the primary data in the field.
+NOTE: **links** is MANDATORY for implementing pagination.
+
+* **links**: [JSON API links](http://jsonapi.org/format/#document-links). Each field of a links object, i.e. a "link",
+  must be either
+  
+  * `null`
+  * a string representing a URI, or
+  * an object ("link object") with fields
```
CasperWA: Remove mention of object, write instead dictionary I think, according to the term definition, this should be fine.
This particular thing has been bugging me for a while - because I would think a JSON API links object is either the dict with `"href"` (and maybe `"meta"`) OR a string. This of course is counter intuitive to the general definition of a JSON object, however, I am fine with this for a link - simply to save having to go down into `"href"` to get the links always...

optimade.md
```diff
@@ -275,17 +275,28 @@ or a _list_ of JSON API resource objects. Every resource object needs the `type`
...
-  containing the JSON API links objects:
-  * **next**: is an URI that represents a suggested way to fetch the
-    next set of results if not all were returned, either directly as a string,
-    or as a link object. The field MUST be null or omitted if there is no additional
-    data, or if there is no way to fetch additional data. The link object can contain
-    the following members:
-    * **href**: a string containing the link’s URL.
-    * **meta**: a meta object containing non-standard meta-information about the link.
+The response MAY OPTIONALLY also return resources related to the primary data in the field.
+NOTE: **links** is MANDATORY for implementing pagination.
```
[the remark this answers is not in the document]
CasperWA: Concerning the "(see section ...)" above for pagination, it seems there is no separate section describing pagination. Maybe it's not needed, but maybe it would be nice to have, where all things pagination can be gathered and referenced? Only the absolutely required fields in the response should be stated here - and any non-special optional fields (such as `"meta"` and others).

optimade.md
```diff
@@ -275,17 +275,28 @@ or a _list_ of JSON API resource objects. Every resource object needs the `type`
...
+  * an object ("link object") with fields
+    * **href**: a string representing a URI
+    * **meta**: (OPTIONAL) a meta object containing non-standard meta-information about the link
+
+  The following fields are REQUIRED for implementing pagination:
+  
+  * **next**: represents a link to fetch the next set of results. When the current response is the last page of data,
+    this field MUST be either omitted or `null`-valued.
+   
+  The following fields are reserved for pagination. An implementation MAY offer these links:
```
CasperWA: Maybe this could be in the separate pagination section suggested above?

CasperWA: There should be a line here stating that the values are similar to `"next"`, in the sense that it should be a link.

optimade.md
```diff
@@ -446,6 +457,36 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
 
 * **filter**: a filter string, in the format described below in section
   [5. API Filtering Format Specification](#h.5).
+  
+* **page_limit**: sets a numerical limit on the number of entries returned. See
+  [JSON API 1.0](https://jsonapi.org/format/1.0/#fetching-pagination). The API
+  implementation MUST return no more than the number specified. It MAY return fewer. The database MAY have a maximum
+  limit and not accept larger numbers (in which case an error code MUST be returned). The default limit value is up to
```
CasperWA: Which error code?

optimade.md
```diff
@@ -446,6 +457,36 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+  [JSON API 1.0](https://jsonapi.org/format/1.0/#fetching-pagination). The API
+  implementation MUST return no more than the number specified. It MAY return fewer. The database MAY have a maximum
+  limit and not accept larger numbers (in which case an error code MUST be returned). The default limit value is up to
+  the API implementation to decide.
+  
+  Example: <http://example.com/optimade/v0.9/structures?page_limit=100>
+  
+* **page_{offset, page, cursor, above, below}**: A server MUST implement pagination in the case of no
+  user-specified `sort` parameter (via the ["links" response field](#h.3.3.2)). A server MAY implement pagination in
+  concert with `sort`. The following parameters, all prefixed by "page_", are RECOMMENDED for use with pagination.
```
CasperWA: nice. "concert". nice.. 😎

optimade.md
```diff
@@ -446,6 +457,36 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+  If an implementation chooses
+  
+  * _offset-based pagination_: using `page_offset` and `page_limit` is RECOMMENDED.
+  * _cursor-based pagination_: using `page_cursor` and `page_limit` is RECOMMENDED.
+  * _page-based pagination_: using `page_number` and `page_limit` is RECOMMENDED (`page_limit` is equivalent to page "size").
+  * _value-based pagination_: using `page_above`/`page_below` and `page_limit` is RECOMMENDED.
+  
+Example (skip 50 structures and fetch up to 100): <http://example.com/optimade/v0.9/structures?page_offset=50&page_limit=100>
```
CasperWA: Please provide more of these excellent simple examples, if it is not too much trouble :)

optimade.md
```diff
@@ -446,6 +457,36 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+
+* **sort**: If supporting sortable queries, an implementation MUST use the `sort` query parameter with format as
+  specified by [JSON API 1.0](https://jsonapi.org/format/1.0/#fetching-sorting).
+  
+  It is not required that an implementation supports multiple sort fields for a single query. However, if it does, it
+  again must conform to the JSON API 1.0 spec.
```
CasperWA: MUST

optimade.md
```diff
@@ -446,6 +457,36 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+  * _cursor-based pagination_: using `page_cursor` and `page_limit` is RECOMMENDED.
+  * _page-based pagination_: using `page_number` and `page_limit` is RECOMMENDED (`page_limit` is equivalent to page "size").
+  * _value-based pagination_: using `page_above`/`page_below` and `page_limit` is RECOMMENDED.
+  
+Example (skip 50 structures and fetch up to 100): <http://example.com/optimade/v0.9/structures?page_offset=50&page_limit=100>
+
+* **sort**: If supporting sortable queries, an implementation MUST use the `sort` query parameter with format as
+  specified by [JSON API 1.0](https://jsonapi.org/format/1.0/#fetching-sorting).
+  
+  It is not required that an implementation supports multiple sort fields for a single query. However, if it does, it
```
CasperWA: NOT REQUIRED. I think? Otherwise, maybe there are other terms that could be used from the definitions in the beginning of the spec?

optimade.md
```diff
@@ -446,6 +457,36 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+  again must conform to the JSON API 1.0 spec.
+  
+  If an implementation supports sorting for an [entry listing endpoint](#h.4.4.2), then the `/<entries>/info` endpoint
+  MUST include, for each field name `<fieldname>` in its "data.properties.`<fieldname>`" response value,
+  the key "sortable" with value `true`. This is in addition to each property description (and optional unit).
```
CasperWA: I can see you have an example later under the info endpoint, however, could you make a very small one here as well? Or perhaps just write an addendum that `"An example is shown in section [4.4.2 title](<link>)"`

optimade.md
```diff
@@ -275,17 +275,28 @@ or a _list_ of JSON API resource objects. Every resource object needs the `type`
...
+  * **next**: represents a link to fetch the next set of results. When the current response is the last page of data,
+    this field MUST be either omitted or `null`-valued.
+   
+  The following fields are reserved for pagination. An implementation MAY offer these links:
+  
+  * **prev**: the previous page of data. `null` or omitted when the current response is the first page of data.
+  * **last**: the last page of data.
+  * **first**: the first page of data.
 
   * **base\_url**: a links object representing the base URL of the implementation. Example:
```
giovannipizzi: Could you please move `base_url` out of the region speaking about pagination?

optimade.md
giovannipizzi: Also, could you make it clear that this is an example of something OPTIONAL?

giovannipizzi:
After discussion with @rartino and @CasperWA, after the minor changes suggested in the edit, we are going to approve this

optimade.md
```diff
@@ -324,17 +324,30 @@ or a _list_ of JSON API resource objects. Every resource object needs the `type`
...
-* **links**: a [JSON API links member](http://jsonapi.org/format/1.0/#document-links)
-  containing the JSON API links objects:
-  * **next**: is an URI that represents a suggested way to fetch the
-    next set of results if not all were returned, either directly as a string,
-    or as a link object. The field MUST be null or omitted if there is no additional
-    data, or if there is no way to fetch additional data. The link object can contain
-    the following members:
-    * **href**: a string containing the link’s URL.
-    * **meta**: a meta object containing non-standard meta-information about the link.
+The response MAY OPTIONALLY also return resources related to the primary data in the field.
```
CasperWA: Colon at the end of this sentence, or a revision, otherwise it doesn't make sense, I think?

optimade.md
```diff
@@ -324,17 +324,30 @@ or a _list_ of JSON API resource objects. Every resource object needs the `type`
...
+
+* **links**: [JSON API links](http://jsonapi.org/format/#document-links) is MANDATORY for implementing pagination.
+(see section [URL Query Parameters `page_*`](#h.4.1.1)) Each field of a links object, i.e. a "link", must be either
+  
+  * `null`
+  * a string representing a URI, or
+  * a dictionary ("link object") with fields
+      * **href**: a string representing a URI
```
CasperWA: indent of 2, not 4.

rartino:
Looks as if this is only waiting for fixes to @CasperWA's small issues. I'l set back the 'minor fixes' label.

optimade.md
```diff
@@ -326,15 +326,28 @@ need to be in a dictionary corresponding to the `attributes` field.
...
-* **links**: a [JSON API links member](http://jsonapi.org/format/1.0/#document-links)
-  containing the JSON API links objects:
-  * **next**: is an URI that represents a suggested way to fetch the
-    next set of results if not all were returned, either directly as a string,
-    or as a link object. The field MUST be null or omitted if there is no additional
-    data, or if there is no way to fetch additional data. The link object can contain
-    the following members:
-    * **href**: a string containing the link’s URL.
-    * **meta**: a meta object containing non-standard meta-information about the link.
+* **links**: [JSON API links](http://jsonapi.org/format/#document-links) is MANDATORY for implementing pagination.
```
CasperWA: Please add back `1.0` in the link

optimade.md
```diff
@@ -326,15 +326,28 @@ need to be in a dictionary corresponding to the `attributes` field.
...
+(see section [URL Query Parameters `page_*`](#h.4.1.1)) Each field of a links object, i.e. a "link", must be either
```
CasperWA: Please add the section number to this link.

optimade.md
```diff
@@ -326,15 +326,28 @@ need to be in a dictionary corresponding to the `attributes` field.
...
+  The following fields are reserved for pagination.  Their values are as with `next`, in the sense that they
+  should be a "link". An implementation MAY offer these links:
+  
+  * **prev**: the previous page of data. `null` or omitted when the current response is the first page of data.
+  * **last**: the last page of data.
+  * **first**: the first page of data.
+  
+  The following fields unrelated to pagination:
 
   * **base\_url**: a links object representing the base URL of the implementation. Example:
```
CasperWA: Could you move this up before you start mentioning pagination, just before `next`?

CasperWA: And remove the immediately previous line here

optimade.md
dwinston: Done

optimade.md
dwinston: Done

optimade.md
dwinston: Good catch

optimade.md
dwinston: Done. I needed to add a line to separate from the "link structure" bullet points. I hope it's okay.

dwinston:
@CasperWA what do you think? 🙏 😄

_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-midnight
```
rartino: It all looks good to go to me, but what is up with this _config.yml with jekyll-theme-midnight? Is someone pushing to the wrong repository? (i.e., this should go to the webpage?)

dwinston: @rartino I had mistakenly merged the `Materials-Consortia:master` branch into `dwinston:sorting-pagination` to ensure consistency with upstream changes. The `master` branch includes this `_config.yml` file. I removed it now via `git rm` and ensured consistency with `Materials-Consortia:develop` instead. This should be good to go now. Please re-review.

optimade.md
```diff
@@ -328,32 +328,45 @@ need to be in a dictionary corresponding to the `attributes` field.
...
-            "_exmpl_db_version": "3.2.1"
-          }
-        }
-        // ...
-      }
-      // ...
-    }
-    ```
+	
+	```json
```
rartino: Sorry for missing this the previous round, but this really should be jsonc for json with comments
```suggestion
	```jsonc
```

dwinston:
Okay @rartino should be good to go with your `json`->`jsonc` change. Please approve.

Issue #101, opened 2019-06-12
Allow web-browsers to issue cross-domain requests


blokhin:
This only concerns JavaScript requests done from the web-browsers via **XMLHttpRequest** and **ActiveXObject**. By default the web-browsers are not allowed to fetch content from an OPTiMaDe server, living on an external domain, unless such header is included into the server's response:

`Access-Control-Allow-Origin: *`

May I kindly ask you to include this header into your server implementation responses.

dwinston:
@blokhin please start a PR to add this as a SHOULD or so to the `optimade.md` spec?

rartino:
Are there security concerns we should worry about with this? The only thing I immediately can see is on the client--- if you are not careful with how you render data fetched via OPTIMaDe, the owner of the database may pull off some cross-site-scripting. But that, to me, is on the client.

Otherwise, I think this is commonly restricted to block man-in-the-middle-type attacks, but I do not yet see how that applies here. I'll think a bit more on this. But I think SHOULD is the right level for it.

sauliusg:
The cross-site scripting is not so much an issue here, AFIU, but setting `Access-Control-Allow-Origin: *` opens slight possibility for subverting database data. If the scenario described in [Quora](https://www.quora.com/What-is-same-origin-policy-Why-is-it-required-How-is-it-a-security-vulnerability?share=1)

sauliusg:
I do not like the `Access-Control-Allow-Origin: *`, considering it a bad design and possible security thread for our servers. Also, it seems to be not necessary: already without the ACAO header, 

> the same-origin policy allows inter-origin HTTP requests with GET and POST methods

[W3C Same Origin Policy](https://www.w3.org/Security/wiki/Same_Origin_Policy), and `GET` is the only request currently relevant for OPTiMaDe.

I therefore suggest either not having any mention of `Access-Control-Allow-Origin` in the OPTiMaDe spec, or at least changing 'SHOULD' to 'MAY'.

blokhin:
Please, see https://github.com/Materials-Consortia/OPTiMaDe/pull/105#issuecomment-502489546

PR #105, opened 2019-06-13
Allow web-browsers to issue cross-domain requests

blokhin:
See https://github.com/Materials-Consortia/OPTiMaDe/issues/101


optimade.md
```diff
@@ -395,6 +396,10 @@ the next course of action SHOULD be to fetch the resource objects under the
 to the corresponding database ID that was originally queried, using the object's
 `base_url` value.
 
+### <a name="h.3.3.4">3.3.4. HTTP Response Headers</a>
+
+The server's response SHOULD include the cross-origin resource sharing header `Access-Control-Allow-Origin: *`. This allows browser-based JavaScript applications (running at the different domains) to access the responses.
```
giovannipizzi:
```suggestion
The server's response SHOULD include the cross-origin resource sharing header `Access-Control-Allow-Origin: *`. This allows browser-based JavaScript applications (running at a different domain) to access the responses.
```

ltalirz:
Let me just try to provide some context here since @sauliusg asked:
One reason to prevent [cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS) is to make it harder for a rogue site A to try and impersonate a site B, e.g. in order to steal users' login credentials of site B.
This mainly applies to AJAX requests, i.e. javascript running in the user's browser, requesting a resource from site B, while browsing on site A.
This should be taken into consideration - but if your OPTiMaDe API server is open and without access restrictions I see no real downside in terms of security for allowing CORS.

Since the CORS policy is enforced by the browser, it is true that - in principle - users can turn it off. In practice, however, this can be difficult. For chrome e.g. you need to [go on the command line](https://stackoverflow.com/questions/3102819/disable-same-origin-policy-in-chrome) or [install an extension](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf?hl=en).

I'm not a security expert, so happy to be corrected.

rartino:
I also dug deeper into online security discussion about this header and came to the same conclusion as @ltalirz. Based on this I think the requirement should be phrased: 
```
If the API implementation is served without use of access restrictions, the server's response SHOULD include the cross-origin resource sharing header `Access-Control-Allow-Origin: *`. This allows browser-based JavaScript applications (running at a different domain) to access the responses.
```

optimade.md
```diff
@@ -395,6 +396,10 @@ the next course of action SHOULD be to fetch the resource objects under the
 to the corresponding database ID that was originally queried, using the object's
 `base_url` value.
 
+### <a name="h.3.3.4">3.3.4. HTTP Response Headers</a>
+
+The server's response SHOULD include the cross-origin resource sharing header `Access-Control-Allow-Origin: *`. This allows browser-based JavaScript applications (running at a different domain) to access the responses.
```
```suggestion
If the API implementation is served without use of access restrictions, the server's response SHOULD include the cross-origin resource sharing header `Access-Control-Allow-Origin: *`. This allows browser-based JavaScript applications (running at a different domain) to access the responses.
```

sauliusg: @ltalirz 
> This should be taken into consideration - but if your OPTiMaDe API server is open and without access restrictions I see no real security downside in allowing CORS. The whole point of an API like OPTiMaDe is that people can use it from elsewhere.

I see a potential problem in future when Optimade n.x (n > 1) will mandate PUT, and a user will get credentials for updating a database via Optimade, then go to a rough site and get a malicous script putting something into a database without the user's knowledge.

rartino: Can't we change the formulation at that point?

sauliusg:
It seems that the proposed `ACAO` header is not necessary for the application intended, as "the same-origin policy allows inter-origin HTTP requests with GET and POST methods"([Same Origin Policy](https://www.w3.org/Security/wiki/Same_Origin_Policy)). Besides that:

- The Same-origin policy is there for a reason. If it servers any purpose at all, why should site administrators start reverting it to their own peril?

- The [W3C](https://www.w3.org/2017/08/16-webappsec-minutes.html#item03) obsoleted CORS; it is now apparently maintained by [WHATWG](https://en.wikipedia.org/wiki/WHATWG) but it is not clear how good the proposals/standards of this group is (CORS definitly seems sub-par to me, for the reasons mentioned below);

- If someone want to implement fetching open data from a site, they should do the work themselves, and not ask admins of the site A to do work for them, especially if the changes that are requested are potentially security-weakening. In particular:

-- even of Cross-origin GET requests are not allowed, the designer of the JS program at site B that wants to use async queries should provide all necessary request relays from their site of the origin; or host the application on the database site – and not ask admins of the site A to do extra work for them;

-- even if designers of the JS program at the site B are plain lazy and do not want to implement request relays, there are sites that relay requests for them (e.g. [crossorigin.me](http://corsproxy.github.io/); thanks to @fawzi for pointing it out to me;

-- If data are open (like from the COD), the site B may cache the data locally for their application.

- Ideally, web pages should provide the main functionality without JS (e.g. with JS switched off); if JS is off then CORS is not a problem :D.

- As pointed out, CORS policies are implemented by browsers and can be bypassed; going to command line does **not** make it difficult :)

- For mandating http headers, OPTiMaDe spec does not seem a proper place, Optimade works on the JSON payload level (so we will mix OSIl levels in our spec if we mandate HTTP headers);
 
Given all this I suggest that we either skip the `Access-Control-Allow-Origin: *` specification altogether, or, if people feel in strong need of having its compatibility level chained from _SHOULD_ to _MAY_.

optimade.md
sauliusg: Of course you do not want CORS to be on if you can PUT. But it is easy to overlook that it was witched ON some N years if we follow the SHOULD recommendation. This is why I suggest using MAY instead of SHOULD (and our site will **not** switch it on, as currently planned)

rartino: Could we use?
```
Access-Control-Allow-Methods: POST, GET
``` 
But I see confusing documentation about this header-- it might only apply in "pre-flight", so I'm not sure if that means it may sometime not apply even if we include it in every response.

blokhin:
There's another side of the coin: saying **MAY** instead of **SHOULD** we discourage any browser-only (pure JavaScript, server-less) implementation. A systematic "Sky-scanner" approach then becomes possible only at the server side.

In general, any third-party thin client or web-based visualization tool (wishing to enjoy the OPTiMaDe specs) should negotiate the access in an ad-hoc manner.

And the security concerns that may be brought by this header are also perfectly valid without this header, so let's not "scapegoat" it.

Colleagues, I urge you, let's make **SHOULD**, having in mind to encourage community for creation of the new web-interfaces to your data.

fawzi:
I agree with @blokhin that SHOULD is a good choice for Optimade, but @sauliusg seems difficult to convince, and this can be changed in the next release.

rartino:
I also like to be able to do things in pure JS, and agree with SHOULD with my reformulation. Based on the above discussion it may even be further clarified as:
```
If the API implementation intend to serve GET and POST requests completely openly, i.e., without any use of access restrictions, the server's response for these methods SHOULD include the cross-origin resource sharing header `Access-Control-Allow-Origin: *`. This allows browser-based JavaScript applications running at a different domain to access the responses. For other methods (e.g, PUT, DELETE) the server SHOULD NOT return any `Access-Control-Allow-Origin` header. If the server supports the OPTIONS method, it SHOULD communicate the same thing, i.e., by setting `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Methods: POST, GET`.
```

But I'm also ok with doing a MAY now and a separate PR to raise it to SHOULD. I think the security-minded clarification above is still relevant, but I don't feel as strongly about it as if we put SHOULD.
 
> And the security concerns that may be brought by this header are also perfectly valid without this header, so let's not "scapegoat" it.

This, on the other hand, is not correct. It may at first seem as this has no security consequences since the header is "just a header" that the client can ignore, but that is a misunderstanding. Someone who sets up a private OPTiMaDe server on their local network protected by a firewall (or NAT), and then browses the web with their usual browser will be vulnerable to data being stolen out of that private server by an external site if this header is set. This vulnerability would not be present without the header.

blokhin:
Again, an attack vector mentioned by @rartino is generally valid, and an elevated CORS is just one way (from many others) to bypass firewall or NAT.

Besides, any scepticism towards my CORS always starts with a presupposition of "private OPTiMaDe". However, OPTiMaDe now refuses of any recommendations on access control. And, at least for reading, "O" stands for "open". Is there an inconsistency?

sauliusg:
I suggest we stop trying to convince each other since this wastes our time in a rather unconstructive way :/
All arguments have been presented by all sides; I think what can be distilled from the discussion is the following: some people want CORS because they want to implement OPTiMaDe clients in one particular way; other's do not want CORS or do no care. Security implications of CORS are muddy, but in any case the server security should not depend on someone's browser honouring or not honouring CORS requests.

What strikes me is that regardless of what is written in the spec, "MAY" or "SHOULD" (short of "MUST" :/), some sites will choose to implement CORS (possibly for reasons unrelated to the Optimade spec.), and others will choose not to implement it (again for reasons not related to the Optimade); all implementations will be compliant. It follows that having the `Access-Control-Allow-Origin: *` specification in the Optimage has no consequences to the implementation. So my questions is does it belong to the specification at all? Technical specs are not the place to promote or advertise your beloved technologies, they should be limited to what is essential for interoperability.

So my preference would be that `Access-Control-Allow-Origin: *` and any other HTTP headers that are not essential for exchanging data are discussed and agreed upon elsewhere. But if you guys think that OPTiMaDe should have `Access-Control-Allow-Origin: *` statement – go ahead. I will not try to block it in any way (actually I even can't ;); but neither will I support it.

rartino:
> [Technical specs] should be limited to what is essential for interoperability. [...] Access-Control-Allow-Origin: * [...] [is] not essential for exchanging data.

Isn't what we are discussing here exactly an interoperability issue? The interoperability between pure JS clients running in popular browsers and OPTiMaDe servers?

For my use case, I'd like to implement that when someone uses my web UI to search OMDB it also shows hits from some other selected databases (clearly marked as such), e.g., materialsproject, materialscloud, and COD, with deep links to their entries. That can be achieved with a far cleaner implementation if I can do it on the client without relaying. With the direction modern web app development is taking, we are likely to see more OPTiMaDe-enhanced websites and database interoperability if we push for API implementations to support pure JS clients.

> regardless of what is written in the spec, "MAY" or "SHOULD" [...] some sites will choose to implement CORS [...] and others will choose not to implement it [...] It follows that [it] has no consequences to the implementation. 

The consequence is that SHOULD will mean almost all API implementations will be interoperable with pure JS clients, MAY will mean almost none will be. 

> But if you guys think that OPTiMaDe should have Access-Control-Allow-Origin: * statement – go ahead. I will not try to block it in any way (actually I even can't ;); but neither will I support it.

I'd like to think that this isn't how OPTiMaDe development works. When someone raises a strong objection, we don't just take a majority vote, we listen and we discuss to find common ground. I think that approach is necessary to reach a common API for interoperability that actually gets implemented across databases.

dwinston:
I am swayed by @sauliusg saying in effect that mention of the `Access-Control-Allow-Origin` response header would be better suited to companion documentation, i.e. what can help folks implement OPTiMaDe, rather than in the specification. This header would certainly be mentioned in the documentation for "Skyscanner" (the tool suggested to survey different live OPTiMaDe servers).

rartino:
Trying to think in more concrete terms about a 'separate document' as suggested by @sauliusg and @dwinston that means adopting a statement that is not about standard conformity, but just clarifies when and how to use ACAO. Perhaps such a statement is indeed what we need. I.e., something like this:
```
There are relevant use-cases for allowing data served via OPTiMaDe to be 
accessed from in-browser javascript, e.g., to enable web-based search 
engines that aggregate data, and for enhancing web pages unaffiliated 
by the provider with data served via OPTiMaDe. For this to be possible, 
the access restrictions for code running this way needs to be considered. 
For most browsers, such use requires the server to include an 
`Access-Control-Allow-Origin: *` header in its responses. 
```
This is essentially meant to say: if you want your results used in these ways, you should include the header. 

If we can agree on something like this, I suppose it could go in a separate document, but perhaps it just as well could be included in the spec?

Jsonapi does already have a few things to say about headers, so it isn't as the spec presently does not at all deal with headers.

fawzi:
I find @rartino proposal very good, I would slightly reword it to
 
For most browsers, such use requires the server to include the header `Access-Control-Allow-Origin: <site-of-the-javascript-page>` or `Access-Control-Allow-Origin: *`(to allow any site) in its responses.

And I would put it in the spec. As info without AY/SHOULD/MUST it *should* ;) be acceptable

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

sauliusg:
> ... adopting a statement that is not about standard conformity, but just clarifies when and how to use ACAO. Perhaps such a statement is indeed what we need. I.e., something like this:
> 
> ```
> There are relevant use-cases for allowing data served via OPTiMaDe to be 
> accessed from in-browser javascript, e.g., to enable web-based search 
> engines that aggregate data, and for enhancing web pages unaffiliated 
> by the provider with data served via OPTiMaDe. For this to be possible, 
> the access restrictions for code running this way needs to be considered. 
> For most browsers, such use requires the server to include an 
> `Access-Control-Allow-Origin: *` header in its responses. 
> ```
> 
> This is essentially meant to say: if you want your results used in these ways, you should include the header.

I like this wording. I definitely think about the issue this way, and I would endorse @rartino suggestion to include ACAO suggestion as an implementation hint (for currently used Web methods), instead of making this a conformity staement.

sauliusg:
@rartino : 
Will you make a change request, or should I do that? The text could read:
> There are relevant use-cases for allowing data served via OPTiMaDe to be 
accessed from in-browser javascript, e.g., to enable web-browser based search 
engines that aggregate data, and for enhancing web pages unaffiliated 
by the provider with data served via OPTiMaDe. For this to be possible, 
the access restrictions for code running this way needs to be considered. 
For most browsers, such use requires the server to include an 
`Access-Control-Allow-Origin: *` header in its responses. 

(Same as you text with the change I suggested:  `web-based` -> `web-browser based`)

rartino:
I'm waiting for input from @blokhin on the proposed version.

rartino:
No feedback from @blokhin? In that case I'll submit a new PR in which I'll put the new formulation and suggest we close this one. (Hopefully I have time within the next few days.)

blokhin:
Hmmm, the diff is too big. In short, I shifted two new sections after this one (also in ToC), and tried to find a balance between all the suggested edits, also being terse:
```
There are relevant use-cases for allowing data served via OPTiMaDe to be accessed from in-browser JavaScript, e.g. to enable the server-less data aggregations. Such use requires the server to include the header `Access-Control-Allow-Origin: *` in all its responses. Without this header, the browser-based JavaScript applications (running at the different domains) cannot access the OPTiMaDe responses.
```

blokhin:
@rartino should I close this?

Issue #102, opened 2019-06-13
Data(base) licenses


merkys:
OPTiMaDe responses should contain the data license indicators. The implementation details depend on the scope of data licensing we want to use:
1. Per-database license. This means that one license covers all the data in a database. Example: [Wikipedia](https://en.wikipedia.org), which uses CC-BY-SA for all its content.
2. Per-entry license. This allows data of heterogeneous licenses coexist in a database without breaching licensing requirements. Example: [Wikimedia Commons](https://commons.wikimedia.org), which specifies licenses per-entry.

Which option we choose depends on the nature of databases that will use OPTiMaDe. For instance, COD and TCOD contain only public-domain data, so option 1) would be sufficient. What do others think?

<s>To specify licenses in a standard way I suggest using license names (abbreviations) from [SPDX list of commonly used licenses](https://spdx.org/licenses/), if we deem it exhaustive enough.</s>

**Edit**: SPDX list has only free licenses, and misses `public-domain`, so not exhaustive enough.

giovannipizzi:
If we are not sure, maybe we should postpone this after 1.0

rartino:
@merkys, @giovannipizzi This is marked milestone v1.0. Are we ok to push this to v1.1? I say 'yes'. Right now, if you want to *redistribute* data you've obtained via OPTIMADE you will have to check with the originator database for license information (their website, email, ...). It would be great to be more helpful than this, but IMO not crucial for 1.0.

merkys:
Yes, I agree with @giovannipizzi and @rartino that we can solve this after v1.0.

merkys:
I am revisiting this issue after looking at #364, where licensing of archival data is discussed.

I would like to build on top of my initial proposal, accommodating [subsequent @rartino's comment](https://github.com/Materials-Consortia/OPTIMADE/issues/102#issuecomment-643835996). Thus:

* Base Info Endpoint could have `attributes.license` field providing a license for *all* its data;
* Single Entry Endpoint could have `attributes.license` field providing a license for this particular entry, overriding the license at Base Info Endpoint;
* If none of the above are present, then provider's website is to be consulted for license information.

[...]

PR #107, opened 2019-06-13
[WIP] Introducing 'license' field for entries

merkys:
Fixes (partially) #102 by introducing per-entry field to specify licensing details.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1332,7 +1332,8 @@ This section defines standard entry types and their properties.
 ### <a name="h.6.1.4">6.1.4. license</a>
 
 * **Description**: Dictionary describing the license of the entry.
-* **Requirements/Conventions**: MUST contain one or both of the following fields:
+* **Requirements/Conventions**: MUST be a dictionary containing one or both of
+  the following fields:
   * **type**: A short string identifying the license. For free/open licenses the
     abbreviations defined in [SPDX list of commonly used licenses](https://spdx.org/licenses/)
     are RECOMMENDED.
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -1332,6 +1333,25 @@ This section defines standard entry types and their properties.
   * \_exmpl\_trajectory
   * \_exmpl\_workflow\_id
 
+### <a name="h.6.1.4">6.1.4. license</a>
+
+* **Description**: Dictionary describing the license of the entry.
+* **Requirements/Conventions**: MUST contain one or both of the following fields:
```
CasperWA: (MUST) be a dictionary containing (~contain~ one or both ...)

CasperWA: Perhaps state whether this is a mandatory field or optional?

giovannipizzi:
Marked as WIP because we are having some discussions.
It's not every single structure that is under license, but the license applies to the DB, i.e., to the work that puts together many structures bringing additional knowledge. The single structure, even if it comes from a proprietary DB, is still under the license of the original paper.

optimade.md
merkys: Fixed.

merkys: This seems to be a subject of discussion. Nevertheless, the requirement level might be added via #2.

rartino:
This seems quite dangerous. 

If we allow every single data entry in a database to carry a license of its own, then---separate from the consideration of whether it really is possible for entries to be individually licensed---it would fall onto users to verify every single data entry to see whether further processing breaks its license terms. One may receive a single entry with a more restrictive license hidden among millions of "openly" licensed entries.

Hence, it makes more sense to only allow a single per-base_url license info. That license info can, however, provide a url that explains a more complex licensing situation that applies when, e.g., two databases have been merged.

I do see the issue for a sky-scanner type service. However, such a service can explain in its license field that it returns structures out of multiple databases and that the license depend on the data from which the data is fetched. Then it would add references relations to everything it returns that point to the source. (Which it should be doing anyway.)

rartino:
So, how do we want to go about with this one? Should it be closed as "we won't do this". Or is @merkys willing to re-edit it to instead allow a license on the database as a whole?

My wishlist for a database-wide license field:
- Some standard way of saying "its complicated" and a URL. 
- Some standard way of saying "all our data may be further distributed via downstream OPTiMaDe APIs." to allow a skyscanner-type implementation to automatically detect whether it is allowed to redistribute data.

merkys:
> So, how do we want to go about with this one? Should it be closed as "we won't do this". Or is @merkys willing to re-edit it to instead allow a license on the database as a whole?

I am closing this PR. Per-database license is a different thing. I will open another PR once it's ready.

PR #115, opened 2019-06-14
Removing all references to the entrypoint `/all`

### LINKED ISSUE #21: Correct how the all endpoint is specified (or remove it completely from the specs)
What does the /all endpoint presently return?

Without a filter it should return all entries of all entry types, using the jsonapi 'type' member to specify which entry_type each record is.

So, e.g.,
```
   /all?filter=id="_exmpl_42"
```
Looks for any entry in the database with the id _exmpl_42, regardless of which type it is.

giovannipizzi:
Fixes #21

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -21,18 +21,15 @@
 &nbsp;&nbsp;&nbsp;&nbsp;[4.2. Single Entry Endpoints](#h.4.2)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[4.2.1. URL Query Parameters](#h.4.2.1)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[4.2.2. JSON API Response Schema](#h.4.2.2)  
-&nbsp;&nbsp;&nbsp;&nbsp;[4.3. General Entry Listing All Endpoint](#h.4.3)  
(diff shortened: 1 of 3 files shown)


optimade.md
```diff
@@ -990,16 +961,15 @@ of the provider's other OPTiMaDe databases.
...
 
-## <a name="h.4.6">4.6. Custom Extension Endpoints</a>
+## <a name="h.4.5">4.5. Custom Extension Endpoints</a>
 
 API implementors can provide custom endpoints under the Extensions endpoint.
 They should have the form "&lt;base\_url&gt;/extensions/&lt;custom paths&gt;".
 
 # <a name="h.5">5. API Filtering Format Specification</a>
 
 An OPTIMaDe filter expression is passed in the parameter `filter`
```
CasperWA: Randomly spotted that OPTiMaDe is written wrongly here - although it has nothing to do directly with this PR, you should be okay to edit it :)

optimade.md
```diff
@@ -990,16 +961,15 @@ of the provider's other OPTiMaDe databases.
...
-either as an URL query parameter as specified by jsonapi, or as part
-of a POST request as described in [4.3. General entry listing 'All' endpoint](#h.4.3).
+either as an URL query parameter as specified by JSON API.
```
CasperWA: Remove "either"

CasperWA: And perhaps make JSON API a link to: https://jsonapi.org/format/1.0/#fetching-filtering

giovannipizzi: Done, can be re-reviewed now

PR #120, opened 2019-06-14
Making a pair of intended RFC2119 keywords uppercase

merkys:
There were a couple of improperly typeset RFC 2119 keywords (`*optional*` instead of `OPTIONAL`).

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -183,7 +183,7 @@ request. If no specific response format is specified in the request by
 use of the `response_format` URL query parameter (see below), the default
 response format is [JSON API v1.0](http://jsonapi.org/format/1.0) specification.
 All endpoints MUST support at least the JSON API format.
-Each endpoint MAY OPTIONALLY support multiple formats,
+Each endpoint MAY support multiple formats,
 and declare these formats in their `info` endpoints
 (see section [4.3.2. Entry Listing Info Endpoints](#h.4.3.2)).
 
@@ -224,7 +224,7 @@ Every response SHOULD contain the following fields, and MUST contain at least on
     * **description**: a longer description of the database provider.
     * **prefix**: database-provider-specific prefix as found in [Appendix 1](#h.app1).
 
-    `provider` MAY OPTIONALLY include these fields:
+    `provider` MAY include these fields:
 
     * **homepage**: a [JSON API links object](http://jsonapi.org/format/1.0/#document-links),
     pointing to the homepage of the database provider, either directly as a string,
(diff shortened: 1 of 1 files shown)


merkys:
> [ ] section 3.3.4. Unset optional properties
> 
>     * All cases of optional here should be OPTIONAL, even the title, I think?

Hmm... I'm not sure. I feel that *optional property* is a term definition here.

merkys:
>     * [ ]  section 3.3.4. Unset optional properties
>       
>       * All cases of optional here should be OPTIONAL, even the title, I think?

Not sure whether the *optional propery* is not a term by itself here.

>     * [ ]  section 4.4. Links endpoint
>       
>       * There is one sentence that starts like "Alternatively, it MAY optionally ..." Should be OPTIONALLY.

I think "MAY" already specifies the level of requirement here. Probably "optionally" might be removed without changing the meaning.

>     * [ ]   section 5.2. The Filter Language Syntax
>       
>       * Found under the sub-section Comparisons of multi-valued properties, for a sentence starting: "The construct is optional ..." Should be OPTIONAL.

Fixed.

>       * Also under the Optional filter features (last sub-section). Several cases, including the title, where it should be OPTIONAL.

Fixed.

>     * [ ]  Appendix 2: The Filter Language EBNF Grammar.
>       
>       * Simply remove "optional" from `ValueZip`, making the comment similar to other comments here.

Removed.

CasperWA:
> * [ ]  section 3.3.4. Unset optional properties
>   
>   * All cases of optional here should be OPTIONAL, even the title, I think?
>>
>> Not sure whether the *optional propery* is not a term by itself here.

Not sure either. I am willing to leave this out now, but I don't know when it would be considered again in the future, if at all, i.e., we should probably figure it out now. Calling @ctoher to answer this :)

> * [ ]  section 4.4. Links endpoint
>   
>   * There is one sentence that starts like "Alternatively, it MAY optionally ..." Should be OPTIONALLY.
>>
>> I think "MAY" already specifies the level of requirement here. Probably "optionally" might be removed without changing the meaning.

Several places (if I remember correctly) we have MAY OPTIONALLY. But I am okay with removing optionally/OPTIONALLY completely here.

CasperWA:
> (I have a feeling that after the workshop one or more of us will have to proofread the compound document anyway, so it seems to me that we can try to catch the rest at that point, if there are any.)

Good point. With this I only request a stance from @merkys on the second remaining change suggestion (section 4.4. MAY OPTIONALLY, versus simply MAY).

merkys:
> Several places (if I remember correctly) we have MAY OPTIONALLY. But I am okay with removing optionally/OPTIONALLY completely here.

Converted all "MAY OPTIONALLY" -> "MAY".

PR #122, opened 2019-06-14
Small spelling and grammar fixes

sauliusg:
Ran spelling checker and switched to US spelling.


.words.lst
```diff
@@ -0,0 +1,93 @@
+personal_ws-1.1 en 92
```
giovannipizzi: Can we limit this list of words to the only ones we actually have in the main document?

giovannipizzi: Also, is there a way to tell aspell to skip code snippets? It's a bit sub-ideal, I think, to add here words that are put as example of mistakes, or that anyway are not English words but just pieces of code or similar

.words.lst
```diff
@@ -0,0 +1,93 @@
...
+PCRE
+lt
+KnownOpRhs
+LengthComparison
+UppercaseLetter
+nl
+zeo
+denox
+deNOx
+cartesian
```
giovannipizzi: I think this is an error, it should be `Cartesian` (capitalized).
I imagine this is because this appears as a part of a string of a key name - to solve this problem (and avoid having someone mistyped) maybe see my comment about avoiding codes within `single ticks` or
```
triple ticks
```

optimade.md
```diff
@@ -1443,7 +1443,7 @@ the Cartesian x, y, z directions.
 
 ### <a name="h.6.2.6">6.2.6. lattice\_vectors</a>
 
-* **Description**: The three lattice vectors in Cartesian coordinates, in ångström (Å).
+* **Description**: The three lattice vectors in Cartesian coordinates, in Ångström (Å).
```
giovannipizzi: the units name should always be lower case, see e.g. https://www.nist.gov/pml/weights-and-measures/writing-metric-units
Then, whether writing this as `ångström` or `angstrom` is a choice - if you want to keep it, maybe change this in the other places in the spec where it's written as `angstrom` (there should be at least one other place)

optimade.md
```diff
@@ -256,7 +256,7 @@ Every response SHOULD contain the following fields, and MUST contain at least on
       email address.
   * **warnings**: a list of warning resource objects representing non-critical errors or warnings.  
     A warning resource object is defined similarly to a [JSON API error object](http://jsonapi.org/format/1.0/#error-objects), but MUST also include the field `type`, which MUST have the value `"warning"`.
-    The field `detail` MUST be present and SHOULD contain a non-critical message, e.g., reporting unrecognised search attributes or deprecated features.  
+    The field `detail` MUST be present and SHOULD contain a non-critical message, e.g., reporting unrecognized search attributes or deprecated features.
```
CasperWA: Is this the US spelling? I thought this was the UK spelling.

optimade.md
```diff
@@ -1722,11 +1722,11 @@ API implementations SHOULD NOT make up and use new prefixes not included in this
 but SHOULD rather work to get such prefixes included in a future revision of this API specification.
 
 **Example**:  
-Database-provder-specific prefix: `"exmpl"`  
+Database-provider-specific prefix: `"exmpl"`  
 Use as a field name in a response: `_exmpl_custom_field`
 
 The initial underscore indicates an identifier that is under a separate namespace under the ownership
-of that organisation. Identifiers prefixed with underscores will not be used for standardized names.
+of that organization. Identifiers prefixed with underscores will not be used for standardized names.
```
CasperWA: Again, I thought the "z" was the UK way, and the more "simple" "s" was the US way?

optimade.md
```diff
@@ -1918,7 +1918,7 @@ Expression` definition as well, and the remaining grammar rules could
 then be used as a parser generator (like yacc, bison, antlr) input.
 
 ## <a name="h.app3">Appendix 3. Regular Expressions for OPTiMaDe Filter Tokens.</a>
-The string below contains Perl-Compatible Regular Expressions to recognise
+The string below contains Perl-Compatible Regular Expressions to recognize
```
CasperWA: see above.

optimade.md
giovannipizzi: Yes, it is: https://global.oup.com/academic/authors/author-guidelines/house-style/?lang=en&cc=ch#ize

optimade.md
merkys: No, on contrary :)

CasperWA: Right! Thanks

.words.lst
sauliusg: I'm not sure if I understand the problem, but if a word ended up in the .words.lst list but was removed in the later revision of the document, it is extra work to remove it from the .words.lst as well, and I think this is not necessary. The word list will not prevent `aspell` from checking the spelling as intended, and the word may be inserted back at some stage :)

sauliusg: I am not aware if `aspell` can skip text in specific contexts; AFAIK it only has options for certain input types (TeX, HTML, etc.). The easiest way to avoid having them flagged on each run of `aspell` is to put them in a private word list. BTW the suggested word list is intended only for this particular document, 'optimade.md', so it will **not** contain words from other unrelated spell checks.

.words.lst
sauliusg: Indeed, `cartesian` in small letters is included because it was in a property name.
But we need to check misspelled property names as well :)

optimade.md
sauliusg: Reverted to "ångström".

sauliusg: There are actually two places where "angstrom" is mentioned:
```
saulius@varanas sauliusg-optimade/ $ grep --color=yes -E '(angstrom|angstroem|ångström)' optimade.md 
* **Description**: The three lattice vectors in Cartesian coordinates, in ångström (Å).
    Otherwise, it should be a float value, expressed in angstrom. Note that if at least one
```
I have no preference to any notation; let's decide and chose whatever we like.

sauliusg:
@CasperWA : I would like to have `.words.lst` in the repo since it helps checking spelling and also can be easily converted, using *x tools, to any other word list for other programs.

The wish is to be able to run the following command in the working copy:

`aspell -x -p ./.words.lst -l en_US -c optimade.md`

and to get **no** misspelled word indications.

CasperWA:
> @CasperWA : I would like to have `.words.lst` in the repo since it helps checking spelling and also can be easily converted, using *x tools, to any other word list for other programs.
> 
> The wish is to be able to run the following command in the working copy:
> 
> `aspell -x -p ./.words.lst -l en_US -c optimade.md`
> 
> and to get **no** misspelled word indications.
> 
> To make the `.words.lst` file less jumping into our eyes I started its name with the dot ('.'). We already have several such "hidden" files in the repo.

Excellent. Thank you for the clear explanation. I'm on board! :)

giovannipizzi:
I agree. If we want to include this at this stage, ok.
But I would suggest switching to [pyspelling](https://facelessuser.github.io/pyspelling/) that internally uses `aspell` and is able to convert first to HTML and then skip certain HTML tags.

The following configuration is already very close to a good one:

```yaml
matrix:
- name: markdown
  sources:
  - optimade.md
  aspell:
    lang: en
  dictionary:
    wordlists:
    - .words.lst
  pipeline:
  - pyspelling.filters.markdown:
      markdown_extensions:
        - markdown.extensions.fenced_code
  - pyspelling.filters.html:
      comments: false
      attributes:
        - title
        - alt
      ignores:
        - code
        - pre
```

and then you run with `pyspelling -c spelling_config.yml`

I would suggest restarting then from an empty words list, and 
- either adding words that are not recognised but are supposed to be in the main text
- or putting in fenced areas words that are instead supposed to be in there (e.g. if we say `doi` as a field, we should write it in a fenced area and not as *doi* nor as **doi** or other variants).

With this I think the list of words will be very minimal, and this can be easily added to travis.

rartino:
@giovannipizzi You are still marked as requesting changes. Do you want to approve this, and then you write up the change to pyspelling as a separate issue, or do you want @sauliusg to update this request to do what you said in the last comment?

sauliusg:
> With this I think the list of words will be very minimal, and this can be easily added to travis.

The `aspell` and .words.lst was not intended for Travis, although you could do that as well. The way I do it is usually to either run aspell directly, or have a Make rule so that one can run `make spell`

sauliusg:
> But I would suggest switching to [pyspelling](https://facelessuser.github.io/pyspelling/) that internally uses `aspell` and is able to convert first to HTML and then skip certain HTML tags.

I do not see a merit of adding yet another python wrapper around a working program, IMHO it just adds extra complexity and unwanted dependence.

What I am definitely against is to exclude words from the .words.lst that are keywords. I think wee **must** check keywords for spelling mistakes as well, and a list of keywords that we currently assume correct is definitely a useful one.  We do not want 'cratesian_sites' in the documentation, do we?

We _may_ want to spell-check keywords separately from the main text; but I would not use `pyspell` or anything like this for the purpose: it is too complicated for the task with its _ad-hoc_ yaml configuration, and I suspect will nor be powerful enough for the task. Converting everything to HTML sounds like and overkill in the first place and raises questions about how faithful the conversion will be.

I also feel that the suggested changes of tools would be more complicated than they should. What is the problem with the current `aspell`?

n

rartino:
> What is the problem with the current `aspell`?

Since aspell doesn't understand the context of what it is spell-checking it will fault, e.g., `"id":"txkoijejwo"`. It seems that forcing everyone with a PR to handle this would get annoying. And also, if almost every PR has to add new strings to the end of .words.lst we'll get unnecessary conflicts. I'd rather add to our release workflow that before a tagged release may be made, it MUST have a clean aspell spell-check.

As I said, I'm OK with the present PR. As long as we DON'T add aspell to the travis tests and/or refuse to merge PRs until they've updated the wordlist.

giovannipizzi:
I think this PR can be merged now, but I am still think that using pyspelling is a good idea for the future, and in particular I agree with @rartino's last comment. I would personally try to do a new PR at some point in the future with pyspelling and we will rediscuss there. I instead prefer to accept the risk of having typos in the keywords but not listing all of them including things like `kvac` that exists because it's under "Examples of incorrect property names", or names of authors of example references.

BTW, @sauliusg you cannot really complain about adding a very useful check using a python library - you already added a Java code to check the grammar!!! ;-) [and anyway the contributors are not required to install any of these (even if it's just a pip install), they just have to commit, and the travis tests will run it for them]

Issue #116, opened 2019-06-14
Querying on relationship, dot operator in grammar, change of sites definition


rartino:
After some discussion in the provenance group, we want to do a PR with the following:
- Enable querying on relationships using the identifier syntax `<endpoint name>.id` that should for the purpose of formulating filters be thought of as a list of ids of entries under that endpoint.
- Conceptually the `<endpoint>` object (without the dot) should be thought of as its own "data type": it is a relationship. API Implementations do not need to include that object in the response alongside other properties, but MUST include information to indicate these relationships in some way. For example, in jsonapi output they are included under the `relationships` key in the standard jsonapi way.
- This phrasing opens the possibility that in the future we could take the route to allow, e.g. `references.author.last_name = "John"`, but we do not put that in the specification at this time.
- The `<identifier>.<identifier>.<...>` syntax is generally adopted to make queries on subfields of hierarchical data. Furthermore, if any segment in such a construct is a list, there is an implicit map operation to create a list out of those subelements; and, in the final step the list is always regarded as flattened (this needs to be formulated more precisely in the spec.) (partially addressed by #123)
- **However**, this suggest OPTIMaDe should not standardize on hierarchical data which uses data values as keys, which there also are other reasons to avoid. (#112)
- Hence, we also suggest to redefine `sites` to instead of begin a dictionary with the site name as a key, it should be a list of objects, with `name` as one of the keys alongside `chemical_elements`, etc.

**Edit by** @merkys: fixing typos, referencing #112

merkys:
Overlaps with #96

rartino:
This issue has, to a large degree, been addressed by PR #123 and #112, and is affected by #119, but it still remains when they have been merged to write up as a PR that relationships can be queried using `<type>.id` and explain why that works (`<type>` access a "relationship" and `<type>.id` means a list of ids of related `<type>` entries.) + the implicit map meaning of `list.id`.

PR #123, opened 2019-06-14
Filtering by nested property names and relationships

### LINKED ISSUE #96: Implement filtering by relationships
[JSON API Relationships](https://jsonapi.org/format/#document-resource-object-relationships) will be used by OPTiMaDe at some point (as discussed with @rartino and @muhrin). It would be useful to allow filtering entries by their relationships (as noted at least in #95).

In other words, we'd like to use filter strings to query the relationships. This requires some sort of namespace merge for the attributes and the relationships.

merkys:
Fixes #96. Fixes partially #116. <s>Blocked by #95.</s>


optimade.md
```diff
@@ -1843,7 +1861,7 @@ Operator = ( '<', [ '=' ] | '>', [ '=' ] | '=' | '!', '=' ), [Spaces] ;
 
 (* Identifier syntax *)
 
-Identifier = LowercaseLetter, { LowercaseLetter | Digit }, [Spaces] ;
+Identifier = LowercaseLetter, { LowercaseLetter | Digit }, { '.', LowercaseLetter, { LowercaseLetter | Digit } }, [Spaces] ;
```
rartino: Is this the best way to do this in the grammar. Anyone doing parsing will now manually have to break parsed identifiers on dots. May be better to introduce another layer here.

merkys: True. Pinging @sauliusg for his opinion.

sauliusg: This is workable but the `LowercaseLetter, { LowercaseLetter | Digit }` phrase is now repeated twice. To maintain SPOT (DRY), we may want to write:

```EBNF
IdentifierComponent = LowercaseLetter, { LowercaseLetter | Digit } ;
Identifier = IdentifierComponent, { '.', IdentifierComponent }, [Spaces] ;
```

This means that white-space are **not** allowed around the dots (which is the usual practice and maybe is not so bad to enforce after all).

rartino: Couldn't we do:
```
IdentifierComponent = LowercaseLetter, { LowercaseLetter | Digit } ;
Identifier = IdentifierComponent, { DOT, IdentifierComponent }, [Spaces] ;

DOT = '.', [Spaces] ;
```
I'm a bit unsure if we should do this, though, or if it really makes sense to forbid spaces inside Identifiers.

merkys: > I'm a bit unsure if we should do this, though, or if it really makes sense to forbid spaces inside Identifiers.

But was it ever allowed to have spaces in `Identifier`? I think it unnecessarily complicates the grammar. For instance I've never seen a programming language which would allow spaces inside variable names.

rartino: Ah, but the crux is that we are trying to make the formal grammar and the simplest token-based parsing that throws away whitespace between tokens to do the same thing. If my token-based parser is to parse `Identifier` into sub-production rules, it should be allowed to throw away whitespace between those.

I agree that most programming languages does not allow whitespace inside identifiers. On the other hand, I think we should more look at '.' as an operator for sub-indexing. So, the naming can be improved a bit as well. I also see I did a misstake not adding [spaces] onto IdentifierCompoinent.

So, here is my suggestion:
```EBNF
Property = Identifier, { DOT, Identifier } ;

Identifier = LowercaseLetter, { LowercaseLetter | Digit }, [Spaces];

DOT = '.', [Spaces] ;
```
Where Property goes into the "grammar part" of the grammar, and Identifier and DOT goes into the "token part" of the grammar. And then all occurences in the grammar of `Identifier` are replaced by `Property`

(I'd be fine with a better name for `Property`, it was the best I could come up with.)

sauliusg: I am neural on the point, but I see some merit in @merkys arguments: it is always easier to relax requirements later and allow spaces around dots (a non-breaking change) than to forbid spaces in the next version of we allow them now (a compat-breaking change). So I would suggest:

```EBNF
Property = Identifier, { DOT, Identifier }, [Spaces] ;

Identifier = LowercaseLetter, { LowercaseLetter | Digit };

DOT = '.';
```

And then `Property` stays in the TOKENS section.

Some current languages (I guess all mainstream ones) allow spaces around '.' separators for readability; but in Filter language where filters are expected to be short this will not  be a big issue. On the other hand, as Andrius has commented, we also disallow spaces within identifiers, although Algol60 and Fortran would allow them. So what is "usual" dependes on which programming languages you look at. We can find examples both in favour of allowing spaces in identifiers, and against it.

merkys: OK, I now see that I've missed @rartino's initial point. I agree that it's better to be homogeneous about handling the whitespace, but as @sauliusg said, we can always relax the syntax. Forbidding previously allowed constructs is harder. That all said, I have no strong opinion now.

rartino: @sauliusg Your suggestion means my token-based parser will still parse the whole `Property` into a token and I will have to add code to manually split identifiers on dots. In that case I prefer @merkys initial suggestion because in this case the additional level makes no difference.

To clarify: in my view, the central subjective decision we are discussing here is not "do we allow spaces in properties or not". It is whether someone with a token-based parser should:
- Get the full Property (dots and all) as a terminator that has to be broken up manually **OR**
- Get a sub-parse-tree with Properties broken down as Identifiers + dots.

If we go for the former, then `Property` is a token. If so, @merkys original suggestion is fine. I see no reason to add more levels within the token, as in the suggestion of @sauliusg. 

If we go with the latter, then `Property` is not a token and cannot be in the tokens section. In this case  we have to go with something similar to my suggestion. In this case we must allow whitespace after '.', otherwise people with token-based parsers would have to specifically encode the special case that whitespace can be dropped after any token EXCEPT for `Identifier` and `DOT` inside a property, which completely defeats the nice redesign of the grammar of @sauliusg where such whitespace now always can be dropped.

fawzi: identifiers should allow also “_”:

     IdentifierComponent = LowercaseLetter, { LowercaseLetter | Digit | Underscore } ;
 
or did I miss something?

rartino: @fawzi The way things are defined in the spec right now, LowercaseLetter contains "_".

rartino:
So, present status (in my opinion): we need to come to a consensus on which of the three alternatives for Identifier/Property to go in the discussion to go with. I'd be happy to see others' input, but otherwise I think we are waiting for input from @sauliusg.

optimade.md
sauliusg: @fawzi, @rartino : we could single out Underscore as a separate named token in the grammar, to be more explicit and less surprising.

sauliusg:
> To clarify: in my view, the central subjective decision we are discussing here is not "do we allow spaces in properties or not". It is whether someone with a token-based parser should:

>    Get the full Property (dots and all) as a terminator that has to be broken up manually OR
>   Get a sub-parse-tree with Properties broken down as Identifiers + dots.

Here I am a bit worried about mixing up specification and implementation. Implementations should be free to use whatever parsing method their implementers choose, the only concern being that they accept the same language as specified in the OPTiMaDe grammar.

The specification grammar should specify the language without implying any specific parse method.

Thus, to me is more "whether we allow spaces" than "whether to put Property to the production section or to the token section". The former is specification; the latter is implementation.

But pls. see below...

sauliusg:
We all find it difficult to resolve, but it seems to me that:

- @merkys is fine having spaces around dots, as long as spaces are not allowed inside identifiers;
- @rartino suggested at some point that spaces around would be natural, and would make his parsing life a somewhat easier ;)
- @sauliusg (myself) is neutral on the point, and there seem to be merits of being consistent with the current Filter spec (spaces are allowed everywhere around words) and the common practice of programming languages (spaces are allowed around dots), I suggest taking @rartino grammar:

```EBNF
Property = Identifier, { DOT, Identifier } ;

(* TOKENS: *)

Identifier = LowercaseLetter, { LowercaseLetter | Digit }, [Spaces];

Dot = '.', [Spaces] ;
```
And put `Identifier` and `Dot` into `(* TOKENS *)` section, while `Property` would go into the productions section. Giving it name `Property` indicates that it is higher-level, structured concept, not just a token like `Identifier` or `Dot`. It would also give an intended structure for parsing `Property`, although I admit that sentence structure interpretation are not specified formally.

I would take a capitalised word spelling `Dot` since we use all-caps for operators like `NOT` or `HAS`.

merkys:
> Everything seems fine now, but I suggest renaming the test case:
> 
> `tests/cases/Filter_500.inp` -> `tests/cases/Filter_071.inp`, since we are about to merge and conflicts are unlikely here.

Done

optimade.md
```diff
@@ -2215,7 +2238,7 @@ identifiers, number, and string values as specified in this specification.
 
 ```
 #BEGIN PCRE identifiers
-[a-z_][a-z_0-9]*
+[a-z_][a-z_0-9]*(\.[a-z_][a-z_0-9]*)*
```
rartino: Wait, this is no longer right, right? This should now still be:
```suggestion
[a-z_][a-z_0-9]*
```

optimade.md
```diff
@@ -2232,7 +2255,7 @@ as specified in this specification.
 
 ```
 #BEGIN ERE identifiers
-[a-z_][a-z_0-9]*
+[a-z_][a-z_0-9]*(\.[a-z_][a-z_0-9]*)*
```
rartino: And this should be
```suggestion
[a-z_][a-z_0-9]*
```

optimade.md
merkys: Fixed

sauliusg: To me, looks OK now.

merkys:
Thanks for approving and merging!

Issue #109, opened 2019-06-13
Make sure it's clear about which properties MUST be queryable


giovannipizzi:
Now, a `property` is `anything that can be in the filtering of results`.
Instead it should be what is returned by the API, and instead we should clearly mark which properties MUST be queryable by *any* implementation (the rest does not have to be).

CasperWA:
We have a bit of a conflict now when defining `field` and `property`.
Since `field` is any "key" in the output, anywhere, shouldn't `property` be defined as a subset of `field`? In the sense that it can be any `field` under the top-level `data` field`?

CasperWA:
As an addendum, this also makes `attributes` in and of itself a property. Is this what is intended?

CasperWA:
According to the discussion in issue #12, it now seems that JSON has defined a `field` as the key of a JSON object (a Python dict) and a property is when you have both key and value together.

If we confine properties to only be a valid interpretation for `field`s (and their values) in the top-level `data` `field`, then it will be easy to reference "property fields" to be the only queryable options.

ltalirz:
I would vote for adopting the rather general way the term "property" is used in JSON:

> Objects are the mapping type in JSON. They map “keys” to “values”. In JSON, the “keys” must always be strings. Each of these pairs is conventionally referred to as a “property”.

And add the "queryable" wherever applicable.

coreyoses:
Should there also be a qualifier for "filterable", as a JSON value need not be filterable (e.g., a structure vs. cell volume)?

giovannipizzi:
I think the meaning we intend for "queryable" is exactly what you mean with "filterable" - and indeed since we call it `filter` it might be a good idea to rename the field to "Filterable" and the sections "Filtering"

rartino:
What is the status here: does @giovannipizzi or @CasperWA plan to create a second PR to match #126 to address this issue? Personally I consider this issue blocking for v0.10. We shouldn't make a release were we knowingly are vague about what properties can and cannot be used in filters.

giovannipizzi:
I agree that this should be done as soon as possible. However, sadly I won't have time in the next 1-2 weeks, to draft this...

rartino:
Good to know; unless I see anyone else take this one I will try to make an attempt at it.

CasperWA:
I also do not have the time, unfortunately - at least for this week. However, I can dedicate time in next week to write it up - or just review your PR in detail, @rartino :)

[...]

PR #126, opened 2019-06-14
(Re)define `property`

### LINKED ISSUE #12: conflicting definitions of content of "entries"

CasperWA:
Fixes #12.

This was originally intended to fix issue #109, however, the issue has been split into two PRs, one (re)defining `property` (this one) and one to define which resource property fields are queryable through `filter`.

Newly defined:
* Resource property

Redefined:
* Property
* Field

Extended name of `Property Types` to be `Property Value Types` in accordance with the new definition of `property`.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -102,8 +102,8 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Field**: A key in a JSON object.
-* **Property**: A field and the field's value collectively in a JSON object.
+* **Field**: A string key (similar to a key in a hash table, with the limitation that it MUST be a string).
+* **Property**: A field-value pair.
 * **Property Value Types**:
   * **string**, **integer**, **float**, **boolean**, **null value**: Base data
     types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
--- a/optimade.md
+++ b/optimade.md
@@ -101,10 +101,19 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
+* **Field**: A key in a JSON object.
+* **Property**: A field and the field's value collectively in a JSON object.
+* **Property Value Types**:
+  * **string**, **integer**, **float**, **boolean**, **null value**: Base data
+    types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
+  * **list**, **dictionary**: Collection of base types with the meaning they have in the JSON
+    data-interchange format, i.e., an ordered list of elements
+    (where each element can have a different type) or a hash table
+    (with the limitation that the hash key MUST be a string), respectively.
 * **Resource object**: Represent resources. MUST contain at least the following top-level fields:
   `id`, `type`.
+* **Resource property**: Special property for resource objects.
+  The only kind of property, whose field can be queried (using `filter`, see section [5. API Filtering Format Specification](#h.5)).
 * **ID**: A unique identifier referencing a specific resource in the database.
   Together with **Entry**, the ID MUST uniquely identify the **Resource object**.
   IDs MUST be URL-safe; in particular, they MUST NOT contain commas.
@@ -114,13 +123,6 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
   database that MUST be immutable.
 * **Reserved words**: The list of reserved words in this standard is:
   `info`.
-* **Property Types**:
-  * **string**, **integer**, **float**, **boolean**, **null value**: Base data
-    types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
-  * **list**, **dictionary**: Collection of base types with the meaning they have in the JSON
-    data-interchange format, i.e., an ordered list of elements
-    (where each element can have a different type) or a hash table
-    (with the limitation that the hash key MUST be a string), respectively.
 
 # <a name="h.3">3. General API Requirements and Conventions</a>
 
--- a/optimade.md
+++ b/optimade.md
@@ -102,15 +102,15 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Field**: A string key (similar to a key in a hash table, with the limitation that it MUST be a string).
+* **Field**: A key of an associative-array-type data structure.
+  A field MUST be a string, which MAY exclusively contain lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
 * **Property**: A field-value pair.
-* **Property Value Types**:
+* **Property value types**:
   * **string**, **integer**, **float**, **boolean**, **null value**: Base data
     types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
-  * **list**, **dictionary**: Collection of base types with the meaning they have in the JSON
-    data-interchange format, i.e., an ordered list of elements
-    (where each element can have a different type) or a hash table
-    (with the limitation that the hash key MUST be a string), respectively.
+  * **list**, **dictionary**: Collections of base types, defined in the samme manner as a JSON [array](https://json-schema.org/understanding-json-schema/reference/array.html) and [object](https://json-schema.org/understanding-json-schema/reference/object.html), respectively.
+* **Queryable property**: A property that can be in the filtering of results.
+  For more information see section [3.5. Queryable Properties](#h.3.5).
 * **Resource object**: Represents resources. MUST contain at least the following top-level fields:
   `id`, `type`.
 * **ID**: A unique identifier referencing a specific resource in the database.
@@ -489,10 +489,10 @@ The value for `is_index` MUST be `true`.
 
 ## <a name="h.3.5">3.5. Queryable Properties</a>
 
-Queryable properties are resource properties.
+_Only_ properties of resource objects can be queried.
 It is understood that when one queries a property, one queries on the property's field.
 In other words, one can _only_ query on property fields from resource objects' properties.  
-A query is performed using `filter` see section [5. API Filtering Format Specification](#h.5)).
+A query is performed using `filter` (see section [5. API Filtering Format Specification](#h.5)).
 
 > **For implementers**: To get an understanding of which properties MUST be queryable and which are RECOMMENDED, please see section [6. Entry List](#h.6).
 
--- a/optimade.md
+++ b/optimade.md
@@ -111,7 +111,7 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
     data-interchange format, i.e., an ordered list of elements
     (where each element can have a different type) or a hash table
     (with the limitation that the hash key MUST be a string), respectively.
-* **Resource object**: Represent resources. MUST contain at least the following top-level fields:
+* **Resource object**: Represents resources. MUST contain at least the following top-level fields:
   `id`, `type`.
 * **ID**: A unique identifier referencing a specific resource in the database.
   Together with **Entry**, the ID MUST uniquely identify the **Resource object**.
--- a/optimade.md
+++ b/optimade.md
@@ -14,6 +14,7 @@
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[3.3.4. Unset optional properties](#h.3.3.4)  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[3.3.5. Warnings](#h.3.3.5)  
 &nbsp;&nbsp;&nbsp;&nbsp;[3.4. Index Meta-Database](#h.3.4)  
+&nbsp;&nbsp;&nbsp;&nbsp;[3.5. Queryable Properties](#h.3.5)  
 
 [4. API endpoints](#h.4)  
 &nbsp;&nbsp;&nbsp;&nbsp;[4.1. Entry Listing Endpoints](#h.4.1)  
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -101,10 +101,19 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
+* **Field**: A key in a JSON object.
```
coreyoses: Can we keep the Term Definitions file-type agnostic?
We can remove "in a JSON object" around, and refer to key, value, and hash tables.

optimade.md
```diff
@@ -101,10 +101,19 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
+* **Field**: A key in a JSON object.
+* **Property**: A field and the field's value collectively in a JSON object.
```
coreyoses: Remove "in a JSON object"

optimade.md
```diff
@@ -101,10 +101,19 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
+* **Field**: A key in a JSON object.
+* **Property**: A field and the field's value collectively in a JSON object.
+* **Property Value Types**:
+  * **string**, **integer**, **float**, **boolean**, **null value**: Base data
+    types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
+  * **list**, **dictionary**: Collection of base types with the meaning they have in the JSON
```
coreyoses: dictionary -> hash table

optimade.md
merkys: Could be simplified to "key-value pair"

optimade.md
CasperWA: done

optimade.md
CasperWA: done

optimade.md
CasperWA: Not here, I think. Since here we define a dictionary as a hash table, with the special limitation that the key MUST be a string. So here it should be fine, no?

optimade.md
```diff
@@ -101,8 +102,15 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
+* **Field**: A string key (similar to a key in a hash table, with the limitation that it MUST be a string).
+* **Property**: A field-value pair.
+* **Property Value Types**:
+  * **string**, **integer**, **float**, **boolean**, **null value**: Base data
+    types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
+  * **list**, **dictionary**: Collection of base types with the meaning they have in the JSON
+    data-interchange format, i.e., an ordered list of elements
+    (where each element can have a different type) or a hash table
+    (with the limitation that the hash key MUST be a string), respectively.
 * **Resource object**: Represent resources. MUST contain at least the following top-level fields:
```
merkys: Represent**s** resources

optimade.md
coreyoses: Ah I see, ok

CasperWA:
> Looks good. But by accepting this PR don't we break the querying? Term "Resource property" is not used anywhere, rendering everything that we used for querying before no longer queryable.

Completely true, this PR sparks a larger document-wide rewrite of every instance of `property` and of what is queryable.
I intend to open an issue for this. However, we first need a firm definition (this PR), before being able to do the revision (another PR, right?).

merkys:
> However, we first need a firm definition (this PR), before being able to do the revision (another PR, right?).

Understood. Just expressed my fear for having the specs slightly broken for a while :)

optimade.md
CasperWA: well spotted - fixed.

CasperWA:
> > However, we first need a firm definition (this PR), before being able to do the revision (another PR, right?).
> 
> Understood. Just expressed my fear for having the specs slightly broken for a while :)

Yeah... yeah 😞 I also thought about doing it all in this PR. And then kind of broke inside by the thought of this rather "small and simple" change snowballing.

optimade.md
```diff
@@ -101,9 +102,16 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
-* **Resource object**: Represent resources. MUST contain at least the following top-level fields:
+* **Field**: A string key (similar to a key in a hash table, with the limitation that it MUST be a string).
```
ltalirz:
```suggestion
* **Field**: A key of an associative-array-type data structure. A field MUST be a string.
```
hash tables are a specific implementation of associative arrays

optimade.md
```diff
@@ -101,9 +102,16 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
-* **Resource object**: Represent resources. MUST contain at least the following top-level fields:
+* **Field**: A string key (similar to a key in a hash table, with the limitation that it MUST be a string).
+* **Property**: A field-value pair.
+* **Property Value Types**:
+  * **string**, **integer**, **float**, **boolean**, **null value**: Base data
+    types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
+  * **list**, **dictionary**: Collection of base types with the meaning they have in the JSON
```
ltalirz:
```suggestion
  * **list**, **dictionary**: Collections of base types, with the same meaning they have in JSON.
```

optimade.md
```diff
@@ -101,9 +102,16 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
-* **Resource object**: Represent resources. MUST contain at least the following top-level fields:
+* **Field**: A string key (similar to a key in a hash table, with the limitation that it MUST be a string).
+* **Property**: A field-value pair.
```
dwinston: I would like **Queryable Property** to be introduced here, as it replaces the previous definition of **Property** (anything one can filter on).

dwinston: Otherwise this is fine.

optimade.md
```diff
@@ -486,6 +487,15 @@ The value for `is_index` MUST be `true`.
 > **Open Databases Integration for Materials Design** consortium can be found in [Appendix 1](#h.app1).
 > This list is also machine-readable, optimizing the automatic discoverability.
 
+## <a name="h.3.5">3.5. Queryable Properties</a>
+
+Queryable properties are resource properties.
```
ltalirz:
```suggestion
_Only_ properties of a _resource_ can be queried.

It is understood that "querying a property" refers to querying the property _field_ of a _resource_ object using the `filter` URL parameter described in section [5. API Filtering Format Specification](#h.5)).
Not all fields of a resource need to be queryable - see section [6. Entry List](#h.6) for information on which properties MUST be queryable and which are RECOMMENDED to be queryable.
```

optimade.md
ltalirz: @CasperWA  If one sees this as the definition of the term "queryable" when used in conjunction with "property", this would be fine with me as well.

optimade.md
fawzi: @rartino and @sauliusg added more constraints to identifiers that can be queried: starting with a letter, alphanumeric and _, and just lowercase letters. Do we want to put something about that here?

rartino: @fawzi Do you mean, to add to the definition of `property` that it is always a string of lowercase + underscore? That may be a good idea.

optimade.md
merkys: According to JSON specification ([RFC 8259](https://tools.ietf.org/html/rfc8259) and [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)), what we call `list` is `array`, and `dictionary` is `object`. If we say refer to JSON, we should use the names from the JSON specification.

merkys: No terminating period in @ltalirz's suggestion.

CasperWA: > According to JSON specification ([RFC 8259](https://tools.ietf.org/html/rfc8259) and [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)), what we call `list` is `array`, and `dictionary` is `object`. If we say refer to JSON, we should use the names from the JSON specification.

I agree with this sentiment, however, I always thought of these definitions here as our own way of sub-defining a dict and list - since (at least) object is a very specific thing in JSON. Sometimes we simply want a dict, e.g., under the `meta` field.
But in the end, if I have misunderstood and object is just the exact same as a dict, then its all fine - I may have mixed up _resource_ object and a regular ol' object sometimes in my mind.

This would then leave us with even more subsequent rewrites in the spec, since dictionary and list is mentioned a bunch of times. For simplicity we could just state that list and dictionary is equivalent to the array and object definitions in JSON?

merkys: > For simplicity we could just state that list and dictionary is equivalent to the array and object definitions in JSON?

Yes, this should do!

optimade.md
CasperWA: Done

optimade.md
CasperWA: Done

optimade.md
CasperWA: Done - however, this should be in the definition for **Field**, so I have added it here.

optimade.md
CasperWA: Done. Added **Queryable property** with the old **Property** defintion and also referencing the reader to section 3.5. Queryable Properties, which defines it more specifically. But is this not redundant, or?

optimade.md
```diff
@@ -101,9 +102,16 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
...
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
-* **Resource object**: Represent resources. MUST contain at least the following top-level fields:
+* **Field**: A key of an associative-array-type data structure.
+  A field MUST be a string, which MAY exclusively contain lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
+* **Property**: A field-value pair.
+* **Property value types**:
+  * **string**, **integer**, **float**, **boolean**, **null value**: Base data
+    types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
+  * **list**, **dictionary**: Collections of base types, defined in the samme manner as a JSON [array](https://json-schema.org/understanding-json-schema/reference/array.html) and [object](https://json-schema.org/understanding-json-schema/reference/object.html), respectively.
```
merkys: "samme" -> "same"

optimade.md
```diff
@@ -101,9 +102,16 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
-* **Resource object**: Represent resources. MUST contain at least the following top-level fields:
+* **Field**: A key of an associative-array-type data structure.
+  A field MUST be a string, which MAY exclusively contain lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
```
fawzi: Isn't MAY here too weak? I would put at least SHOULD, otherwise the keys might not be queriable...

CasperWA: I first put MUST, but then there would be two right after each other, which sounds a bit weird.
I think maybe instead I'll put something like: "A field MUST be a string, exclusively containing lowercase ..."

optimade.md
CasperWA: done

CasperWA:
> BTW, a PR with explicit listing of queryable properties should follow, ideally.

Hmm, I don't know. Personally, I think it's best to keep it defined as "all _resource_ properties", since one may have implementation-specific resource properties or other. Listing them would create too much overhead, I think.
However, we _do_ already have an issue for adding **Queryable** to all the fields under section 6. Entry List, and whether they MUST be queryable by a client or are only "SHOULD"s.

optimade.md
CasperWA: done

CasperWA:
> Good to go! Sorry for accidentally clicking "Unresolve/Resolve" many a time.

No problem - however, I would need another "Accepted" from you, since I always write "done" in review comments when I implement them in my local repo, and then later on push the commit(s).
Ah - thanks! :)

optimade.md
dwinston: Thanks @CasperWA. This is indeed redundant, but I think it is good to introduce key terms (even in not defining them fully) in this glossary.

CasperWA:
@ltalirz, you're free to merge if/when you approve :)

optimade.md
```diff
@@ -101,9 +102,16 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database**: An implementation that serves materials information.
 * **Entry**: A type of resource, over which a query can be formulated using the API
   (e.g., structure or calculation).
-* **Property**: Anything that can be in the filtering of results.
-* **Field**: A property that can be requested as partial output from the API.
-* **Resource object**: Represent resources. MUST contain at least the following top-level fields:
+* **Field**: A key of an associative-array-type data structure.
+  A field MUST be a string, exclusively containing lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
```
sauliusg: Sorry to re-raise an issue on the merged PR but lest I forget:

Should't we also say that "A field MUST be a string, exclusively containing lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`), **and start with a lowercase letter or an underscore**" ? The latter addition would make them consistent with the filter grammar, without  limiting identifier choices to much (`_1234_` would be a valid identifier :)

CasperWA: Done, see PR #132. And please review/approve :)

PR #127, opened 2019-06-14
Species redesign

sauliusg:
Redesigning the `species` property so that it does not contain dictionaries where data values are used as keys.


optimade.md
```diff
@@ -1517,13 +1514,14 @@ species are found in the [6.2.9. `species`](#h.6.2.9) property.
...
+* **Description**: A list describing the species of the sites of this structure. Species can be
 pure chemical elements, or virtual-crystal atoms representing a statistical occupation of a given site
 by multiple chemical elements.
 * **Requirements/Conventions**:
   * This property is REQUIRED.
-  * It MUST be a dictionary, where keys represent the species' name, and values are themselves
-  dictionaries with the following keys:
+  * It MUST be a list of dictionaries, and the each list member MUST be a
+  dictionary with the following keys:
+    * **name**: REQUIRED; gives the name of the species;
```
rartino: I'd like to see it repeated that name MUST be unique within all species in the list, since the automatic enforcement of this was a major reason that the old syntax was adopted.

dwinston:
I love this proposal.

optimade.md
```diff
@@ -1502,10 +1502,7 @@ species are found in the [6.2.9. `species`](#h.6.2.9) property.
   * This property is REQUIRED.
   * It MUST be a list of strings, which MUST have length equal to the number of sites in the structure
     (first dimension of the [6.2.7. `cartesian_site_positions`](#h.6.2.7) list).
-  * Each string MUST be a valid key of the dictionary specified by the [6.2.9. `species`](#h.6.2.9)
```
merkys: Do we want to drop the requirements for species names?

rartino: Well-spotted, I think this should be replaced with "MUST be present under the key `name` in one item in the [6.2.9. `species`](#h.6.2.9) list.

sauliusg: @merkys : since now the species name is just another value and will not be used anywhere in the code as identifier names, I suggest dropping all limitations on species names -- any string that JSON can support should be acceptable.

optimade.md
sauliusg: @rartino : The reason I did not put it here is that it is specified in the `species_at_site` description. But I see your point: uniqueness in the `species_at_site` list does **not** imply uniqueness of the name in the `species` list. So we need a separate uniqueness specification here as well. Will add it.

optimade.md
sauliusg: The equivalent of this statement is currently included in the point "Each species mentioned in the `species_at_sites` list MUST be described (i.e. for each value...", lines 1556-1560. I think this definition should be present only in one place, so that if we edit the requirements later the two separate definitions do not end up contradicting each other (SPOT/DRY). If you find this requirement easier to find in the `species_at_site` definition (probably it is, indeed), I'll move it there.

optimade.md
sauliusg: A side node: what we actually discussed can be formally expressed as follows:
`species_at_site`: constraint unique; foreign key references `species` (`name`); not null;
`species`.`name`: constraint unique; primary key, not null;

Feels to me like we are (re)designing SQL encoded in JSON ;), and eventually will arrive at a relational schema.

optimade.md
sauliusg: I have moved the phrase and pushed the new 'optimade.md' source, but for some strange reason it does not apper here. Can you please check if you see my latest changes?

optimade.md
```diff
@@ -1502,10 +1502,11 @@ species are found in the [6.2.9. `species`](#h.6.2.9) property.
...
   * It MUST be a list of strings, which MUST have length equal to the number of sites in the structure
     (first dimension of the [6.2.7. `cartesian_site_positions`](#h.6.2.7) list).
-  * Each string MUST be a valid key of the dictionary specified by the [6.2.9. `species`](#h.6.2.9)
-    property. The requirements on this string are the same as for property names, i.e., it can be of any
-    length, may use upper and lower case letters, underscore, and digits 0-9, but MUST NOT begin with a
-    digit.
+  * Each species MUST have a unique name.
+  * Each species name mentioned in the `species_at_sites` list MUST be
+    described in the `species` list (i.e. for each value in the `species_at_sites` list
+    there MUST exist a dictionary in the `species` list with the `name`
```
```suggestion
    there MUST exist exactly one dictionary in the `species` list with the `name`
```
I see your comment about having requirements in only one place, but still, I prefer using as strict language as possible in statements like this.

optimade.md
merkys: > @merkys : since now the species name is just another value and will not be used anywhere in the code as identifier names, I suggest dropping all limitations on species names -- any string that JSON can support should be acceptable.

Makes sense. I'd mark this as "resolved" if @rartino is OK with it.

optimade.md
```diff
@@ -1566,16 +1570,16 @@ by multiple chemical elements.
   * It is NOT RECOMMENDED that a structure includes species that do not have at least one
   corresponding site.
 * **Examples**:
-  * `"Ti": {"chemical_symbols": ["Ti"], "concentration": [1.0]}`: any site with this species is
+  * `"species": [ {"name": "Ti", "chemical_symbols": ["Ti"], "concentration": [1.0]} ]`: any site with this species is
```
fawzi: Add something like “Examples of species objects that need to be listed in the array of the species key:” just below **Examples**, and replace ‘“species”[ ... ]’ with '...’ ?

rartino:
Just a merge conflict left to fix.

merkys:
@sauliusg, could you please resolve the conflicts? It would be great to get this PR accepted before the freeze (this Friday).

optimade.md
sauliusg: I think it is useful to have a bit of context to specify where the species dictionary is placed; also, all our examples so far contain only one array element, but in principle the can be more – and so, `[ ... ]` braces indicate this fact.

Issue #103, opened 2019-06-13
Move from Markdown to RST?


giovannipizzi:
As discussed in [this comment](https://github.com/Materials-Consortia/OPTiMaDe/pull/100#discussion_r293129319) by @rartino, managing numbering of sections is becoming a big burden. Should we move so some different language (e.g. RST that supports automatic numbering and much more?)
There is probably a small barrier for users that are not used to RST but I think it's probably ok?

merkys:
RST would allow us to handle cross references easier, right? If so, I am in favor of this change.

giovannipizzi:
> RST would allow us to handle cross references easier, right?

Yes

dwinston:
I would like to bring attention to [MkDocs](https://www.mkdocs.org/) as an alternative solution to the problem of section link management.

Pros:
- still Markdown
- still a static site (i.e., hostable by github pages. convenient [`mkdocs gh-deploy`](https://www.mkdocs.org/user-guide/deploying-your-docs/) command.)
- sections can be broken into separate files for fewer merge conflicts, without sacrificing document search (built-in search with context-highlighted matches)

Cons:
- Rendered markdown may differ from github preview. Recommended to `mkdocs serve` and preview in local browser before git commit and push.

rartino:
Do I understand correctly that with an MkDocs-based solution there would be no way to see a rendered file with section numbers and cross references inside github? (I assume that we do not want to check in a rendered copy, e.g., generated by a hook, since that would trigger unnecessary conflicts?)

dwinston:
With a MkDocs-based solution, one would not have the convenience of the built-in Github website editor for instant feedback, i.e. editing a Markdown file and clicking a tab to render and preview your changes. Instead, one would need to edit that Markdown file locally and check the rendered preview at e.g. `http://localhost:8000` (regenerated automatically on file changes).

[...]

PR #131, opened 2019-06-18
Fix in filter language examples: 'HAS, EXACTLY' -> 'HAS EXACTLY'.

merkys:
There seems to be a pair of mistakes (typos?) in the filter language examples, for instance:
```
identifier HAS, EXACTLY values
```
The grammar does not allow commas between `HAS` and `EXACTLY`. This PR fixes these examples.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1289,7 +1289,7 @@ The following constructs MUST be supported:
 
 * `identifier HAS value`: matches if the given value is present in the list property (i.e., set operator IN).
 * `identifier HAS ALL values`: matches when all the values given are present in the list property (i.e., set operator >=).
-* `identifier HAS, EXACTLY values`: matches when the property contains all the values given and none other (i.e., set operator =).
+* `identifier HAS EXACTLY values`: matches when the property contains all the values given and none other (i.e., set operator =).
 * `identifier HAS ANY values`: matches when any one of the values given are present in the property (i.e., equivalent with a number of HAS separated by OR).
 * `LENGTH identifier <operator> value`: applies the numeric comparison operator for the number of items in the list property. 
 
@@ -1312,7 +1312,7 @@ element count.
 
 * `id1:id2:... HAS val1:val2:...`
 * `id1:id2:... HAS ALL val1:val2:...`
-* `id1:id2:... HAS, EXACTLY val1:val2:...`
+* `id1:id2:... HAS EXACTLY val1:val2:...`
 * `id1:id2:... HAS ANY val1:val2:...`
 * `id1:id2:... HAS ONLY val1:val2:...`
 


sauliusg:
As a note for the future: to avoid such mistakes in the future, since we have a machine readable grammar, wouldn't it be useful to extract all examples from the specification and run automatic checks (also in Travis) to see if they conform to the specs? (for filter examples that would be compilation with the Filter grammar derived from the EBNF in the spec; for response examples that would be validating against the schemas).

For this to work, all examples must be syntactically complete (ellipses will not work in general), and filter examples must be tagged somehow. They are now in `code` quotes, but this is not enough; many other code snippets that are mentioned are not filters.

Such specific tagging, however, would also be a step towards introducing a more detailed spelling checker, as was suggested by @giovannipizzi earlier.

merkys:
> to avoid such mistakes in the future, since we have a machine readable grammar, wouldn't it be useful to extract all examples from the specification and run automatic checks (also in Travis) to see if they conform to the specs?

Indeed. Invalid examples are real show-stoppers for adopters.

rartino:
> For this to work, all examples must be syntactically complete (ellipses will not work in general), and both filter and response examples must be tagged somehow. They are now in `code` quotes, but this is not enough; many other code snippets that are mentioned are not filters.

If we go for reStructuredText as suggested in #103 there are text roles for this, e.g.,
```
:optimade-filter:`identifier HAS EXACTLY values` 
```

sauliusg:
I would like to see code tagging functionality available, so if reStructuredText does that I am in favour.

But before switching we need to make sure that we do not loose any important functionality of MD, and that we actually can do everything that we need in the future – we do not want to switch to yet another markup in a couple of months, do we?

giovannipizzi:
My understanding is that markdown is basically a way to type easier HTML - using dashes instead of `<li>` etc. - and indeed one could directly write HTML inside a markdown document. It's much less standardised.
RST is a much more complete and more standardized language. I'm not aware (but I might be wrong) of features of MD missing in RST (while it's very much true for the opposite).

Some interesting resources (but there are many more on the web):
- https://eli.thegreenplace.net/2017/restructuredtext-vs-markdown-for-technical-documentation/
- https://gist.github.com/dupuy/1855764
- https://www.jungledisk.com/blog/2017/01/20/restructuredtext-vs-markdown/

sauliusg:
As I have commented elsewhere (https://github.com/Materials-Consortia/OPTiMaDe/issues/103#issuecomment-503431989), if we are happy to have a compilation step performed to get a readable text, then other possibilities are available:

- LaTeX (https://www.tug.org/texshowcase/);
- DocBook (https://tdg.docbook.org/tdg/5.2/);

PR #134, opened 2019-06-21
Forbidding values in field names

### LINKED ISSUE #112: Migrate away from using values in dictionary keys
An agreement was reached between @fawzi, @rartino and @sauliusg that values MUST NOT be used as dictionary keys. The main argument of such decision is the upcoming advent of queryable nested properties (e.g. dictionaries inside dictionaries; partially covered in #96). The following problems have been identified:
1. values in nested names might break identifier syntax of filtering language;
2. the user would need to query for the dictionary keys (suppose the user wants to query for all structures with sites with concentrations less than 0.001, but queries like `species.*.concentration < 0.001` are not going to be supported).

merkys:
PR #127 redesigned `species` so as not to allow dynamic field names, effectively removing the last occurrence of such names from the specification. In order to fully implement the spirit of #112, this policy has to be hard-coded into the specification. Here I suggest adding a restricting sentence to the definition of the term **Field**.

Fixes #112.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -109,7 +109,8 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Field**: A key of an associative-array-type data structure.
   A field MUST be a string, exclusively containing lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
   A field MUST start with a lowercase letter (`[a-z]`) or an underscore (`"_"`).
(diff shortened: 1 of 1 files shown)


merkys:
In live discussion with @sauliusg I started leaning towards not including meta-specifications. It might happen, of course, that in future we'll forget to maintain our decision of #112. Nevertheless the price of not being able to query such data structures would be minimal. Hence, I retract this PR.

PR #137, opened 2019-06-23
clarify JSON API connection

ltalirz:
* Replacing `jsonapi` => `json` to fix disagreement between spec and response examples.
 * Replace mentions of the "JSON API format" with "default `json` response format" (in order to avoid confusion with the many cases in the document, where the actual JSON API spec is referenced)

Other changes:
 * intro: update sentence on workshops, removing explicit dates, introducing API
 * intro: shorten section on JSON API, clarifying that this spec builds
   *on top of* the JSON API unless otherwise noted
 * intro: remove sentence on version number (seems obvious?)
 * response format: reorder, first explaining the default behavior
   and then moving on to what can be modified by the `response_format`
   query parameter


optimade.md
```diff
@@ -103,6 +94,7 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
 * **Database-provider-specific prefix**: This specification defines a set of
   database-provider-specific prefixes in [Appendix 1](#h.app1).
 * **Implementation**: An instance serving the OPTiMaDe API.
+* **JSON API response schema**: Schema for the default response format, which MUST be *compatible* with the [JSON API v1.0](http://jsonapi.org/format/1.0) (see also [3.3 Responses](#h.3.3)).
```
rartino: Is the definition of the term JSON API response schema the right place to insert a side remark that the default response format MUST be compatible with jsonapi? Because at a first quick read I was confused: it sounds as if the sentence is trying to say that a JSON API response schema itself must be compatible with the jsonapi specification somehow. I'd like to see this side remark moved (integrated with what is already there) to 3.3.

optimade.md
```diff
@@ -185,27 +177,23 @@ API implementations MUST decode URLs according to [RFC 3986](http://tools.ietf.o
...
-and declare these formats in their `info` endpoints
-(see section [4.3.2. Entry Listing Info Endpoints](#h.4.3.2)).
+By default, all API responses MUST comply with the [JSON API v1.0](http://jsonapi.org/format/1.0) specification.
+In particular, all endpoints MUST be able to respond in the JSON API format.
 
-An API implementation MAY return other formats than specified here.
-These can be implemented and documented according to the database provider.
-However, they MUST be prefixed by a database-provider-specific prefix as defined in
-[Appendix 1](#h.app1).
+Each endpoint MAY support additional formats, and declare these formats in its own `info` endpoint
```
```suggestion
Each endpoint MAY support additional formats, but these formats SHOULD be declared in the corresponding entry `info` endpoint
```
- Add a conformity keyword to the requirement (I'm going with SHOULD here as I don't want database-specific-prefixed formats temporarily being added for testing to formally break the spec.)
- Slight rephrase to clarify which info endpoint is meant.

optimade.md
```diff
@@ -185,27 +177,23 @@ API implementations MUST decode URLs according to [RFC 3986](http://tools.ietf.o
...
+(see section [4.3.2. Entry Listing Info Endpoints](#h.4.3.2)).
+Clients can request these formats using the `response_format` URL query parameter (see below).
+Specifying a `response_format` URL query parameter different from `json`
```
rartino: Oh, wow, I see now that we are inconsistent throughout the spec whether the magical string is 'json' or 'jsonapi'. (Text says 'jsonapi', examples of info output uses 'json'). This needs to be fixed (not necessarily in this PR.)

optimade.md
```diff
@@ -185,27 +177,23 @@ API implementations MUST decode URLs according to [RFC 3986](http://tools.ietf.o
...
+allows the API to break conformance with the JSON API specification not only
+in response format but also, e.g., in terms of how content negotiation is implemented.
 
-Specifying a `response_format` URL query parameter different from JSON API,
-allows the implementation to break conformance with the JSON API specification.
-Not only in response format, but also in, e.g., how content negotiation is implemented.
+API implementations MAY support additional formats beyond the ones specified in this document.
```
rartino: I'm confused; it seems the changes in this PR make this piece of text say the same thing twice, i.e. "Each endpoint MAY support additional formats" and "API implementations MAY support additional formats"? Can you try to integrate everything into one sentence?

optimade.md
```diff
@@ -185,27 +177,23 @@ API implementations MUST decode URLs according to [RFC 3986](http://tools.ietf.o
...
+In this case, the corresponding `response_format` identifiers MUST be include a
+database-provider-specific prefix as defined in [Appendix 1](#h.app1).
```
rartino: I don't like that this PR drops this text:
*"Specifying a `response_format` URL query parameter different from JSON API, allows the implementation to break conformance with the JSON API specification. Not only in response format, but also in, e.g., how content negotiation is implemented."*
This clarification seems quite crucial to me, in relation to the requirements of the jsonapi spec.

rartino:
One day away from the projected release of v0.10, this PR appears stalled? I think the changes of the introduction would be useful for v0.10, but if the PR isn't updated, I don't see this as release-blocking.

optimade.md
ltalirz: This is not a definition of the default response format. 
It is an explanation of what is meant by "JSON API response schema", which is used as a subtitle in many sections of the document (e.g. 4.1.2, 4.2.2 etc.)
This title can be confusing, since it can seem to imply that what follows below the title is something exclusively related to JSON API (while actually what follows will be a description of the default response format).

However, I agree with @rartino that this confusion would be better resolved by replacing all of these headlines with "JSON Response Schema" (or even "Default Response Schema"), in which case this term definition is no longer needed.
@rartino Should I go ahead and make the change?

optimade.md
ltalirz: Agreed, see my comment above.

optimade.md
ltalirz: Well... the distinction here is between endpoint-specific additional formats (that are nevertheless specified in this document) and provider-specific additional formats (which are not specified in this document).

But I agree that this can be shortened. Will do.

ltalirz:
@rartino Should be ready for re-review

optimade.md
rartino: I think I'm still confused about what these edits try to do, sorry.

A "JSON API response" is one out of several possible response formats. The "JSON API response schema" is a 'schema'-type description of the expectations of a response delivered in that format. Response formats are discussed in Sec 3, and there is some general general text about them in Section 3.3.1. As it seems you are concluding at the end in your reply, that means "JSON API response" / "JSON API Response Schema"  isn't something one would normally put as a **term definition**, just like we don't have a **term definition** of `chemical_formula_Hill`, but rather place all property definitions in a section of its own.

Now, with that background I don't quite understand the proposed rename: "JSON API response schema" -> "JSON Response Schema". If we in section 3.3.1 say that the default format is based on the "JSON API specification" and then refer back to that in headings and elsewhere as a "JSON API Response". Doesn't the headings "JSON API Resposne schema" make sense as written?

ltalirz: @rartino It reads like you didn't notice that my changes removed the term definition?
Did you refresh the "Files" tab?

> Doesn't the headings "JSON API Resposne schema" make sense as written?

I don't think so because it is confusing as mentioned in my comment above.
The schema we provide there is a JSON schema for the response that we expect when specifying `response_format='json'`.

rartino: @ltalirz I see them; but given your suggested rename, I was still not sure if we agree on the reason for why to do that.

> The schema we provide there is a JSON schema for the response that we expect when specifying response_format='json'.

Right now, Section 4.1.1. specifies that the string is `response_format=jsonapi`...

"JSON" is just a serialized data format for general data. In the specification, as it is written right now, we thus talk about the "JSON API response format" to be more specific than "JSON response format". In that case, I'd also expect section titles to be, e.g., "JSON API Response schema" rather than "JSON Response schema", etc.

I suppose it would be OK to consistently drop the "API" everywhere. But then all mentions of "JSON API Response format", "JSON API format" etc. throughout the text needs to be edited to drop the "API". And the selection string should be ` response_format=json.`

(But I also don't think what we have right now is in error, or even confusing. It is just using a slightly more specific name for the default output format.)

ltalirz: > Right now, Section 4.1.1. specifies that the string is response_format=jsonapi...

Ah - now I start to understand your comments better... I saw your previous comment on 
`jsonapi` vs `json` but did not realize that the **only** definition of the format string in the text was indeed `jsonapi` - sorry.
Do you know whether there is agreement on whether `json` and `jsonapi` should be used?

I would personally vote for `json` since it is something you can type in the URL and `json` might be more familiar to users of optimade (most of whom will have never heard of JSON API).

rartino: I wouldn't go as far as to say there is an agreement. But I think at some point, some of us, during the switch-over from the older json BUT NOT json api format into the json api format agreed to name the new format "jsonapi" to avoid confusion in the transition period, "json" would remain an obsolete reference to the old format.

BUT, with the latest major overhaul of the spec, I think that point is now moot. I do not feel strongly against overhauling the spec to consistently call the format "JSON resonse format".

ltalirz: Alright, will do!

ltalirz:
@rartino I've made the changes discussed.

Reading through the spec reminded me again that there are numerous instances where "JSON API" is used to refer to the **actual** JSON API spec, so I think calling our format the `json` response format does reduce the potential for confusion significantly.

rartino:
Ok, this seems internally consistent so I'm ready to approve it. But I would also really like to see someone else comment on whether they think this change is a good idea. The change of the output format specifier is "very backwards incompatible". At the release of v0.10, we may want to actually say that it is no longer recommended for clients or servers to support anything below v0.10...

However, you missed one occurrence of "JSON API response format" in section 3.6; (I cannot mark it in github because it is too far from anything else you've edited; perhaps it isn't even in your version and you need to merge upstream first.)

ltalirz:
> However, you missed one occurrence of "JSON API response format" in section 3.6; (I cannot mark it in github because it is too far from anything else you've edited; perhaps it isn't even in your version and you need to merge upstream first.)

Thanks - fixed.

> The change of the output format specifier is "very backwards incompatible". At the release of v0.10, we may want to actually say that it is no longer recommended for clients or servers to support anything below v0.10...

Depending on what one regarded as the spec, though...
If you went by the examples, nothing changes ;-)

optimade.md
```diff
@@ -188,27 +179,25 @@ API implementations MUST decode URLs according to [RFC 3986](http://tools.ietf.o
...
+This document defines a `json` response format that complies with the [JSON
+API v1.0](http://jsonapi.org/format/1.0) specification.
+All endpoints of an API implementation MUST be able to provide responses in the `json` format specified below
+and MUST respond in this format by default.
 
-An API implementation MAY return other formats than specified here.
-These can be implemented and documented according to the database provider.
-However, they MUST be prefixed by a database-provider-specific prefix as defined in
-[Appendix 1](#h.app1).
+Each endpoint MAY support additional formats, and SHOULD declare these formats under `endpoint/info`
```
CasperWA: To keep with convention, please encapsulate "endpoint" in your URI example with <> to show it is a placeholder.

optimade.md
```diff
@@ -188,27 +179,25 @@ API implementations MUST decode URLs according to [RFC 3986](http://tools.ietf.o
...
-An API implementation MAY return other formats than specified here.
-These can be implemented and documented according to the database provider.
-However, they MUST be prefixed by a database-provider-specific prefix as defined in
-[Appendix 1](#h.app1).
+Each endpoint MAY support additional formats, and SHOULD declare these formats under `endpoint/info` 
+(see section [4.3.2. Entry Listing Info Endpoints](#h.4.3.2)).
+Clients can request these formats using the `response_format` URL query parameter.
+Specifying a `response_format` different from `json` (e.g.
+`response_format=xml`) allows the API to break conformance not only with the
+`json` response format specification, but also, e.g., in terms of how content
```
CasperWA: Maybe it should here be JSON instead of \`json\`, or?

optimade.md
```diff
@@ -188,27 +179,25 @@ API implementations MUST decode URLs according to [RFC 3986](http://tools.ietf.o
...
+negotiation is implemented.
 
-Specifying a `response_format` URL query parameter different from JSON API,
-allows the implementation to break conformance with the JSON API specification.
-Not only in response format, but also in, e.g., how content negotiation is implemented.
+Database-provider-specific `response_format` identifiers MUST be include a
```
CasperWA: remove "be"

optimade.md
```diff
@@ -522,7 +511,7 @@ A query is performed using `filter` (see section [5. API Filtering Format Specif
 
 The API implementation MAY describe many-to-many relationships between entries along with OPTIONAL human-readable descriptions that describe each relationship. These relationships can be to the same, or to different, entry types.
 
-In responses that use the JSON API response format, such relationships MUST be communicated using [JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships) encoded in the `"relationships"` field of the response. The OPTIONAL human-readable description is provided in the `"description"` field inside the `"meta"` dictionary of a relationships object.
+In responses that use the `json` response format, such relationships MUST be communicated using [JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships) encoded in the `"relationships"` field of the response. The OPTIONAL human-readable description is provided in the `"description"` field inside the `"meta"` dictionary of a relationships object.
```
CasperWA: So, you are making it convention to write \`json\` instead of JSON, even though this is "only" the value to be queried for, when using `response_format`, while the "name" of the format is still JSON?

optimade.md
```diff
@@ -643,11 +631,11 @@ Examples:
 > URL tokens, and the above example is not to be taken as a recommendation for
 > such a mechanism.
 
-### <a name="h.4.1.2">4.1.2. JSON API Response Schema</a>
+### <a name="h.4.1.2">4.1.2. JSON Response Schema</a>
 
 "Entry listing" endpoint response dictionaries MUST have a `data`
 key. The value of this key MUST be a list containing dictionaries that
-represent individual entries. In the JSON API format every dictionary
+represent individual entries. In the default `json` response format every dictionary
```
CasperWA: Here, I would say JSON again, but this should be discussed in the first comment. I will, however, keep stating throughout the review, where I think it makes more sense to replace \`json\` with JSON, if you agree with me in the first place :)

optimade.md
```diff
@@ -714,7 +702,7 @@ Example:
 A client can request a specific entry by appending an ID component to the URL of an entry listing
 endpoint. This will return properties for the entry with that ID.
 
-If using the JSON API format, the ID component MUST be the content of the `id` field.
+In the default `json` response format, the ID component MUST be the content of the `id` field.
```
CasperWA: JSON

optimade.md
```diff
@@ -731,11 +719,11 @@ URL query parameters not recognized MUST be ignored. While the following URL que
 are OPTIONAL for clients, API implementations MUST accept and handle them:
 **response\_format**, **email\_address**, **response\_fields**. The meaning of these URL query parameters are as defined above in section [4.1.1. URL Query Parameters](#h.4.1.1).
 
-### <a name="h.4.2.2">4.2.2. JSON API Response Schema</a>
+### <a name="h.4.2.2">4.2.2. JSON Response Schema</a>
 
 The response for a 'single entry' endpoint is the same as for 'entry listing'
 endpoint responses, except that the value of the `data` field MUST have only one or zero entries.
-If using the JSON API format, this means that the response type of the `data` field MUST be
+In the default `json` response format, this means the value of the `data` field MUST be
```
CasperWA: JSON

optimade.md
```diff
@@ -2079,12 +2067,11 @@ of that organization. Identifiers prefixed with underscores will not be used for
 The database-provider-specific prefixes currently assigned are listed in the `providers.json` file
 provided in the main repository. This file serves as a machine-readable list of OPTiMaDe providers.
 
-The content of the `providers.json` file follows the same JSON API specifications as the rest of the
-API, in particular the resource objects under the top-level `data` field are defined to be valid
+The content of the `providers.json` file complies with the default `json` format specification for API responses.
```
CasperWA: JSON

ltalirz: All don, thanks @CasperWA

optimade.md
```diff
@@ -1,5 +1,4 @@
 # OPTiMaDe API specification v0.9.8-develop
-
```
CasperWA: Shouldn't there be a new line after a header in markdown?

ltalirz: This change was not intentional. I've put it back.

CasperWA:
For me, you're welcome to merge this @rartino :smiley:

PR #138, opened 2019-06-25
Set operators

### LINKED ISSUE #117: Clarify description of set operators when multiple elements can match
Based on a discussion with @giovannipizzi I have convinced myself that there is some ambiguity in how we describe what the set operators HAS, HAS ALL, HAS ONLY, HAS EXACTLY does when there can be multiple elements in the list/set that matches.

rartino:
This PR fixes #117.

Summary:
- Rephrase definitions of HAS operators to be logically consistent for duplicate matches.
- Remove `HAS EXACTLY` (including in grammar tests)
- Clarify in the phrase definitions what a `property name` is, because I needed it (but I also see it was used throughout the spec.)

During discussions on the workshop it became clear that we phrased the definitions of the HAS operators in the filer language with sets in mind. But, the specification has evolved to be very explicit about that these operators work on 'lists' and also the operator syntax, e.g., `HAS < 3`, means multiple elements may match even in a set. So, our formulations have to be improved so it is unambiguously clear how these operators work when multiple matches are present. This PR rewrites the definitions to do that.

However, it appears "HAS EXACTLY" is troublesome in this case. I don't think there is a single definition of "HAS EXACTLY" that is self-evident when multiple matches are present. Some attempts may be to define it as:
- `list HAS EXACTLY 3,4,5` := `list HAS ALL 3,4,5 AND list LENGTH 3` - This works for sets, but, e.g., with `list=[3,4,5]` do you expect `list HAS ALL 3,3,5` to match?
- `list HAS EXACTLY 3,4,5` := `list HAS ALL 3,4,5 AND list HAS ONLY 3,4,5` - This is probably the most logically consistent definition, **however** if API implementations are supposed to implement it this way, they need `HAS ONLY`, which is optional. `EXACTLY` was marked as mandatory.

So, for now, with < 1 week left to the planned release of OPTiMaDe v0.10 I think it is better for now to remove `HAS EXACTLY` from the specification completely and revisit it later to possibly bring it back. No functionality is lost, as a client can phrase their queries using one of the two constructs above.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1300,7 +1300,7 @@ Examples:
 
 ### Comparisons of list properties
 
-In the following, `values` is one or more `value` separated by commas (","), i.e., strings or numbers. An implementation MAY also support property names and nested property names in `values`.
+In the following, `list` is a list-type property, and `values` is one or more `value` separated by commas (","), i.e., strings or numbers. An implementation MAY also support property names and nested property names in `values`.
 
 The following constructs MUST be supported:
 
@@ -1338,7 +1338,7 @@ values on the right-hand side MAY allow `<operator> value` in each
 place a value can appear. In that case, a match requires that the
 `<operator>` comparison is fulfilled instead of equality. Strictly,
 the definitions of the `HAS`, `HAS ALL`, `HAS ANY` and
-`HAS ONLY` operators as written above applies, but with the word
+`HAS ONLY` operators as written above apply, but with the word
 'equal' replaced with the `<operator>` comparison.
 
 For example:
--- a/optimade.md
+++ b/optimade.md
@@ -110,7 +110,7 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
   A field MUST be a string, exclusively containing lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
   A field MUST start with a lowercase letter (`[a-z]`) or an underscore (`"_"`).
   MUST NOT match any of the entry names.
-* **Property**: A field-value pair. 
+* **Property**: A field-value pair.
 * **Property value types**:
   * **string**, **integer**, **float**, **boolean**, **null value**: Base data
     types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
--- a/optimade.md
+++ b/optimade.md
@@ -110,7 +110,7 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
   A field MUST be a string, exclusively containing lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
   A field MUST start with a lowercase letter (`[a-z]`) or an underscore (`"_"`).
   MUST NOT match any of the entry names.
-* **Property**: A field-value pair.
+* **Property**: A field-value pair. The field is called the **property name**, the value is called the **property value**.
 * **Property value types**:
   * **string**, **integer**, **float**, **boolean**, **null value**: Base data
     types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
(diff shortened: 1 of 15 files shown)


optimade.md
```diff
@@ -110,7 +110,7 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
   A field MUST be a string, exclusively containing lowercase alphanumerics (`[a-z0-9]`) and underscores (`"_"`).
   A field MUST start with a lowercase letter (`[a-z]`) or an underscore (`"_"`).
   MUST NOT match any of the entry names.
-* **Property**: A field-value pair.
+* **Property**: A field-value pair. The field is called the **property name**, the value is called the **property value**.
```
merkys: I'd suggest against introducing an additional alias for **field** (no need for **property name**). Or is **field** different from **property name**?

optimade.md
```diff
@@ -1300,55 +1300,62 @@ Examples:
 
 ### Comparisons of list properties
 
-List properties can be thought of as lists or sets of strings or numbers. 
-In the following, a set of `values` is one or more strings or numbers separated by a comma (",").
-An implementation MAY also support identifiers in the value set.
+In the following, `values` is one or more `value` separated by commas (","), i.e., strings or numbers. An implementation MAY also support property names and nested property names in `values`.
```
merkys: An explanation of `list` is needed here, providing its connection to the identifier.

optimade.md
```diff
@@ -1300,55 +1300,62 @@ Examples:
...
 
 Finally, all the above constructs that allow a value or lists of
-values on the right-hand side MAY allow `<operator> value`
-in each place a value can appear. In that case, a match requires that
-the equality or inequality is fulfilled. For example:
+values on the right-hand side MAY allow `<operator> value` in each
+place a value can appear. In that case, a match requires that the
+`<operator>` comparison is fulfilled instead of equality. Strictly,
+the definitions of the `HAS`, `HAS ALL`, `HAS ANY` and
+`HAS ONLY` operators as written above applies, but with the word
```
merkys: applies -> apply (plural)

optimade.md
rartino: My understanding is that "field" means a key in any associative array. In particular, it is used a bunch of times for referring to keys in the json objects that are returned, e.g., the 'attributes' field, the 'query' field, etc., not only for property names.

"Property name" specifically refers to the name of a property. 

"Property name" is already used a bunch of times in the specification with this meaning, so I thought it was nice to add it among the term definitions?

optimade.md
rartino: Right, I did cut a bit too much. Fixed.

optimade.md
merkys: Right, but from the Term Definition I see no distinction between a property and a key-value pair of any associative array.

optimade.md
rartino: Indeed, thanks.

optimade.md
rartino: I see what you are saying. "Property" used to be reserved for actual data properties, in the sense of things in the output and/or queryable. But you are right that since PR #126 it now appears *any* key-value entry in an associative array is a property. Yet, we never use the phrase in the more general sense in the document, but rather in several places seem to rely on the more strict meaning. Was this change to make 'property' completely generic really intentional? (Pinging @ltalirz who made that PR).

ltalirz: > Property" used to be reserved for actual data properties, in the sense of things in the output and/or queryable.

This seems like a slightly vague definition and to my knowledge it was not present in the spec in explicit form before (e.g. is it anything under `data` or anything in the response (e.g. `meta`), ...) . The current definition of "property" is its [usual meaning in JSON](https://json-schema.org/understanding-json-schema/reference/object.html) and I would discourage adopting a different meaning in our spec.

As such I agree with @merkys to not introduce further terms if not necessary.

rartino: Thanks for the quick reply! The definition prior to #126 was that a property was anything that was queryable, which, at the time that was written was all data that we had standardized, so I don't think it was overly vague. I am a bit concerned that a lot of use of "property" in the document was written with that definition in mind, and now it is changed. But that has to be addressed in a separate PR. I'll simply remove the attempted definition of "Property name" from this PR as the definition under Lexical tokens for the same term is sufficient for the needs of this PR anyway.

But I think all this means we need to re-think some use of terminology; either in the definitions, or in the text.

rartino:
Ok, property name definition is removed. This PR is read for re-review.

optimade.md
merkys: I agree with @rartino. I think that properties under `attributes` of the entry types have to be distinguished from the others, because it's only these `attributes` properties that the filters act upon (not talking about relationships now), and that are allowed to be extended using the DB-specific prefixes.

rartino: I suppose we could call the properties I talk about "entry properties" for clarity. That can be abbreviated as "properties" in segments of text where it is obvious that it is what is being referred to.

But in that case, I would change our main term definition to define *entry property* rather than *property*. I'm OK with either a term definition or just a comment elsewhere that says that *in the context of json objects* a property is a field-value pair.

rartino:
Waiting for one more approval to get this through.

optimade.md
merkys: I like "entry properties". For JSON objects, I'd prefer "field-value pair", or "name-value pair" as per [RFC 8259](https://tools.ietf.org/html/rfc8259).

ltalirz: Before continuing this discussion, I think it would be helpful if @rartino or @merkys could provide a crystal clear definition of what subset of field-value pairs you mean by "entry properties" (or whatever name we end up choosing).
I.e., given any possible response of the OPTiMaDe API, how do I decide whether a given field-value pair is an "entry property"?

I ask this because I get the impression that the two of you already mean different things. The whole issue that led to the explicit definition of "property" is that e.g. section 6.1 lists field-value pairs like "type", which is *not* an attribute, and field-value pairs like `last_modified` which *is* an attribute.

rartino: Lets move this discussion to #141 as we are merging (and thus closing) this PR.

rartino:
With two approvals and one day to v0.10 I am merging this. If someone finds remaining issues, we'll handle them as another PR.

PR #140, opened 2019-06-25
Updated: allow web-browsers to issue cross-domain requests

### LINKED ISSUE #101: Allow web-browsers to issue cross-domain requests

rartino:
I don't dare to merge the original PR #105 since it shows a diff that is the whole file.

So, here is a PR rebased on the latest devel that shows a sane diff. I also *slightly* reworded the phrasing in a way I hope is uncontroversial. 

Closes #105, closes #101.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -453,7 +453,7 @@ to the corresponding database ID that was originally queried, using the object's
 
 ### <a name="h.3.3.4">3.3.4. HTTP Response Headers</a>
 
-There are relevant use-cases for allowing data served via OPTiMaDe to be accessed from in-browser JavaScript, e.g. to enable the server-less data aggregations. For such use, many browsers need the server to include the header `Access-Control-Allow-Origin: <site-of-the-javascript-page>` or `Access-Control-Allow-Origin: *` (to allow any site) in its responses. 
+There are relevant use-cases for allowing data served via OPTiMaDe to be accessed from in-browser JavaScript, e.g. to enable the server-less data aggregations. For such use, many browsers need the server to include the header `Access-Control-Allow-Origin: *` in its responses, which indicates that in-browser JavaScript access is allowed from any site. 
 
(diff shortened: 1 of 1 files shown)


merkys:
I still feel there's a need to include at least one of [RFC 2119 keywords](http://tools.ietf.org/html/rfc2119) to let the adopters know how strong this paragraph is. (I suspect `MAY/OPTIONALLY` is needed here)

rartino:
@merkys The whole point in the discussion in #105 was that what we could come to an agreement on is to leave this as an implementors note and not a statement of conformity to the standard.

My intent is that in the hypotetical conversion to RST after the release of v0.10 we separate out implementors notes from the specification with this markup: https://github.com/rartino/OPTiMaDe/blob/restructuredtext/demo.rst#15notes

merkys:
Sorry, I didn't follow the discussion on #105 to the end. I suppose we can add it for now then.

rartino:
That addition was @fawzi's request. @fawzi, what are your thoughts?

fawzi:
The idea was that if we explain the issue an solutions, I would explain things neutrally, and for paranoids listing only one site is an option, but I do not feel strongly about it.

rartino:
I've tried my best to rectify the final concerns. Is the present version acceptable?

rartino:
Since the opinions expressed about `Access-Control-Allow-Origin: <site-of-the-javascript-page>` both come with "not very import", and we have 3 approvals, I'm going to merge this. Still, we may want to revisit to this question for v1.0.

PR #142, opened 2019-06-25
Filtering using nested property names and on relationships

### LINKED ISSUE #116: Querying on relationship, dot operator in grammar, change of sites definition

rartino:
This is the final remaining part of #116, with content that was agreed on in the final two days of the workshop and which I agreed on writing up. Sorry for the delay, but it wasn't so easy to write up. Closes #116.

In summary:
- Describes filtering on nested property names. Much of this was already present in the lexical tokens subsection. It is now moved alongside other discussion of filtering.
- Describes filtering on relationships. This was vaguely hinted at alongside the discussion on filters on nested property names, but is now in the filtering section and clarified.
- Some other cleanup with regards to relationships necessary to concisely formulate these parts. This includes disintegrating some jsonapi-response-format-specifics from the more abstract concept of relationships in OPTiMaDe (and existing also for other potential output formats.); including moving up the general discussion of relationships to section 3, rather than down in section 6.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1372,7 +1372,7 @@ For example: a client queries the `structures` endpoint with a filter that refer
 > For example, to find all structures with bibliographic references where one of the authors has the last name "Schmit" is performed by the following two steps:
 >
 > - Query the `references` endpoint with a filter `authors.lastname HAS "Schmit"` and store the `id` values of the returned entries. 
-> - Query the `structures` endpoint with a filter `references.id HAS ANY <list-of-IDs>`, where `<list-of-IDs>` is the IDs retrieved from the first query separated by commas. 
+> - Query the `structures` endpoint with a filter `references.id HAS ANY <list-of-IDs>`, where `<list-of-IDs>` are the IDs retrieved from the first query separated by commas. 
 >
 > (Note: the type of query discussed here corresponds to a "join"-type operation in a relational data model.)
 
--- a/optimade.md
+++ b/optimade.md
@@ -519,9 +519,9 @@ A query is performed using `filter` (see section [5. API Filtering Format Specif
 
 The API implementation MAY describe many-to-many relationships between entries along with OPTIONAL human-readable descriptions that describe each relationship. These relationships can be to the same, or to different, entry types.
 
-In responses that use the jsonapi format, such relationships MUST be communicated using the `"relationships"` field of the response encoded as jsonapi relationship objects in accordance with the [jsonapi specification](https://jsonapi.org/format/1.0/#document-resource-object-relationships). The OPTIONAL human-readable description is provided in the `"description"` field inside the `"meta"` dictionary of a relationship.
+In responses that use the JSON API response format, such relationships MUST be communicated using [JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships) encoded in the `"relationships"` field of the response. The OPTIONAL human-readable description is provided in the `"description"` field inside the `"meta"` dictionary of a relationships object.
 
-Other response formats (e.g., ones using database-specific prefixes) have to encode these relationships in ways appropriate for each format. If the format has no dedicated mechanism to indicate relationships, it is suggested that they are encoded alongside other properties. For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the ids of the referenced entries alongside the respective human-readable description of the relationships. It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
+Other response formats (e.g., ones using database-specific prefixes) have to encode these relationships in ways appropriate for each format. If the format has no dedicated mechanism to indicate relationships, it is suggested that they are encoded alongside other properties. For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the IDs of the referenced entries alongside the respective human-readable description of the relationships. It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
 
 # <a name="h.4">4. API Endpoints</a>
 
@@ -1225,8 +1225,8 @@ All filtering expressions MUST follow the
 [EBNF](http://standards.iso.org/ittf/PubliclyAvailableStandards/s026153_ISO_IEC_14977_1996(E).zip)
 grammar of [Appendix 2](#h.app2) of this specification. The appendix
 contains a complete machine-readable EBNF, including the definition
-of the lexical tokens described above in [section '5.1. Lexical
-tokens'](#h.5.1). The EBNF is enclosed in special strings constructed
+of the lexical tokens described above in [5.1. Lexical
+tokens](#h.5.1). The EBNF is enclosed in special strings constructed
 as `BEGIN` and `END`, both followed by `EBNF GRAMMAR Filter`, to enable automatic
 extraction.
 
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -515,6 +515,14 @@ A query is performed using `filter` (see section [5. API Filtering Format Specif
 
 > **For implementers**: To get an understanding of which properties MUST be queryable and which are RECOMMENDED, please see section [6. Entry List](#h.6).
 
+## <a name="h.3.6">3.6. Relationships</a>
+
+The API implementation MAY describe many-to-many relationships between entries along with OPTIONAL human-readable descriptions that describe each relationship. These relationships can be to the same, or to different, entry types.
+
+In responses that use the jsonapi format, such relationships MUST be communicated using the `"relationships"` field of the response encoded as jsonapi relationship objects in accordance with the [jsonapi specification](https://jsonapi.org/format/1.0/#document-resource-object-relationships). The OPTIONAL human-readable description is provided in the `"description"` field inside the `"meta"` dictionary of a relationship.
```
merkys: * jsonapi format -> `jsonapi` format
* jsonapi relationship objects in accordance with the [jsonapi specification](https://jsonapi.org/format/1.0/#document-resource-object-relationships) -> [JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships)

optimade.md
```diff
@@ -515,6 +515,14 @@ A query is performed using `filter` (see section [5. API Filtering Format Specif
...
+
+In responses that use the jsonapi format, such relationships MUST be communicated using the `"relationships"` field of the response encoded as jsonapi relationship objects in accordance with the [jsonapi specification](https://jsonapi.org/format/1.0/#document-resource-object-relationships). The OPTIONAL human-readable description is provided in the `"description"` field inside the `"meta"` dictionary of a relationship.
+
+Other response formats (e.g., ones using database-specific prefixes) have to encode these relationships in ways appropriate for each format. If the format has no dedicated mechanism to indicate relationships, it is suggested that they are encoded alongside other properties. For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the ids of the referenced entries alongside the respective human-readable description of the relationships. It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
```
merkys: To use the proper term:
* ids -> IDs

optimade.md
```diff
@@ -1352,6 +1346,35 @@ Examples:
 * OPTIONAL: `_exmpl_element_counts HAS < 3 AND _exmpl_element_counts HAS ANY > 3, = 6, 4, != 8` (note: specifying the = operator after HAS ANY is redundant here, if no operator is given, the test is for equality.)
 * OPTIONAL: `elements:_exmpl_element_counts:_exmpl_element_weights HAS ANY > 3:"He":>55.3 , = 6:>"Ti":<37.6 , 8:<"Ga":0`
 
+### Nested property names
+
+Everywhere in a filter string where a property name is accepted, the API implementation MAY accept nested property names as described in [section '5.1. Lexical tokens'](#h.5.1), consisting of fields separated by periods ('.'). A filter on a nested property name consisting of two fields `field1.field2` matches if either one of these points are true:
```
merkys: To be consistent with the rest of the document:
* [section '5.1. Lexical tokens'](#h.5.1) -> section [5.1. Lexical tokens](#h.5.1)

optimade.md
```diff
@@ -1352,6 +1346,35 @@ Examples:
...
+
+The API implementation MAY allow this notation to generalize to arbitary depth. 
+A nested property name that combines more than one list MUST, if accepted, be interpreted as a completely flattened list.
+
+### Relationships
+
+As described in section [3.6. Relationships](#h.3.6), it is possible for the API implementation to describe relationships between entries of the same, or different, entry types. 
+The API implementation MAY support queries on relationships with an entry type `<entry type>` by using special nested property names:
+
+- `<entry type>.id` references a list of ids of relationships with entries of the type `<entry type>`.
```
merkys: ids -> IDs

optimade.md
```diff
@@ -1352,6 +1346,35 @@ Examples:
...
+- `<entry type>.description` references a correlated list of the human-readable descriptions of these relationships.
+
+Hence, the filter language acts as, for every entry type, there is a property with that name which contains a list of dictionaries with two fields, `id` and `description`.
+For example: a client queries the `structures` endpoint with a filter that references `calculations.id`. For a specific structures entry, the nested property may behave as the list `["calc-id-43", "calc-id-96"]` and would then, e.g., match the filter `calculations.id HAS "calc-id-96"`. This means that the structures entry has a relationship with the calculations entry of that id.
```
merkys: id -> ID

optimade.md
```diff
@@ -1352,6 +1346,35 @@ Examples:
...
+
+> **Note:** formulating queries on relationships with entries that have specific property values is a multi-step process.
```
merkys: Might be good to mention that this is known as "join" in a relational data model

optimade.md
```diff
@@ -1352,6 +1346,35 @@ Examples:
...
+
+Hence, the filter language acts as, for every entry type, there is a property with that name which contains a list of dictionaries with two fields, `id` and `description`.
+For example: a client queries the `structures` endpoint with a filter that references `calculations.id`. For a specific structures entry, the nested property may behave as the list `["calc-id-43", "calc-id-96"]` and would then, e.g., match the filter `calculations.id HAS "calc-id-96"`. This means that the structures entry has a relationship with the calculations entry of that id.
+
+> **Note:** formulating queries on relationships with entries that have specific property values is a multi-step process. 
+> For example, to find all structures with bibliographic references where one of the authors have the last name "Schmit" is performed by the following two steps:
+>
+> - Query the `references` endpoint with a filter `authors.lastname HAS "Schmit"` and store the `id` values of the returned entries. 
+> - Query the `structures` endpoint with a filter `references.id HAS ANY <list-of-ids>`, where `<list-of-ids>` is the ids retrieved from the first query separated by commas.
```
merkys: ids -> IDs

optimade.md
```diff
@@ -1943,10 +1966,8 @@ in section [6.1. Properties Used by Multiple Entry Types](#h.6.1).
 
 ## <a name="h.6.6">6.6. Relationships Used by Multiple Entry Types</a>
 
-[JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships)
-MAY be used to describe the relations between entries. A human-readable description
-of a relationship MAY be provided using the `"description"` field inside the
-`"meta"` dictionary of a relationship.
+In accordance with section [3.6. Relationships](#h.3.6), all entry types may use
```
merkys: may -> MAY

optimade.md
rartino: I agree that "jsonapi" is sloppy. For the first point, I didn't mean to reference the string to select the format, so maybe better to change it to "JSON API response format". For the second point I had to reword a bit, so this is what I arrived at: 
```
In responses that use the JSON API response format, such relationships MUST be communicated using [JSON API Relationships](https://jsonapi.org/format/1.0/#document-resource-object-relationships) encoded in the `"relationships"` field of the response. The OPTIONAL human-readable description is provided in the `"description"` field inside the `"meta"` dictionary of a relationships object.
```

optimade.md
rartino: Consistent indeed - as I copy-pasted that exact link format from what was already in 5.2 :-). I'll go slightly outside the mandate of the present PR to update both places.

rartino:
Many thanks for the feedback @merkys and @ml-evs, please re-review.

optimade.md
merkys: OK, thanks

ml-evs:
Go ahead @rartino!

PR #145, opened 2019-06-28
Introducing timestamp data type

### LINKED ISSUE #139: Date data type
The current description of comparison operators does not cover the handling of the following cases:
1. <s>String vs. numeric types. My proposition: if one of the operands is numeric, then the string has to be converted to number, and a numeric comparison should be carried out. If the string does not convert to a number, emit error.</s>
2. Dates. Currently, the value of `time_stamp` is a date (expressed as `string` in JSON), thus I believe the API would benefit from being able to compare dates. My proposition: when filtered on date fields, the backend has to convert the query operand (in [ISO 8601](https://www.iso.org/standard/40874.html) with time zone) to internal date representation and compare to see which of the dates is earlier and which is later.
3. <s>Booleans. My proposition: if one of the operands is boolean, then the other has to be converted to boolean (`0`, `"0"`, `""` are `False`, rest being `True`), and `True` is greater than `False`.</s>

What do you think?

**Edit**: first case is already covered in "Type handling and conversions in comparisons", and the third case is not important for the specification right now; retitling the issue to match the second case.

merkys:
Introducing `timestamp` data type, currently used only for `last_modified` property.

Summary of changes:
* Added `timestamp` to the Term Definitions (as the rest of data types are defined there)
* Switched from [ISO 8601](https://www.iso.org/standard/40874.html) to [RFC 3339 Internet Date/Time Format](https://tools.ietf.org/html/rfc3339#section-5.6): ISO is non-free and too vague. I suggest using open and well-defined property.
* Forbidding conversions from numeric values to timestamps.

Fixes #139

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -116,6 +116,8 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
   * **string**, **integer**, **float**, **boolean**, **null value**: Base data
     types as defined in more detail in section [5.1. Lexical Tokens](#h.5.1).
   * **list**, **dictionary**: Collections of base types, defined in the same manner as a JSON [array](https://json-schema.org/understanding-json-schema/reference/array.html) and [object](https://json-schema.org/understanding-json-schema/reference/object.html), respectively.
+  * **timestamp**: string representation of date and time as defined in
+    [RFC 3339 Internet Date/Time Format](https://tools.ietf.org/html/rfc3339#section-5.6).
 * **Queryable property**: A property that can be in the filtering of results.
   For more information see section [3.5. Queryable Properties](#h.3.5).
 * **Resource object**: Represents resources. MUST contain at least the following top-level fields:
@@ -217,12 +219,8 @@ Every response SHOULD contain the following fields, and MUST contain at least on
     * **representation**: a string with the part of the URL following the base URL.
 
   * **api\_version**: a string containing the version of the API implementation.
-  * **time\_stamp**: a string containing the date and time at which the query
-    was executed, in [ISO 8601](https://www.iso.org/standard/40874.html)
-    format.  Times MUST be timezone-aware (i.e., MUST NOT be local times),
-    in one of the formats allowed by [ISO 8601](https://www.iso.org/standard/40874.html)
-    (i.e., either be in UTC, and then end with a Z, or indicate explicitly
-    the offset).
+  * **time\_stamp**: a timestamp containing the date and time at which the query
+    was executed.
   * **data\_returned**: an integer containing the number of data objects returned for the query.
   * **more\_data\_available**: `false` if all data for this query has been
     returned, and `true` if not.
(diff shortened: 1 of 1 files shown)


merkys:
I completely agree with @rartino. We should separate the terms, format-specific conventions and filter syntax. I hope this PR does not make this separation more difficult for the future.

optimade.md
```diff
@@ -1483,8 +1482,8 @@ This section defines standard entry types and their properties.
 
 ### <a name="h.6.1.3">6.1.3. last\_modified</a>
 
-* **Description**: Date representing when the entry was last modified.
-* **Requirements/Conventions**: String with [ISO 8601](https://www.iso.org/standard/40874.html) format.
+* **Description**: date representing when the entry was last modified.
```
CasperWA: Please keep the capital "D" here, to make it similar to other entries.

Furthermore, perhaps this should not only be date, but "Date and possibly time"?

optimade.md
```diff
@@ -1483,8 +1482,8 @@ This section defines standard entry types and their properties.
 
 ### <a name="h.6.1.3">6.1.3. last\_modified</a>
 
-* **Description**: Date representing when the entry was last modified.
-* **Requirements/Conventions**: String with [ISO 8601](https://www.iso.org/standard/40874.html) format.
+* **Description**: date representing when the entry was last modified.
+* **Requirements/Conventions**: a timestamp.
```
CasperWA: Capital "A", to keep with the conventions of the other entries here.

Possibly either repeat the RFC here for easy look-up access, or refer to the "Term Definitions". It can be confusing for someone who goes here, but then don't know where to turn to afterwards, when they don't know what "timestamp" is defined as.
Personally, I prefer the way it was done before, i.e., link to the RFC :)

optimade.md
merkys: > Please keep the capital "D" here, to make it similar to other entries.

OK.

> Furthermore, perhaps this should not only be date, but "Date and possibly time"?

I'd say "Date and time", as `timestamp` contains both date and time.

optimade.md
merkys: OK, thanks for the review!

optimade.md
CasperWA: Sure, but I added the "possibly", since I don't know if we should require a time here as well? If a database does not keep track of the time their entries are changed, but only the date, this should be a valid value I think? They shouldn't have to update all their entries starting to record time as well, or? Does RFC 3339 not support _only_ writing a date?

Sorry for the many questions here, but it should be able to answer with a single answer almost :)

optimade.md
```diff
@@ -657,8 +657,7 @@ MUST have the following fields:
 * **attributes**: a dictionary, containing key-value pairs representing the
   entry's properties and the following fields:
   * **local\_id**: the entry's local database ID (having no OPTiMaDe requirements/conventions)
-  * **last\_modified**: an [ISO 8601](https://www.iso.org/standard/40874.html)
-    representing the entry's last modification time
+  * **last\_modified**: an timestamp representing the entry's last modification date and time
```
CasperWA: Well caught!

rartino:
@CasperWA has a point in that as far as I can see RFC3339 dates require times, and it may be worthwhile to consider if OPTiMaDe timestamps conceptually must adhere to the same granularity or is allowed to represent full days. But for now I'm going to approve this.

merkys:
Thanks for reviews, @rartino and @CasperWA!

I wrote this PR having both date and time in mind (timestamp MUST contain both date and time, as per `date-time` production in [RFC 3339 Internet Date/Time Format](https://tools.ietf.org/html/rfc3339#section-5.6)). Previous definition of `time_stamp` was:

> a string containing the date and time at which the query was executed, in [ISO 8601](https://www.iso.org/standard/40874.html) format.  Times MUST be timezone-aware [...]

The same was in all the examples. Therefore I assumed the same definition for all the occurrences. We can in principle allow date-only timestamps, but I think we can do this on-demand.

PR #146, opened 2019-06-28
Adding AiiDA and improving description of Materials Cloud

giovannipizzi:
Similar to #143, where `httk` was added for the software
in parallel to `omdb` for the database, here we add `aiida`
as the prefix for the AiiDA software (www.aiida.net) in parallel
to the `mcloud` prefix for the Materials Cloud (www.materialscloud.org)
online database.

Moreover, I'm clarifying and improving the description of the
Materials Cloud entry.

### DIFF
--- a/providers.json
+++ b/providers.json
@@ -2,6 +2,15 @@
   "data": [
     {
       "type": "provider",
+      "id": "aiida",
+      "attributes": {
+        "name": "AiiDA",
+        "description": "AiiDA: Automated Interactive Infrastructure and Database for Computational Science (http://www.aiida.net)",
+        "base_url": null
+      }
+    },
+    {
+      "type": "provider",
       "id": "aflow",
       "attributes": {
         "name": "aflow.org",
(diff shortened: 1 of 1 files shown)


CasperWA:
> But maybe we want to define a separate field for `description_url`?

To what end? The `description` is meant as a short human-readable description of the provider, and the `base_url` should be (from our consortiums side) the URL of an Index meta-database, right? Or have I jumped into a previous discussion now that I didn't read up on? :)

rartino:
To the end that one may also want to provide a URL to the home page of the provider that isn't the base_url of the OPTiMaDe API implementation.

If you are asking why that URL cannot just be part of the human readable string, e.g., a UI that presents the list of providers may want to separate out the clickable links with info about them.

giovannipizzi:
Yes, I think that @rartino suggestion is very good - actually this was part of my original suggestion in    #46, see e.g. `'db_provider_homepage': 'http://example.com',` in point 4 of the description.
I vote for adding this in a new PR. If you agree, we should open an issue for this.

CasperWA:
> To the end that one may also want to provide a URL to the home page of the provider that isn't the base_url of the OPTiMaDe API implementation.

Ah, thanks for clarifying. This makes more sense to me. And I agree to adding this.
@giovannipizzi, if you want to add an issue for this, please go ahead. We can flip a coin over who makes the PR :)

giovannipizzi:
Ok, I do the issue and you do the PR ;-)

giovannipizzi:
Done, it's #147

PR #148, opened 2019-06-28
fix example json (remove spurious commas)

fawzi:


merkys:
Well spotted!

CasperWA:
Sorry @merkys for involuntarily dismissing your review. I simply wanted to add some more fixes I found, when probing a bit deeper into the document, when I was doing my review.

Thanks for this though @fawzi.

PR #149, opened 2019-06-28
format: json->jsonapi

fawzi:
The specification claims the format to be 'jsonapi', but the examples
use 'json'


CasperWA:
I believe this is already being addressed in PR #137. @fawzi, if you agree, then please close this PR (not merging).

fawzi:
yep handled in #137 (going the other way jsonapi -> json)

PR #154, opened 2019-07-03
Fix available_api_versions so the same version can be at more than one URL

### LINKED ISSUE #65: Is the dictionary of available_api_versions in the wrong order?
I'm running in some trouble with the `available_api_versions` property in the info endpoint. The example specify it like this:
```
        "available_api_versions": {
          "0.9.5": "http://db.example.com/optimade/v0.9/",
          "0.9.2": "http://db.example.com/optimade/v0.9.2/",
          "1.0.2": "http://db.example.com/optimade/v1.0/"
        },
```
This makes it impossible to serve the same API version on two different URLs. But, that seems a very natural thing to want to do. E.g., if one's implementation supports up to v0.9.5, then to let both `http://db.example.com/optimade/v0.9/` and `http://db.example.com/optimade/v0.9.5/` both serve `0.9.5` seems the right thing to do. 

So, would it not make more sense if we defined the dictionary the other way around?, i.e.,
```
        "available_api_versions": {
          "http://db.example.com/optimade/v0.9/": "0.9.5",
          "http://db.example.com/optimade/v0.9.5/": "0.9.5",
          "http://db.example.com/optimade/v0.9.2/": "0.9.2",
          "http://db.example.com/optimade/v1.0/": "1.0.2"
        },
```
Or, if we want the dictionary in this order, perhaps we can let the URL be a list?,
```
        "available_api_versions": {
          "0.9.5": ["http://db.example.com/optimade/v0.9/, http://db.example.com/optimade/v0.9.5/"],
          "0.9.2": ["http://db.example.com/optimade/v0.9.2/"],
          "1.0.2": ["http://db.example.com/optimade/v1.0/]"
        },
```

rartino:
Fixes #65 that at least I considered a release blocker for v0.10. 

(I'd really prefer not going back and messing with the version negotiation when implementations start to appear...)

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -777,9 +777,11 @@ The single resource object's response dictionary MUST include the following fiel
 * **id**: `"/"`
 * **attributes**: Dictionary containing the following fields:
   * **api\_version**: Presently used version of the OPTiMaDe API.
-  * **available\_api\_versions**: Dictionary where values are the base URLs for the versions of the
-  API that are supported, and the keys are strings giving the full version number provided by that
-  base URL. Provided base URLs MUST adhere to the rules in section [3.1. Base URL](#h.3.1).
+  * **available\_api\_versions**: A list of URLs where the same data is served using different versions of this API.
+    It MUST be a list of dictionaries, each containing the fields:
+	* **url**: a string specifying a base URL that MUST adhere to the rules in section [3.1. Base URL](#h.3.1)
+	* **version**: a string containing the full version number of the API served at that base URL. 
+	  The version number string SHOULD NOT be prefixed by, e.g., "v".
   * **formats**: List of available output formats.
   * **entry\_types\_by\_format**: Available entry endpoints as a function of output formats.
   * **available\_endpoints**: List of available endpoints (i.e., the string to be appended to the
@@ -819,11 +821,12 @@ Example:
     "id": "/",
     "attributes": {
       "api_version": "v0.9",
-      "available_api_versions": {
-        "0.9.5": "http://db.example.com/optimade/v0.9/",
-        "0.9.2": "http://db.example.com/optimade/v0.9.2/",
-        "1.0.2": "http://db.example.com/optimade/v1.0/"
-      },
+      "available_api_versions": [
+	    {"url": "http://db.example.com/optimade/v0.9/", "version": "0.9.5"},
+        {"url": "http://db.example.com/optimade/v1.0/", "version": "1.0.2"},
+        {"url": "http://db.example.com/optimade/v0.9.2/", "version": "0.9.2"},
+        {"url": "http://db.example.com/optimade/v0.9.5/", "version": "0.9.5"}
+      }},
       "formats": [
         "json",
         "xml"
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -777,9 +777,11 @@ The single resource object's response dictionary MUST include the following fiel
 * **id**: `"/"`
 * **attributes**: Dictionary containing the following fields:
   * **api\_version**: Presently used version of the OPTiMaDe API.
-  * **available\_api\_versions**: Dictionary where values are the base URLs for the versions of the
-  API that are supported, and the keys are strings giving the full version number provided by that
-  base URL. Provided base URLs MUST adhere to the rules in section [3.1. Base URL](#h.3.1).
+  * **available\_api\_versions**: A list of URLs where the same data is served using different versions of this API.
```
merkys: "list of URLs" and "list of dictionaries" sound confusing. Could the first sentence be left out?

optimade.md
```diff
@@ -777,9 +777,11 @@ The single resource object's response dictionary MUST include the following fiel
...
 * **attributes**: Dictionary containing the following fields:
   * **api\_version**: Presently used version of the OPTiMaDe API.
-  * **available\_api\_versions**: Dictionary where values are the base URLs for the versions of the
-  API that are supported, and the keys are strings giving the full version number provided by that
-  base URL. Provided base URLs MUST adhere to the rules in section [3.1. Base URL](#h.3.1).
+  * **available\_api\_versions**: A list of URLs where the same data is served using different versions of this API.
+    It MUST be a list of dictionaries, each containing the fields:
+	* **url**: a string specifying a base URL that MUST adhere to the rules in section [3.1. Base URL](#h.3.1)
+	* **version**: a string containing the full version number of the API served at that base URL. 
+	  The version number string SHOULD NOT be prefixed by, e.g., "v".
```
merkys: I'd say MUST. It's a simple transformation on the server-side, but the client-side will still need to check and add/drop the `v` if needed.

optimade.md
```diff
@@ -819,11 +821,12 @@ Example:
...
-        "0.9.5": "http://db.example.com/optimade/v0.9/",
-        "0.9.2": "http://db.example.com/optimade/v0.9.2/",
-        "1.0.2": "http://db.example.com/optimade/v1.0/"
-      },
+      "available_api_versions": [
+	    {"url": "http://db.example.com/optimade/v0.9/", "version": "0.9.5"},
+        {"url": "http://db.example.com/optimade/v1.0/", "version": "1.0.2"},
+        {"url": "http://db.example.com/optimade/v0.9.2/", "version": "0.9.2"},
+        {"url": "http://db.example.com/optimade/v0.9.5/", "version": "0.9.5"}
+      }},
```
merkys: Should be `],`.

optimade.md
rartino: I wanted an "explanation" for what the purpose was of this field beyond the formal schema, but it wasn't there before, so never mind. I've followed the suggestion.

optimade.md
rartino: Oooops

PR #156, opened 2019-07-11
Fix UnicodeHighChar comment

### LINKED ISSUE #153: UnicodeHighChar comment
I see that this comment has remained in the EBNF grammar:
```
(* The 'UnicodeHighChar' specifies all Unicode characters above 0x7F;
   the syntax used is the one compatible with Grammatica: *)

UnicodeHighChar = ? [^\x00-\x7F] ? ;
```

But, if I remember what @sauliusg  told me: we *thought* this syntax was compatible with Grammatica, but it is not due to what appears to be a bug in Java's regex implementation (present at least in the version of Java that @sauliusg was running...). This regex syntax is however very broadly supported across programming languages with regex support.

Hence, I suggest that we change that comment into something like:
```
(* A 'UnicodeHighChar' is any Unicode character above 0x7F. 
It is specified in this grammar by an extension to EBNF that allows a regular expression. *)
```

merkys:
Fixes #153.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -2344,8 +2344,9 @@ Space = ' ' | tab | nl | cr | vt | ff ;
 
 Spaces = Space, { Space } ;
 
-(* The 'UnicodeHighChar' specifies all Unicode characters above 0x7F;
-   the syntax used is the one compatible with Grammatica: *)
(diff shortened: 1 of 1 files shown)


merkys:
Thanks!

PR #158, opened 2019-07-15
Restructuring the entry listing info endpoint

### LINKED ISSUE #70: Extending "reserved keywords"
There may be a need/want to add reserved words. Now, only `info` is a reserved keyword.

> Technically, I think only `info` needs to be reserved in the sense described under `ID`, since the point is to not get collisions between, e.g., /structures/info and /structures/42. If we intend to actually reserve our other endpoint names here, then `extensions` should also be in this list. (In fact, it might be a very good idea to at least also reserve `extensions`, since it facilitates future use cases where someone needs to add endpoint-specific extensions...)

_Originally posted by @rartino in https://github.com/Materials-Consortia/OPTiMaDe/pull/68_

merkys:
Fixes #70

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -213,7 +213,7 @@ API v1.0](http://jsonapi.org/format/1.0) specification.
 All endpoints of an API implementation MUST be able to provide responses in the JSON format specified below
 and MUST respond in this format by default.
 
-Each endpoint MAY support additional formats, and SHOULD declare these formats under `info/<endpoint>` 
+Each endpoint MAY support additional formats, and SHOULD declare these formats under `/info/<entry type>`
 (see section [4.3.2. Entry Listing Info Endpoints](#h.4.3.2)).
 Clients can request these formats using the `response_format` URL query parameter.
 Specifying a `response_format` different from `json` (e.g.
--- a/optimade.md
+++ b/optimade.md
@@ -118,7 +118,6 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
   The following constraints and conventions apply to IDs:
   * Taken together, the ID and entry type MUST uniquely identify the entry.
   * IDs MUST be URL-safe strings; in particular, they MUST NOT contain commas.
-  * IDs MUST NOT be the word `info`.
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -547,6 +546,7 @@ following endpoints:
 * an "entry listing" endpoint
 * a "single entry" endpoint
 * an introspection `info` endpoint
+* an "entry listing" introspection `info` endpoint
```
CasperWA: Shouldn't this be "merged" with the line above instead? E.g.:

```suggestion
* a base and an "entry listing" introspection `info` endpoint
```

Or

```suggestion
* a common introspection `info` endpoint
```

merkys: Well, we have separate "entry listing endpoint" and "single entry endpoint", which, I believe, are `/<entry_type>` and `/<entry_type>/<id>`, respectively. Thus I'd either merge both, or none. What'd you say?

optimade.md
```diff
@@ -118,7 +118,6 @@ interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119).
   The following constraints and conventions apply to IDs:
   * Taken together, the ID and entry type MUST uniquely identify the entry.
   * IDs MUST be URL-safe strings; in particular, they MUST NOT contain commas.
```
rartino: Was not the idea to let this PR also remove the requirement that IDs MUST be URL-safe?

merkys: Right. I will address this in a separate PR.

merkys:
I have merged this PR for now despite the ongoing discussion about endpoint list (by @CasperWA) and remaining constraints on IDs (by @rartino). I see both remaining issues as somewhat logically separate from what has been done in this PR. I will open a follow-up PR for clearing up things concerning IDs.

optimade.md
CasperWA: After writing a long(ish) response here, I have deleted it and can appreciate your point of view now. The reason why is that the client will have to deal differently with an "entry listing" endpoint and a "single entry" endpoint, due to the way the response is set up for the `data` field.
The same is true for the `info` endpoint when dealing with the "base URL" `info` endpoint or the "entry listing" `info` endpoint(s).
So it makes sense to divide them here as well.

PR #159, opened 2019-07-15
Define returning properties with unknown values

### LINKED ISSUE #152: Rules for returning properties with unknown values
From PR #150 by @rartino, we have found a definition void, concerning when to return properties with unknown values, depending on whether they are REQUIRED in the response or OPTIONAL, and whether they have been specifically requested or not (using `response_fields`).

The following quote is my suggested policy:

>If a property is mandatory, but the database does not have the information, the property MUST be returned with value `null`, unless `response_fields` is used to omit it. However, if an OPTIONAL property has a `null` value it SHOULD be omitted from the response, unless specifically requested (using `response_fields`).

_Originally posted by @CasperWA in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjU2MjI1MTY2/pull_request_reviews/more_threads_

### LINKED ISSUE #163: Changes in 'Allowing for unknown/null coordinates' (eec3b0d) lost !
As I was responding to the conversation in #159, I discovered that we somehow have lost the changes introduced by @giovannipizzi about how `null` is allowed in `cartesian_site_positions` in eec3b0d !

**Edit:** I see that three commits show these lines as removed, 254e446, 1deb6e68dcdb4accc727dc7a5c2cb96ece775c0b, c0a72939644bf354608f55e5fac0d855c7d0442c (one of them a commit by me). It is *possible* I mis-merged this in a rebase in the chemical formula PR since that affected the same region of the text. I will go through those changes again and check that nothing else was lost.

CasperWA:
Fixes #152 
Fixes #163

REQUIRED properties with an unknown value MUST be returned (unless it is left out through the use of, e.g., `response_fields`).
OPTIONAL properties with an unknown value MAY be returned, but if they left out in a _full_ response, the client MUST assume they have an unknown value, i.e., is `null`.

Furthermore, I removed some EOL single trailing spaces and split up a line to keep with the sentiment of having a single line per complete sentence.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1617,8 +1617,9 @@ This section defines standard entry types and their properties.
 
 ### <a name="h.6.2.6">6.2.6. chemical\_formula\_hill</a>
 
-* **Description**: The chemical formula for a structure as a string in [Hill form](https://dx.doi.org/10.1021/ja02046a005) with element symbols followed by integer chemical proportion numbers.
+* **Description**: The chemical formula for a structure in [Hill form](https://dx.doi.org/10.1021/ja02046a005) with element symbols followed by integer chemical proportion numbers.
   The proportion number MUST be omitted if it is 1.
+* **Type**: string
 * **Requirements/Conventions**:
   * **Response**: OPTIONAL in the response.
   * **Query**: Support for queries on these properties are OPTIONAL.
--- a/optimade.md
+++ b/optimade.md
@@ -481,8 +481,9 @@ There are relevant use-cases for allowing data served via OPTiMaDe to be accesse
 Many databases allow specific data values to exist for some of the entries, whereas for others, no data value is present.
 This is referred to as the property having an *unknown* value, or equivalently, that the property value is `null`.
 
-The text in this section describes how the API handles properties that are `null`.
-The use of `null` values inside property data structures are described in the definitions of those data structures elsewhere in the specification, see section [6. Entry List](#h.6).
+The text in this section describes how the API handles properties with the value `null`.
+The use of `null` values inside nested property values (such as, e.g., lists or dictionaries) are described in the definitions of those data structures elsewhere in the specification, see section [6. Entry List](#h.6).
+For these properties, `null` MAY carry a special meaning.
 
 REQUIRED properties with an unknown value MUST be returned in the response, unless explicitly left out (e.g., by using `response_fields`, see section [4.4.1. URL Query Parameters](#h.4.4.1)).
 
--- a/optimade.md
+++ b/optimade.md
@@ -1941,24 +1941,21 @@ by multiple chemical elements.
     0.8\*0.3 = 24 % probability; and the pair (1, 3) with 0.8\*0.7 = 56 % probability).
 
 ### <a name="h.6.2.15">6.2.15. structure\_features</a>
+
 * **Description**: A list of strings that flag which special features are used by the structure.
-* **Type**: list of strings 
-* **Requirements/Conventions**: 
+* **Type**: list of strings
+* **Requirements/Conventions**:
   * **Response**: REQUIRED in the response (SHOULD be absent if there are no partial occupancies).
-  * **Query**: MUST be a queryable property. Filters on the list MUST support all mandatory HAS-type queries. 
-    Filter operators for comparisons on the string components MUST support equality, support for 
-    other comparison operators are OPTIONAL.
+  * **Query**: MUST be a queryable property. Filters on the list MUST support all mandatory HAS-type queries.
+    Filter operators for comparisons on the string components MUST support equality, support for other comparison operators are OPTIONAL.
   * MUST be an empty list if no special features are used.
   * MUST be sorted alphabetically.
   * If a special feature listed below is used, the list MUST contain the corresponding string.
   * If a special feature listed below is not used, the list MUST NOT contain the corresponding string.
   * **List of strings used to indicate special structure features**:
-    * `disorder`: This flag MUST be present if any one entry in the `species` list has a 
-      `chemical_symbols` list that is longer than 1 element.
-    * `unknown_positions`: This flag MUST be present if at least one component of the
-      `cartesian_site_positions` list of lists has value `null`.
-    * `assemblies`: This flag MUST be present if the [`assemblies`](#h.6.2.14)
-      list is present.  
+    * `disorder`: This flag MUST be present if any one entry in the `species` list has a `chemical_symbols` list that is longer than 1 element.
+    * `unknown_positions`: This flag MUST be present if at least one component of the `cartesian_site_positions` list of lists has value `null`.
+    * `assemblies`: This flag MUST be present if the [`assemblies`](#h.6.2.14) list is present.  
 * **Examples**: A structure having unknown positions and using assemblies:
 
   ```jsonc
--- a/optimade.md
+++ b/optimade.md
@@ -1699,13 +1699,19 @@ the Cartesian x, y, z directions.
 
 * **Description**: Cartesian positions of each site.
   A site is an atom, a site potentially occupied by an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual crystal approximation).
-* **Type**: list of list of floats
+* **Type**: list of list of floats and/or unknown values
 * **Requirements/Conventions**:
   * **Response**: REQUIRED in the response unless explicitly excluded.
   * **Query**: Support for queries on this property is OPTIONAL.
     If supported, filters MAY support only a subset of comparison operators.
   * It MUST be a list of length N times 3, where N is the number of sites in the structure.
-  * An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see e.g., the [6.2.14. `assemblies`](#h.6.2.14) property).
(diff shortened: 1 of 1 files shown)


optimade.md
```diff
@@ -477,23 +477,27 @@ to the corresponding database ID that was originally queried, using the object's
...
 ### <a name="h.3.3.5">3.3.5. Properties with unknown value</a>
 
-Many databases allow specific data values to exist for some of the entries, whereas for others, no data value is present. 
-This is referred to as the property having an unknown value, or equivalently, that the property value is `null`.
+Many databases allow specific data values to exist for some of the entries, whereas for others, no data value is present.
+This is referred to as the property having an *unknown* value, or equivalently, that the property value is `null`.
 
-Properties with an unknown value MUST NOT be returned in the response, unless explicitly requested in the search query. 
+The text in this section describes how the API handles properties that are `null`.
+It does not regulate the handling of values inside property data structures that can be `null`.
```
merkys: I would suggest remove this sentence, as the following one describes where to look for `null` regulation.

optimade.md
```diff
@@ -477,23 +477,27 @@ to the corresponding database ID that was originally queried, using the object's
...
+The use of `null` values inside property data structures are described in the definitions of those data structures elsewhere in the specification.
```
merkys: I would explicitly add a link to [6. Entry List](#h.6).

optimade.md
sauliusg: I would also agree on removing the "It does not regulate the handling of values inside property data structures that can be `null`." sentence; see my comment below.

optimade.md
sauliusg: I agree, the link would be helpful.

CasperWA:
> > It does not regulate the handling of values inside property data structures that can be `null`.
> 
> I'm not exactly sure what this means. In my understanding, the specification describes what values MUST, SHOULD or MAY be returned when a valid request is submitted. These values then MUST be expected and correctly handled by a client; in particular, the client MAY NOT "hang" or otherwise crash, or complain with an error message, when a property with a `null` value is returned by an OPTiMaDe server. To that extent, the specification mandates the handling of `null` values; it can not be otherwise.
> 
> Of course the specification does not say what the client should do with these `null` values once they are recognised; the client my ignore them, or store them explicitly, or substitute by defaults, or what not. I think this is application dependent. If this is what the above sentence means, then to me it is an pbvious consequence of the scope of the OPTiMaDe specification.
> 
> Thus, I would suggest to remove this sentence.

The sentence is originally written by @rartino, I merely moved it further up in this section (it was last before) and may have no meaning anymore with my changes?
I will agree it is not easy to parse for me either, since I am also unsure what exactly is meant by "property data structures". I will wait a bit for @rartino to respond and then I will try and clarify the sentence - or remove it :)

optimade.md
CasperWA: I have answered this below as well, calling on the org. author of this, @rartino :)

rartino:
> It does not regulate the handling of values inside property data structures that can be null.

This formulation does not originate from me, but (I think), @ctoher. 

Nevertheless, I believe I understand what it says: the directives in this section only describe conformity with regards to "the whole property" being unknown, as in: `chemical_formula_hill` = `null`. In addition to what is being said in this section, one should be prepared on that section 6 defines properties as lists or dictionaries, in which `null` can carry a special meaning. E.g., the definition of `cartesian_site_positions` may say that a null value indicates an unknown position.

I'd be happy to see a reformulation in this PR of the sentence that explains this better. (Or as a separate PR.)

CasperWA:
I have tried to divide my additions and minor corrections into separate commits, i.e., you should be able to see the _actual_ changes, simply disregard the changes in commit 34157ee.

CasperWA:
> > It does not regulate the handling of values inside property data structures that can be null.
> 
> This formulation does not originate from me, but (I think), @ctoher.

My bad, sorry.

> I'd be happy to see a reformulation in this PR of the sentence that explains this better. (Or as a separate PR.)

Hereby done with the last commit, 67beb06.

rartino:
I have approved, and this PR only waits for @sauliusg to dismiss his change request, because I think the PR is a great improvement of what is in the spec right now.

However, before merging, maybe we should thinking about this:

> REQUIRED properties with an unknown value MUST be returned in the response, unless explicitly left out (e.g., by using `response_fields`, see section [4.4.1. URL Query Parameters](#h.4.4.1)).

Does this mean that any REQUIRED property may be returned with the value `null`? Because then that means a client cannot rely on any property being set to the expected value. Or does this only apply to REQUIRED properties that specify that they may be set to `null`? (Because we have none of those...)

This seems quite unclear presently, which isn't great for the specification.

sauliusg:
@rartino wrote:
> this PR puts back all_coordinates_known, which it should not. The text about unknown coordinates should be returned to the state it had in commit d1b7c1e

I did not dig deep into the discussions about the `all_coordinates_known` property, but I agree that it is a kind of obsolete and duplicates information of the coordinate section (and raises the question what do we do if the `all_coordinates_known` value contradicts what is actually presented in the coordinates).

I suggest that this is sorted out and then @rartino gives the final review.

CasperWA:
> This version looks good to me. Thanks for fixing #163 in this (right, that is intentional?)

Yes that is intentional. I will add it to the "Fixes xxx" in the first comment.

CasperWA:
I have now changed "Giovanni's" commit, removing `all_coordinates_known`.
It seems all should be okay now according to the PRs mentioned in my previous comment. However, if any of you find an even newer change, please don't refrain from making yourselves heard.

CasperWA:
> So is "the database just can't provide this piece of information" == "value for this entry is unknown"? I feel this is true, but want to clarify.

Your feeling do not betray you. This is what the spec should say now (and probably even did before) :)

CasperWA:
I will just wait to hear from @rartino if I fixed the issue of removing again `all_coordinates_known` properly and whatever else I may have cooked up :sweat_smile:

optimade.md
```diff
@@ -1704,24 +1699,29 @@ the Cartesian x, y, z directions.
...
-  * **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
+  * **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
   * It MUST be a list of length N times 3, where N is the number of sites in the structure.
-  * An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see
-  e.g., the [6.2.14. `assemblies`](#h.6.2.14) property).
+  * An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see e.g., the [6.2.15. `assemblies`](#h.6.2.15) property).
+  * If a component of the position is unknown, the `null` value should be provided instead (see section [3.3.5. Properties with unknown value](#h.3.3.5)).
+    Otherwise, it should be a float value, expressed in angstrom.
+    (See also the property `all_coordinates_known` under section [6.2.11](#h.6.2.11).)
```
rartino: This text still references `all_coordinates_known` under `cartesian_site_positions`....

CasperWA:
> Thanks for the continued work on this! But, sorry, it is not quite there yet. Your fix of #163 still leaves references to `all_coordinates_known` in the description of `cartesian_site_positions`.

Thanks for catching this!

> Note that apart from bringing in the relevant change in this section from [d1b7c1e](https://github.com/Materials-Consortia/OPTiMaDe/commit/d1b7c1e707685bb8992b0b98b235c337f2996c10), I have here reverted your own addition of the type being specified as "list of list of floats and/or unknown values" and gone back to "list of list of floats". This is because the way I look at types the **type** of this property still is "list of list of floats", only, just like in any list of floats (according to our term definitions), elements in that list are allowed to be `null`/unknown. (In my opinion this is the the most correct way to do it, but I also don't feel extremely strongly against the explicit mentioning of unknowns if others want that.)

Since `null` or unknown value has a very specific meaning in `cartesian_site_positions`, I felt it was appropriate to add it.
Whereas for other properties, an unknown value, is exactly that, a value to represent the property's value is unknown. However, for `cartesian_site_positions` one may actively use the unknown value to represent a specific type of species.
What do you think?

rartino:
> Since `null` or unknown value has a very specific meaning in `cartesian_site_positions`, I felt it was appropriate to add it.
> Whereas for other properties, an unknown value, is exactly that, a value to represent the property's value is unknown. However, for `cartesian_site_positions` one may actively use the unknown value to represent a specific type of species.

But the importance and meaning of `null` has very little to do with the **type**. In a type-strong language when declaring types, they are not influenced by concerns like this. I could see your point if we were labeling this "structure". But in my opinion whether items can be null or not is not part of the type.

From a more practical standpoint, I will probably want to use these in a data structure in my implementation to be used for automatic type conformity checks. For this the "comment" about nulls is a distraction I will have to remove when I build that data structure.

But, I've already approved the PR because I don't feel very strongly about this. So, if you disagree, go ahead and merge.

sauliusg:
> Philosophical question: suppose a database provider decides/is forced to not provide values for REQUIRED property, say `chemical_formula_anonymous`. So is "the database just can't provide this piece of information" == "value for this entry is unknown"? I feel this is true, but want to clarify.

I think in this case it is true if we clarify: "value for this entry is unknown **to the database**".

Issue #91, opened 2019-06-12
Introduce compatibility levels


sauliusg:
Currently, some features of the OPTiMaDe specification are mandatory, and some are optional. For N optional features that databases choose to implement or not implement, we have 2^N combinations of unimplemented features...

To minimise the burden for a client of finding out which queries are compatible with which implementations, I suggest specifying **compatibility levels** of Optimade implementations. Each compatibility level would be designated by a whole number (0,1,2,3,...). It is to be decided how many levels will be specified (probably there should not be too many, probably no more than 3-6).

Each compatibility level will have a list of features that MUST be implemented by an Optimade engine to claim this particular compatibility level. Each higher level will suppose implementing features of all lower levels. Features that are grouped into particular levels should be related, and should be approximately equally easy/diffucult to implement.

The 0-th compatibility level implementation MUST implement only mandatory features.

merkys:
I am a bit skeptical about this proposal:
1. it will be difficult to decide on the level boundaries;
2. a database implementing all features of levels up to 20, but lacking a single feature in level 2, will be classified as of level 1.

One may define complexity levels having our current *N* backend databases in mind, but these levels might be completely different for the (*N+1*)th backend.

sauliusg:
@andrius: an update
I agree. After our discussion and the talk with @rartino I am convinced by you that a much better strategy is to assign not __compatibility levels__, but **feature sets** (FS). There might be a feature set "database versioning", another feature set "trajectory querries" and so on. The feature sets would list features that an OPTiMaDe implementation that implements the particular FS has all implemented according to the Optimade spec. The FSs need not be disjoint. They are not hierarchical, and any implementer can choose which feature sets to implement (or if to implement any at all). On the "core set" of mandatory features MUST be provided by all implementations.

In this way we can easily add or reshape  feature sets when new needs of new technological possibilities arise in the future.

The feature sets are intended to group features that are useful for particular application area, e.g. mol. dynamics, DFT computations, experimental DB validation, versioned querries, etc., etc.,

Issue #87, opened 2019-06-12
LIKE operators


sauliusg:
It has been suggested (Issue #42, PR #69) to have operator similar to SQL "LIKE" in a query language.
Since there are reservations about implementing 'LIKE "%string%"' construct as being too much geared towards a particular DB implementation, and 'LIKE "*string*"' would be inventing something superfluously similar to the SQL LIKE but different enough to be confusing, what about the  following syntax:r

`some_property MATCHES "*C*"`

and interpret it as, say a Shell GLOB?

The idea is to have a query expression that is simpler and more widely supported in database engines than full-featured RegExps, but still useful for general substring searches,

sauliusg:
It seems from the issue #42 discussions that we have the following possibilities for querrying substrings:

- STARTS WITH/ENDS WITH/CONTAINS;

- LIKE with SQL-like semantics;

- REGEXP of different kinds (PCRE, ERE, with or without Unicode support).

each has increasing expressive power, but also may be increasingly different to implement on back-ends without the native support (for instance, MySQL at some stage required a plug-in to support regexp matches).

I suggest defining Filter language syntax and the following compatibility requirements:

- STARTS WITH/ENDS WITH/CONTAINS – **required** (as it is now);

- LIKE with the SQL-like semantics – **optional**;

- MATCH GLOB, MATCH PCRE, MATCH ERE (e.g. `_cod_chemical_name MATCH PCER "µ.*[Oo]cta[^\s]+[23]"`) – **optional**.

In this way, all backends will provide the same basic STARTS/ENDS/CONTAINS functionality (which is easy to implement using LIKE of SQL, REGEXPs or even a post-filtering); but we will also be able to introduce efficient extensions in a compatible way.

What others think?

merkys:
This issue was brought up by @ml-evs in today's Web meeting. @rartino asked the providers to express their opinions about it.

Personally, I like the proposal to support different string matching languages (SQL LIKE, GLOB, PCRE etc). However, I think that here we again encounter the same dilemma where by allowing diversity among providers we transfer the burden onto the clients. I easily image every provider preferring a different string matching language, and only that one.

SQL LIKE seems least demanding and, as such, least powerful. Maybe we can start from that and move on to the next ones?

By the way, I would rename `LIKE` operator to `MATCHES SQLLIKE` (not sure if `_` in operator name would work) just to have simpler grammar. `UNLIKE` is the same as `NOT LIKE`, right?

[...]

PR #160, opened 2019-07-15
Like operator

sauliusg:
Offering implementation of the "LIKE", "NOT LIKE" and "UNLIKE" string matching operators in the Filter grammar, and briefly describing their semantics.

Alternatively, offering implementation of the "MATCH /regexp/" operator.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -1300,7 +1300,8 @@ The `x` string MUST be interpreted as a pattern where an underscore character ('
 matches any single character and a percent character ('%', ASCII DEC 37, HEX 25) matches an arbitrary
 sequence of characters (including zero characters).
 
-If operator "UNLIKE" is supported, the bahavior of this oprtator MUST be the negation of the "LIKE" operator; i.e.
+If the `UNLIKE` operator is supported, the behavior of this operator MUST be the negation of the "LIKE" operator; i.e.
+
 an expression `(property UNLIKE "value")" must behave exactly as `(NOT (property LIKE "value"))`.
 
 Examples:
--- a/optimade.md
+++ b/optimade.md
@@ -1298,7 +1298,21 @@ In addition to the standard equality and inequality operators, matching of parti
 
 OPTIONAL features: 
 
+The following comparison operators are OPTIONAL:
+
+* `identfier LIKE x`
+
+* `identfier UNLIKE x`
+
 * Support for x to be an identifier, rather than a string is OPTIONAL.
+
+If implemented, the "LIKE" operator MUST behave as the correspoding standard SQL operator. In particular,
+The `x` string MUST be interpreted as a pattern where an underscore character ('_', ASCII DEC 95, HEX 5F)
+matches any single character and a percent character ('%', ASCII DEC 37, HEX 25) matches an arbitrary
+sequence of characters (including zero characters).
+
+If operator "UNLIKE" is supported, the bahavior of this oprtator MUST be the negation of the "LIKE" operator; i.e.
+an expression `(property UNLIKE "value")" must behave exactly as `(NOT (property LIKE "value"))`.
 
 Examples:
 
@@ -2218,7 +2232,8 @@ ValueOpRhs = Operator, Value ;
 
 KnownOpRhs = IS, ( KNOWN | UNKNOWN ) ; 
 
-FuzzyStringOpRhs = CONTAINS, String | STARTS, [ WITH ], String | ENDS, [ WITH ], String ;
+FuzzyStringOpRhs = CONTAINS, String | STARTS, [ WITH ], String | ENDS, [ WITH ], String |
+                   LIKE, String | NOT, LIKE, String | UNLIKE, String ;
 
 SetOpRhs = HAS, ( [ Operator ], Value | ALL, ValueList | ANY, ValueList | ONLY, ValueList ) ;
 (* Note: support for ONLY in SetOpRhs is OPTIONAL *)
@@ -2265,6 +2280,9 @@ HAS = 'H', 'A', 'S', [Spaces] ;
 ALL = 'A', 'L', 'L', [Spaces] ;
 ONLY = 'O', 'N', 'L', 'Y', [Spaces] ;
 ANY = 'A', 'N', 'Y', [Spaces] ;
+
+LIKE = 'L', 'I', 'K', 'E', [Spaces];
+UNLIKE = 'U', 'N', 'L', 'I', 'K', 'E', [Spaces];
 
 (* OperatorComparison operator tokens: *)
 
--- a/optimade.md
+++ b/optimade.md
@@ -2207,8 +2207,8 @@ StringProperty = String | Property ;
 FuzzyStringOpRhs = CONTAINS, StringProperty |
                    STARTS, [ WITH ], StringProperty |
                    ENDS, [ WITH ], StringProperty |
-                   MATCH, RegularExpression |
-                   NOT, MATCH, RegularExpression ;
+                   MATCH, ( RegularExpression | StringProperty ) |
+                   NOT, MATCH, ( RegularExpression | StringProperty ) ;
 
 SetOpRhs = HAS, ( [ Operator ], Value | ALL, ValueList | ANY, ValueList | ONLY, ValueList ) ;
 (* Note: support for ONLY in SetOpRhs is OPTIONAL *)
--- a/optimade.md
+++ b/optimade.md
@@ -2202,8 +2202,13 @@ ValueOpRhs = Operator, Value ;
 
 KnownOpRhs = IS, ( KNOWN | UNKNOWN ) ; 
 
-FuzzyStringOpRhs = CONTAINS, String | STARTS, [ WITH ], String | ENDS, [ WITH ], String |
-                   LIKE, String | NOT, LIKE, String | UNLIKE, String ;
+StringProperty = String | Property ;
+
+FuzzyStringOpRhs = CONTAINS, StringProperty |
+                   STARTS, [ WITH ], StringProperty |
+                   ENDS, [ WITH ], StringProperty |
+                   MATCH, RegularExpression |
+                   NOT, MATCH, RegularExpression ;
 
 SetOpRhs = HAS, ( [ Operator ], Value | ALL, ValueList | ANY, ValueList | ONLY, ValueList ) ;
 (* Note: support for ONLY in SetOpRhs is OPTIONAL *)
@@ -2251,8 +2256,7 @@ ALL = 'A', 'L', 'L', [Spaces] ;
 ONLY = 'O', 'N', 'L', 'Y', [Spaces] ;
 ANY = 'A', 'N', 'Y', [Spaces] ;
 
-LIKE = 'L', 'I', 'K', 'E', [Spaces];
-UNLIKE = 'U', 'N', 'L', 'I', 'K', 'E', [Spaces];
+MATCH = 'M', 'A', 'T', 'C', 'H', [Spaces];
 
 (* OperatorComparison operator tokens: *)
 
@@ -2280,15 +2284,33 @@ LowercaseLetter =
 
 String = '"', { EscapedChar }, '"', [Spaces] ;
 
+UnescapedChar = Letter | Digit | Space | '/' |
+                Punctuator | RegexpMetacharacter |
+                UnicodeHighChar ;
+
 EscapedChar = UnescapedChar | '\', '"' | '\', '\' ;
 
-UnescapedChar = Letter | Digit | Space | Punctuator | UnicodeHighChar ;
-
 Punctuator =
-    '!' | '#' | '$' | '%' | '&' | "'" | '(' | ')' | '*' | '+' | ',' |
-    '-' | '.' | '/' | ':' | ';' | '<' | '=' | '>' | '?' | '@' | '[' |
-    ']' | '^' | '`' | '{' | '|' | '}' | '~'
+    '!' | '#' | '%' | '&' | "'" | ',' |
+    '-' | ':' | ';' | '<' | '=' | '>' | '@' |
+    '`' | '~'
 ;
+
+RegexpMetacharacter =
+    '(' | ')' | '[' | ']' | '+' | '*' | '?' | '.' | '{' | '}' |
+    '|' | '^' | '$'
+;
+
+(* Regular expressions: *)
+
+UnescapedREChar = Letter | Digit | Space | '"' |
+                  Punctuator | RegexpMetacharacter |
+                  UnicodeHighChar ;
+
+EscapedREChar = UnescapedREChar | '\', '/' | '\', '\' |
+                '\', RegexpMetacharacter ;
+
+RegularExpression = '/', { EscapedREChar }, '/', [Spaces] ;
 
 (* BEGIN EBNF GRAMMAR Number *)
 (* Number token syntax: *)
--- a/tests/cases/Filter_072.inp
+++ b/tests/cases/Filter_072.inp
@@ -1,0 +1,1 @@
+chemical_formula LIKE "H2 O2"
--- a/tests/cases/Filter_072.opt
+++ b/tests/cases/Filter_072.opt
@@ -1,0 +1,1 @@
+Filter_001.opt
--- a/tests/cases/Filter_073.inp
+++ b/tests/cases/Filter_073.inp
@@ -1,0 +1,1 @@
+chemical_formula NOT LIKE "C6 H12 O6"
--- a/tests/cases/Filter_073.opt
+++ b/tests/cases/Filter_073.opt
@@ -1,0 +1,1 @@
+Filter_001.opt
--- a/tests/cases/Filter_074.inp
+++ b/tests/cases/Filter_074.inp
@@ -1,0 +1,1 @@
+chemical_formula UNLIKE "C6 H12 O6"
--- a/tests/cases/Filter_074.opt
+++ b/tests/cases/Filter_074.opt
@@ -1,0 +1,1 @@
+Filter_001.opt
--- a/tests/cases/Filter_075.inp
+++ b/tests/cases/Filter_075.inp
@@ -1,0 +1,1 @@
+property MATCH /"^.?abc+[a-z0-9]*$"/
--- a/tests/cases/Filter_075.opt
+++ b/tests/cases/Filter_075.opt
@@ -1,0 +1,1 @@
+Filter_001.opt
--- a/tests/cases/Filter_076.inp
+++ b/tests/cases/Filter_076.inp
@@ -1,0 +1,1 @@
+property MATCH "\"^.?abc+[a-z0-9]\\*$\""
--- a/tests/cases/Filter_076.opt
+++ b/tests/cases/Filter_076.opt
@@ -1,0 +1,1 @@
+Filter_001.opt
--- a/tests/outputs/Filter_019.out
+++ b/tests/outputs/Filter_019.out
@@ -56,7 +56,9 @@ Filter(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 26
 Error: in tests/cases/Filter_019.inp: line 1:
-    unexpected token "4", expected """
+    unexpected token "4", expected one of """, "a", "b", "c", "d",
+    "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q",
+    "r", "s", "t", "u", "v", "w", "x", "y", "z", or "_"
 
 chemical_formula CONTAINS 42
                           ^
--- a/tests/outputs/Filter_020.out
+++ b/tests/outputs/Filter_020.out
@@ -56,7 +56,9 @@ Filter(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 26
 Error: in tests/cases/Filter_020.inp: line 1:
-    unexpected token "S", expected """
+    unexpected token "S", expected one of """, "a", "b", "c", "d",
+    "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q",
+    "r", "s", "t", "u", "v", "w", "x", "y", "z", or "_"
 
 chemical_formula CONTAINS STARTS "Al"
                           ^
--- a/tests/outputs/Filter_021.out
+++ b/tests/outputs/Filter_021.out
@@ -55,22 +55,23 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 26
-              String(9999)
-                TOKEN_3(9999): """, line: 1, col: 27
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      UppercaseLetter(9999)
-                        TOKEN_33(9999): "A", line: 1, col: 28
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      LowercaseLetter(9999)
-                        TOKEN_76(9999): "l", line: 1, col: 29
-                TOKEN_3(9999): """, line: 1, col: 30
-                Spaces(9999)
-                  Space(9999)
-                    TOKEN_1(9999): " ", line: 1, col: 31
+              StringProperty(9999)
+                String(9999)
+                  TOKEN_3(9999): """, line: 1, col: 27
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Letter(9999)
+                        UppercaseLetter(9999)
+                          TOKEN_33(9999): "A", line: 1, col: 28
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Letter(9999)
+                        LowercaseLetter(9999)
+                          TOKEN_76(9999): "l", line: 1, col: 29
+                  TOKEN_3(9999): """, line: 1, col: 30
+                  Spaces(9999)
+                    Space(9999)
+                      TOKEN_1(9999): " ", line: 1, col: 31
       AND(9999)
         TOKEN_33(9999): "A", line: 1, col: 32
         TOKEN_46(9999): "N", line: 1, col: 33
@@ -130,22 +131,23 @@ Filter(9999)
                   Spaces(9999)
                     Space(9999)
                       TOKEN_1(9999): " ", line: 1, col: 59
-                String(9999)
-                  TOKEN_3(9999): """, line: 1, col: 60
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Letter(9999)
-                        UppercaseLetter(9999)
-                          TOKEN_33(9999): "A", line: 1, col: 61
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Letter(9999)
-                        LowercaseLetter(9999)
-                          TOKEN_76(9999): "l", line: 1, col: 62
-                  TOKEN_3(9999): """, line: 1, col: 63
-                  Spaces(9999)
-                    Space(9999)
-                      TOKEN_1(9999): " ", line: 1, col: 64
+                StringProperty(9999)
+                  String(9999)
+                    TOKEN_3(9999): """, line: 1, col: 60
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Letter(9999)
+                          UppercaseLetter(9999)
+                            TOKEN_33(9999): "A", line: 1, col: 61
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Letter(9999)
+                          LowercaseLetter(9999)
+                            TOKEN_76(9999): "l", line: 1, col: 62
+                    TOKEN_3(9999): """, line: 1, col: 63
+                    Spaces(9999)
+                      Space(9999)
+                        TOKEN_1(9999): " ", line: 1, col: 64
         AND(9999)
           TOKEN_33(9999): "A", line: 1, col: 65
           TOKEN_46(9999): "N", line: 1, col: 66
@@ -203,20 +205,21 @@ Filter(9999)
                     Spaces(9999)
                       Space(9999)
                         TOKEN_1(9999): " ", line: 1, col: 90
-                  String(9999)
-                    TOKEN_3(9999): """, line: 1, col: 91
-                    EscapedChar(9999)
-                      UnescapedChar(9999)
-                        Letter(9999)
-                          UppercaseLetter(9999)
-                            TOKEN_33(9999): "A", line: 1, col: 92
-                    EscapedChar(9999)
-                      UnescapedChar(9999)
-                        Letter(9999)
-                          LowercaseLetter(9999)
-                            TOKEN_76(9999): "l", line: 1, col: 93
-                    TOKEN_3(9999): """, line: 1, col: 94
-                    Spaces(9999)
-                      Space(9999)
-                        nl(9999)
-                          SPECIAL_1(9999): "(...)", line: 1, col: 95
+                  StringProperty(9999)
+                    String(9999)
+                      TOKEN_3(9999): """, line: 1, col: 91
+                      EscapedChar(9999)
+                        UnescapedChar(9999)
+                          Letter(9999)
+                            UppercaseLetter(9999)
+                              TOKEN_33(9999): "A", line: 1, col: 92
+                      EscapedChar(9999)
+                        UnescapedChar(9999)
+                          Letter(9999)
+                            LowercaseLetter(9999)
+                              TOKEN_76(9999): "l", line: 1, col: 93
+                      TOKEN_3(9999): """, line: 1, col: 94
+                      Spaces(9999)
+                        Space(9999)
+                          nl(9999)
+                            SPECIAL_1(9999): "(...)", line: 1, col: 95
--- a/tests/outputs/Filter_022.out
+++ b/tests/outputs/Filter_022.out
@@ -44,9 +44,12 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 18
+            FuzzyStringOpRhs(9999)
+              UNLIKE(9999)
+                TOKEN_53(9999): "U", line: 1, col: 19
+                TOKEN_46(9999): "N", line: 1, col: 20
 Error: in tests/cases/Filter_022.inp: line 1:
-    unexpected token "U", expected one of "<", ">", "=", "!", "I",
-    "C", "S", "E", "H", or ":"
+    unexpected token "K", expected "L"
 
 prototype_formula UNKNOWN
-                  ^
+                    ^
--- a/tests/outputs/Filter_022.out
+++ b/tests/outputs/Filter_022.out
@@ -44,12 +44,9 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 18
-            FuzzyStringOpRhs(9999)
-              UNLIKE(9999)
-                TOKEN_53(9999): "U", line: 1, col: 19
-                TOKEN_46(9999): "N", line: 1, col: 20
 Error: in tests/cases/Filter_022.inp: line 1:
-    unexpected token "K", expected "L"
+    unexpected token "U", expected one of "<", ">", "=", "!", "I",
+    "C", "S", "E", "M", "N", "H", or ":"
 
 prototype_formula UNKNOWN
-                    ^
+                  ^
--- a/tests/outputs/Filter_023.out
+++ b/tests/outputs/Filter_023.out
@@ -28,7 +28,7 @@ Filter(9999)
                     TOKEN_1(9999): " ", line: 1, col: 9
 Error: in tests/cases/Filter_023.inp: line 1:
     unexpected token "f", expected one of "<", ">", "=", "!", "I",
-    "C", "S", "E", "H", or ":"
+    "C", "S", "E", "L", "N", "U", "H", or ":"
 
 chemical formula IS KNOWN 42
          ^
--- a/tests/outputs/Filter_023.out
+++ b/tests/outputs/Filter_023.out
@@ -28,7 +28,7 @@ Filter(9999)
                     TOKEN_1(9999): " ", line: 1, col: 9
 Error: in tests/cases/Filter_023.inp: line 1:
     unexpected token "f", expected one of "<", ">", "=", "!", "I",
-    "C", "S", "E", "L", "N", "U", "H", or ":"
+    "C", "S", "E", "M", "N", "H", or ":"
 
 chemical formula IS KNOWN 42
          ^
--- a/tests/outputs/Filter_024.out
+++ b/tests/outputs/Filter_024.out
@@ -44,7 +44,7 @@ Filter(9999)
                     TOKEN_1(9999): " ", line: 1, col: 17
 Error: in tests/cases/Filter_024.inp: line 1:
     unexpected token "K", expected one of "<", ">", "=", "!", "I",
-    "C", "S", "E", "H", or ":"
+    "C", "S", "E", "L", "N", "U", "H", or ":"
 
 chemical_formula KNOWN
                  ^
--- a/tests/outputs/Filter_024.out
+++ b/tests/outputs/Filter_024.out
@@ -44,7 +44,7 @@ Filter(9999)
                     TOKEN_1(9999): " ", line: 1, col: 17
 Error: in tests/cases/Filter_024.inp: line 1:
     unexpected token "K", expected one of "<", ">", "=", "!", "I",
-    "C", "S", "E", "L", "N", "U", "H", or ":"
+    "C", "S", "E", "M", "N", "H", or ":"
 
 chemical_formula KNOWN
                  ^
--- a/tests/outputs/Filter_028.out
+++ b/tests/outputs/Filter_028.out
@@ -26,9 +26,11 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 9
+            FuzzyStringOpRhs(9999)
+              LIKE(9999)
+                TOKEN_44(9999): "L", line: 1, col: 10
 Error: in tests/cases/Filter_028.inp: line 1:
-    unexpected token "L", expected one of "<", ">", "=", "!", "I",
-    "C", "S", "E", "H", or ":"
+    unexpected token "E", expected "I"
 
 elements LENGTH 42
-         ^
+          ^
--- a/tests/outputs/Filter_028.out
+++ b/tests/outputs/Filter_028.out
@@ -26,11 +26,9 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 9
-            FuzzyStringOpRhs(9999)
-              LIKE(9999)
-                TOKEN_44(9999): "L", line: 1, col: 10
 Error: in tests/cases/Filter_028.inp: line 1:
-    unexpected token "E", expected "I"
+    unexpected token "L", expected one of "<", ">", "=", "!", "I",
+    "C", "S", "E", "M", "N", "H", or ":"
 
 elements LENGTH 42
-          ^
+         ^
--- a/tests/outputs/Filter_059.out
+++ b/tests/outputs/Filter_059.out
@@ -57,21 +57,22 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 27
-              String(9999)
-                TOKEN_3(9999): """, line: 1, col: 28
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      UppercaseLetter(9999)
-                        TOKEN_35(9999): "C", line: 1, col: 29
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Digit(9999)
-                      TOKEN_18(9999): "2", line: 1, col: 30
-                TOKEN_3(9999): """, line: 1, col: 31
-                Spaces(9999)
-                  Space(9999)
-                    TOKEN_1(9999): " ", line: 1, col: 32
+              StringProperty(9999)
+                String(9999)
+                  TOKEN_3(9999): """, line: 1, col: 28
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Letter(9999)
+                        UppercaseLetter(9999)
+                          TOKEN_35(9999): "C", line: 1, col: 29
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Digit(9999)
+                        TOKEN_18(9999): "2", line: 1, col: 30
+                  TOKEN_3(9999): """, line: 1, col: 31
+                  Spaces(9999)
+                    Space(9999)
+                      TOKEN_1(9999): " ", line: 1, col: 32
       AND(9999)
         TOKEN_33(9999): "A", line: 1, col: 33
         TOKEN_46(9999): "N", line: 1, col: 34
@@ -141,15 +142,16 @@ Filter(9999)
                   Spaces(9999)
                     Space(9999)
                       TOKEN_1(9999): " ", line: 1, col: 66
-                String(9999)
-                  TOKEN_3(9999): """, line: 1, col: 67
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Letter(9999)
-                        UppercaseLetter(9999)
-                          TOKEN_33(9999): "A", line: 1, col: 68
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Digit(9999)
-                        TOKEN_18(9999): "2", line: 1, col: 69
-                  TOKEN_3(9999): """, line: 1, col: 70
+                StringProperty(9999)
+                  String(9999)
+                    TOKEN_3(9999): """, line: 1, col: 67
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Letter(9999)
+                          UppercaseLetter(9999)
+                            TOKEN_33(9999): "A", line: 1, col: 68
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Digit(9999)
+                          TOKEN_18(9999): "2", line: 1, col: 69
+                    TOKEN_3(9999): """, line: 1, col: 70
--- a/tests/outputs/Filter_060.out
+++ b/tests/outputs/Filter_060.out
@@ -55,21 +55,22 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 25
-              String(9999)
-                TOKEN_3(9999): """, line: 1, col: 26
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      UppercaseLetter(9999)
-                        TOKEN_34(9999): "B", line: 1, col: 27
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Digit(9999)
-                      TOKEN_18(9999): "2", line: 1, col: 28
-                TOKEN_3(9999): """, line: 1, col: 29
-                Spaces(9999)
-                  Space(9999)
-                    TOKEN_1(9999): " ", line: 1, col: 30
+              StringProperty(9999)
+                String(9999)
+                  TOKEN_3(9999): """, line: 1, col: 26
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Letter(9999)
+                        UppercaseLetter(9999)
+                          TOKEN_34(9999): "B", line: 1, col: 27
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Digit(9999)
+                        TOKEN_18(9999): "2", line: 1, col: 28
+                  TOKEN_3(9999): """, line: 1, col: 29
+                  Spaces(9999)
+                    Space(9999)
+                      TOKEN_1(9999): " ", line: 1, col: 30
       AND(9999)
         TOKEN_33(9999): "A", line: 1, col: 31
         TOKEN_46(9999): "N", line: 1, col: 32
@@ -137,15 +138,16 @@ Filter(9999)
                   Spaces(9999)
                     Space(9999)
                       TOKEN_1(9999): " ", line: 1, col: 62
-                String(9999)
-                  TOKEN_3(9999): """, line: 1, col: 63
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Letter(9999)
-                        UppercaseLetter(9999)
-                          TOKEN_36(9999): "D", line: 1, col: 64
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Digit(9999)
-                        TOKEN_18(9999): "2", line: 1, col: 65
-                  TOKEN_3(9999): """, line: 1, col: 66
+                StringProperty(9999)
+                  String(9999)
+                    TOKEN_3(9999): """, line: 1, col: 63
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Letter(9999)
+                          UppercaseLetter(9999)
+                            TOKEN_36(9999): "D", line: 1, col: 64
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Digit(9999)
+                          TOKEN_18(9999): "2", line: 1, col: 65
+                    TOKEN_3(9999): """, line: 1, col: 66
--- a/tests/outputs/Filter_064.out
+++ b/tests/outputs/Filter_064.out
@@ -74,22 +74,23 @@ Filter(9999)
                 Spaces(9999)
                   Space(9999)
                     TOKEN_1(9999): " ", line: 1, col: 35
-              String(9999)
-                TOKEN_3(9999): """, line: 1, col: 36
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      UppercaseLetter(9999)
-                        TOKEN_33(9999): "A", line: 1, col: 37
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      LowercaseLetter(9999)
-                        TOKEN_76(9999): "l", line: 1, col: 38
-                TOKEN_3(9999): """, line: 1, col: 39
-                Spaces(9999)
-                  Space(9999)
-                    TOKEN_1(9999): " ", line: 1, col: 40
+              StringProperty(9999)
+                String(9999)
+                  TOKEN_3(9999): """, line: 1, col: 36
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Letter(9999)
+                        UppercaseLetter(9999)
+                          TOKEN_33(9999): "A", line: 1, col: 37
+                  EscapedChar(9999)
+                    UnescapedChar(9999)
+                      Letter(9999)
+                        LowercaseLetter(9999)
+                          TOKEN_76(9999): "l", line: 1, col: 38
+                  TOKEN_3(9999): """, line: 1, col: 39
+                  Spaces(9999)
+                    Space(9999)
+                      TOKEN_1(9999): " ", line: 1, col: 40
       AND(9999)
         TOKEN_33(9999): "A", line: 1, col: 41
         TOKEN_46(9999): "N", line: 1, col: 42
@@ -149,22 +150,23 @@ Filter(9999)
                   Spaces(9999)
                     Space(9999)
                       TOKEN_1(9999): " ", line: 1, col: 68
-                String(9999)
-                  TOKEN_3(9999): """, line: 1, col: 69
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Letter(9999)
-                        UppercaseLetter(9999)
-                          TOKEN_33(9999): "A", line: 1, col: 70
-                  EscapedChar(9999)
-                    UnescapedChar(9999)
-                      Letter(9999)
-                        LowercaseLetter(9999)
-                          TOKEN_76(9999): "l", line: 1, col: 71
-                  TOKEN_3(9999): """, line: 1, col: 72
-                  Spaces(9999)
-                    Space(9999)
-                      TOKEN_1(9999): " ", line: 1, col: 73
+                StringProperty(9999)
+                  String(9999)
+                    TOKEN_3(9999): """, line: 1, col: 69
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Letter(9999)
+                          UppercaseLetter(9999)
+                            TOKEN_33(9999): "A", line: 1, col: 70
+                    EscapedChar(9999)
+                      UnescapedChar(9999)
+                        Letter(9999)
+                          LowercaseLetter(9999)
+                            TOKEN_76(9999): "l", line: 1, col: 71
+                    TOKEN_3(9999): """, line: 1, col: 72
+                    Spaces(9999)
+                      Space(9999)
+                        TOKEN_1(9999): " ", line: 1, col: 73
         AND(9999)
           TOKEN_33(9999): "A", line: 1, col: 74
           TOKEN_46(9999): "N", line: 1, col: 75
@@ -222,30 +224,31 @@ Filter(9999)
                     Spaces(9999)
                       Space(9999)
                         TOKEN_1(9999): " ", line: 1, col: 99
-                  String(9999)
-                    TOKEN_3(9999): """, line: 1, col: 100
-                    EscapedChar(9999)
-                      UnescapedChar(9999)
-                        Letter(9999)
-                          UppercaseLetter(9999)
-                            TOKEN_33(9999): "A", line: 1, col: 101
-                    EscapedChar(9999)
-                      UnescapedChar(9999)
-                        Letter(9999)
-                          LowercaseLetter(9999)
-                            TOKEN_76(9999): "l", line: 1, col: 102
-                    TOKEN_3(9999): """, line: 1, col: 103
-                    Spaces(9999)
-                      Space(9999)
-                        TOKEN_1(9999): " ", line: 1, col: 104
-                      Space(9999)
-                        TOKEN_1(9999): " ", line: 1, col: 105
-                      Space(9999)
-                        TOKEN_1(9999): " ", line: 1, col: 106
-                      Space(9999)
-                        TOKEN_1(9999): " ", line: 1, col: 107
-                      Space(9999)
-                        TOKEN_1(9999): " ", line: 1, col: 108
-                      Space(9999)
-                        nl(9999)
-                          SPECIAL_1(9999): "(...)", line: 1, col: 109
+                  StringProperty(9999)
+                    String(9999)
+                      TOKEN_3(9999): """, line: 1, col: 100
+                      EscapedChar(9999)
+                        UnescapedChar(9999)
+                          Letter(9999)
+                            UppercaseLetter(9999)
+                              TOKEN_33(9999): "A", line: 1, col: 101
+                      EscapedChar(9999)
+                        UnescapedChar(9999)
+                          Letter(9999)
+                            LowercaseLetter(9999)
+                              TOKEN_76(9999): "l", line: 1, col: 102
+                      TOKEN_3(9999): """, line: 1, col: 103
+                      Spaces(9999)
+                        Space(9999)
+                          TOKEN_1(9999): " ", line: 1, col: 104
+                        Space(9999)
+                          TOKEN_1(9999): " ", line: 1, col: 105
+                        Space(9999)
+                          TOKEN_1(9999): " ", line: 1, col: 106
+                        Space(9999)
+                          TOKEN_1(9999): " ", line: 1, col: 107
+                        Space(9999)
+                          TOKEN_1(9999): " ", line: 1, col: 108
+                        Space(9999)
+                          nl(9999)
+                            SPECIAL_1(9999): "(...)", line: 1, col: 109
--- a/tests/outputs/Filter_065.out
+++ b/tests/outputs/Filter_065.out
@@ -49,19 +49,20 @@ Filter(9999)
                 TOKEN_41(9999): "I", line: 1, col: 22
                 TOKEN_46(9999): "N", line: 1, col: 23
                 TOKEN_51(9999): "S", line: 1, col: 24
-              String(9999)
-                TOKEN_3(9999): """, line: 1, col: 25
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      UppercaseLetter(9999)
-                        TOKEN_33(9999): "A", line: 1, col: 26
-                EscapedChar(9999)
-                  UnescapedChar(9999)
-                    Letter(9999)
-                      LowercaseLetter(9999)
-                        TOKEN_76(9999): "l", line: 1, col: 27
-                TOKEN_3(9999): """, line: 1, col: 28
+              StringProperty(9999)
+                String(9999)
+                  TOKEN_3(9999): """, line: 1, col: 25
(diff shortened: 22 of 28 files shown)


optimade.md
```diff
@@ -1298,8 +1298,22 @@ In addition to the standard equality and inequality operators, matching of parti
...
+* `identfier LIKE x`
+
+* `identfier UNLIKE x`
+
 * Support for x to be an identifier, rather than a string is OPTIONAL.
 
+If implemented, the "LIKE" operator MUST behave as the correspoding standard SQL operator. In particular,
+The `x` string MUST be interpreted as a pattern where an underscore character ('_', ASCII DEC 95, HEX 5F)
+matches any single character and a percent character ('%', ASCII DEC 37, HEX 25) matches an arbitrary
+sequence of characters (including zero characters).
```
giovannipizzi: Shouldn't then we also explain how to escape these characters, and how to escape the escape character? (should be `\` IIRC

merkys: Good catch. [MySQL specification](https://dev.mysql.com/doc/refman/5.5/en/string-comparison-functions.html) says that to search for `\` in `LIKE` one would have to write `\\\\` (instead of `\\`, which is common way to escape the escaping sequences in C and the like). Which will we choose?

giovannipizzi:
They say that this is just because they first C-unescape the string, eg to accept `\n` as a newline. We don't, so we should just say `\\` to escape a backslash.
Whoever converts the string to MySQL must then first C-escape whatever its obtained from the query string.

giovannipizzi:
Maybe it's still good to point this out to implementers though, as I'm sure this will be a common error (but anyway if you don't want SQL injection you should encode your string, is the user passes a quote in itts string....)

rartino:
Is the hope to get this in for v0.10? Because, after having mulled over this a bit since the meeting last Friday, I think this is something we may want to discuss a bit more before merging.

*(To me, the point with the several rushed PRs this week is to make sure the spec is in a sane state for a v0.10 release, i.e., internally consistent and preferably without having rough completely backwards-incompatible changes just around the corner--- not so much to get new features in last-minute. I'd be happy to see a release v0.11 with this feature in shortly after v0.10.)*

Now, about the content of this PR:

OPTiMaDe filters are designed to be user friendly for our target audience of 'client' end users, right? Hence, even as someone who knows SQL quite well, I'm not super stoked on building in the SQL convention into OPTiMaDe:

- This choice has a very non-backend-agnostic "feel". Won't this antagonize those without SQL backends? *"So, to implement OPTiMaDe we have to translate wildcard expressions from the SQL convention to our backend?"*
- I expect most client users of OPTiMaDe to not know SQL and the conventions of its LIKE operator.
- I expect those users to find SQL's `%` and `_` unexpected for a wildcard mini-language. Just Google "wildcard" and disregard the hits about SQL; you'll find a wide array of descriptions of built-in functions across various software that very universally uses `*` for multi-character wildcards, and often `?` for single character matches. Two notable examples includes:
  - Shell glob patterns in [Unix shells](https://en.wikipedia.org/wiki/Glob_(programming)), [Windows powershell](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_wildcards?view=powershell-6), and even [MS-DOS](https://devblogs.microsoft.com/oldnewthing/20071217-00/?p=24143). 
  - [Microsoft excel, access, etc.](https://support.office.com/en-us/article/using-wildcard-characters-in-searches-ef94362e-9999-4350-ad74-4d2371110adb) as well as all software that mimics the same functions (google sheets, libreoffice) (though, libreoffice apparently also supports regex syntax.)

  I suspect users are likely to have come across one of these.
- Maybe worst: as coming out from the point @giovannipizzi brought up, I don't think there is a single escape syntax standard for SQL LIKE valid across SQL backends, so this is not a great standard anyway... *And*, we must make sure escaping in our LIKE does not interfere with string escaping in our string tokens. Reading the specification we cannot say that `identifier LIKE "\%"` is a search for a literal percent sign, because the right hand side is not a valid string token. A valid string token can only contain `\` followed by a `"` or a `\`. So, for us it would have to be "\\%", and how do we then encode a literal backslash + a wildcard `%`?

So, my suggestion is for us to skip SQL `LIKE` entirely, and give the operator a different name, e.g., `PATTERN` with the conventions:
- `*` for any number of arbitrary characters (including zero).
- `?` for exactly one arbitrary character.
- `**` for a literal `*`   
- `??` for a literal `?`
- the use of `\` remains as for any string token: `\\` for literal `\` and `\"` for literal `"`.

This seems somewhat user-friendly, and would be trival to search-replace into a corresponding SQL LIKE operator for us with SQL backends.

giovannipizzi:
I see @rartino's point and I tend to agree that, if there is no agreement, it's better not to merge this in 0.10 yet.
And to admit, I also think `*` and `?` are more intuitive (and should be easy to replace for SQL syntax in the backend).

A note on the choice of escaping: I think this is not yet "perfect":
- `**` as a literal `*`: would then `***` mean [a] "a literal `*` followed by any sequence of characters" or [b] "any sequence of characters followed by a literal `*`"? Might seem a little detail, but for instance the string `ab*cd` would match the pattern `ab***` in interpretation [a] but not in interpretation [b].
- Similar concern for `???`
- For `??` there is also an additional concern: how then can we specify that we want *exactly two characters*?

I think the only solution is to have a third character for escaping. I see that using backslash is counterintuitive because one has to escape the string again (indeed to encode `\\` one would have to write `\\\\` e.g. in JSON) and multiple escaping is the hardest thing ever to get right ;-) )

optimade.md
sauliusg: Thanks @giovannipizzi & @merkys for spotting this deficiency; I'll look up the existing escape rules and suggest a version in the text.

I agree with @giovannipizzi that we can use single backslash to escape characters, but we need to think over all possible combinations. I'll propose a variant in a while.

rartino:
@giovannipizzi are indeed correct about the quoting scheme not yet being perfect.

*(For anyone confused by the discussion about `??`, `???`, these are based on the first version of my comment; shortly after posting it I noticed the problem with `??` myself and re-edited it to use `\*?`, thinking no one would notice, sorry. But even with those edits, `***` remains broken.)*

I have a different suggestion for quoting now: going forward, even with other functionality, I think it may be very useful if we can extend our string tokens to represent "literal characters" aside "normal characters" using a single backslash. 

It would require a change in the grammar, so `\<character>` is allowed for any character in a string token, marking that character as "literal". (Right now it is forbidden by the grammar for `<character>` to be anything else but a `"` and `\`.)

We then need some carefully formulated text in the specification, where we say that for all regular uses of a string token, a literal character `\<character>` means the same thing as `<character>`. However, in other situations where some special meaning has been assigned to characters, a literal character refers to that character without that special meaning.

Then in a `PATTERN`, `?` and `*` are assigned the special meaning as specified above. It then follows from the paragraph above that `\?` and `\*` are used to represent the literal `?` and `*`. I think this would be elegant.

Note: this solution allows marking any character in an OPTiMaDe string as "literal" with the following exception: the characters `"` and `\` are *always* literal, one cannot indicate non-literal versions of these characters. I think that is an OK trade-off. Practically this *could* mean limitations on what syntax we can allow for other mini-languages embedded in OPTiMaDe strings, e.g., if we want to adopt a standard language for time formatting that uses `"` and `\` with their own special meanings, etc.

We might want to think forward to make sure this allows embedding a reasonable regex format. I think it does.

sauliusg:
@rartino wrote:

> So, my suggestion is for us to skip SQL LIKE entirely, and give the operator a different name, e.g., PATTERN with the conventions:

>    * for any number of arbitrary characters (including zero).
>    ? for exactly one arbitrary character.
>    ** for a literal *
>    *? for a literal ? (whereas ?? indicates two arbitrary characters, and ?* is the standard way to indicate an arbitrary character sequence of at least one character)
>    the use of \ remains as for any string token: \\ for literal \ and \" for literal ".

I though about this in the beginning (Issue https://github.com/Materials-Consortia/OPTiMaDe/issues/87), and then I realised that this is a very bad idea. Essentially we put on our shoulders a burden of developing yet another wildcard language, with the explicit goal to be **not** similar to existing ones, and with no apparent gains. This is a large work; as @giovannipizzi has just pointed out it is very easy to get it wrong; it gives no clear benefit, and adds extra burden on *every* implementer (because now everyone will have to implement, test and support this notation), and on every OPTiMaDe user (since everyone will have to learn a **new** wildcard language). I think this is very inefficient and unnecessary, since we will hardly give in the new wildcard syntax any new essential features that one does not have in SQL patterns, globs or regexps.

And then, what is the reason to be **not** similar to SQL, or any other pattern/regexp standard? One might object to it if it were a MUST feature and one would have to re-implement it in an SQL-unfriendly backend; but since it is supposed to be an **optional** feature, those backends that find it difficult to implement can simply skip it, and offer their versions of 'MATCH REGEXP' instead. I think sticking to a known standard, if that standard is well documented, has clear benefits: a/ one can find working and tested implementations; b/ the documentation is out there c/ many people will already know the syntax d/ those who learn it anew will be able to use it also in other contexts (thus will have more incentives to learn it)

I therefore suggest the following strategy:

- let's settle on several widespread standards with accessible implementations. I suggest picking SQL LIKE patterns, ERE regular expressions, and PCRE regular expressions. It covers a large area of useful cases. 

- All such matches will be optional. Thus, the implementer can choose to implement those matching operators that are easy to match to their respective backend, and skip all the rest. Nobody should be put at disadvantage by this.

- let's have in mind possible extensions of the patterns. This is the reason I suggested (in https://github.com/Materials-Consortia/OPTiMaDe/issues/87) to have regexp type specified in the match expression (e.g. `MATCH PCRE`); this will allow adding new flavours of regexps would such extensions be needed in the future (I personally doubt that it will be ever needed, but you never know what new "revolutionary" DBs will be rolled out with incompatible and non-standard interfaces ;). One possible backend which we up to now did not discuss much is SparQL; I'm not sure if it uses its own match expressions or some standard ones. Still, it fits into the proposed scheme, either with 'MATCH ERE' if SparQL supports ERE, or with 'MATCH SPARQL' if SparQL has (will have) its own matching syntax, and we decide to support it in the future.

- OPTiMaDe in itself is neutral as to the implementation of the matches, but we need to specify what the semantics of the matches will be in case they are implemented. Thus I think we can do it without extension of the existing Filter grammar.

In such setting, which is very simple, and is actually implied in this PR, I think we could make it even into v0.10 :).

Regards,
Saulius

sauliusg:
> It would require a change in the grammar, so `\<character>` is allowed for any character in a string token, marking that character as "literal". (Right now it is forbidden by the grammar for `<character>` to be anything else but a `"` and `\`.)

It could be done, but at the moment I think it is totally unnecessary. Filter language itself does not define or handle regular expressions, just strings; strings can encode arbitrary character sequences, including backslash with any characters; we just need to escape the backslash as a double backslash. Thus, an ERE matching arbitrary number of stars after the `a` character would be `a\**`, and embedded into a Filter string it will become `"a\\**"`. The Filter language quotes and backslashes are one encapsulation level higher than those of regexps. The situation is similar as with Java regular expressions.

sauliusg:
@rartino 

>     * Reading the specification we cannot say that `identifier LIKE "\%"` is a search for a literal percent sign, because the right hand side is not a valid string token. A valid string token can only contain `\` followed by a `"` or a `\`. 

True

> So, for us the right hand side would have to be `"\\%"`, 

True, this is a logical consequence of the current grammar. Java does the same with regexps in strings, awk does it the same way, and I suspect Python has no other option but to behave the same.

> and how do we then encode a literal backslash + a wildcard `%`?

`a LIKE "\\\\%"` Not the nicest looking but usable (and in fact used in many occasions).

I still need to double-check how SQL escapes `%`.

sauliusg:
>     * This choice has a very non-backend-agnostic "feel". Won't this antagonize those without SQL backends?

Why should it? As said, if the people find it difficult to implement, they just skip it (the feature is supposed to be optional). Unless they feel bitter envious that SQL has better pattern matching than their beloved super-modern backend, there should be no problems :P. And then again, I suggest also introducing (again optional) match operator syntax that would be easy to implement in alternative, non-SQL backends.

sauliusg:
> * I expect most client users of OPTiMaDe to not know SQL and the conventions of its LIKE operator.

Why? At least currently, most of the OPTiMaDe consrtium members do know about at least something SQL :)

> * I expect those users to find SQL's `%` and `_` unexpected for a wildcard mini-language.

You can always learn :). You search for "SQL LIKE", and in 5 mins you have the answer. Also, we will include a short description and a couple of examples into the OPTiMaDe spec (I'll update my PR with these features).

> Just Google "wildcard" and disregard the hits about SQL; you'll find a wide array of descriptions of built-in functions across various software that very universally uses `*` for multi-character wildcards, and often `?` for single character matches. Two notable examples includes:
>   
>   * Shell glob patterns in [Unix shells](https://en.wikipedia.org/wiki/Glob_(programming)), [Windows powershell](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_wildcards?view=powershell-6), and even [MS-DOS](https://devblogs.microsoft.com/oldnewthing/20071217-00/?p=24143).

That is why I suggested also describing the `MATCH GLOB` operator, with shell glob semantics; I can include it into the PR; I just had impression that we decided to have only LIKE and no more advanced operators when we skyped in Friday.

>   * [Microsoft excel, access, etc.](https://support.office.com/en-us/article/using-wildcard-characters-in-searches-ef94362e-9999-4350-ad74-4d2371110adb) as well as all software that mimics the same functions (google sheets, libreoffice) (though, libreoffice apparently also supports regex syntax.)

Again, this would be an argument to include an **optional** `MATCH ERE` and `MATCH PCRE` operator.

What does MongoDB and/or CoachDB support, actually? These would be probably the most likely backends that some of us actually use.

rartino:
I'll get to the individual points, but I see an important philosophical difference here, so lets start with sorting that out:
> - let's settle on several widespread standards with accessible implementations. I suggest picking SQL LIKE patterns, ERE regular expressions, and PCRE regular expressions. It covers a large area of useful cases.
> All such matches will be optional. Thus, the implementer can choose to implement those matching operators that are easy to match to their respective backend, and skip all the rest. Nobody should be put at disadvantage by this.

When we agreed on the need for optional features in the filtering language, you were very concerned about the fragmentation in support for filters that this would lead to. My argument was that optional features actually helps avoid fragmentation, because instead of every provider inventing their own extension for feature X, we now say in the standard that "you don't have to support feature X, but if you do, *this* is how you do it." That way, all providers that support feature X do it the same way, and queries are at least universal across all databases that support that feature. 

However, to get this effect, we need to be conservative with adding several optional features that are overlapping, i.e. exactly what you propose to do here. If we do, we get back to the same fragmentation we were trying to avoid. Now all databases will support only their own favorite brand of partial text matching operators, and partial text matching queries are not going to be universal even across databases that support partial text matching. Clients will constantly have to rewrite their queries to fit the text matching capabilities of each database. This, to me, is a too high cost for everyone to get their favorite syntax in.

So, I'm very much against supporting e.g. multiple regex flavors, and rather want us to standardize on a flavor that is translatable into the specific ones used by the backends. I realize I put more work on implementors here, but this is necessary to achieve the main goal of a universal api: to provide a common interface for access.

rartino:
On the specific points:
> Essentially we put on our shoulders a burden of developing yet another wildcard language, with the explicit goal to be not similar to existing ones, and with no apparent gains. This is a large work; as @giovannipizzi has just pointed out it is very easy to get it wrong;

I'm with you if we are talking about full regex support, I don't think we want to invent a new flavor of regexes. But, if we talk about patterns with character wildcards, that is trivial *apart* from how to do escaping. And, unless I am mistaken, there is no SQL standard for that escaping. If we pick one, e.g. MySQL, we still have to do the same amount of work to make sure we embedd it in a way consistent with our string quoting mechanism. This isn't easier than starting from scratch; and for all intents and purposes, we are creating a new pattern language in the eyes of users, because the actual strings they need to input are different in, e.g., how many backslashes are needed.

 In fact, what is likely to happen with the SQL approach is that users wondering how to represent a literal % will find a stack overflow question about, e.g., postgree SQL, try that syntax with whatever number of backslashes they use, and then be confused over the exact same string behaving differently in OPTiMaDe.

> Filter language itself does not define or handle regular expressions, just strings; strings can encode arbitrary character sequences, including backslash with any characters; we just need to escape the backslash as a double backslash. 

Right, but the point in much of the design we do is for the filter language to be user friendly. I know from heavy amount of experience that no one appreciates this "backslash hell" where several quoting mechanisms are stacked to yield unreadable expressions. If we can support single backslash qouting, is it not nice to do so?

> Ehhh... why should we disregard SQL? 

My point here was that it is only SQL that uses `%` and `_`. There are far more pattern mini-language using `*` and `?`. Hence, for users not familiar with SQL LiKE, this is going to be a surprise.

> At least currently, most of the OPTiMaDe consrtium members do know about at least something SQL :)

We are all database providers. Who are you designing OPTiMaDe for? I'm having the PhD students at our department in mind who do various projects that would benefit from easy aggregating of data across databases. I think few of them know SQL.

sauliusg:
> I'm with you if we are talking about full regex support,

ACK. And so, one flavour of matches should include RE support; but we do not make it mandatory and do not want to exclude the use of simpler matching mechanisms, do we? The proposed solution, with 'LIKE', 'MATCH ERE', 'MATCH PCRE' would permit *standard* implementation of various levels of patterns.

> I don't think we want to invent a new flavor of regexes. 

I agree, we don't :)

> But, if we talk about patterns with character wildcards, that is trivial _apart_ from how to do escaping. 

I don't find it trivial; it's not just regexp search/replace, since we will need to deal with escaped characters and character ranges within [] brackets. I feel that a client will have to do a full pattern parsing, and then map it into the backend's patterns/regexps if we want to have specific OPTiMaDe pattern syntax.

> And, unless I am mistaken, there is no SQL standard for that escaping.

I think there is; I would be very much surprised if it did not.

I do not have the latest SQL standard at hand (costs 180$ with ISO :/), but open sources on the Web (https://modern-sql.com/standard, http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt) specify that the syntax is `column LIKE 'pattern%'` or `column LIKE '30\%pattern% ESCAPE '\'`, where ESCAPE allows to specify any escape character one wishes. MySQL defaults ESCAPE to '\'; other SQL description do not mention any default, so probably default backslash escape is MySQL specific extension. That does not cause difficulties if we specify backslash as an escape character for OPTIMADE since all SQL databases should support the standard ESCAPE clause.

> If we pick one, e.g. MySQL, we still have to do the same amount of work to make sure we embedd it in a way consistent with our string quoting mechanism. 

As said, SQL `ESCAPE` mechanism should allow to specify escape character for any SQL engine.

> This isn't easier than starting from scratch; and for all intents and purposes, 

I would say it is much easier than starting from scratch, we just take a (subset) of a standard implementation.

> we are creating a new pattern language in the eyes of users, 

Not if we take an existing standard(s) and reference it/them.

> because the actual strings they need to input are different in, e.g., how many backslashes are needed.

We should not confuse defining the meaning of patterns, and defining the way how to encode patterns into strings. Currently the OPTiMaDe string definition allows encoding *any* character sequence (using escapes); so we have no problem here. If we think that too many backslashes in escaped strings are awkward to read (I agree with this), then we can think about how to improve OPTiMaDe string escaping; this is an issue independent from match-pattern semantics, IMHO.

> In fact, what is likely to happen with the SQL approach is that users wondering how to represent a literal % will find a stack overflow question about, e.g., postgree SQL, try that syntax with whatever number of backslashes they use, and then be confused over the exact same string behaving differently in OPTiMaDe.

I don't find this argument convincing. If people take Postgress SQL *specific* (non-standard) rules and try to apply to something else, be it OPTiMaDe, MySQL, MogoDB or what not, of course they will get nonsense, and here you can not help them in any way – except maybe by writing decent OPTiMaDe documentation (that specified how OPTiMaDe patterns work) and advising them to use '-stackoverflow' in their Google searches :P.

sauliusg:
> My point here was that it is only SQL that uses `%` and `_`. There are far more pattern mini-language using `*` and `?`. Hence, for users not familiar with SQL LiKE, this is going to be a surprise.

SQL seems to be still the most widely used database query language. Its patterns, as I see now after some days of searching on the Web, are very well described and present in all searches. Disregarding this as "only SQL" does not seem wise to me. Nobody will be surprised at SQL pattern languages if they do a short search on the Web.

Yes, there are essentially two match syntaxes – SQL LIKE and PCRE. All the rest is variations and subsets of these two. SQL seems to be a fairly large chunk of DB implementations; probably SQL `LIKE` is the most frequently supported backend operator, if we take by number of installations, no?

Another is ERE/PCRE. Also fairly popular. That's why I suggest having possibility to use either or both of them in OPTiMaDe queries.

merkys:
> and advising them to use '-stackoverflow' in their Google searches :P.

This won't help. Web is full of other how-tos. If we refer to some other standard in our specification, then we have to follow it as closely as possible. It's the "SQL" in "SQLish" that may get the user searching the Web and trying first hits of SQL syntax how-tos.

sauliusg:
> When we agreed on the need for optional features in the filtering language, you were very concerned about the fragmentation in support for filters that this would lead to. My argument was that optional features actually helps avoid fragmentation, because instead of every provider inventing their own extension for feature X, we now say in the standard that "you don't have to support feature X, but if you do, _this_ is how you do it." That way, all providers that support feature X do it the same way, and queries are at least universal across all databases that support that feature.

I agree, and this argument convinced me that optional features are not bad. They will be even better if supplemented by machine-readable descriptions of "feature sets" (see https://github.com/Materials-Consortia/OPTiMaDe/issues/91, second suggestion).

> However, to get this effect, we need to be conservative with adding several optional features that are overlapping,

I agree if the features are completely overlapping, but here we have different features with different trade-offs between search power and implementation complexity.

> i.e. exactly what you propose to do here.

This is not true. I propose three well-defined optional features, so that, as you say, "if you do [them], _this_ is how you do it.". Implementers then can choose which ones to implement.

> If we do, we get back to the same fragmentation we were trying to avoid.

Not really.

> Now all databases will support only their own favorite brand of partial text matching operators, 

NO! If compliant to OPTiMaDe, they will support the ones described in the spec.

> and partial text matching queries are not going to be universal even across databases that support partial text matching. Clients will constantly have to rewrite their queries to fit the text matching capabilities of each database.

With optional features you will have to adapt your queries to different implementation, no way around it as long as you introduce options. This could only be made somewhat easier if each OPTiMaDe endpoint advertises in a machine-readable form which optional features they support (the above-mentioned "feature sets:"). 

And then, you do not need to re-write constantly, there will be no constant changes once the specs are out, will there?

> This, to me, is a too high cost for everyone to get their favorite syntax in.

Its not about favourite syntax, its about easier implementation. We can not require all implementers say "look, we have invented our own matching semantics, no go and implement it in your backend".

> So, I'm very much against supporting e.g. multiple regex flavors, and rather want us to standardize on a flavor that is translatable into the specific ones used by the backends. 

At the moment I think that "translatable into the specific ones used by the backends" is too difficult to be reasonably implementable.

> I realize I put more work on implementors here, but this is necessary to achieve the main goal of a universal api: to provide a common interface for access.

I think the burden you suggest for implementers is unacceptably high. In that case, we are probably better of with just what is in the specs right now, STARTS/ENDS/CONTAINS. Regexp filtering can then be done at the client side if needed...

sauliusg:
> > and advising them to use '-stackoverflow' in their Google searches :P.
> 
> This won't help. Web is full of other how-tos. If we refer to some other standard in our specification, then we have to follow it as closely as possible. It's the "SQL" in "SQLish" that may get the user searching the Web and trying first hits of SQL syntax how-tos.

This would be a good reason to make `LIKE` behaviour as close to the SQL's as possible, this is my main point.

sauliusg:
@rartino :

>  If we do, we get back to the same fragmentation we were trying to avoid.

One more though on avoiding fragmentation.

Lets consider the following scenario. I want to send a query to find entries that match RE 'Saulius.*Grazulis' to multiple OPTiMaDe databases. I want to use pattern match, but I do not know which databases implement which features.

I could compose the following Filter expression: `(author LIKE "%Name%Surname%" OR author MATCH GLOB "*[Nn]ame*[Ss]urname*" OR author MATCH REGEXP "[Nn]ame.*[Ss]urname")`. Let's assume that  in an OR expression, unimplemented operators return FALSE, and implemented ones the expected values. Then, if a database implements at least one of the queries, I will get the expected answer; and I could send this query to all databases. 

The only problem is when an OPTiMaDe endpoint does not implement **any** of these operators; then I would get no data at all instead of getting the right answer.

Another approach would be, as @rartino suggests, to have the `author MATCH "*[Ss]aulius*[Gg]ražulis*" operator, with wildcards specified by the OPTiMaDe, which each API implementation will have to translate to the appropriate backend query.

The difference between the two approaches are: in the first case, we put translation burden on the user; in the second case we put the translation burden on the API implementers. Arguably, it is easier to compose a query in different languages, sincne you know what you are looking for, than to parse a formal definition and translate it **in all cases** to another language. So, assuming qualification of both users and implementers is the same, composing multiple queries is easier than translating. It is only if we target very inexperienced users that composing queries becomes an issue.

rartino:
> I could compose the following Filter expression: `(author LIKE "%Name%Surname%" OR author MATCH GLOB "*[Nn]ame*[Ss]urname*" OR author MATCH REGEXP "[Nn]ame.*[Ss]urname")`. Let's assume that in an OR expression, unimplemented operators return FALSE, and implemented ones the expected values. Then, if a database implements at least one of the queries, I will get the expected answer; and I could send this query to all databases.

So, in effect, you propose that the syntax in OPTiMaDe for partial string matching is to give manually translated versions of your pattern to cover ALL relevant flavors of pattern languages? You don't think this is overly cumbersome to use, even for experienced users? To me, your example shows precisely why supporting all these overlapping options all as individually optional is a bad idea.

The translator approach is not as bad as you suggest. If we are talking about patterns (and not regexes) it is going to be a more or less straightforward string transform. And as soon as someone implements these, the code can be converted to other languages and distributed as support libraries. Whereas if we put all of the burden on the user as you propose, you can bet every second query is going to have one of the redundant patterns translated in error, and users will act on incomplete search results without knowing it.

> > However, to get this effect, we need to be conservative with adding several optional features that are overlapping,
> 
> I agree if the features are completely overlapping, [...]

If you agree with that statement; can we perhaps discuss a "ladder" of complexity levels? This is my view:
1. Basic operations: STARTS WITH, ENDS WITH, CONTAINS: *required*
2. Basic pattern with arbitrary character string (e.g., * or %), arbitrary single character (e.g., ? or _): *optional*
3. Full regex matching support: *optional*

What I am trying to say is that it would be optimal to only provide ONE facility for text matching on each level. E.g., SQL LIKE *or* GLOB-like *or* something else for level 2. But, I agree that Level 3 is trickier. Maybe we can postpone the discussion of REGEX flavors for a while, because I'd like to think more on the question "translating" in this context. And this PR is about level 2. 

With reference to this ladder of complexity, I propose that if an API implementer supports a higher level, they *MUST* also support all lower levels. (This should be ok, as the lower levels are translatable into higher level expressions.) This means that we can recommend a user to use the lowest level at which a query is expressible, and it will then be handled by the highest possible number of databases.

 Now, specifically about whether the SQL LIKE standard is a good idea for OPTiMaDe level 2:

> > if we talk about patterns with character wildcards, that is trivial _apart_ from how to do escaping.
> 
> I don't find it trivial; it's not just regexp search/replace, since we will need to deal with escaped characters and character ranges within [] brackets. I feel that a client will have to do a full pattern parsing, and then map it into the backend's patterns/regexps if we want to have specific OPTiMaDe pattern syntax.

Why are you talking about character ranges? I'm talking about LIKE-equivalent (level 2) patterns here as you have described it in your PR, so one arbitrary length wildcard, and one single-character wildcard + escaping rules for those. And the mapping I am talking about then is from our "own" syntax (`?` and `*`) into the LIKE syntax you want (`%` and `_`), not into the backends regex format. You are certainly allowed to implement these patterns via SQL LIKE, and those who don't will have an equally hard time from this point as with an SQL LIKE-based syntax.
 
> I do not have the latest SQL standard at hand (costs 180$ with ISO :/), 

Isn't this a major problem in itself that not even we have access to the standard you want to base the OPTiMaDe pattern operator on? Especially since we conclude the need to firmly point users to it, since, there are a LOT of helpful sources on the Internet that cover database-specific flavors that isn't exactly implementing the standard. (E.g., MS SQL Server supports `[` and `]` for ranges, whereas it seems few others do.)

I'm already running into this issue, as for my response above, I tried to figure out if the SQL standard LIKE actually supports character ranges with `[` and `]`, but I find no place I can confirm of deny it with certainty. At least in the 1992 draft you linked, they are not supported.

> the syntax is `column LIKE 'pattern%'` or `column LIKE '30\%pattern% ESCAPE '\'`, where ESCAPE allows to specify any escape character one wishes. MySQL defaults ESCAPE to ''; other SQL description do not mention any default, so probably default backslash escape is MySQL specific extension. That does not cause difficulties if we specify backslash as an escape character for OPTIMADE since all SQL databases should support the standard ESCAPE clause.
> 
> As said, SQL `ESCAPE` mechanism should allow to specify escape character for any SQL engine.

After researching this, I'll accept that `LIKE` *with* `ESCAPE` behaves fairly standard across SQL backends, aside that some add their own pattern extensions... Without `ESCAPE`, not so much. In my reading of the 1992 draft, it seems the default should be no escape symbol at all. This is apparently what Oracle uses (according to a help page for v10.1 I found online). But then many other implementations (MySQL, postgresql) takes backslash as the default, which leads to confusing "help" webpages.

What is the concrete proposal for OPTiMaDe? That we also include the ESCAPE keyword? I.e., to search for a literal backslash + wildcard one have these options for filter: `identifier LIKE "\\\\%" ESCAPE "\\"` or, e.g., `identifier LIKE "#%" ESCAPE "#"`? What is our default escape char?

My end conclusion is that I suppose it wouldn't be a catastrophe to standardize on LIKE + ESCAPE, but I'm far from convinced it is the best choice, since:
1. There is no openly accessible standard to point users to.
2. Many actual SQL implementations do not follow the standard, which makes it confusing for end users who seek help (which is worsened by point 1)
3. For those who want to support LIKE without having an SQL backend, the construct with arbitrary escape character adds extra work with not very much benefit.
4. End users with little experience in SQL will find `%` and `_` unusual as wildcards, and the syntax with LIKE + ESCAPE bulky.

And what is the gain compared to the proposed alternative?:  
`identifier MATCHES <PATTERN>` where in `<PATTERN>`: 
* `*` means any number of arbitrary characters (including zero).
* `?` means exactly one arbitrary character.
* `\*` is a literal `*`
* `\?` is a literal `?` 
* `\\` is a literal `\`

which you can translate to an SQL LIKE query for your backend this way: 
1. Make these substitutions in `<PATTERN>`:
   * `%` -> `\%`
   * `_` -> `\_`
   * `?` -> `_` 
   * `*` -> `%`
2. Query your SQL backend with `identifier LIKE <PATTERN> ESCAPE "\"`

sauliusg:
> My end conclusion is that I suppose it wouldn't be a catastrophe to standardize on LIKE + ESCAPE, but I'm far from convinced it is the best choice, since:

I agree. I also think we both agree that ESCAPE would be too bulky and unnecessary for OPTiMaDe.

>     1. There is no openly accessible standard to point users to.

Well, not quite true, one is (SQL-92)[http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt] that describes the `LIKE` predicate, and from the documents that we discussed privately we can see that SQL-99 did not change that definition significantly.

>     2. Many actual SQL implementations do not follow the standard, which makes it confusing for end users who seek help (which is worsened by point 1)

True, but those implementations only extend the standard, so 'column LIKE pattern ESCAPE '\'` should be portable across SQL implementations.

>     3. For those who want to support LIKE without having an SQL backend, the construct with arbitrary escape character adds extra work with not very much benefit.

I think we all agree that OPTiMaDe should specify one fixed escape character (or one fixed escape mechanism, such as character doubling).

>     4. End users with little experience in SQL will find `%` and `_` unusual as wildcards, and the syntax with LIKE + ESCAPE bulky.

After all the research that we have done on the Web, I strongly disagree that "`%` and `_` [are] unusual as wildcards". Rather, I would say, `*` and `?` are limited to the shell glob community... :)

sauliusg:
> And what is the gain compared to the proposed alternative?:
> `identifier MATCHES <PATTERN>` where in `<PATTERN>`:
> 
>     * `*` means any number of arbitrary characters (including zero).
> 
>     * `?` means exactly one arbitrary character.
> 
>     * `\*` is a literal `*`
> 
>     * `\?` is a literal `?`
> 
>     * `\\` is a literal `\`

So essentially we arrive at the same SQL `LIKE` convention, with the backslash ('\\') as an escape character :). I'm not against that, since, in addition to being a reworded SQL LIKE, this is superficially similar to shell GLOBS, but I want to point out that '*' and '?' have no clear advantage over '%' and '_'; as said, I strongly **disagree** that '%' are "unusual"; moreover, '?' is confusing since REs use '.' instead, and `?` means something slightly different there (I myself was confused by this in my young days ;).

sauliusg:
> which you can translate to an SQL LIKE query for your backend this way:
> 
>     1. Make these substitutions in `<PATTERN>`:
>        
>        * `%` -> `\%`
>        * `_` -> `\_`
>        * `?` -> `_`
>        * `*` -> `%`

This is not that straighforward as you write, since `\*` does not need to be translated to `\%`, it translates to `*`, and `\\*` would not translate to `\*`, what is meant presumably 'backslash-any_character' which would be `\\%', whereas `\\a` is 'backslash-a' which translates to `\\a` in LIKE...

In other words, doable but requires extra work to get it right. My original instinct is that I want to avoid that work if possisble.

sauliusg:
> If you agree with that statement; can we perhaps discuss a "ladder" of complexity levels? This is my view:
> 
>     1. Basic operations: STARTS WITH, ENDS WITH, CONTAINS: _required_
> 
>     2. Basic pattern with arbitrary character string (e.g., * or %), arbitrary single character (e.g., ? or _): _optional_
> 
>     3. Full regex matching support: _optional_

Yes, I agree on this complexity ladder, this is what I meant. We can settle on this :)

sauliusg:
> After researching this, I'll accept that `LIKE` _with_ `ESCAPE` behaves fairly standard across SQL backends,

I also find it so.

>  aside that some add their own pattern extensions... 

I have the same impression.

> Without `ESCAPE`, not so much. 

Probably yes...

> In my reading of the 1992 draft, it seems the default should be no escape symbol at all. 

True, I think.

> This is apparently what Oracle uses (according to a help page for v10.1 I found online).

I thought Oracle has a (non-standard?) feature `SET ESCAPE ...` that changes the behaviour, bu we do not need to rely on that.

> But then many other implementations (MySQL, postgresql) takes backslash as the default, 

True

> which leads to confusing "help" webpages.

OPTiMaDe should be the only Help :)

> What is the concrete proposal for OPTiMaDe? That we also include the ESCAPE keyword?

`property LIKE pattern` with default backslash ('\\') as an escape character, translatable to SQL as:

```
property_column LIKE pattern ESCAPE '\'
```

with `pattern` taken more-or-less verbatim after resolving OPTiMaDe string double-quote and backslash escapes). More-or-less dependes on whether we take `%` and `_` as wildcards (then it will be verbatim), or `*` and `?` as wildcards, then it will need translation with good chances that people will get it incorrectly :)

sauliusg:
> Why are you talking about character ranges? I'm talking about LIKE-equivalent (level 2) patterns here as you have described it in your PR

I meant the most general case, level 3, with regexps.

This is the whole point. Do we design syntax **only** for the level 2 patterns (`LIKE`), or do we want it to be general enough to be usable for level 3 as well? If it is only for `LIKE`, then regexps will have different metacharacters and different escape rules anyway (that is why I mentioned `'[]'` brackets). This would imply different syntax constructs for different complexity levels, as I suggest now. And since SQL `LIKE` seems to be the only existing standardised implementation giving the level 2 complexity, I see no reason why we should deviate from it.

If we want just a **single** OPTiMaDe operator, `MATCH`, that covers all complexity levels (essentially, 2 and 3), then we need to think how we convey them in the syntax. Should brackets `'[]'` be escaped by a back-slash even at the level 2? Should bare `[]` be reported at level 2 as errors even though they are not used for matching? Should `a?` be a regexp "maybe-a", or is it "a with arbitrary character"? Wil it be the same at level 2 and level 3? We need to decside all these issues before we move on.

My suggestion is essentially to segregate level 2 and level 3 syntactically, by having `predicated LIKE pattern` (for level 2) and `predicate MATCH ERE pattern` (for level 3). True, this will require the user to know all these different pattern encodings; but a serious computer user will have to know them anyway, IMHO, and indicating syntactically which complexity level we mean will simplify job for specification writers (meaning us ;), for implementers, and arguably also for users (there will be no confusion what `?` means in `p LIKE 'xa?'` and `p MATCH ERE `xa?`)

We can, of course, say that level 3 is not supported by OPTiMaDe at all, but that would be a pity, since some backends allow to implement RE search easily, and people will start doing this in their own incompatible ways, which we wanted to avoid by describing optional (but standardised) features.

As a side note: SQL (SQL-99) has `SIMLAR TO` predicate that supports bracketed character ranges in a standard way. Not all backends seem to implement it, though (MySQL that we have does not). I am not suggesting it for OPTiMaDe :)

sauliusg:
After talking with Andrius over the lunch, we came up with an alternative (compromise?) solution:

Let's leave just to "complexity levels":

- `STARTS/ENDS/CONTAINS` operators _(required)_;

- ERE `MATCH` operator _(optional)_.

That is, we skip `LIKE` with all its "unusual" wildcards altogether.

The REs used could be a common subset of expressions supported by ERE and PCRE.

The RE strings can be encoded as OPTiMaDe strings (in that case, the RE backslash will be encoded as two back-slashes in a string), or we can provide a `/`-delimited specialised RE syntax – in that case, we can use single physical backslash to escape RE meta-characters, improving readability at the expense of extra grammar rules in the Filter language.

Rationale:

- not supporting `LIKE` only makes problems for backend engines that implement SQL LIKE but do not implement RE matches. Arguably, there are not many of such engines;

- `MATCH`, if implemented, covers all functionality of `LIKE`, and more;

- when a backend supports (PC|E)REs, the implementation will be simple – just pass the RE to the backend RE engine;

- since `MATCH` is optional, anyone who encounters difficulties implementing RE match can simply skip it;

- the mandatory `STARTS/ENDS/CONTAINS` operators offer some of the `LIKE` functionality;

- since we have only one optional feature, there will be minimal fragmentation and no room for confusion;

- setting on EREs allows to reuse the existing ERE standard (POSIX), with increased likelihood that it will be implemented in backends in a compatible way.

sauliusg:
A Regexp survey:

>  MySQL uses the extended version to support regular expression pattern-matching operations in SQL statements
(https://dev.mysql.com/doc/refman/5.7/en/regexp.html)

> MongoDB uses Perl compatible regular expressions (i.e. “PCRE” ) version 8.41 with UTF-8 support.
(https://docs.mongodb.com/manual/reference/operator/query/regex/)

> PostgreSQL supports the same three regular expressions flavors: Tcl Advanced Regular Expressions, POSIX Extended Regular Expressions and POSIX Basic Regular Expressions.
(https://www.postgresql.org/docs/9.3/functions-matching.html)

A word of caution:
> While most regular-expression searches can be executed very quickly, regular expressions can be contrived that take arbitrary amounts of time and memory to process. Be wary of accepting regular-expression search patterns from hostile sources. If you must do so, it is advisable to impose a statement timeout.

> LIKE searches, being much simpler than the other two options, are safer to use with possibly-hostile pattern sources.
(https://www.postgresql.org/docs/9.3/functions-matching.html)

rartino:
I like your latest suggestions for how to support regexes. I see no issue with implementing regex support first and leave a hypothetical level 2 until later.

But, can you clarify a homework-not-done question: is PCRE a strict superset of functionality of EREs so we can simply say that OPTiMaDe uses the POSIX ERE format, and databases that uses PCRE can then simply plug those ERE expressions into their PCRE engines without issue? Or, do we need to say that we only support the common subset of features supported by both EREs and PCREs? Because for the latter, we will have to do some work to define the precise syntax.

Furthermore, I'm in support (but it is not an overly strong opinion) of reducing the number of backslashes via your suggestion of a regex token on format `/... regex pattern.../`, OR, alternatively picking up my idea further up in this thread to extend our string tokens to carry meta information about per-character literal-ity. (Which is conceptually a bit more tricky, but would resolve this issue more generally for all kinds of embedded mini-languages.) 

Now onto the basic pattern level (level 2): 

I'm not quite convinced of the idea of completely dropping the basic pattern level. Two reasons from the top of my mind:
- OPTiMaDe support for basic patterns allows backends that do not support regexes (or, at least not, exactly EREs/PCREs), but which can do other decent partial text matching, to provide at least that level of support. (As far as I can find online, an example of such a backend is MS Access.)
- Novice users may find the basic pattern syntax easier than the full regex syntax.

I agree with most of your latest comments about level 2 support, so we may be converging on a design here. In particular we seem to agree that:
- level 2 and level 3 partial text matching should use segregated syntax. 
- OPTiMaDe basic pattern support should work equivalently to standard SQL LIKE with ESCAPE `\` (but possibly with other wildcard characters.)

So, perhaps then all that is left is to decide is between these alternatives?:
- SQL LIKE-based basic patterns with wildcard characters `%` and `_` as it is in that standard.
- SQL-LIKE-like basic patterns with wildcard characters `*` and `?`.
- Skip basic pattern support, have only regex.

I see good and bad things with each of these options, so I propose to bring this up for discussion at the next meeting.

sauliusg:
@rartino :

just a quick answer to the ERE<->PCRE compatibility question:

> But, can you clarify a homework-not-done question: is PCRE a strict superset of functionality of EREs so we can simply say that OPTiMaDe uses the POSIX ERE format, and databases that uses PCRE can then simply plug those ERE expressions into their PCRE engines without issue? Or, do we need to say that we only support the common subset of features supported by both EREs and PCREs? Because for the latter, we will have to do some work to define the precise syntax.

I have DuckDucked yesterday on this subject and could not find a definitive answer, but it seems that ERE and PCRE are **not** in the subset-superset relation. The demonstration can be found in character set (`[a-z]`) behaviour:

- in EREs, "An escape inside a character class declaration shall match itself: in other words the escape character is not "special" inside a character class declaration; so `[\^]` will match either a literal '`\`' or a '`^`' " (https://www.boost.org/doc/libs/1_64_0/libs/regex/doc/html/boost_regex/syntax/basic_extended.html);

- in Perl flavour of PCER, 'Characters that may carry a special meaning inside a character class are: "`\`", "`^`", "`-`", "`[`" and "`]`" /.../  The sequences "`\a`", "`\c`", "`\e`", "`\f`", "`\n`", "`\N{NAME}`", "`\N{U+hex char}`", "`\r`", "`\t`", and "`\x`" are also special and have the same meanings as they do outside a bracketed character class' (`man perlrecharclass`), which in my understanding is incompatible with the ERE definition.

This is attested by the behaviour of GNU `grep`:

```
saulius@koala ~/ $ echo -e 'one\ntwo\none and a half\n1\\2\nTAB\tSEPARATED\nthree' | grep --color -E '[ \t]'
two
one and a half
1\2
three

saulius@koala ~/ $ echo -e 'one\ntwo\none and a half\n1\\2\nTAB\tSEPARATED\nthree' | grep --color -P '[ \t]'
one and a half
TAB	SEPARATED
```

That is, with the `-E` (ERE) option, the GNU `grep` treats `\` in square brackets as an ordinary character (so both `\` and `t` are matched), while in the `-P` (PCRE) mode the `\t` in the bracketed character set is treated as a code for the TAB character, and is matched accordingly.

So yes, we will have to define a common subset...

(More on you remaining comments later; I think we are converging :)

sauliusg:
> I like your latest suggestions for how to support regexes. I see no issue with implementing regex support first and leave a hypothetical level 2 until later.

I agree.

> Furthermore, I'm in support (but it is not an overly strong opinion) of reducing the number of backslashes via your suggestion of a regex token on format `/... regex pattern.../`, OR, 

Fine, I will add more tests to check the current grammar more thoroughly.

We also need to decide on two approaches:

- do we want the Filter grammar to accept any RE character and any RE escape int the `/.../` construct

- or do we want it to specify the syntax of the supported REs in full.

The second approach is more flexible – to add a RE extension in the future, you do no need to change the Filter grammar, just amend the documentation. The downside is that now Filter language will permit any sequence of characters in the REs, even if they do not make sense as REs, and rely on (informal) description for the back-end implementers to detect these;

The approach is more strict – we will have a definite grammar to specify what a filter RE is; implementations accepting this (or compatible) grammars will reject incorrect REs at parsing stage, and all will do it in the same way; and we can attach more formal meaning to RE constructs. However, future extensions would need extension to Filter grammar, an may potentially conflict with the current parsing engine capabilities.

> alternatively picking up my idea further up in this thread to extend our string tokens to carry meta information about per-character literal-ity. (Which is conceptually a bit more tricky, but would resolve this issue more generally for all kinds of embedded mini-languages.)

I am not quote sure what you mean by literal-ity here, and I am afraid that this approach already contains conflicting purposes (conflicting == not implementable at the same time in the same syntax).

The current filter strings currently can encode arbitrary printable character sequence (which is probably enough for search language). To make it fully transparent, we can add `\n`, `\t`, `\012` and `\x1BCD` style escapes, to specify arbitrary character sequences (both printable and non-printable). 

Regular expressions are encoded **by** those character strings. The strings serve only as a **wrapper** layer for REs. This means, for example, that `\s` will not have any special meaning in strings (since it does not stand for any specific Unicode character); but in PCREs the **two** characters `\` and `s` for a special wildcard `\s` matching arbitrary space. We could say that `\` and a letter `s` in strings mean just those two characters if it is not a special escape, but what will happen when at some stage one will need `\b` to denote back-tab control character and it will already be taken by a `\` and `b` character sequence?

We can of course say that in Filter strings any combination of a backslash and a character is allowed, both these characters continue a string even if the second character is a double-quote, and we do not prescribe any interpretation to these characters; i.e. in `"some \strange string"` the `\s` pair might mean "arbitrary space" in REs and "literal s character" in strings... I am not sure if this is a good approach since we leave implementations to interpret the meaning of the escapes.

/.../

> I see good and bad things with each of these options, so I propose to bring this up for discussion at the next meeting.

OK.

sauliusg:
> Now onto the basic pattern level (level 2):
> 
> I'm not quite convinced of the idea of completely dropping the basic pattern level.

I was originally in favour of this idea, and so am I now :)

> Two reasons from the top of my mind:
> 
>     * OPTiMaDe support for basic patterns allows backends that do not support regexes 
(or, at least not, exactly EREs/PCREs), but which can do other decent partial text matching, 
to provide at least that level of support. (As far as I can find online, an example of such a backend is MS Access.)

>     * Novice users may find the basic pattern syntax easier 
than the full regex syntax.

In addition to these, the security warning mentioned by Postgresql people should be considered as well.

Also, simpler patterns allow easier and more uniform implementation.

> I agree with most of your latest comments about level 2 support, so we may be converging on a design here. In particular we seem to agree that:
> 
>     * level 2 and level 3 partial text matching should use segregated syntax.

Agreed.

>     * OPTiMaDe basic pattern support should work equivalently to standard SQL LIKE with ESCAPE `\` (but possibly with other wildcard characters.)

I agree on the first part, but I do not see a reason to change wildcard characters.

Moreover, if we say that OPTiMaDe `LIKE` is designed and should work in the same way as the SQL LIKE, changing wildcards will introduce more confusion, not less. As an experiment, I have search in DuckDuckGo for "LIKE SQL patterns", and the whole first page points to SQL tutorials and manuals than plainly mention '%' and '_' as the wildcards, with no mention of '*' and '?'. Google will probably give the same result. Imagine how confused and irritated the users will be if the documented SQL '%' wildcard does not work, and instead they find out that OPTiMaDe has changed them to '*' for... clarity!

> So, perhaps then all that is left is to decide is between these alternatives?:
> 
>     * (1) SQL LIKE-based basic patterns with wildcard characters `%` and `_` as it is in that standard.

If we implement `LIKE` and explicitly say that it should behave like in SQL, then I'm very much in favour of this option;

>     * (2) SQL-LIKE-like basic patterns with wildcard characters `*` and `?`.

I rather strongly opinionated against this, since here we invent yet another match pattern that needs to be translated into SQL (extra work, extra bugs), needs to be learned and remembered by end users (extra confusion), and gives no clear benefit.

The shell globs that this is modelled did enter computer folklore, for sure (in form of "*nix" systems ;), but this does not mean that inexperienced users will understand them better than SQL patterns, and for experienced users like us, I think, uniformity with existing standards and tools is more important.

>     * (3) Skip basic pattern support, have only regex.

This could be an option if we want to bet on simplicity of the OPTiMaDe specs, non-proliferation of features and if we want to avoid fragmentation of implementations. For this to work, we need to very carefully describe what we mean by regexps and how they are intended to work, so that implementations are simple enough and the searches are powerful enough. I see a common subset of ERE and PCRE as a possibility, but we need to look of this does not exclude too many useful patterns (as I have mentioned before, even `[\t\n]` character ranges pose a problem). We need to be careful not to reduce the resulting REs to the SQL `LIKE` functionality :)

In my perception, options would be grouped from most favourable to least favourable as (1), (3), (2), with (1) and (3) pretty close in my view (if we do them right). An yes, I am strongly against (2).

Let's discuss during the Skype meeting. It would be interesting also to have opinions of @giovannipizzi , @merkys , @fawzi and others.

rartino:
This has been sitting for a while, perhaps long enough for others to solve part of the problem. We discovered in #436 that the JSON Schema standard now may have defined a reasonable "common" regex subset here:

  https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-regular-expressions

This PR includes adding "MATCH" to the filter language with a regex expression. If the PR is updated to also document "MATCH" with info on that the regex syntax is the SHOULD-level limited one documented in the JSON Schema text I think that part of this PR could be merged. (The limited regex syntax could either be spelled out in our spec or referenced; however, I'm a bit wary of referencing a draft specification; maybe including it as a quote is the best solution?).

I suppose we also just should agree on the LIKE syntax. I suggest we send out an email to the full OPTIMADE list to get a bit of broader feedback from the community, and then take a decision on the next web meet based on the resulting discussion (if any).

optimade.md
```diff
@@ -1287,8 +1287,22 @@ In addition to the standard equality and inequality operators, matching of parti
...
+
+If operator "UNLIKE" is supported, the bahavior of this oprtator MUST be the negation of the "LIKE" operator; i.e.
```
vaitkus, line 1303:
```suggestion
If operator "UNLIKE" is supported, the behavior of this operator MUST be the negation of the "LIKE" operator; i.e.

```

optimade.md
```diff
@@ -1287,8 +1287,22 @@ In addition to the standard equality and inequality operators, matching of parti
...
+an expression `(property UNLIKE "value")" must behave exactly as `(NOT (property LIKE "value"))`.
```
vaitkus, line 1304:
```suggestion
expression `(property UNLIKE "value")` must behave exactly as `(NOT (property LIKE "value"))`.

```

merkys:
> This has been sitting for a while, perhaps long enough for others to solve part of the problem. We discovered in #436 that the JSON Schema standard now may have defined a reasonable "common" regex subset here:
> 
> https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-regular-expressions
> 
> This PR includes adding "MATCH" to the filter language with a regex expression. If the PR is updated to also document "MATCH" with info on that the regex syntax is the SHOULD-level limited one documented in the JSON Schema text I think that part of this PR could be merged. (The limited regex syntax could either be spelled out in our spec or referenced; however, I'm a bit wary of referencing a draft specification; maybe including it as a quote is the best solution?).

The syntax in JSON Schema draft seems to me to be the very core of many regular expression syntaxes I am aware of, thus I support this proposal. On the other hand, I would prefer something more stable than draft specification, but spelling the syntax out in our spec would suffice.

Alternatively we may select one of the [other syntaxes](https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines), preferably stable and having wider support. For example [PCRE2](https://pcre.org/current/doc/html/pcre2pattern.html) which has implementations at least in Perl, Python and JavaScript.

rartino:
> Alternatively we may select one of the [other syntaxes](https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines), preferably stable and having wider support. For example [PCRE2](https://pcre.org/current/doc/html/pcre2pattern.html) which has implementations at least in Perl, Python and JavaScript.

The issue is not what is supported by programming languages, but across database backends, since it is your database that needs to be able to search efficiently for matches of the regex.

merkys:
> > Alternatively we may select one of the [other syntaxes](https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines), preferably stable and having wider support. For example [PCRE2](https://pcre.org/current/doc/html/pcre2pattern.html) which has implementations at least in Perl, Python and JavaScript.
> 
> The issue is not what is supported by programming languages, but across database backends, since it is your database that needs to be able to search efficiently for matches of the regex.

Right. Just checked MariaDB (backend for COD) and Mongo, both seem to support PCRE2 (although COD would have to update MariaDB to v10.5.1 to have PCRE2 instead of PCRE). It seems to me that JSON Schema draft is a subset of PCRE2.

rartino:
> Just checked MariaDB (backend for COD) and Mongo, both seem to support PCRE2 (although COD would have to update MariaDB to v10.5.1 to have PCRE2 instead of PCRE). It seems to me that JSON Schema draft is a subset of PCRE2.

Just the fact that one of the most active OPTIMADE participants would need to update their backend is perhaps itself an argument against, e.g., PCRE2. 

Nevertheless, also consider, e.g.:

- https://learn.microsoft.com/en-us/sql/ssms/scripting/search-text-with-regular-expressions?view=sql-server-ver16
- https://docs.oracle.com/cd/B13789_01/appdev.101/b10795/adfns_re.htm
- https://www.ibm.com/docs/en/db2/11.5?topic=sql-regular-expression-control-characters
- https://www.elastic.co/guide/en/elasticsearch/reference/current/regexp-syntax.html
- https://docs.snowflake.com/en/sql-reference/functions-regexp.html
- https://docs.splunk.com/Documentation/Splunk/9.0.3/SearchReference/Regex (PRCE1 I think?)
- https://duckdb.org/docs/sql/functions/patternmatching.html#regular-expressions

etc., etc.

(To be honest, I didn't explicitly check that none of the above doesn't happen to exactly be equal to PCRE2, just that they don't explicitly say so.)

merkys:
> Just the fact that one of the most active OPTIMADE participants would need to update their backend is perhaps itself an argument against, e.g., PCRE2.

Makes sense. However I have to check to make sure JSON Schema regex is really a subset of the old PCRE that our backend currently uses.

sauliusg:
> This has been sitting for a while, perhaps long enough for others to solve part of the problem. We discovered in #436 that the JSON Schema standard now may have defined a reasonable "common" regex subset here:
> 
> https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-regular-expressions

After a quick look at the JSON Schema REs, I have the question similar to that one we had with the LIKE operator strings: how do we escape special characters? E.e. what is the regular expression according the the JSON schema variety to match "(", ")", "|" and similar characters?

Maybe I overlooked in the https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-keyword-independence document, but I could not found the escape methods there...

vaitkus:
@sauliusg, according to the regular expression examples given in https://json-schema.org/understanding-json-schema/reference/regular_expressions.html, special characters can be escaped using a double backslash.

optimade.md
```diff
@@ -1287,8 +1287,22 @@ In addition to the standard equality and inequality operators, matching of parti
 
 OPTIONAL features: 
 
+The following comparison operators are OPTIONAL:
+
+* `identfier LIKE x`
+
+* `identfier UNLIKE x`
```
ml-evs, line 1294:
```suggestion
* `identifier LIKE pattern`: Is true if the property matches the provided `pattern`.

* `identifier UNLIKE x`: Is true if the property does not match the provided `pattern`.

```

Adds a bit of explanation matching the format for `CONTAINS` etc. above. I think `pattern` is clearer than `x` here, as `x` is used as a placeholder for the substrings above. Also fixes a typo in "`identifier`" - will also add a separate suggestion for this in isolation.

optimade.md
```diff
@@ -1287,8 +1287,22 @@ In addition to the standard equality and inequality operators, matching of parti
 
 OPTIONAL features: 
 
+The following comparison operators are OPTIONAL:
+
+* `identfier LIKE x`
```
ml-evs, line 1292:
```suggestion
* `identifier LIKE x`

```

optimade.md
ml-evs, line 1294:
```suggestion
* `identifier UNLIKE x`

```

ml-evs, line 1294: Is there a strong preference for `UNLIKE` rather than just `NOT(identifier LIKE pattern)`? Is `UNLIKE` is an SQL construct that I have not come across? 

I guess the precedent for this is our inclusion of `KNOWN` and `UNKNOWN` -- I just don't think I like having two ways of doing different optional things, e.g., an implementation could choose to implement only `LIKE` and therefore has to implement `~LIKE` but can still choose to ignore `UNLIKE`. At least `KNOWN`/`UNKNOWN` are both mandatory. I guess we could add to the spec that if `LIKE` is implemented then `UNLIKE` must also be implemented.

optimade.md
```diff
@@ -1287,8 +1287,22 @@ In addition to the standard equality and inequality operators, matching of parti
...
+
+* `identfier UNLIKE x`
+
 * Support for x to be an identifier, rather than a string is OPTIONAL.
 
+If implemented, the "LIKE" operator MUST behave as the correspoding standard SQL operator. In particular,
+The `x` string MUST be interpreted as a pattern where an underscore character ('_', ASCII DEC 95, HEX 5F)
+matches any single character and a percent character ('%', ASCII DEC 37, HEX 25) matches an arbitrary
+sequence of characters (including zero characters).
```
ml-evs, line 1301:
```suggestion
If implemented, the `LIKE` operator MUST behave as the corresponding standard SQL operator.
The `x` string MUST be interpreted as a string-matching pattern where an underscore character ('_', ASCII DEC 95, HEX 5F) matches any single character and a percent character ('%', ASCII DEC 37, HEX 25) matches an arbitrary sequence of characters (including zero characters).

```

Fix a few typos and enforces one line per sentence

optimade.md
ml-evs:
```suggestion
If the `UNLIKE` operator is supported, the behavior of this operator MUST be the negation of the "LIKE" operator; i.e.

```


based on the comment above, one more reformatting/wording change

ml-evs (changes requested):
Thanks for this PR @sauliusg, I definitely support the inclusion of such an operator once we've ironed the technicalities out. Sorry I didn't have a chance to look through this before the last meeting!

I can easily add this construct to the optimade-python-tools reference server if we want a live version to play around with (perhaps if COD has it too we can do some empirical tests on the regexp engines rather than just trusting the docs!)

My review below mostly focuses on wording/typos, though I do have some concerns about adding both `LIKE` and `UNLIKE` as separate optional features (based on my current reading of the spec).

rartino:
@sauliusg @vaitkus:

@sauliusg makes a good point about the escapes. The text in the json schema standard points to the syntax standardized in [ECMA-262, section 21.2.1](https://262.ecma-international.org/11.0/#sec-regexp-regular-expression-objects) which says the syntax is "modeled" after Perl 5. However, the json schema standard then says - on SHOULD level - that the expressions should be "limit[ed] [...] to the following regular expression tokens", which doesn't actually include any description of the escape token.

This is likely an oversight. In the ECMA standard it is handled in the grammar where an `Atom` can be `\` plus an `AtomEscape`, which is then further discussed in [21.2.2.8.1](https://262.ecma-international.org/11.0/#sec-regexp-regular-expression-objects).

If we end up quoting this subset of ECMA into our standard (as we discussed above that we may want to do), we probably want to clarify this. 

Furthermore, @sauliusg did question whether the lazy wildcards was widely supported; so we may want to consider to also exclude those, making our subset even more stricter than the one described by json schema.

sauliusg:
> @sauliusg, according to the regular expression examples given in https://json-schema.org/understanding-json-schema/reference/regular_expressions.html, special characters can be escaped using a double backslash.

Thanks, @vaitkus for finding this document! AFAIR, this is the same "random page" that I tried to refer to during the last meeting, or at least it looks very similar :).

Under these circumstances I have no more objections to include a subset of these regexps, or the subset of regexps listed in https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-regular-expressions, in OPTIMADE.

The only point I'd like to stress rather strongly:

I would not reference any JSON document as a source of the description. I would rather explicitly list the permissible regexps in the OPTIMADE standards; we can (and should) reference the rationale, where these definitions come from – PCRE, JSON Schema, JSON Schema draft, etc., but the _definitive_ list should be in the OPTIMADE standard.

All RE construct listed in the mentioned documents seem to be modeled after the PCRE, and PCRE is probably the most convenient form of REs, so I suggest we go for it.

That would leave us with one 'string MATCH "re"' or 'string MATCH /re/' operator, RECOMMENDED, to implement RE matches.

optimade.md
sauliusg: > @sauliusg makes a good point about the escapes. The text in the json schema standard points to the syntax standardized in [ECMA-262, section 21.2.1](https://262.ecma-international.org/11.0/#sec-regexp-regular-expression-objects) which says the syntax is "modeled" after Perl 5. However, the json schema standard then says - on SHOULD level - that the expressions should be "limit[ed] [...] to the following regular expression tokens", which doesn't actually include any description of the escape token.
> 
> This is likely an oversight. In the ECMA standard it is handled in the grammar where an `Atom` can be `\` plus an `AtomEscape`, which is then further discussed in [21.2.2.8.1](https://262.ecma-international.org/11.0/#sec-regexp-regular-expression-objects).

If we describe, as suggested, the RE syntax in the OPTIMADE independently, we can strengthen the requirements as we see it fit. This is a good opportunity to correct the oversight, if it really was.

optimade.md
sauliusg: > Is there a strong preference for `UNLIKE` rather than just `NOT(identifier LIKE pattern)`? Is `UNLIKE` is an SQL construct that I have not come across?

UNLIKE is shorter, which is significant for URL-embedded querries...
 
> I guess the precedent for this is our inclusion of `KNOWN` and `UNKNOWN` -- I just don't think I like having two ways of doing different optional things, e.g., an implementation could choose to implement only `LIKE` and therefore has to implement `~LIKE` but can still choose to ignore `UNLIKE`. At least `KNOWN`/`UNKNOWN` are both mandatory. I guess we could add to the spec that if `LIKE` is implemented then `UNLIKE` must also be implemented.

I would say that if LIKE is implemented, UNLIKE also MUST be implemented; i.e. `x UNLIKE "%s%"` and `NOT(x LIKE "%s%"` MUST be synonyms – can we decide on that?  If the `LIKE` is implemented, then implementing `UNLIKE` does not require any extra effort.

rartino:
> I would not reference any JSON document as a source of the description. I would rather explicitly list the permissible regexps in the OPTIMADE standards; we can (and should) reference the rationale, where these definitions come from – PCRE, JSON Schema, JSON Schema draft, etc., but the definitive list should be in the OPTIMADE standard.

Yes, also from prior discussion above I think we are generally in agreement on that the rules need to be stated in our standard.

However - this a technicality - it is relevant for us to say that what we end up with is not just based on/inspired by the PCRE regex subset in the JSON Schema draft, but is equal to, or a subset (whichever is true) of the JSON Schema draft standard. The reason is that this matters for the other place we want to invoke this regex standard, namely when specifying RE string validations in property definitions.

Question: what should a server do if a client uses REGEX that are valid PCRE2 but NOT constrained as specified? Since we've concluded that the ECMA standard is based on Perl 5 (i.e., PCRE2 if I'm not mistaken) we may as well say that supporting the subset is a MUST, but full support of PCRE2 is allowed, and on MAY level?

> That would leave us with one 'string MATCH "re"' or 'string MATCH /re/' operator, RECOMMENDED, to implement RE matches.

Right, about that - the PR right now implements them as "/"-delimited. I was initially surprised by that decision, but realized I had no reason to oppose it. Does anyone see things differently?

sauliusg:
> I agree with most of your latest comments about level 2 support, so we may be converging on a design here. In particular we seem to agree that:
> 
>     * level 2 and level 3 partial text matching should use segregated syntax.
> 
>     * OPTiMaDe basic pattern support should work equivalently to standard SQL LIKE with ESCAPE `\` (but possibly with other wildcard characters.)
> 
> 
> So, perhaps then all that is left is to decide is between these alternatives?:
> 
>     * SQL LIKE-based basic patterns with wildcard characters `%` and `_` as it is in that standard.
> 
>     * SQL-LIKE-like basic patterns with wildcard characters `*` and `?`.
> 
>     * Skip basic pattern support, have only regex.

So to make things converge – is there a consensus, or at least not a strong opposition, to the following OPTIONAL features:

- The LIKE operator: `string LIKE "%s_"`. Characters of the SQL standard are used (`%` for any sequence of characters, `_` for any single character; alternatively – `*` for any sequence of characters, `?` for any single character, which is essentially a shell glob); the `UNLIKE` operator MUST be a synonym of `NOT ... LIKE` (i.e. `NOT (string LIKE "*s?")` <=> `string UNLIKE "*s?"). Escape character is backslash (`'\'`), which escapes itself ("\\" → a string with a single character `'\'`);

- The MATCH operator that uses a subset of PCRE (based on [1] and [2]); `string MATCH /abc.*xyz[0-9]{3,7}/`; I'll check if slashes `'/'` can be easily accommodated into out Filter language grammar – otherwise we use the standard Filter strings. Escape character is backslash (`'\'`), which escapes itself ("\\" → a string with a single character `'\'`).

Shall we go for it?

sauliusg:
> > That would leave us with one 'string MATCH "re"' or 'string MATCH /re/' operator, RECOMMENDED, to implement RE matches.
> 
> Right, about that - the PR right now implements them as "/"-delimited. I was initially surprised by that decision, but realized I had no reason to oppose it. Does anyone see things differently?

You are right, I start forgetting what is in the PR :). 

The `/re/` syntax should minimize the escape backslashes that are needed.

merkys:
> Question: what should a server do if a client uses REGEX that are valid PCRE2 but NOT constrained as specified? Since we've concluded that the ECMA standard is based on Perl 5 (i.e., PCRE2 if I'm not mistaken) we may as well say that supporting the subset is a MUST, but full support of PCRE2 is allowed, and on MAY level?

If we do so, OPTIMADE RE will have to stay a subset of PCRE2 at least until the next major release. I am not against it, just stating that all additions/changes will have to be PCRE2 compatible. In addition, PCRE2 standard has life of its own, so if we tie OPTIMADE RE to PCRE2, we have to tie to some specific version.

rartino:
> If we do so, OPTIMADE RE will have to stay a subset of PCRE2 at least until the next major release. I am not against it, just stating that all additions/changes will have to be PCRE2 compatible. In addition, PCRE2 standard has life of its own, so if we tie OPTIMADE RE to PCRE2, we have to tie to some specific version.

This is all true. And, arguably (but I suspect this will be argued...) our reference should be [ECMA-262 v11](https://262.ecma-international.org/11.0/) just like JSON Schema rather than a specific version of the PCRE2 library.

Nevertheless, I feels strange about *forcing* those with full PRCE2 implementations in their backends to put in the extra work to lock away useful features. I suspect people will be reluctant to do that in practice.

We *could* say (not saying it is a good idea) something along the lines of: "For *non-compliant REGEXs* servers MAY behave in any undefined way, including serving matches according to the PCRE2 standard (or any other standard). Non-compliant REGEXs SHOULD trigger a warning. Servers are free to document their behavior for non-compliant REGEXs."

Such a statement does work against general compatibility, but will make implementing MATCH for various underlying REGEX standards a lot easier. Also for those backends where one need to translate REGEXs it would be nice to only have to worry about what the translator is doing for compliant REGEXs.

sauliusg:
> > * The MATCH operator that uses a subset of PCRE (based on [1] and [2]); `string MATCH /abc.*xyz[0-9]{3,7}/`; I'll check if slashes `'/'` can be easily accommodated into out Filter language grammar – otherwise we use the standard Filter strings. Escape character is backslash (`'\'`), which escapes itself ("" → a string with a single character `'\'`).
> 
> Using `/` in grammar might clash with math operator `/` if we want to introduce math operators in future. SQL seems to support math operators.

I think the fact that the RE slash follows the `MATCH` keyword and REs never participate in arithmetic expressions should resolve the ambiguity...

sauliusg:
> > The MATCH operator that uses a subset of PCRE (based on [1] and [2]); string MATCH /abc.*xyz[0-9]{3,7}/; I'll check if slashes '/' can be easily accommodated into out Filter language grammar – otherwise we use the standard Filter strings. Escape character is backslash (''), which escapes itself ("" → a string with a single character '').
> 
> Doesn't the grammar change in the PR already do this with "/"?

Yes it does; I forgot it (had to review the PR again ;)

merkys:
> > Using `/` in grammar might clash with math operator `/` if we want to introduce math operators in future. SQL seems to support math operators.
> 
> I think the fact that the RE slash follows the `MATCH` keyword and REs never participate in arithmetic expressions should resolve the ambiguity...

Indeed. I suppose our LALR parsers will be able to handle that.

sauliusg:
> I suggest to skip `LIKE` entirely and just add `MATCH`. For simple matching we already provide `BEGINS WITH`, `ENDS WITH`, and `CONTAINS` which are user friendly in their definitions (i.e., no need to remember any particular wildcard syntax). Those with more sophisticated needs for general wildcard matching can use `MATCH`. Lets try to keep down the implementation work for servers.
> 
> >

Agreed. So we are converging on:



-   The MATCH operator that uses a subset of PCRE (based on [1] and [2]); string MATCH /abc.*xyz[0-9]{3,7}/; I'll check if slashes '/' can be easily accommodated into out Filter language grammar – otherwise we use the standard Filter strings. Escape character is backslash ('\\'), which escapes itself ("\\\\" → a string with a single character '\\').

rartino:
> The /re/ syntax should minimize the escape backslashes that are needed.

This is an excellent point. This is why we *have* to use `/` as the delimiter. Otherwise users would be exposed to figuring out our own double-escaping rules for regexs inside strings. As someone who has occasionally been exposed to preparing regexs through >= 4 layers of escaping, extra escape layers is something that absolutely should be avoided.

merkys:
To test for the possible grammar conflicts with putative `/` operator I made the following change (which I think reflects a possible use-case):
```diff
diff --git a/optimade.md b/optimade.md
index 030be71..a364197 100644
--- a/optimade.md
+++ b/optimade.md
@@ -2164,7 +2164,7 @@ Filter = [Spaces], Expression ;
 
 Constant = String | Number ;
 
-Value = String | Number | Property ;
+Value = String | Number | Property | Number, '/', Number ;
 (* Note: support for Property in Value is OPTIONAL *)
 
 ValueList = [ Operator ], Value, { Comma, [ Operator ], Value } ;

```
With it, every grammar test fails with the following output:
```
> Error: in tests/generated/Filter.g: line 110:
>     inherent ambiguity in production 'Value' starting with tokens "+",
>     "-", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", and "."
> 
> Value = String | Number | Property | Number '/' Number;
> 
```
Thus I assume `/` operator and current RE syntax cannot co-exist. Which is not a show-stopper, unless we foresee introduction of math operators in OPTIMADE.

rartino:
> Thus I assume / operator and current RE syntax cannot co-exist. Which is not a show-stopper, unless we foresee introduction of math operators in OPTIMADE

This strikes me as strange, and may be a consequence on the details of how `/regex/` is introduced in the grammar, because it should be possible to formulate the grammar so that a division `/` can ONLY follow a `<number>` whereas `<non-number>/` puts you in a state where the following `<anything>/` completes the regexp, and thus cannot be ambiguous with a division.

sauliusg:
> > Thus I assume / operator and current RE syntax cannot co-exist. Which is not a show-stopper, unless we foresee introduction of math operators in OPTIMADE
> 
> This strikes me as strange, and may be a consequence on the details of how `/regex/` is introduced in the grammar, because it should be possible to formulate the grammar so that a division `/` can ONLY follow a `<number>` whereas `<non-number>/` puts you in a state where the following `<anything>/` completes the regexp, and thus cannot be ambiguous with a division.
> 
> **edit** Actually, looking at your error message, isn't that just because it is impossible to distinguish `Number` and `Number '/' Number`? What if you rewrite the rule as:
> 
> ```
> Value = String | Number, ['/', Number] | Property
> ```

This sounds strange to me as well, since I thing we tried things similar to REs, I guess. I wold like to look closer into the grammar and see if we can transform it in such a way that the language stays the same (with REs) and the ambiguity goes away.

rartino:
@sauliusg Why did you close this? If we get #490 merged, this PR can be updated to just reference that format for the MATCH operator and we will be done?

Issue #92, opened 2019-06-12
Diagnostics of misspelled property names in Filter queries


sauliusg:
If property name is not recognised by the queried end-point, an error diagnosis would be appropriate. However, we also want to formulate the following queries:

`filter=_aida_bandgap<2.0 OR _mp_bandgap<2.5`

and then send the same query to AiiDA-supported and MP OPTiMaDe endpoints.

For these queries to succeed, the following behaviour is suggested:

- if a database receives a query filter with universal Optimade properties (i.e. defined without a database prefix, and deemed universal for all databases), or a property with database private prefix (e.g. _cod_ for the COD database), then the database MUST check if the provided property names are known, and MUST return an appropriate error code if they are not known the database.

- if a database received a query filter with a property from an **unknown** database, the end-point MUST behave as if that property is unknown, i.e. is if it has the value `null`. If the database prefix of the unknown property is similar to that of a known database, the response MAY issue a corresponding warning.

merkys:
Therefore:

* unrecognized universal property results in an error (HTTP 400 or the like);
* unhandled database-specific property results in a warning.

Sounds good to me.

merkys:
Fixed by #162

PR #162, opened 2019-07-15
Adding description of unknown property name handling

sauliusg:
Adding text as discussed in the Issue https://github.com/Materials-Consortia/OPTiMaDe/issues/92.


optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
+formulate a single query suitable for multiple databases, even when
+querying database-sepcific properties. For example, we would like to
+construct the following query:
+
+`filter=_aida_bandgap<2.0 OR _mp_bandgap<2.5`
+
+and then send it query to AiiDA-backed and MP-backed OPTiMaDe
+endpoints.
+
+For these queries to succeed, the following behavior is suggested:
```
giovannipizzi: There are a few other minor typos or minor things that I would change, but I start with the main question: do we want to SUGGEST this, or enforce this? Because below we then say MUST and I think this is now confusing.

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
+  universal for all databases), or a property with the database's
+  private prefix (e.g. cod for the COD database), then the database
+  MUST check if the provided property names are known, and MUST return
+  an appropriate error code if they are not known the database.
+
+* if a database received a query filter with a property from an
+  unknown database, the end-point MUST behave as if that property is
+  unknown, i.e. is if it has the value null. If the database prefix of
+  the unknown property is similar to that of a known database, the
+  response MAY issue a corresponding warning.
```
giovannipizzi: Why do we also want to issue a warning? I understand if this is possible mistype of *your* db (let's say you find `_cdo_` and you are `_cod_` and you don't have `_cdo_` in the list of optimade providers, but in general, is this useful?

optimade.md
sauliusg: The formal conformance level is MUST; 'suggested' was a milder expression for humans ;)

Should we change 'suggested' -> 'required'?

optimade.md
sauliusg: I think warnings are very useful in this case: imagine you query the COD, and instead of `_cod_deposition_date > "2019-01-01"` you mistype `_cdo_deposition_date > "2019-01-01"`. The OPTiMaDe back end assumes that `_cdo_deposition_date` is always `null`, thus the comparison is always `false`, and you end up with no structures instead of a required ones. This can be very confusing, especially of you query has another ORed components that **do** return some structures – you would en up in incorrect selection, but this would be hard to detect, and the typo is difficult to notice as well.

A warning would draw your attention to a possible mistake; it is supposed to detect typos similar to a known database names ('cod' -> 'cdo' would most probably be considered "similar"). Still, the formal similarity level is not specified, and the implementation of similar string identification might be tricky, so I left this requirement at the "MAY" level, although "SHOULD" could be discussed since the feature is very useful (we do not want OPTiMaDe to be used to retrieve garbage data, do we?)

optimade.md
giovannipizzi: even better, use proper RFC e.g. `MUST`

optimade.md
giovannipizzi: I agree - but my point is that you need to check only for your own name and similar ones - e.g. if COD finds a `_aidia` instead of `_aiida` it shouldn't complain - it could just be a different, recent/new DB that it does not know. If instead you find `_cdo` you should warn. 

By reducing the warnings, we make sure these are not very often printed, that will in turn mean that the user will start not even reading them

rartino: I'm in favor of that implementations SHOULD issue a warning for any unknown property name encountered that does not trigger an error.

Sending queries that include other databases' fields is going to be something quite unusual anyway, where a user should know that what they are doing is "odd" and therefore they shoulde expect warnings. I don't think this is an issue of over-using the warning system.

giovannipizzi: Ok. But then should we then just trigger a warning anyway, independent of string "closeness" or "similarity" to known strings?

rartino: I'd say no to string closeness comparisons.

To clarify, in my opinion:
* Global properties and properties with the providers own prefix, which the implementation *can* verify as non-existing should generate errors, just as described in this PR.
* Properties with a different db prefix, which the implementation thus *cannot* verify as existing or non-existing, are to be handled in filters as unknowns as described in this PR, but SHOULD always generate warnings about being handled this way. 

When *not* trying to do the type of query this PR tries to handle, these warnings signal that the user has misspelled the database's own prefix (perhaps badly so). They are thus highly relevant.

In queries of the type this PR tries to handle, these warnings will be somewhat useful in confirming that the databases does what the user intends. E.g. *"Warning: unrecognized property \_aida\_bandgap handled as unknown"*. I would very much expect that warning if I sent the example query given in the text of this PR to the MP OPTiMaDe implementation.

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
   For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the IDs of the referenced entries alongside the respective human-readable description of the relationships. 
   It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
 
+## <a name="h.3.6">3.6. Handling unknown property names</a>
+
+If a property name is not recognised by the queried end-point, an
```
```suggestion
If a property name is not recognized by the queried endpoint, an
```
(endpoint is written like this throughout the rest of the text)

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
   For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the IDs of the referenced entries alongside the respective human-readable description of the relationships. 
   It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
 
+## <a name="h.3.6">3.6. Handling unknown property names</a>
+
+If a property name is not recognised by the queried end-point, an
+error diagnosis would be appropriate. However, we also want to
+formulate a single query suitable for multiple databases, even when
+querying database-sepcific properties. For example, we would like to
```
```suggestion
querying database-specific properties. For example, we would like to
```

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
   For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the IDs of the referenced entries alongside the respective human-readable description of the relationships. 
   It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
 
+## <a name="h.3.6">3.6. Handling unknown property names</a>
+
+If a property name is not recognised by the queried end-point, an
+error diagnosis would be appropriate. However, we also want to
+formulate a single query suitable for multiple databases, even when
```
```suggestion
support the formulation of a single query suitable for multiple databases, even when
```

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
   For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the IDs of the referenced entries alongside the respective human-readable description of the relationships. 
   It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
 
+## <a name="h.3.6">3.6. Handling unknown property names</a>
+
+If a property name is not recognised by the queried end-point, an
+error diagnosis would be appropriate. However, we also want to
+formulate a single query suitable for multiple databases, even when
+querying database-sepcific properties. For example, we would like to
+construct the following query:
```
```suggestion
support the following query:
```

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
 
+## <a name="h.3.6">3.6. Handling unknown property names</a>
+
+If a property name is not recognised by the queried end-point, an
+error diagnosis would be appropriate. However, we also want to
+formulate a single query suitable for multiple databases, even when
+querying database-sepcific properties. For example, we would like to
+construct the following query:
+
+`filter=_aida_bandgap<2.0 OR _mp_bandgap<2.5`
```
rartino: I'd like this anonymized. In my opinion the specification should be neutral in its 'endorsement' of specific databases, even in examples. E.g.,
```suggestion
`filter=_exmpl_bandgap<2.0 OR _exmpl2_bandgap<2.5`
```
with some additional text defining that `_exmpl_` and `_expl2_`  are supposed to be a database-specific prefix of two different databases.

Also, if this text anyway remains, the `_aiida_` prefix is misspelled.

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
+
+and then send it query to AiiDA-backed and MP-backed OPTiMaDe
+endpoints.
+
+For these queries to succeed, the following behavior is suggested:
+
+* if a database receives a query filter with universal Optimade
```
```suggestion
* if a database receives a query filter with universal OPTiMaDe
```

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
+  properties (i.e. defined without a database prefix, and deemed
```
```suggestion
  properties (i.e., defined without a database prefix, and deemed
```

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
+  universal for all databases), or a property with the database's
+  private prefix (e.g. cod for the COD database), then the database
+  MUST check if the provided property names are known, and MUST return
+  an appropriate error code if they are not known the database.
```
```suggestion
  an appropriate error code if they are not known to the API implementation.
```

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
+
+and then send it query to AiiDA-backed and MP-backed OPTiMaDe
+endpoints.
+
+For these queries to succeed, the following behavior is suggested:
+
+* if a database receives a query filter with universal Optimade
+  properties (i.e. defined without a database prefix, and deemed
+  universal for all databases), or a property with the database's
+  private prefix (e.g. cod for the COD database), then the database
```
```suggestion
  private prefix (e.g. `_exmpl_`), then the API implementation
```

optimade.md
```diff
@@ -539,6 +540,34 @@ In the default response format, relationships are encoded as [JSON API Relations
...
+  MUST check if the provided property names are known, and MUST return
+  an appropriate error code if they are not known the database.
+
+* if a database received a query filter with a property from an
+  unknown database, the end-point MUST behave as if that property is
+  unknown, i.e. is if it has the value null. If the database prefix of
```
rartino: (For rationale, see separate comment)

In my opinion 
> If the database prefix of the unknown property is similar to that of a known database, the response MAY issue a corresponding warning.

should be replaced with

> The API implementation SHOULD issue warnings for every unrecognized property and specify that they have been handled as unknown.

optimade.md
dwinston: I agree that "suggested" should be proper RFC, e.g. one of `MAY`, `SHOULD`, `MUST`.

optimade.md
sauliusg: Maybe better: "support construction of the following query"?

optimade.md
sauliusg: If we change to "API implementation" then we will probably have to change it elsewhere in the suggested text as well; but is it really the API implementation that decides on the names or the database? It is actually the interaction of both; without the database providing a property API an API implemenatation can not support the name; but API can also fail to support some property name even if the database has corresponding data...

optimade.md
sauliusg: After some though I realise that issuing warning on every unrecognised property (even if we know that it is not "our" property, i.e. we recognise another database's prefix) would **always** issue warnings on mixed-database queries. These expected warnings will then obscure the "real" (mistyped) ones; user (including myself) will tend to ignore warnings if they are too numerous and mostly uninformative. (Let's not shout "wolf!" too often). This consideration would point to issuing warnings only when something indeed might be wrong.

What about such behaviour:

- if a database prefix is **recognised** by an API implementation as belonging to a different existing database, no warning is issued, or different kind of warning is issued (e.g. "property 'xyz' for database 'example9' was assumed to be `null`"); this warning could have MAY or SHOULD conformance level;

- if a database prefix is **not** recognised, we issue a warning (e.g. "database 'exxxxample' was not recognised"); conformance level for this warning would be SHOULD.

Is that OK? If you are fine with it, I'll change the text accordingly.

sauliusg:
> **Edit:** Also, @CasperWA is trying to push us all towards the formatting of one sentence per line in the raw

This will tend to give rather long lines which are not so comfortable...
What about folding the sentences to some reasonable width (80 symbols) and starting each sentence on a new line? I can probably even get an Emacs mode for this...

CasperWA:
> > **Edit:** Also, @CasperWA is trying to push us all towards the formatting of one sentence per line in the raw
> 
> This will tend to give rather long lines which are not so comfortable...

It's much worse detecting minor changes in a long sentence (git-wise) if the sentence is split on several lines.

> What about folding the sentences to some reasonable width (80 symbols) and starting each sentence on a new line? I can probably even get an Emacs mode for this...

Most editors should be capable of wrapping long single-lines already, I believe?

I feel like I must be adamant about this particular thing. It will significantly reduce minor errors and/or decrease time spent on PR reviews, since the diff for a change in every sentence will be clearly stated. So in my opinion the benefits far outweigh the drawbacks :)

optimade.md
rartino: I see your rationale, and can accept this solution. I assume @giovannipizzi prefers this over my suggestion as well, since it will give less warnings.

Do we have a facility for INFO level messages? I may want to still trigger those in this case (but, completely optionally for my implementation; I wouldn't want to mandate INFO level messaging.)

optimade.md
rartino: We define what API Implemenation means in term definitions, and elsewhere in the specification text we use this term with the meaning you are looking for here, i.e., when it is the "interaction" between the API implementation + the database backend.

optimade.md
rartino: Can you try explain what are you aiming to convey with the extra word "construct" here? 

I replaced it with "support" because formally the API specification doesn't deal with the "construction" of anything, only what is a valid and invalid API request and API response.

Also, while we are discussing the formulations here: I don't find it very clear who 'we' refers to in this paragraph, so I'd also suggest rephrasing those.

sauliusg:
> It's much worse detecting minor changes in a long sentence (git-wise) if the sentence is split over several lines.

No really; actually line-wise diff (the usual vcs system default) will work better with folded lines. With 1 sentence/line we will have to resort to 'wdiff' (word diff) to see that actual changes.

> > What about folding the sentences to some reasonable width (80 symbols) and starting each sentence on a new line? I can probably even get an Emacs mode for this...

> Most editors should be capable of wrapping long single-lines already, I believe?

They should :)

> I feel like I must be adamant about this particular thing. It will significantly reduce minor errors and/or decrease time spent on PR reviews, since the diff for a change in every sentence will be clearly stated. So in my opinion the benefits far outweigh the drawbacks :)

OK, lets give it a try and see. Actually I was myself advocating 1 sentence/line policy, however not for manual editing but for automatic processing :). We can try and see if it works for editing as well.

But currently the text is in a mixed mode – some sentences are folded, and some long sentences start on a line after another long sentence. Will you fix that?

sauliusg:
> > > **Edit:** Also, @CasperWA is trying to push us all towards the formatting of one sentence per line in the raw
> > 
> > 
> > This will tend to give rather long lines which are not so comfortable...
> 
> It's much worse detecting minor changes in a long sentence (git-wise) if the sentence is split over several lines.

Actually, does Github fold long lines in diffs? As I currently see, long lines introduce a horizontal scroll-bar; If this is always the case, then I would say this is totally unusable and a complete show stopper for long lines.

Can you confirm that Github interface handles long lines decently?

sauliusg:
> **Edit:** Also, @CasperWA is trying to push us all towards the formatting of one sentence per line in the raw markdown. This has advantages when looking at text changes and for suggesting changes, etc. I think this is a good idea. (Which I think we should update the whole spec to when/if converting to RST)

BTW, should we open a separate issue on that and move the discussion there?

sauliusg:
> > It's much worse detecting minor changes in a long sentence (git-wise) if the sentence is split over several lines.
> 
> No really; actually line-wise diff (the usual vcs system default) will work better with folded lines. With 1 sentence/line we will have to resort to 'wdiff' (word diff) to see that actual changes.

One more thought: it is indeed awkward to detect changes in a sentence when the whole **paragraph** gets refolded; changes, which are only cosmetic, can then propagate arbitrarily far.

But what I would suggest is a combination of a current approach: we fold lines **in each sentence independently**, and start each sentence on a separate line. In this way, when you change a sentence, you will only see the corresponding lines of this sentence changed. Easy to read, easy diff, easy to merge, no unnecessary merge conflicts. What you think?

E.g.:

folded sentences, each sentence starts on its own line (no scroll-bars on wide-enough displays):

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua. 
Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip
ex ea commodo consequat. 
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu 
fugiat nulla pariatur. 
Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt
mollit anim id est laborum.

 * Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor 
   incididunt ut labore et dolore magna aliqua. 

 * Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip 
   ex ea commodo consequat. 
   Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat 
   nulla pariatur. 
   Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt 
   mollit anim id est laborum.
```

long line sentences suggestion (inevitable scroll-bars in Github and some other tools):

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. 
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. 
Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

 * Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 

 * Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. 
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. 
Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```

optimade.md
sauliusg: OK, I'll have a careful look at the current wording and adapt the text accordingly.

optimade.md
sauliusg: 1b10d49 Rewording the phrase 'behavior is suggested' using the RFC2119 "MUST" keyword.

optimade.md
sauliusg: > Sending queries that include other databases' fields is going to be something quite unusual anyway, where a user should know that what they are doing is "odd" and therefore they should expect warnings. I don't think this is an issue of over-using the warning system.

I agree with @rartino that the database SHOULD issue a warning (or a note), but for somewhat different reason :).

I think the situation when you send requests with multiple database prefixes is quite normal and there is nothing "odd" about it, but the warnings (or even note-level messages) will not obscure other ones since one can (and should ;) filter them out. These messages just explicitly state what has happened in a backend, making the situation explicit.

optimade.md
sauliusg: @rartino : your suggestion "support" does not fit into the current text: it is the backend that "supports" the query; but the client (as is implied in the text) "constructs" the query (in a very general sense – composes the query from characters). I leave "construct" at the moment for the lack of better word; "maybe "compose a query" would be a better alternative, but definitely not the "support".

sauliusg: > Also, while we are discussing the formulations here: I don't find it very clear who 'we' refers to in this paragraph, so I'd also suggest rephrasing those.

Changing "we" to "an OPTiMaDe client".

optimade.md
sauliusg: See commits https://github.com/sauliusg/OPTiMaDe/commit/7d8d3cc54a92631d2ea94f39e4e06d451f2a0fb1, https://github.com/sauliusg/OPTiMaDe/commit/b9630a5dc2af680f0d989b3b5fb3bf4961a585d3 and https://github.com/sauliusg/OPTiMaDe/commit/906d27aec803dd3cd89ea4322b11bb7b9e1a8ec3 for a suggested wording.

optimade.md
sauliusg: The solution is, I believe, implemented in the latest commits of this PR.

sauliusg: > Do we have a facility for INFO level messages? I may want to still trigger those in this case (but, completely optionally for my implementation; I wouldn't want to mandate INFO level messaging.)

I'm not sure if the current implementation specifies message levels formally, but it would be helpful to have ERROR, WARNING and NOTE/INFO level messages. The distinction would be as follows:

- ERROR would mean that the API implementation could not server the request. No data is provided, and the computations can not reasonably continue on the response.

- WARNING would mean that the response contains valid data, and formally the computations can continue, but the result is probably not what the client expects. For example, the query might contain a misspelled foreign database property and return 0 entries as a response (which is valid), but there is a hight probability that the property name was mistyped and the client expects query on a different property;

- NOTE/INFO would mean that the query ran as expected by the API implementation, and the message just states what has been done to make the situation explicit in cases where it otherwise might be ambiguous.

sauliusg:
Hi, folks, for some reason I can not find a text of any new reviews on the GitHub, so I found a button "re-request review" and clicked on it. If this does not appear on the Github Web, please e-mail me the links/text.

optimade.rst
```diff
@@ -560,6 +560,23 @@ In the default response format, relationships are encoded as `JSON API Relations
...
+`filter=_exmpl_bandgap<2.0 OR _exmpl2_bandgap<2.5`
+
+and then send it query to OPTiMaDe endpoints backed by the engines `Exmpl` and `Exmpl2`.
+
+For these queries to succeed, the backend MUST implement the following behavior:
+
+* if an API implementation receives a query filter with universal OPTiMaDe properties (i.e. defined without a database prefix, and deemed universal for all databases), or a property with the database's own prefix (e.g. `_exmpl_` for the `Exmpl` engine), then the API implementation MUST check if the provided property names are known, and MUST return an appropriate error code if the property names are not known to the API implementation.
+
+* if the API implementation received a query filter with a property from an unknown database (e.g. `_exmpl2_` for the `Exmpl` engine), the API implementation MUST behave as if that property is unknown, i.e. is if it has the value null. If the database prefix of the unknown property is similar to that of a known database, the response MAY include a corresponding warning.
+
```
rartino: The text does not appear to say what we in the end agreed on with regards to SHOULD conformance. 
It also asks implementations to return "error" without specifying what error. 
And the mixed tenses seem odd to me, e.g., "receives" in the first clause and "received" in the second. 

Here is my attempt at a re-write to address the issues I see. Either adopt this or use it as inspiration to adapt the existing version. Note my suggestion to use a comment block at the end to give the rationale, rather than put it at the beginning.

```reStructuredText

When an implementation receives a request with a query filter that refers to an unknown property name it is handled differently depending on the database-specific prefix:

* If the property name has no database-specific prefix, or the database-specific prefix that belongs to the implementation itself, the error :http-error:`400 Bad Request` MUST be returned with a message indicating the offending property name. 

* If the property name has a database-specific prefix that does not belong to the implementation itself, it MUST NOT treat this as an error, but rather MUST evaluate the query with the property treated as unknown, i.e., comparisons are evaluated as if the property has the value :val:`null`. Furthermore, if the implementation does not recognize the prefix at all, it SHOULD return a warning that indicates that the property has been handled as unknown. However, if the prefix is recognized, i.e., as belonging to a known database provider, the implementation SHOULD NOT issue a warning but MAY issue diagnostic output with similar information. 

    The rationale for treating properties from other databases as unknown rather than triggering an error is for OPTiMaDe to support queries using database-specific properties that can be sent to multiple databases.
    For example, the following query can be sent to API implementations exmpl1 and exmpl2 without generating any errors:

    :filter:`filter=_exmpl1_bandgap<2.0 OR _exmpl2_bandgap<2.5`

```

optimade.rst
```diff
@@ -560,6 +560,23 @@ In the default response format, relationships are encoded as `JSON API Relations
     For each entry type, the relationships with entries of that type can then be encoded in a field with the name of the entry type, which are to contain a list of the IDs of the referenced entries alongside the respective human-readable description of the relationships.
     It is the intent that future versions of this standard uphold the viability of this encoding by not standardizing property names that overlap with the entry type names.
 
+Handling unknown property names
+===============================
+
+If a property name is not recognized by the queried endpoint, an error diagnosis would be appropriate. However, we also want to support the formulation of a single query suitable for multiple databases, even when querying database-specific properties.
```
merkys: 'an error diagnosis would be appropriate' -> 'an error diagnosis SHOULD be appropriate'

'we also want to support': We don't seem to use 'we' in the specification. Could we get rid of this instance too? Just for consistency.

optimade.rst
```diff
@@ -560,6 +560,23 @@ In the default response format, relationships are encoded as `JSON API Relations
...
+
+For example, an OPTiMaDe client might compose the following query:
+
+`filter=_exmpl_bandgap<2.0 OR _exmpl2_bandgap<2.5`
+
+and then send it query to OPTiMaDe endpoints backed by the engines `Exmpl` and `Exmpl2`.
```
merkys: 'send it query' -> 'send the query'

optimade.rst
```diff
@@ -560,6 +560,23 @@ In the default response format, relationships are encoded as `JSON API Relations
...
+
+For these queries to succeed, the backend MUST implement the following behavior:
```
merkys: 'these' where only one query is given: to be replaced either with 'this' or 'such'

optimade.rst
merkys: I like @rartino's wording. It reads more like a pseudocode (`if <condition> then <effect>`), but provides the required justification and examples for the reader looking for details.

sauliusg: I also like the @rartino version, it indeed reads more like a specification (while mine was too much discussion-like) and, I believe addresses all points rased by @rartino and @merkys .

The only point remaining: the text mentions:

> if the prefix is recognized, /.../, the implementation SHOULD NOT issue a warning but MAY issue diagnostic output with similar information

which implies that we have warnings and something else, called here 'diagnostic output'. To me warnings, as well as error messages, are all ''diagnostic output', so the text becomes unclear.

Should we explicitly differentiate severity levels to `errors`, `warnings` and `notes` (or `info`, but `notes` is the noun similar to the previous two)? If we do so, should this be stated in the specification explicitly, describing what the intended meanings of these three levels are? (briefly: errors – can't produce any output, can't continue; `warnings` – produced some output, syntactically correct but double check if this is really what you wanted; `notes` – everything OK, just stating explicitly what has been done for your records).

merkys: > The only point remaining: the text mentions:
> 
> > if the prefix is recognized, /.../, the implementation SHOULD NOT issue a warning but MAY issue diagnostic output with similar information

I overlooked this. It seems to me that we don't have agreed means to convey the diagnostic output yet, do we? If it's not in the specification yet, this should come in separate issue/PR.

rartino: Feel free to continue in your PR, I will not do a PR.

You can edit the text to define 'note'-level messages (but as @merkys says maybe in a separate PR). I think it needs a new section 3.3.7 for the general facility and an edit of 3.3.2 to defines how to communicate `notes` in the json-formatted output.

However, until such an edit is in, I personally don't see a problem with the suggested text. As you say, "diagnostic output'" is a fully generic concept. It refers to any information that can be output inline using a field that allows such info in the output data format used or in side-channels (e.g., console messages). And the text just hints the implementer that it is ok to supply this info using such facilities, BUT, it says explicitly, NOT as a warning or error.

But sure, when an official info facility exists, then this is better reformulated to say that this info MAY go there.

merkys: I like generic terms being used throughout the document. However, when our primary format (JSON API) lacks mechanisms covering these terms, we risk loosing the standardization.

sauliusg: > I overlooked this. It seems to me that we don't have agreed means to convey the diagnostic output yet, do we? If it's not in the specification yet, this should come in separate issue/PR.

I think a PR to specify the NOTE/WARNING/ERROR levels and diagnostic details would be appropriate.

optimade.rst
```diff
@@ -563,19 +563,21 @@ In the default response format, relationships are encoded as `JSON API Relations
 Handling unknown property names
 ===============================
 
-If a property name is not recognized by the queried endpoint, an error diagnosis would be appropriate. However, we also want to support the formulation of a single query suitable for multiple databases, even when querying database-specific properties.
+When an implementation receives a request with a query filter that refers to an unknown property name it is handled differently depending on the database-specific prefix:
 
-For example, an OPTiMaDe client might compose the following query:
+* If the property name has no database-specific prefix, or the database-specific prefix that belongs to the implementation itself, the error :http-error:`400 Bad Request` MUST be returned with a message indicating the offending property name.
```
merkys: The sentence

> If the property name has no database-specific prefix, or the database-specific prefix that belongs to the implementation itself, ...

sounds somewhat ambiguous. I suggest

> If the property is recognized as belonging to the implementation itself, or it has no database-specific prefix altogether, ...

sauliusg: I've read both formulations several times. IMHO:

- i can not find a second way to interpret the first (the current) sentence; maybe I'm looking at the same spot for too long. @merkys , could you pls clarify the alternative implementation?

- The first (@rartino 's) text mentions "database-specific prefix" in both places, which is formally correct; in the reformulation, the prefix is implied. I would say that in the specs we better stick to more formal expression, even if it is somewhat repetitive;

- Andriu's version puts the description of the prefixed version in front; but the unprefixed version will probably be encountered more often. It is also no explicitly clear to me how implementation is supposed to recognise that "the property is ... belonging to the implementation itself", other than checking for a prefix; but in this case checking prefix is my guess. We should probably say it explicitly, shouldn't we?

- The new text essentially swaps the two cases, which I find equally easy (or difficult) to read.

With all these thoughts I would leave the current version as it is, unless there is indeed an alternative incorrect interpretation that I have overlooked.

merkys: The sentence

> If the property name has no database-specific prefix, or the database-specific prefix that belongs to the implementation itself, ...

could be understood as

1. If (the property name has no database-specific prefix) OR (the property name has **the** database-specific prefix that belongs to the implementation itself)

or

2. If (the property name has no database-specific prefix) OR (the property name has **no** database-specific prefix that belongs to the implementation itself)

I understand that 1. is supposed, but I need to re-read sentence a couple of times to get it right.

rartino: So the main fix in the reformulation of @merkys is the extra "it has". Given @sauliusg points, the same clarification can be done with a minimal change to the original sentence:

-  If the property name has no database-specific prefix, or if it has the database-specific prefix that belongs to the implementation itself, ...

merkys: I agree with @rartino's suggestion.

sauliusg: I apparently spoiled by C that implements "'not' binds tighter than 'or'" :).
Sure, we can add "it has" to dispel all ambiguity. Will push right now.

optimade.rst
```diff
@@ -560,6 +560,25 @@ In the default response format, relationships are encoded as `JSON API Relations
...
+
+  * Furthermore, if the implementation does not recognize the prefix at all, it SHOULD return a warning that indicates that the property has been handled as unknown.
+
+  * On the other hand, if the prefix is recognized, i.e., as belonging to a known database provider, the implementation SHOULD NOT issue a warning but MAY issue diagnostic output with a note explaining how the request was handled.
+
+The rationale for treating properties from other databases as unknown rather than triggering an error is for OPTiMaDe to support queries using database-specific properties that can be sent to multiple databases.
+
+For example, the following query can be sent to API implementations `exmpl1` and `exmpl2` without generating any errors:
+
+   :filter:`filter=_exmpl1_bandgap<2.0 OR _exmpl2_bandgap<2.5`
```
rartino: Either dedent this line, or indent the line before it so it renders correctly.

sauliusg: Removing the indentation. Hopefully it is OK now (at least it renders like some other examples). PlS have a look.

Actually, there are several more cases of strange rendering in the `optimade.rst` (e.g. at  'Note: the specification presently makes no attempt to standardize access control mechanisms.'). I'm not so familiar with the RST to say in the spot what is going on there. We should probably go through the text and make a separate issue/PR on these formatting peculiarities.

rartino: Let's just handle them in #189. If you don't know how to fix them, just do a review, add comments to the relevant places in the text, and pick the option "comment" when submitting the final review.

The note: lines look ok to me. The formatting is meant to indicate that these statements are clarifications, implementors notes, etc., which just clarifies the standard without being meant as having a formal bearing on it. 

The data: lines is a formatting bug. The point has been indented one step too far, and thus renders as a point in a note instead of just a point.

PR #164, opened 2019-07-16
Dropping restrictions on the IDs

merkys:
This PR is a follow-up on #158. Changes done (from TODO list by @rartino):

* Removed the requirement on ID to be URL-safe so that any character string is a valid ID.
* Pointed out that in URLs of the kind, e.g., `<entry type>/<ID>` the `<ID>` should be URL-encoded.
* Remove the `local_id` property (the main ID now can be the local ID of your database).


merkys:
Oh dear, I must have branched off the `master` instead of `develop`... Closing for now.

PR #128, opened 2019-06-15
structure -> structures, calculation -> calculations

fawzi:
Fixes #119 type and endpoint should be equal.
I converted most uses, there are some places where I left the singular
structure, a more aggressive renameing could replace them with
"structures object"

Please follow [the guidelines here](https://github.com/Materials-Consortia/OPTiMaDe/wiki/How-to-contribute#submit-proposed-changes) to create a new pull request.
If your pull request requires discussion, make sure the OPTiMaDe developer list is alerted (for more info see the link above.)

PR #168, opened 2019-07-17
Fixing description that looks like a grammar (or logic) mistake.

merkys:
Stumbled upon this while reading the specification. This must be a leftover of singular -> plural transition (`structure` -> `structures`), and I argue that this `a structures` in not appropriate here.

### DIFF
--- a/optimade.md
+++ b/optimade.md
@@ -921,7 +921,7 @@ Example:
 ```jsonc
 {
   "data": {
-    "description": "a structure",
+    "description": "a structures entry",
     "properties": {
       "nelements": {
         "description": "Number of elements",
--- a/optimade.md
+++ b/optimade.md
@@ -921,7 +921,7 @@ Example:
 ```jsonc
 {
   "data": {
-    "description": "a structures",
+    "description": "a structure",
     "properties": {
       "nelements": {
         "description": "Number of elements",


CasperWA:
This looks fine.
Going through the spec I also found that the only `type` property example is `"structures"` (see section 6.1.2), is this then incorrect or was this what we decided on?

merkys:
> Going through the spec I also found that the only `type` property example is `"structures"` (see section 6.1.2), is this then incorrect or was this what we decided on?

Examples do not have to show all the possible variants anyway, so I guess this should not be a problem.

rartino:
Ah, this odd grammar was actually discussed in #128 and I believe we settled on that the text would be "a structures entry". But then we had some major issues with a rebase... and I think this was lost there.

merkys:
> Ah, this odd grammar was actually discussed in #128 and I believe we settled on that the text would be "a structures entry". But then we had some major issues with a rebase... and I think this was lost there.

So it is intended to be "a structures entry", right? Or maybe we could provide something more descriptive, i.e. "data that pertain to a single structure" (taken from Term Definitions)?

rartino:
If you look at the discussion in #128, I believe we agreed on "a structures entry" (as in, it was suggested and no one complained). 

Going forward towards v1.0, I think it would be great if we simply specify exactly what these strings are supposed to be for the various endpoints. And, preferably, they should be a couple of sentences explaining what is actually being represented.

merkys:
> Going forward towards v1.0, I think it would be great if we simply specify exactly what these strings are supposed to be for the various endpoints. And, preferably, they should be a couple of sentences explaining what is actually being represented.

+1

I have changed the description to "a structures entry" for now.

PR #171, opened 2019-07-19
Add link to raw develop-branch providers.json

dwinston:
Note: this link should refer to the `v0.10.0` branch raw version rather than the `develop` version of the package.json file for that release, the `master` version in the `master` branch, etc.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2067,7 +2067,7 @@ Appendices
 Database-Provider-Specific Namespace Prefixes
 ---------------------------------------------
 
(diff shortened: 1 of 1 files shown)


CasperWA:
> This also changes the original desired behavior where the link would always be to the current version being read, rather than the latest development version

Yeah, I guess the idea was to change this once v0.10.0 came out, which I guess it has now. So we should change it to be version static, right? Or am I misinterpreting?

rartino:
Rather then this, I'm getting more and more strongly opinionated that we should implement the fix suggested at the end of the now closed #214. I'll start to put things in motion for this.

rartino:
I now suggest closing this and merging #222 instead.

CasperWA:
> I now suggest closing this and merging #222 instead.

I agree with this sentiment.

Superseded by #222.

PR #182, opened 2019-08-27
Conversion of specification to restructuredText / RST

rartino:
Pros:

* Automates, and thus avoids error-prone manual work with:
  - Table of contents.
  - Section numbers.
  - Cross-references (but note that renaming section titles requires a search-replace of all cross-references.)
* Avoids accidental line breaks due to white-space at the end of lines.
* Fully supported by github's diffs.
* Markup that is more end-format-agnostic (i.e., much less need for inline html).
* Text roles for marking up the context of literals, for easier formatting of output (e.g., by css), and makes spelling and syntax checks possible.
* A syntax that really is not that different from markdown.

Cons:
 
* It isn't markdown.

A few notes:
* In the github-rendered output, cross-references will now only contain the section name and the link, not the section number. To render cross-references in html with section numbers will require a post-processing script to put them in.
* I have tried to let this PR stick to the conversion without other editing. However, I have sometimes made deliberate changes that relate to the formatting, e.g., removed "half-paragraphs" (new line without new paragraph).
* I have made text changes in relation to cross-references to make these grammatically clear now that they will display in github without section numbers.

This closes: #103


CasperWA:
This is great work @rartino! I will go through it when I have some time. It's very nice with the top comment already in the spec, providing some info on the new format.
I will be under the assumption that the document is otherwise 1-to-1 with the source markdown (as you also write). I don't feel like going through the new document, having the original side by side :sweat_smile: (i.e., now is the time for you to unnoticed slip in all those little personal changes in the spec :wink:)

rartino:
Ok, tests are fixed now. 

I note that since rst requires code blocks to be indented, extracting data has become a tad more involved if one wants to avoid the indent. 

If this is something that concerns people, we can consider replacing these segments with, e.g.:
```
.. literalinclude:: identifiers.pcre
   :language: pcre
```
which would include the segment from a separate file `identifiers.pcre` instead. The drawback is that github does not render such rst includes. Most normal rst converters (e.g., to html, pdf) does.

Personally, I don't think an extra `sed 's/^ *//' ` is anything to worry about, so I prefer to keep the data inline.

rartino:
In case the approval of this is held back by concerns that the PR could contain "hidden" changes in the text, below is an attached "diff" produced the following way:
1. The rst file in this branch was rendered with rst2html. The html file was then converted to text with lynx.
2. The md file in the OPTiMaDe develop branch was rendered to html with pandoc. That html file was then convered to text with lynx.
3. A diff between the two text files was created using wdiff. 

Here is the result: [diff.pdf](https://github.com/Materials-Consortia/OPTiMaDe/files/3683207/diff.pdf)

PR #187, opened 2019-10-03
Clarify page number for pagination

### LINKED ISSUE #186: Conflicting query parameters 'page_page' and 'page_number'
4.1.1. "URL Query Parameters" mentions two query parameters, `page_page` and `page_number`, which seemingly are synonymous. I assume this to be a bug, and suggest using only `page_number` as it sounds more naturally than `page_page`.

merkys:
The changes:

1. Using `page_number` throughout the document (was `page_page` in one occurrence);
2. Making `page_number` explicitly 1-based.

Fixes #186

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -615,7 +615,7 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
 
   - *offset-based pagination*: using :field:`page_offset` and :field:`page_limit` is RECOMMENDED.
   - *cursor-based pagination*: using :field:`page_cursor` and :field:`page_limit` is RECOMMENDED.
-  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is the maximum number of entries per page). It is RECOMMENDED that the first page has number 1, i.e., that :field:`page_number` is 1-based.
+  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is the number of entries per page). It is RECOMMENDED that the first page has number 1, i.e., that :field:`page_number` is 1-based.
   - *value-based pagination*: using :field:`page_above`/:field:`page_below` and :field:`page_limit` is RECOMMENDED.
 
   Examples (all OPTIONAL behavior a server MAY implement):
--- a/optimade.rst
+++ b/optimade.rst
@@ -615,7 +615,7 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
 
   - *offset-based pagination*: using :field:`page_offset` and :field:`page_limit` is RECOMMENDED.
   - *cursor-based pagination*: using :field:`page_cursor` and :field:`page_limit` is RECOMMENDED.
-  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is equivalent to page "size"). It is RECOMMENDED that the first page has number 1, i.e., that :field:`page_number` is 1-based.
+  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is the maximum number of entries per page). It is RECOMMENDED that the first page has number 1, i.e., that :field:`page_number` is 1-based.
   - *value-based pagination*: using :field:`page_above`/:field:`page_below` and :field:`page_limit` is RECOMMENDED.
 
   Examples (all OPTIONAL behavior a server MAY implement):
--- a/optimade.rst
+++ b/optimade.rst
@@ -615,7 +615,7 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
 
   - *offset-based pagination*: using :field:`page_offset` and :field:`page_limit` is RECOMMENDED.
   - *cursor-based pagination*: using :field:`page_cursor` and :field:`page_limit` is RECOMMENDED.
-  - *page-based pagination*: using :field:`page_number` (1-based) and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is equivalent to page "size").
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -608,14 +608,14 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
-- **page\_{offset, page, cursor, above, below}**: A server MUST implement pagination in the case of no user-specified :query-param:`sort` parameter (via the :field:`links` response field, see section `JSON Response Schema: Common Fields`_).
+- **page\_{offset, number, cursor, above, below}**: A server MUST implement pagination in the case of no user-specified :query-param:`sort` parameter (via the :field:`links` response field, see section `JSON Response Schema: Common Fields`_).
   A server MAY implement pagination in concert with :query-param:`sort`.
   The following parameters, all prefixed by "page\_", are RECOMMENDED for use with pagination.
   If an implementation chooses
 
   - *offset-based pagination*: using :field:`page_offset` and :field:`page_limit` is RECOMMENDED.
   - *cursor-based pagination*: using :field:`page_cursor` and :field:`page_limit` is RECOMMENDED.
-  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is equivalent to page "size").
+  - *page-based pagination*: using :field:`page_number` (1-based) and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is equivalent to page "size").
```
CasperWA: Is the existing parenthesis needed here?

rartino: I think it would help readability to be a bit more clear, both what 1-based refers to, and the conformity level of page_number being 1-based.
```suggestion
  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is equivalent to page "size"). It is RECOMMENDED that the first page has number 1, i.e., that :field:`page_number` is 1-based.
```

merkys: I have accepted @rartino suggestion (bc13f93) removing the parenthesis.

CasperWA: Sure - my question still stands though: Is it necessary to let the reader know that `page_limit` is equivalent to page "size"? And why have we chosen to write it here and not previously, e.g., for offset-based pagination?
I don't know if you know the answers, it just strikes me as odd - unless page "size" is a well-defined concept for page-based pagination.

merkys:
> Only a single comment, since I am unsure of all the different pagination setups. I am guessing that for the "page-based" pagination, page "size" is a specific thing?

Here "size" stands for the number of items on the page, fixed that in 837e83e.

> Is it possible to add a reference to pagination definitions somewhere on www? Or is it considered bad practice?

I am not aware of any RFC or like document describing the pagination strategies. These seem straightforward (IMO), but I fully understand the need for the reference here.

CasperWA:
> Here "size" stands for the number of items on the page, fixed that in [837e83e](https://github.com/Materials-Consortia/OPTiMaDe/commit/837e83ec74e21cb66128b5d2380f79d7bb1b48cc).

Sure - but doesn't it always? I.e., shouldn't this explaining parenthesis be placed earlier?

> > Is it possible to add a reference to pagination definitions somewhere on www? Or is it considered bad practice?
> 
> I am not aware of any RFC or like document describing the pagination strategies. These seem straightforward (IMO), but I fully understand the need for the reference here.

Fair enough - it was also only if we knew of something, since it seemed to me the different kinds of pagination here were listed as if obvious and well-defined at another source.

optimade.rst
merkys: > Sure - my question still stands though: Is it necessary to let the reader know that `page_limit` is equivalent to page "size"? And why have we chosen to write it here and not previously, e.g., for offset-based pagination?

The surplus explanation of `page_limit` sounds odd to me. Is there really a reason to retain it? Can we just remove it?

merkys:
> Sure - but doesn't it always? I.e., shouldn't this explaining parenthesis be placed earlier?

It should, and it seems that it is explained earlier. I suppose we can remove it. It seems @dwinston has touched this text recently. @dwinston maybe you remember why we have this additional clarification of `page_limit`?

> Fair enough - it was also only if we knew of something, since it seemed to me the different kinds of pagination here were listed as if obvious and well-defined at another source.

Again, maybe @dwinston knows a suitable reference?

optimade.rst
CasperWA: My point (and wish) exactly :smile:

optimade.rst
```diff
@@ -608,14 +608,14 @@ Standard OPTIONAL URL query parameters standardized by the JSON API specificatio
...
+- **page\_{offset, number, cursor, above, below}**: A server MUST implement pagination in the case of no user-specified :query-param:`sort` parameter (via the :field:`links` response field, see section `JSON Response Schema: Common Fields`_).
   A server MAY implement pagination in concert with :query-param:`sort`.
   The following parameters, all prefixed by "page\_", are RECOMMENDED for use with pagination.
   If an implementation chooses
 
   - *offset-based pagination*: using :field:`page_offset` and :field:`page_limit` is RECOMMENDED.
   - *cursor-based pagination*: using :field:`page_cursor` and :field:`page_limit` is RECOMMENDED.
-  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is equivalent to page "size").
+  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED (:field:`page_limit` is the number of entries per page). It is RECOMMENDED that the first page has number 1, i.e., that :field:`page_number` is 1-based.
```
```suggestion
  - *page-based pagination*: using :field:`page_number` and :field:`page_limit` is RECOMMENDED. It is RECOMMENDED that the first page has number 1, i.e., that :field:`page_number` is 1-based.
```

merkys:
Thanks for the reviews!

PR #190, opened 2019-11-01
Remove uniqueness constraint for species_at_sites

### LINKED ISSUE #188: Species and species_in_sites
Going back over the issues and PRs I have found that the `species` attribute for structure objects have been updated: #127.
This is fine, the sentiment of not using relevant data as dictionary keys makes complete sense.

However, as @fawzi also pointed out [in the same PR](https://github.com/Materials-Consortia/OPTiMaDe/pull/127#pullrequestreview-250247721), this will result in rather long `species` lists compared to previous versions.

In order to accommodate this, while also not having dictionary keys be data, I have a suggested change to `species_at_sites`, which should also reduce `species`:

Instead of having each data/species in `species_at_sites` be unique, let's not :)
This may be implemented in two ways, depending on the _need_ to have these pieces of data be unique:

1. **(No need for uniqueness)**: Simply have `species_at_sites` be a list of equal length to the first dimension of `cartesian_site_positions`, i.e., be equal to `nsites`. Make it a requirement that each string/piece of data in `species_of_sites` must match up with its corresponding site in `cartesian_site_positions`. Then lose the uniqueness of each string/piece of data in `species_of_sites`, but keep the requirement that for each unique data in `species_of_sites`, there must be a dictionary in the `species` list with its `name` value equal to it.

Example:
```json
{
  "nsites": 9,
  "species_at_sites": ["T1", "T1", "T2", "O1", "O1", "O1", "O1", "O2", "O2"],
  "species": [
    {"name": "T1", "chemical_symbols": ["Si"], "concentration": [1.0]},
    {"name": "T2", "chemical_symbols": ["Si", "Al"], "concentration": [0.8, 0.2]},
    {"name": "O1", "chemical_symbols": ["O"], "concentration": [1.0]},
    {"name": "O2", "chemical_symbols": ["O", "vacancy"], "concentration": [0.9, 0.1]}
  ]
}
```

2. **(Need for uniqueness)**: Do the same as for non-uniqueness, but let each piece of data in `species_at_sites` be a list of exactly length 2, where the first entry is the unique `species_at_sites` name, and the second is a unique `species` `name` value.

Example:
```json
{
  "nsites": 9,
  "species_at_sites": [
    ["T1", "T1"],
    ["T2", "T1"],
    ["T3", "T2"],
    ["O1", "O1"],
    ["O2", "O1"],
    ["O3", "O1"],
    ["O4", "O1"],
    ["O5", "O2"],
    ["O6", "O2"]
  ],
  "species": [
    {"name": "T1", "chemical_symbols": ["Si"], "concentration": [1.0]},
    {"name": "T2", "chemical_symbols": ["Si", "Al"], "concentration": [0.8, 0.2]},
    {"name": "O1", "chemical_symbols": ["O"], "concentration": [1.0]},
    {"name": "O2", "chemical_symbols": ["O", "vacancy"], "concentration": [0.9, 0.1]}
  ]
}
```

In this way we do not end up with `species` being a list of length `nsites`, but of the same length it "used to be", while at the same time, there are no dictionary keys containing data.

CasperWA:
Fixes #188 

Choosing solution number 1 from issue #188.

By removing the requirement on `species_at_sites` that:

- Each species MUST have a unique name.

The bloating of `species` can be stopped.
I've also added an extra example to `species_at_sites`.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1715,7 +1715,6 @@ species\_at\_sites
   - **Response**: REQUIRED in the response unless explicitly excluded.
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
   - MUST have length equal to the number of sites in the structure (first dimension of the list property `cartesian_site_positions`_).
-  - Each species MUST have a unique name.
   - Each species name mentioned in the :property:`species_at_sites` list MUST be described in the list property `species`_ (i.e. for each value in the :property:`species_at_sites` list there MUST exist exactly one dictionary in the :property:`species` list with the :property:`name` attribute equal to the corresponding :property:`species_at_sites` value).
   - Each site MUST be associated only to a single species.
     **Note**: However, species can represent mixtures of atoms, and multiple species MAY be defined for the same chemical element.
@@ -1724,6 +1723,7 @@ species\_at\_sites
 - **Examples**:
   
   - :val:`["Ti","O2"]` indicates that the first site is hosting a species labeled :val:`"Ti"` and the second a species labeled :val:`"O2"`.
+  - :val:`["Ac", "Ac", "Ag", "Ir"]` indicating the first two sites contains the :val:`"Ac"` species, while the third and fourth sites contain the :val:`"Ag"` and :val:`"Ir"` species, respectively.
 
 species
 ~~~~~~~


ml-evs:
I think this change reflects the original intent of the spec (or at least my understanding of it). As we've already discussed offline it I'll leave it for others to review.

CasperWA:
> I think this change reflects the original intent of the spec (or at least my understanding of it). As we've already discussed offline it I'll leave it for others to review.

Since it takes 2 approving reviews, yours can easily be one of them :smile:

PR #195, opened 2019-11-07
Makefile reorganization, validation in docker

### LINKED ISSUE #180: Make sure AUTHORS is alphabetically sorted
When doing the v0.10.0 release, we noted that *some* people (*cough cough*) are not in their correct alphabetical spot in AUTHORS. That should be fixed. (Perhaps also checking with a tool that every author is in the right order).

### LINKED ISSUE #192: Running the test environment in Docker
The `Dockerfile` below can be used to run the test on macOS or other operating systems. This could be useful for Linux users as well because trusty is not the latest distribution...

`cat Dockerfile`
```
FROM ubuntu:trusty

RUN apt-get update \
 && apt-get install -y -q --no-install-recommends \
        locales \
        ant \
        unzip \
        wget \
        make \
        default-jdk \
        subversion \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/* \
 && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8

ENV LANG en_US.utf8

WORKDIR /root/optimade
COPY . .

CMD ["/bin/bash", "-c", "make clean; make tools; make grammars; make tests; make listdiff"]
```

Usage:
- Building the image locally:
  `docker build -t optimade .`
- Running the tests:
  `docker run -it --rm optimade:latest`
- Running the test by using the local filesystem:
  `docker run -it --rm -v $PWD/:/root/optimade optimade:latest`

Note: It is important that the repository should be "clean" because `make clean` keeps some of the files of the tools in place. (a dirty way: `rm -rf tests; git restore tests`)

rartino:
Fixes #192 
~~Overlaps with #194, and thus fixes #180 and closes #194 (because the AUTHORS file is now validated, it needs to be correct in this PR for the travis check to pass.)~~ (Edit: no longer relevant as #194 is now merged.)

I started off wanting to add makefile targets for running the grammar validation inside docker using @fekad's Dockerfile, and also add some other useful targets; e.g., sorting the AUTHORS file, spellcheck, and for extracting the grammar segment from the rst file.

However, the old GNUmakefile was very targeted specifically at running the grammar test suite to the extent that adding other types of targets was difficult. Among other things, regardless of target, it included a large set of other makefiles that produced unrelated console output.

Hence, this PR also reorganizes the build system. The grammar test GNUmakefile has been moved down into test/GNUmakefile. Various build targets are now present in the "topmost" GNUmakefile. Various types of validations can be run with `make validate_<something>`. And, running `make validate` runs all validations inside a docker container. `make validate_without_docker` runs all validations without docker. All places that validate the specification in other containers, e.g., our travis setup, should now do by `make validate_without_docker` rather than the previous set of make calls (`make tools; make grammars; make tests; make difflists`).

For security reasons, I am proposing that docker should be the default as long as we need to fetch and run outside code as part of the validation process.

This PR is not super-important for v1.0, but I would prefer it to be in as it simplifies future contributions to the specification by new people.


rartino:
> Instead of having separate make targets for Docker and Docker-less execution, I'd suggest having unified target plus a variable DOCKER to switch between them. [...] This way we could have includable configuration file for site-specific setup

What use-case is this trying to address? Very generally, I'd like to avoid communication in side-channels (e.g., environment variables) unless there are very good reasons.

What I could see an argument for is to instead have the targets:
- `validate_docker`
- `validate_without_docker`
and then have `make validate` simply echo that one has to choose one of those.

merkys:
> What use-case is this trying to address? Very generally, I'd like to avoid communication in side-channels (e.g., environment variables) unless there are very good reasons.

I prefer direct execution to Docker, thus I suggest supporting site-specific configurations for the validation. Control flag `DOCKER` is not intended to be an environment variable, but Makefile variable, used either on the command line
```
make validate DOCKER=0
```
or specified in an included configuration file in the main Makefile:
```
DOCKER = 1

MAKECONF_FILES = ${filter-out %~, ${wildcard Makeconf*}}

ifneq ("${MAKECONF_FILES}","")
include ${MAKECONF_FILES}
endif
```

sauliusg:
On 2019-11-11 08:52, Andrius Merkys wrote:
> I prefer direct execution to Docker, thus I suggest supporting
> site-specific configurations for the validation. Control flag |DOCKER|
> is not intended to be an environment variable, but Makefile variable,
> used either on the command line
> 
> |make validate DOCKER=0 |
> 
> or specified in an included configuration file in the main Makefile:
> 
> |DOCKER = 1 MAKECONF_FILES = ${filter-out %~, ${wildcard Makeconf*}} |

Sorry for a short notice, but very quickly:

Andrius is right, there should be a Make config variable to invoke docker.

We should use our proven system of setting up a Makeconfig file, with
Makeconfig.example committed to the repository and a rule to make/update
the Makeconfig from the Makeconfig.example. Makeconfig.example would
contain default values that are to be discussed.

I still wonder why Docker is necessary at all? I see a use case of
docker for eventual deployment, but this should be a down-the-stream
(packaging) concern, shouldn't it? And for the documentation, the main
outcome is the documentation itself; the test system is for developing,
so will not be included into the final distribution either (if such
distribution would be necessary, apart from the Web page announcement
and a stable download link).

Do I miss something? (I admit I still have to go through the whole
Docker discussion; as is fashionable to boast nowadays, TL;DR :/).

Regards,
Saulius

-- 
Dr. Saulius Gražulis
Vilnius University Institute of Biotechnology, Saulėtekio al. 7
LT-10257 Vilnius, Lietuva (Lithuania)
fax: (+370-5)-2234367 / phone (office): (+370-5)-2234353
mobile: (+370-684)-49802, (+370-614)-36366

rartino:
>I admit I still have to go through the whole Docker discussion; as is fashionable to boast nowadays, TL;DR :/

It is not a long discussion, just the topmost post in the PR and #192.

> Andrius is right, there should be a Make config variable to invoke docker.
>
> We should use our proven system of setting up a Makeconfig file, with
> Makeconfig.example committed to the repository and a rule to make/update
> the Makeconfig from the Makeconfig.example. Makeconfig.example would
> contain default values that are to be discussed.

But, what is the use case for this? Who needs a single common target with site-configurable behavior instead of just two separate targets?

> I still wonder why Docker is necessary at all?
> [...] the test system is for developing, so will not be included into the final distribution

In my opinion, the "distribution" from this repository is just the standards document. The makefile is there to help developers work with the standard, e.g., for validating the grammar and other parts, run spellcheck, sort AUTHORS etc.

Originally @fekad developed the docker setup to be able to validate grammars on mac osx. I imagine it can be greatly helpful for Linux users as well who don't need to figure out how to install the requirements / what requirements to install, it "just works" if your system supports docker.

My additional point is that the present grammar validation downloads external code from your SVN repository and runs it directly on my machine, and this is a security issue (and also unexpected). This should not be the default behavior for developers who clone the git repository and runs 'make'. The docker setup at least alleviates this security issue to some degree.

sauliusg:
> Hence, this PR also reorganizes the build system. The grammar test GNUmakefile has been moved ...

I will separate the docker issue into a different comment. On the Make system, my comments are below.

Let me explain the rationale of our Makefile system so that we do not drag the code into different directions...

The idea behind using GNU Make is to achieve the state where one can check our the repo (in our case OPTiMaDe spec), change to that directory and type `make` to build everything what is needed by default. Such setup would arguably be convenient for new people and old developers alike. This is currently achieved on Linux systems Ubuntu-16.04 and the latest LTS, Ubuntu-18.04 with java 1.8 (checked right now). There is one caveat: one needs a tool-chain which is not available on default installations; to simplify this we use dependencies/<your-os-name>/install.sh files, that only call standard system package manager under sudo (and your password). This is safe since most systems (at least in Linux world) have robust package managers that do not break systems (an diff they do the system is unusable anyway). Nothing else should be run under sudo by default, including docker. With these set up, all the rest should be automatic. 

In this context, your suggestion to abolish default the make target 'all' and introduce 'validate_whatever' targets is very much against the fur in our world. I would stick to well-known targets ('all', 'clean', 'distclean', 'test/check') and make them work as expected. There can be other targets; I personally use 'make run' when working on grammars, so removing it is not welcome. You can introduce other targets if needed.

To make the things automatic, the code which is not in the standard system repositories needs to be fetched and built. This is why the checkouts from externals repos are done. If you do not want it, you can do that yourself but this is clearly extra work. Thus a balance needs to be found between convenience and what we are used to see in "pure code" repositories. In these repos, developers are expected to fight the "dependency hell" themselves, so no extra code is usually fetched.

As for security, the current setup is no less secure that checking out a foreign repo and building it. If you trust anybody's code at all, then you have pretty much the same security concerns with the current setup as with any repo. Now, I agree that svn: protocol which I used for simplicity, and which was considered secure enough for the last decade, does allow some extravagant attack vectors; if this is a concern I'm happy to put my code under https: on my serve or on GitLab or on your personal machine, as you prefer. Docker, IMHO, does not make it much more secure since you can do the same mischief under docker as you can do under a usual account (I assume you are not running the tests as `root` :).

Make config files are very handy when different people want to keep different environments on different machines and still use the same repo without changes. So if you prefer docker, you would be able to set up your local config file (which is not reset by updates) and 'make', 'make all' or 'make tests' will run docker for you. I see docker as a nuisance in this case and prefer to run tests directly.

To wrap up, my preferred Make testing system would consist of the top-level `GNUmakefile` (pretty much like your changes suggest), and have `make all` to do all the necessary jobs. If we unify our requirements, I guess we arrive at the Makefile which is pretty close to the one that existed. It them has little sense to have yet another Makefile in the tests/ directory, does it?

sauliusg:
> What use-case is this trying to address? Very generally, I'd like to avoid communication in side-channels (e.g., environment variables) unless there are very good reasons.

To be precise: Make variables are not exactly environment variables (you can pass them as command line parameters) and definitely not a side-channel. It is a standard way to communicating parameters to Make.

> What I could see an argument for is to instead have the targets:
> 
>     * `validate_docker`
> 
>     * `validate_without_docker`
> 
> 
> and then have `make validate` simply echo that one has to choose one of those.

Such behaviour is IMHO a nuisance and not needed. I tried to stay away from this pattern as far as possible when constructing the Makefiles for testing.

Extra targets are good when you want to do something very specific, e.g. run a pre-defined test to see how your recent grammar changes behave; `make run` was meant for such use.

sauliusg:
> However, the old GNUmakefile was very targeted specifically at running the grammar test suite to the extent that adding other types of targets was difficult. Among other things, regardless of target, it included a large set of other makefiles that produced unrelated console output.

Well, the "old" GNUMakefile had test targets since we were using them, and adding any new target is straightforward.

I would just advice against modifying ad-hoc the main Makefile and adding extra targets there; this is indeed cumbersome and we are trying back here at home to move away from such habit. That is why we use a modular Make system with Makelocal-* files that are included into the main Makefile. Each new task (with its new targets) is added as a new Makelocal file, quote independent from all others. In his way new functions (targets) can be easily added or removed as necessary, and Makefiles can be reused. We use this system very successfully back at home for quite a while.

sauliusg:
> For security reasons, I am proposing that docker should be the default as long as we need to fetch and run outside code as part of the validation process.

I personally would see Docker more as nuisance in this case, since one can run the tests just on the native machine. For Mac's, the same software should be available as for Linux, just I do not have a spare Mac sitting around to test... Docker could be a solution, but I would see it optional and configurable. As for security, as soon as the docker container has access to the net, it is not so much secure as the main machine, IMHO.

sauliusg:
> All places that validate the specification in other containers, e.g., our travis setup, should now do by `make validate_without_docker` rather than the previous set of make calls (`make tools; make grammars; make tests; make difflists`).

I would say that Travis setup is not what a casual user will do every day, so having several commands not an issue. Moreover, having them explicit is actually showing what happens when Travis is run, unlike `make validate_without_docker` which is more obscure and more work to change.

One more thing that Travis needed is a status code that uniquely indicates the status of the tests. This is not the same as the status of `make tests`! The status of `make tests` indicates if all tests have been launched and run; some of them may have failed but this is still a success of the `make test` system (that is what the system has been set for!). For the Travis, OTOH, success code should indicate that all tests have passed, not just that all tests have been run. That's why an extra analysis was necessary.

rartino:
Thanks @sauliusg for the detailed reply and explanations. I've read it all :-)

> my preferred Make testing system would consist of the top-level GNUmakefile (pretty much like your changes suggest), and have make all to do all the necessary jobs. If we unify our requirements, I guess we arrive at the Makefile which is pretty close to the one that existed. It then has little sense to have yet another Makefile in the tests/ directory, does it?

I did not set out to change the top-level makefile! I just added my targets to the old top one, expecting it to work. However, `make <my_target>` then gave a mass of output about missing files, executions of awk, mkdepend, etc. I did not want to start majorly re-working "your" makefiles, so I kept your build system intact, but diverted it into a separate makefile invoked only for the grammar validation target. Someone who wants to interact directly with it, e.g., to run `make run`, can still do so simply by first `cd tests`.

That said, I'm OK with trying to rework the PR so the old GNUMakefile is at top. But, can you please explain how I can add my targets so they can be invoked completely independently of the grammar validation code? I don't want anything to be run except for what is in my target.

I get that you are trying to explain this in a reply further down, saying that I should add my targets as "Makelocal-* files". But can you please spell out very clearly exactly what I need to do for `make spellcheck` to only invoke precisely the command: `aspell -x -p ./.words.lst -l en_US -c optimade.rst`? (If you explain this one example, I can handle all the other targets myself.)

> One more thing that Travis needed is a status code that uniquely indicates the status of the tests. 

I thought this PR handles this by makiing the validate taget end with:
```
(exit $$($(MAKE) -s -f tests/GNUmakefile listdiff | head -n 1 | wc -l))
```
Meaning the top-level `make validate` is meant to return an error status code if the validation fails. However, I see your point about wanting separate `run_tests` and `validate` targets, where the former does not return an error code even if the tests fails, whereas the latter does. I'll add that when I update the PR.

*To keep things organized, I'll address the "docker or no docker" and "docker by default" in a separate comment.*

rartino:
Now, about docker;

>  [...] if you want the make system to fire docker for you, this can be achieved using your local config file. Or even a 'make docker' rule, but as a special functionality, not a default.

My desire to make the default validation run inside a container is solely based on my concerns for security with our present way of doing grammar validation. If you convince me there are no relevant security concerns with the present setup to be worried about, I'd fully agree with you that docker should not be the default, but an option. 

But, I really, *really*, find it questionable to go public with a OPTiMaDe v1.0 distributed alongside a makefile where `make` downloads external code from a private SVN repository and executes it as the current user on the local machine. Some of the people we hope to use OPTiMaDe work in places where security is of a very high concern and there is a constant threat of attackers searching for attack vectors, e.g., national laboratories in the US.

> As for security, as soon as the docker container has access to the net, it is not so much secure as the main machine, IMHO.

Are you arguing that running untrusted code under your own user is as safe/unsafe as running it in a container with access to the network? I *fundamentally* disagree. 

Just to give a few examples of things that could be done to me if harmful code somehow makes it into your SVN and I run `make`:
* *Anything* with all my files (leaked, deleted, encrypted for a ransomware attack).
* My address book(s) can be grabbed and used for socially engineering my friends. 
* My running programs can be controlled, e.g., emails sent as me from my MUA using its configured mail accounts, my browser snooped on to see where I browse and grab, e.g., passwords. 
* All my network traffic can be man-in-the-middle--attacked. 
* A keylogger can be installed in my init files.
* The equivalent of `alias sudo="sudo hack_this_machine_wrapper"` can be inserted in my init files so that when I use `sudo` my computer is owned, and, e.g., `alias ssh="hack_remote_machine_ssh_wrapper"` to compromise any other machine I log into. (And such config changes can be set to remove themselves after the deed is done...)
* etc, etc...

Whereas, what can really be done by untrusted code in a container with network access (unless that container can be broken out of)? Sure, it can try to attack my internal machines, and set up a proxy for e.g., spam or attacks on others. But unless I also have other security issues in my environment, there is a limit on the permanent damage left after the container stops.

Hence, I try to avoid running untrusted code as my own user, and so should anyone else.

This isn't so much a trust issue with the code you intend to put in your SVN. But how can I completely trust that machine? How secure is that SVN server from attack? Are you putting the same effort in securing it as Debian, Ubuntu, and RedHat are in securing their download servers?

sauliusg:
I'm glad that you have spelled it our explicitly. I fully agree with you that we need to be vigilant, aware of security issues and make all reasonable efforts to secure our systems. What worries me a lot is your apparent (and, IMHO, misplaced) trust in the "big companies" servers, apparent trust in your personal computer, and total mistrust in my SVN server. Personally, this is understandable. Computationally, ungrounded. Let me explain.

> This isn't so much a trust issue with the code you intend to put in your SVN. But how can I completely trust that machine?

You can't. And you (we) should be deeply worried if you "completely trust" *any* computer or system, for that matter.

"Absolute security is a myth" (https://itsfoss.com/linux-mint-hacked/).

> How secure is that SVN server from attack? Are you putting the same effort in securing it as Debian, Ubuntu, and RedHat are in securing their download servers?

I hope this was not a rhetoric question. In any case, I'll answer it :). I hope it will be some interesting reading. For me, it was :).

How secure are the providers' servers from attack? The breach of security in software distribution I do remember and can give you instantly was the MintLinux server compromise:

https://itsfoss.com/linux-mint-hacked/ : "Hackers managed to hack the website and replace the download links of some Linux Mint ISOs to their own, modified ISOs with a backdoor in it"

For the others, I had to DuckDuck. This is what comes out in 5 mins:

Debian: "On July 12th (2006) the host gluck.debian.org has been compromised using a local root vulnerability in the Linux kernel" (https://www.linux.com/news/debian-server-restored-after-compromise/, https://linux.slashdot.org/story/03/11/21/1314238/debian-project-servers-compromised)

Ubuntu: "A security breach at Ubuntu Forums exposed the information of as many as two million users." (https://www.tripwire.com/state-of-security/latest-security-news/ubuntu-forums-hack-potentially-compromised-2m-users-information/)

RedHat: "(August 22, 2008) Last week one or more of Red Hat’s servers got cracked." (https://www.cyberciti.biz/tips/red-hat-fedora-linux-servers-compromised.html)

Google: "5 Million Google Passwords Leaked" (https://www.bankinfosecurity.com/5-million-google-passwords-leaked-a-7299)

Twitter: "33m purported account details were posted for sale on the dark web" (https://www.theguardian.com/technology/2016/jun/10/twitter-locks-millions-accounts-passwords-leak)

And, finally, the web site we are using right now:

GitHub: "Attackers Used Leaked Passwords to Hack GitHub Accounts" (https://www.securityweek.com/attackers-used-leaked-passwords-hack-github-accounts). Not strictly a GitHub fault, but still an attack.
And then another: "GitHub has warned a select number of users that a bug in its password reset functionality has recorded users' passwords in plaintext format inside the company's internal logs" (https://www.securityweek.com/attackers-used-leaked-passwords-hack-github-accounts). Not a break in, but still an unintended leak.

Should we stop using GitHub because of that? I guess your answer, as well as mine, will be "no" (at least not for the mentioned security reasons). Does this mean that the guys are bad at maintaining their servers? Again, no, the incidents were handled frankly, quickly and the damage was minimised.

Now regarding my server. First of all, we all trust our personal computers (including you trusting the machine you are using right now); if that machine were compromised, a malware could theoretically patch any executable script within your Git repo, push it to GitHub on your name, and infect our computers before we notice that. You running the tests on a virtual machine would not help because the intruder would have the same access as you, so could access the container as well.

I have several computers; one of them runs at saulius-grazulis.lt. What would be the reason for you to trust my laptop from which I commit (it also has a name, varanas.saulius-grazulis.lt) but not my server? And what would be the reason for me to trust *your* laptop, for that matter? "how can I completely trust *you* laptop not to push any crazy content without you spotting it"?

How secure is my server? Well, it is behind a firewall with only relevant ports open. It does not run MS Windows :). It does not run any login services over Web that are known to have had issues in the past and that would need increased attention, like WordPress. I use the server regularly so I monitor the process lists, activities, load, logs on a regular basis. I have fail2ban installed with am extended ban period to make automatic password scanning harder. I keep my passwords off-line and/or encrypted, use different password for different domain, use long and randomised passwords. Does that make you slightly more comfortable? Would switching from svn: to https: increase your confidence? (the latter switch could indeed prevent certain name-highjacking attacks). And also it is a very small server with little computing power in it, so there is little gain for the intruders to get there (no user accounts, no passwords, no mails... just scripts). Sure it can be hacked and the harm can be caused, but you will see quite soon a strange script arrived upon you check-out, wouldn't you?

I can put my source on GitLab or GitHub, but will that increase the security if the sources are still pushed from my server? Or from my laptop? I guess no. BTW, in case you wish, you are welcome to perform your security audit of my server :)

So what are our options? Short of not using somebody else's code (including mine) at all, will changing a host increase the security? It may *seem* that it is, but is it really? Given that people have hacked into company hosted Git repos (https://security.stackexchange.com/questions/209448/gitlab-account-hacked-and-repo-wiped, https://github.com/search?o=desc&q=1ES14c7qLb5CYhLMUekctxLgc1FV2Ti9DA&s=indexed&type=Code), I doubt.

To make the matters more secure, I am ready to add extra security measures; say signing my code with my GPG key, so that we can add a verification step to OPTiMaDe repo on 'make test' before running the code. But this is extra work for me, so we need to agree exactely what we need and what is considered secure enough ("Wie unsicher ist sicher genug") before I engage in doing that; I would be unhappy if I invest my time into extra security and then someone says "well, you know, I can not completely trust your server, so ...".

sauliusg:
> > One more thing that Travis needed is a status code that uniquely indicates the status of the tests.
> 
> I thought this PR handles this by making the validate taget end with:
> 
> ```
> (exit $$($(MAKE) -s -f tests/GNUmakefile listdiff | head -n 1 | wc -l))
> ```
> 
> Meaning the top-level `make validate` is meant to return an error status code if the validation fails.

This makes a lot of sense and we can do such target, why not. I am positive.

Just regarding the name: "validate" in our parlour is validating data against some semantic descriptions, e.g. validating CIF agains CIF dictionaries, XML against XML Schema or similar. I would pick another taget name. What about `make test status` or similar?

For just running tests, usually just `make test` or `make check` is enough; I would leave these well-known target names instead of `make run_tests` (`run` seems superfluous to me).

sauliusg:
> I did not set out to change the top-level makefile! I just added my targets to the old top one, expecting it to work. However, `make <my_target>` then gave a mass of output about missing files, executions of awk, mkdepend, etc. 

If I see exactly you are trying to do, most probably I'll figure out what is taking place and why you get these messages. I'll have a look at the code that fails in your hands; do I correctly understand that it is committed to you fork of the OPTiMaDe specs? Could you please give me an URL/commit which I should look at first, that would help me a lot?

The GNU Make is a rather advanced Make system (this is why I advocate it :); it works in several steps to automate things. First of all, it can run shell commands to analyse the file system and to find our the work it needs to do (this is what the `TEST_FILES = ${wildcard ${TEST_DIR}/*.inp}` stanzas are about). While doing these, the current Makefiles run `awk` to analyse `optimade.rst` and to figure out what grammars and REs are there and what needs to be tested.

After that, the GNU Make has a possibility to "make makefiles" (https://www.gnu.org/software/make/manual/html_node/Remaking-Makefiles.html). This allows automatic on the fly dependency generation. With other/older Make systems, you had to run `make depend` or something like that to get your dependencies up to date. This is done in the current make system to update test and grammar dependencies. We use the tool `mkcomdepend` to generate Make-syntax dependencies from shell scripts; the tool is small, has no extra dependencies except Perl so it is included into the OPTiMaDe repo (under `tests/tools`).

All these features work for decades in our hands here in Vilnius, have been used by multiple people, and allow to automate a lot of repeating tasks, avoiding mistakes; therefore I highly recommend it.

If you computer runs the current `make tests` smoothly (with all the downloads, compilations, etc.), then it means that all dependencies are right (including all extra software like awk, Perl, mkcomdepend). The problem then should be in interaction of the new additions with this existing system. Instead of moving the existing test system, which works nice IMHO, let me have a look at the make files and suggest a way to incorporate the new changes into it. I'm rather confident that this is doable.

> I did not want to start majorly re-working "your" makefiles, so I kept your build system intact, but diverted it into a separate makefile invoked only for the grammar validation target. Someone who wants to interact directly with it, e.g., to run `make run`, can still do so simply by first `cd tests`

Well, this *is* a reworking of the test system, be it mine or not, and requiring `cd whatever` changes the intended interface. Maybe before doing this let's try to figure out what you would like to achieve and how to include those targets into the existing system? I'm pretty confident that this is possible and I'm ready to do much of the necessary work and explanation==documentation (which is admittedly scarce at the moment) for that; but I'll need your input regarding what yo would like to acheive.

sauliusg:
> I get that you are trying to explain this in a reply further down, saying that I should add my targets as "Makelocal-* files". But can you please spell out very clearly exactly what I need to do for `make spellcheck` to only invoke precisely the command: `aspell -x -p ./.words.lst -l en_US -c optimade.rst`? (If you explain this one example, I can handle all the other targets myself.)

Indeed, the Makelocal-* files are the intended way to add new functionality. And, incidently, I have a Makelocal file that does pretty much that! You see, the intention of Makelocal-* stuff is that you can have modular Makefiles, each containing a separate function, and you just drop in a Makefile and it works without modification (or with as little modification as possible). In this way you can reuse the same Makefiles for very different projects! All customisation that is peculiar to the project, or to a working copy, should go into Makeconfig-* files. So far this works pretty nice in our hands.

Unfortunately, one needs to observe several conventions for this to work smoothly. For instance, multiple Makefiles may not contain commands for the same targtet; so you can not say:

```
# Makelocal-spell
all:
<tab>spell ${TEXT_FILES}
```

in one makefile and later 

```
# Makelocal-check
all:
<tab>check-grammar ${GRAMMARS}
```

in another. By one can instead write:

```
# Makelocal-spell
all: spell

spell: ${TEXT_FILES}
<tab>spell $^
```

and then later:

```
# Makelocal-check
all: check

ckeck: ${GRAMMARS}
<tab>check-grammar $^
```

and this works fine!

So let's have a look how we can apply this method here; I'm pretty sure we can integrate all you targets into the existing system. Let me propose a solution and then we discuss.

Apart from 'spell', what would be other targets that you indent to put in?

sauliusg:
> For security reasons, I am proposing that docker should be the default as long as we need to fetch and run outside code as part of the validation process.

Why not running just

```
docker run make test
```

and have it "dockerised"? In this way, the Makefiles do not need to know anything about the containers. The separation of concerns happens. In other circumstances, someone may want to run OPTiMaDe tests in a virtual machine, so the will use something like `ssh my-virtual-host make -C OPTiMaDe test`. Extra secure if you wish, just close all outgoing TCP/IP conections :).

Alternatively, one could set up a extra Make configuration, DOCKER=1, in a Makeconfig-* file specific to his/her machine, and have `make test` run under docker automagically. But for the rest of us, if we do not use docker, why should we care about it? A good principle would be "if you do not use it, you do  not pay for it", as much as possible.

rartino:
Let us start with the pure practicalities.

I try to follow your instructions. You can see the code here:
  https://github.com/rartino/OPTiMaDe/tree/makefile_reorg_v2
where the only commit I've made is this:
  https://github.com/rartino/OPTiMaDe/commit/bb535f9e6f4940d6ffb28a5d7c1c295dccdae437

When I run `make spell` on a freshly checked out repository, I get this console output:
```
tests/makefiles/Makelocal-grammars:27: .grammars.d: No such file or directory
awk '/^ *\(\* BEGIN EBNF GRAMMAR/{print "tests/generated/"$5".ebnf:", FILENAME}' \
	optimade.rst > .grammars.d
awk '/^ *#BEGIN ERE/{print "tests/generated/"$3".ere:", FILENAME}' \
	optimade.rst | tr -d "\r" >> .grammars.d
awk '/^ *#BEGIN PCRE/{print "tests/generated/"$3".pcre:", FILENAME}' \
	optimade.rst | tr -d "\r" >> .grammars.d
aspell -x -p ./.words.lst -l en_US -c optimade.rst
```
and then aspell is run, as it should.

This output is tied to the validation of the grammar in the specification, so I find it very confusing that these things are shown when I am executing a target that has nothing to do with validation of the grammar. Clearly these are commands to collect data from optimade.rst into a .grammars.d file. Why aren't they only run for targets that depend on .grammars.d?

So, what do I need to fix the dependency handling of .grammars.d?

rartino:
It should be apparent from up-thread that I am trying to work with the GNU Makefile framework we have in the repo. Nevertheless, I'll make a brief comment on the discussion about its merits.

I have nothing against a build framework like this, rather, I recognize that the complexity is *highly warranted* for large-ish software projects. I have created similar setups of scripts and makefiles myself in various projects. 

We are presently working with a repository of *one single text file*. We want a few targets for some validation (grammar, spelling, author order) and some other shell helpers, e.g., render the text to pdf, html. For this, do we *really* need the complexity of this build framework over, say, a more normally formatted posix-style makefile? There is a non-trivial cost associated with the complexity in that new people who want to edit the make targets need to learn how it works.

rartino:
Now, about the security situation of `make` fetching code from an SVN server and executing it as my user on the local machine.

Again, my main point remains that the present setup is less secure than a reasonable person would expect from running `make`, which is why I advocate running the grammar checks in docker as the default. I wasn't so much trying to get us into the rabbit hole of discussing relative security, but here we go...

I agree with the point that security is never absolute, but I don't quite see the relevance of some of your security links in the present context. It would seem most relevant if you can point at attacks where binaries executed on end-user's machines were altered, not, e.g., leaked forum identities. I maintain that some distributions spend quite a bit of effort in trying to avoid that scenario with signatures, etc. 

Do they spend more effort than you spend on securing your SVN server? I can indeed not know that. 

However, that isn't so crucial for my argument. I indeed *have to* (to some degree) trust the binaries on my local machine. (Which is why some security-sensitive places spend a lot of effort on securing that attack vector.) But, beyond that, I simply like to extend as little trust as possible. 

The argument that "system A is at least as secure as system B, so if you trust B, you can as well trust A" is flawed. One is still better off trusting only B. Your argument mirrors that argument, with B = my distro, and A = your SVN server.

Them you mention "so what about github?", but the elegant thing with git is that I don't need to trust the servers hosting my git repositories, as long as I have local copies that I, at some point, had trust in. When I `git pull`, I'll either see the changes in the changelog OR my git client will scream that someone has tampered with the remote repository. The same thing would apply to the code downloaded from your SVN at `make`-time, if it instead was on github and downloaded at `clone`-time. But with the present setup, if someone breaks into your SVN and replaces something, there is no way for me to notice.

Your final question is: so what can we do? My answers are:
1. For now, lets make the default grammar validation run in docker, as pushed by this PR.
2. Work to find a better solution where all scripts and tools needed to validate the grammar are either in the main repository or a subrepository.

In a final attempt to illustrate my point, what would you say if tomorrow we merge a bunch of PRs from people you don't know which add instructions to dynamically download code from their servers at `make`-time? E.g., for spellcheck of rst we need a patched aspell, so why not let `make spell` do this:
```
wget http://johns.private.server/bins/my_patched_aspell_binary \
  && chmod +x ./my_patched_aspell_binary \
  && ./my_patched_aspell_binary optimade.rst 
```
To properly sort authors we need a modified 'sort' that can handle name affixes (actually very tricky), so `make sort_authors` now do this:
```
wget http://bennys.private.server/bins/better_sort \
  && chmod +x ./better_sort \
  && ./better_sort AUTHORS 
```
etc., etc. Would you be comfortable with running these on your machine? Why/why not?

sauliusg:
> When I `git pull`, I'll either see the changes in the changelog OR my git client will scream that someone has tampered with the remote repository. The same thing would apply to the code downloaded from your SVN at `make`-time, if it instead was on github and downloaded at `clone`-time.

Actually, the make downloads the code only once and then reuses it, so you get the packages at the pretty much same time and in the same schedule as if it were only from Github...

But, to the point: would you see it as more secure if we commit *all* code that is needed for the grammar checks into the OPTiMaDe repo, say under test/tools/external, with specific versions tagged, and get all the code at the `clone`-time? This would then have only one point potential of failure, the GitHub (and, actually, our personal computers, but that's inevitable).

The 'make' would then build the tools immediately after the cloning (as it does now in the main branch) and run the tests; subsequent tests would re-use the code.

That would solve the issue of "untrusted" svn server, wouldn't it?

sauliusg:
> * For now, lets make the default grammar validation run in docker, as pushed by this PR.

There are a couple of issues with that:

1. The code in the current GNUMakefile attempts to run commands under `sudo`:

```
saulius@koala OPTiMaDe/ $ make -n validate
sudo docker build -t optimade . -f tests/Dockerfile
sudo docker run -it --rm optimade:latest
```

Now, we can debate whether it is safe to run a code downloaded from somebody's server as a regular user is secure or not; but running code from make under `sudo` is IMHO an absolute no go!

Moreover, it is not necessary! Just add yourself, and the users who are supposed work with Docker to a `docker` group, and the `dockerd` will handle everything fine. At least it works so on ubuntu 16.04 and 18.04.

Thus, I suggest removing the `sudo` commands and possibly adding explanation about the `docker` group.

2. The provided Dockerfile does not build image on Ubuntu-16.04:

```
saulius@koala OPTiMaDe/ $ docker build -t optimade . -f tests/Dockerfile
Sending build context to Docker daemon   8.97MB
Step 1/6 : FROM ubuntu:trusty
 ---> d5e7f166b406
...
E: Failed to fetch https://esm.ubuntu.com/ubuntu/pool/main/t/tzdata/tzdata_2019c-0ubuntu0.14.04_all.deb  HttpError401

E: Failed to fetch https://esm.ubuntu.com/ubuntu/pool/main/t/tzdata/tzdata-java_2019c-0ubuntu0.14.04_all.deb  HttpError401

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

The Docker build works of the FROM is set to 18.04 in `tests/Dockerfile` on both Ubuntu-16.04 and Ubuntu-18.04; the current 'trusty' (14.04) is rather dated and not even under maintenance support; why do we want to use it? I guess 16.04 or 18.04 would work nicely (there may be Java issues on 18.04 but I know how to resolve them).

I can currently not afford changing the OS from 16.04 which is still supported; my upgrade is planned to 20.04 next year. In any case, the Dockerfile that works only on some of the supported systems is too fragile, IMHO, to be offered to broader audience.

3. The current `make validate` builds all the tools each time it is run anew, and runs all the tests:

```
time run make
cd tests/tools/ && unzip /root/optimade/tests/tools/grammatica-1.6.zip
Archive:  /root/optimade/tests/tools/grammatica-1.6.zip
611daf4d1efeeca9fb740d128ed8b4940540759f
   creating: grammatica-1.6/
  inflating: grammatica-1.6/.classpath  
 ...
real	0m54.699s
```

The current system runs just the tests, and just those tests that are necessary:

```
time make clean tests
...
real	0m28.102s
```

Twice the difference. Not a spectacular one, but still – why should I wait for the Docker to do the same work over and over again?

Finally, I can run the tests selectively:

```
saulius@kukutis sauliusg-optimade-specification/ $ time make tests -W tests/cases/Filter_010.inp 
tests/cases/Filter_010.inp:    OK

real	0m0.201s
```

And this is now *really* fast; very handy when changing the grammar (I even have special target, `make run`, for such quick checks). Now, how can I do the same under Docker?

If there is no possibility to run selective tests, and to run just the necessary tests automatically, the whole point of using `make` is moot, it reduces to just another shell script. A step backwards, IMHO.

sauliusg:
> And, running `make validate` runs all validations inside a docker container. `make validate_without_docker` runs all validations without docker

To make it consistent, it should then be called `make validate_with_docker` :P

sauliusg:
> However, the old GNUmakefile was very targeted specifically at running the grammar test suite to the extent that adding other types of targets was difficult.

Actually, this is totally untrue. The existing system is not at all targeted to tests, it has just one of the Makefiles/target set for that purpose, and it is very easy to extend – much easier than the monolithic Makefile you propose. I'll show how in the upcoming comments.

sauliusg:
> I try to follow your instructions. You can see the code here:
> https://github.com/rartino/OPTiMaDe/tree/makefile_reorg_v2
> where the only commit I've made is this:
> [rartino@bb535f9](https://github.com/rartino/OPTiMaDe/commit/bb535f9e6f4940d6ffb28a5d7c1c295dccdae437)

OK, now I see! Thanks for sharing the link.

The Makefile you have added is fine, and works as expected :). There is the only thing that I would do differently: I would not let `all:` target depend on `spell`.

This is because `make spell` does not produce any output, and therefore it is re-run each time you invoke `make`. The desired behaviour from my point of view would be:

```
saulius@koala sauliusg-optimade-specification/ $ make -n
make: Nothing to be done for 'all'.
```
In this way one can easily establish that all targets are rebuilt (and `make listdiff` can show which tests have failed, without running them once more).

> When I run `make spell` on a freshly checked out repository, I get this console output:
> 
> ```
> tests/makefiles/Makelocal-grammars:27: .grammars.d: No such file or directory
> awk '/^ *\(\* BEGIN EBNF GRAMMAR/{print "tests/generated/"$5".ebnf:", FILENAME}' \
> 	optimade.rst > .grammars.d
> awk '/^ *#BEGIN ERE/{print "tests/generated/"$3".ere:", FILENAME}' \
> 	optimade.rst | tr -d "\r" >> .grammars.d
> awk '/^ *#BEGIN PCRE/{print "tests/generated/"$3".pcre:", FILENAME}' \
> 	optimade.rst | tr -d "\r" >> .grammars.d
> aspell -x -p ./.words.lst -l en_US -c optimade.rst
> ```
> 
> and then aspell is run, as it should.
> 
> This output is tied to the validation of the grammar in the specification, so I find it very confusing that these things are shown when I am executing a target that has nothing to do with validation of the grammar. Clearly these are commands to collect data from optimade.rst into a .grammars.d file. Why aren't they only run for targets that depend on .grammars.d?

This is because the target that depends on `.grammar.d` is the GNUMakefile itself :). Please, please have a look into "making makefiles" (https://www.gnu.org/software/make/manual/html_node/Remaking-Makefiles.html).

The behaviour you observe is correct, intended one, causes no harm and is informative, showing us that the dependences were generated as expected. The dependencies are only generated the very first time you run `make` after cloning, or after you change the *.rst file(s) in the directory. After that they are not executed any more:

```
saulius@koala OPTiMaDe/ $ make -n spell 
tests/makefiles/Makelocal-grammars:27: .grammars.d: No such file or directory
awk '/^ *\(\* BEGIN EBNF GRAMMAR/{print "tests/generated/"$5".ebnf:", FILENAME}' \
	optimade.rst > .grammars.d
awk '/^ *#BEGIN ERE/{print "tests/generated/"$3".ere:", FILENAME}' \
	optimade.rst | tr -d "\r" >> .grammars.d
awk '/^ *#BEGIN PCRE/{print "tests/generated/"$3".pcre:", FILENAME}' \
	optimade.rst | tr -d "\r" >> .grammars.d
aspell -x -p ./.words.lst -l en_US -c optimade.rst

saulius@koala OPTiMaDe/ $ make -n spell 
aspell -x -p ./.words.lst -l en_US -c optimade.rst
```

The dependency generation will be rerun automatically when the `optimade.rst` file is changed:

```
saulius@koala OPTiMaDe/ $ make -n spell -W optimade.rst 
awk '/^ *\(\* BEGIN EBNF GRAMMAR/{print "tests/generated/"$5".ebnf:", FILENAME}' \
	optimade.rst > .grammars.d
awk '/^ *#BEGIN ERE/{print "tests/generated/"$3".ere:", FILENAME}' \
	optimade.rst | tr -d "\r" >> .grammars.d
awk '/^ *#BEGIN PCRE/{print "tests/generated/"$3".pcre:", FILENAME}' \
	optimade.rst | tr -d "\r" >> .grammars.d
aspell -x -p ./.words.lst -l en_US -c optimade.rst
```

(the `-W optimade.rst ` option tells GNU Make to pretend that `optimade.rst` was modified without actually changing or touching it).

Automatic dependency generation was a big step forward compared to old Makes, I guess some two decades ago; before that, you had to rebuild dependency files each time your sources are changed, by running `make depend` or something like that, and if you forget (which would happen more often that one would wish ;), you would end up with botched build. With the 'making makefiles' feature, the Make itself takes care for this. and you always (well, nearly always ;) get correct dependencies.

In our case, if you add a new grammar/RE group to the `optimade.rst`, e.g.:

```
    #BEGIN PCRE identifiers
    [a-z_][a-z_0-9]*
    #END PCRE identifiers

    #BEGIN PCRE operators -- added just for testing
    [A-Z][A-Z_]*
    #END PCRE operators

    #BEGIN PCRE numbers
    [-+]?(?:\d+(\.\d*)?|\.\d+)(?:[eE][-+]?\d+)?
    #END PCRE numbers
```
and run `make -n`, you'll see the dependency:

```
tests/generated/operators.pcre: optimade.rst
```
appearing in `.grammars.d`, and, since the provided Makefiles contain a general rule to extract REs, the file can be automatically generated:

```
saulius@koala OPTiMaDe/ $ make tests/generated/operators.pcre; cat tests/generated/operators.pcre
awk '/^ *#BEGIN PCRE operators/,/^ *#END PCRE operators/' optimade.rst \
	| sed 's/^ *//' | tr -d "\r" > tests/generated/operators.pcre
#BEGIN PCRE operators
[A-Z][A-Z_]*
#END PCRE operators
```

So you can just drop in a test that tests `operators.pcre` and it starts working, no config or changing Makefiles is necessary; and this file will be automatically updated each time you change `optimade.rst`, exactly before the tests are run, without us coding a single line!

The current system is geared towards having multiple *.rst files; while this is not necessary at the moment, it does not harm either, and having generic system is useful since you can reuse it.

> So, what do I need to fix the dependency handling of .grammars.d?

As you see, there is no need to fix it, everything works just fine :)

If you are absolutely uncomfortable with `make` printouts, we can add `@` characters before commands that generate `.grammars.d`, and they will not be printed (but will be executed nontheles).

```
diff --git a/tests/makefiles/Makelocal-grammars b/tests/makefiles/Makelocal-grammars
index 590d9d5..9e11cc4 100644
--- a/tests/makefiles/Makelocal-grammars
+++ b/tests/makefiles/Makelocal-grammars
@@ -41,11 +41,11 @@ ${OUTP_FILES}: ${GRAMMAR_FILES}
 #------------------------------------------------------------------------------
 
 ${GRAMMAR_DEPENDENCIES}: ${RST_FILES}
-	awk '/^ *${GRAMMAR_START_STRING}/{print "${GRAMMAR_DIR}/"$$5".ebnf:", FILENAME}' \
+	@awk '/^ *${GRAMMAR_START_STRING}/{print "${GRAMMAR_DIR}/"$$5".ebnf:", FILENAME}' \
 		$^ > $@
-	awk '/^ *${RE_START_STRING} ERE/{print "${GRAMMAR_DIR}/"$$3".ere:", FILENAME}' \
+	@awk '/^ *${RE_START_STRING} ERE/{print "${GRAMMAR_DIR}/"$$3".ere:", FILENAME}' \
 		$^ | tr -d "\r" >> $@
-	awk '/^ *${RE_START_STRING} PCRE/{print "${GRAMMAR_DIR}/"$$3".pcre:", FILENAME}' \
+	@awk '/^ *${RE_START_STRING} PCRE/{print "${GRAMMAR_DIR}/"$$3".pcre:", FILENAME}' \
 		$^ | tr -d "\r" >> $@
```
And then you only see the one-line "no such file or directory" message once, after you clone the repo:

```
saulius@koala OPTiMaDe/ $ rm .grammars.d 

saulius@koala OPTiMaDe/ $ make -n spell 
tests/makefiles/Makelocal-grammars:27: .grammars.d: No such file or directory
aspell -x -p ./.words.lst -l en_US -c optimade.rst

saulius@koala OPTiMaDe/ $ make -n spell 
aspell -x -p ./.words.lst -l en_US -c optimade.rst

saulius@koala OPTiMaDe/ $ make -n spell -W optimade.rst 
aspell -x -p ./.words.lst -l en_US -c optimade.rst
```

But I would not do that, since it is useful to look at the dependency generation every now and then to make sure it happens as expected.

sauliusg:
> We are presently working with a repository of _one single text file_. We want a few targets for some validation (grammar, spelling, author order) and some other shell helpers, e.g., render the text to pdf, html. 

The targets you list are far from being few. To the existing `test`, `all`, `clean`, `distclean` you plan to add `pdf`, `spell`, `author-order`, .... what else? These not so few. Not too many either, but, since several of these functions are unrelated to each other (say `test`, `spell` and `pdf`), there is a good reason to keep them separated in distinct files.

> For this, do we _really_ need the complexity of this build framework over, say, a 

The framework is *not* complex; I wouldn't even call it a framework. It is very straightforward.

Adding new set of related targets can now be done completely independently, without changing existing working Makefiles, by just adding another makefile, as you have nicely done for Makelocal-spell. Was that a problem?

> more normally formatted posix-style makefile? 

I'm strongly advocating GNU Make instead of POSIX make. GNU Make has a full set of relevant features that older Makes lack.

Having all targets in one Makefile makes it difficult to maintain and error-prone. We have gone through that. This is so regardless whether the Make is POSIX or GNU.

Otherwise the proposed Makefiles are formatted in a pretty usual way.

> There is a non-trivial cost associated with the complexity in that new people who want to edit the make targets need to learn how it works.

The people will need to learn how Make works, anyway. If they know this, adding new target is as difficult/easy as in any Makefile. Of course you need to know how the GNU Make works and not to get confused by extra messages it outputs; but this has nothing to do with the current Make system; any Makefile with the same rules will give identical messages, even if you merge everything into one file.

Editing large Makefiles is a hassle; I've tried this many times, and we have abandoned such approach long ago. The problem is that a long Make has interactions of rules and variables all over the file, and there is no way to easily separate them (apart from manually picking rules an pasting them into another file). Given in addition that Make does not diagnose misspelled variables, it is easy to make mistakes. Searching such mistakes in *long* Makefiles is a nightmare. Keeping Makefiles short and focused on one task helps a lot. Moreover, you can easily test each Makefile separately (by running `make -f tests/makefiles/Makelocal-spell -n`, for example), and the combine them by simply placing/linking them into a single directory.

IMHO, this justifies introducing several conventions, e.g. not using recipes for `all` target; this is widely used anyway and the Make will tell you if you make this mistake.

And if modular make with 8 Makefiles (including your Makelocal-spell) seems complex, then Docker with its client/server, snapshots, downloads of the whole OSes from remote servers and close to 20000 files in its setup working is *way* more complex. And you still are not against it :). Let's be consistent – if Docker is not too complex, the GNU Make is sure not too complex as well.

rartino:
@sauliusg while we seem to have a few fundamental disagreements on some of these topics, I see no problem getting in agreement over the change request list you posted. (Since github allows it, I took the liberty to edit that post to turn the itemized list into numbered points, so I can reference them.)

## Point 1, 2, 4
I'll update the PR to confirm with points 1, 2, 4.

I have no trouble at all understanding the immense usefulness of automatic dependency generation for software projects. However, in specifically the present project, can you please explain *why* `.grammars.d` needs to be created as part of "making makefiles"? I don't see how we have any dynamic dependencies at all in this project? Why can we not just make all grammar testing targets depend on `.grammars.d`?

## Point 3
Are you OK with the solution of adding the SVN code and Grammatica as git subrepositories to the OPTiMaDe one? It is my impression that this feature is precisely indended to address the situation we are having. But, it means that to run the tests one needs an extra step after `git clone` of:
```
git submodule update --init --recursive
```
However, I think I can add code to the makefile to detect that this hasn't been run and give the user a helpful hint about the need to run this command.

If you approve this solution, can you please open a new repository under Materials-Consortia and upload your SVN repository tools there? More or less exactly as you did with Grammatica.

## Point 5
With the solutions implemented to point 3, I'm happy to back off my request of docker being default. Would you then be ok with me just adding a target `test_in_docker` that builds the docker image and runs the tests inside that, and leave all other testing targets as they are?

## Point 6
With the suggested solution to point 5, the "normal" way for running the tests would be on the local machine just as it has so far. Is it then so important if the Makefile helper for docker wastes 30 seconds by re-running the tests from scratch each time it is invoked? A fix to this can certainly be implemented, but it seems unnecessary to require this work for what now will be a helper target for those that have difficulty with getting all the dependencies in place.

rartino:
@sauliusg Ok, with the reverts + new commits I've believe I have addressed the points brought forward here, except point 3.

Point 3 requires your help in checking in your SVN code in a subrepository of Materials-Consortia so that I can add it as a subrepository dependency.

On the other hand, perhaps this PR can already be merged, and we can address point 3 in a separate PR? 

From your point of view, I don't think the present PR should make anything worse? Everything from the original build system is still left intact, all I've done is add targets. (From my point of view, the security issue I see still remains even after merging this PR.)

sauliusg:
> I have no trouble at all understanding the immense usefulness of automatic dependency generation for software projects. However, in specifically the present project, can you please explain _why_ `.grammars.d` needs to be created as part of "making makefiles"? 

> I don't see how we have any dynamic dependencies at all in this project? 

The short answer is in the `.grammars.d` itself:

```
tests/generated/Filter.ebnf: optimade.rst
tests/generated/Number.ebnf: optimade.rst
# ... the rest skipped for brevity
```

which lists files that need to be generated from `optimade.rst` because the corresponding grammars are mentioned in the `optimade.rst` contents. 

The long and more detailed answer is below:

The current automated testing sequence runs as follows:

1. From the `optimade.rst`, *all* (EBNF, PCRE and ERE) grammars are automatically extracted;

2. The EBNF grammars are automatically converted to the form suitable for the currently used parser generator (currently, Grammatica/LL languages);

3. All tests that depend on grammar files are run.

For this to work, Make needs to know somehow *which* grammar files are generated from the `optimade.rst`. Our convention is that *all* grammar files mentioned in that file should be extracted and available for testing.

One way (and the only way available if you do not use automatic dependency generators) would be to list *by* *hand* all dependencies of all grammars, e.g.:

```
tests/generated/Filter.ebnf: optimade.rst
tests/generated/Number.ebnf: optimade.rst
```

and so on. Now, the first instinct might be to put the rules next to these dependencies:

```
tests/generated/Filter.ebnf: optimade.rst
<tab>awk '/^ *${GRAMMAR_START_STRING} $*/,/^ *${GRAMMAR_END_STRING} $*/'  optimade.rst \
                | sed 's/^ *//' >  Filter.ebnf

tests/generated/Number.ebnf: optimade.rst
<tab>awk '/^ *${GRAMMAR_START_STRING} $*/,/^ *${GRAMMAR_END_STRING} $*/'  optimade.rst \
                | sed 's/^ *//' > Filter.ebnf
```

But this, of course, is very bad since it violates the [SPOT](https://en.wikipedia.org/wiki/Single_source_of_truth)/[DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principles.  The mistake in the above code is semi-intentional (I first did it inadvertently by copying the text, and when spotted it on the second reading and decided to leave it as an example).

Now, I am pretty sure you are aware of these principles and for sure most of you code follows them, but I point this out once more since, IMHO, they are also applicable to `Make` files even though these files might be seen as auxiliary tools only. Also, duplications of exactly the same kind were present in the Makefile that was suggested as a replacement of the existing one (`extract_..._grammar:` target rules in GNUMakefile of the commit fc9aec5). I suggest that the Makefiles also follow SPOT/DRY for the ease of maintenance.

Thus the better solution IMHO would be:

```
tests/generated/Filter.ebnf: optimade.rst
tests/generated/Number.ebnf: optimade.rst

# from tests/makefiles/Makelocal-grammars:
${GRAMMAR_DIR}/%.ebnf:
        awk '/^ *${GRAMMAR_START_STRING} $*/,/^ *${GRAMMAR_END_STRING} $*/' $< \
                | sed 's/^ *//' > $@
```

which removes the code duplication by using the same rule for all grammars. The only remaining point is that in this solution (which I do not advocate) we would have to update dependencies by hand, and in a while the fact that these dependencies are necessary will be forgotten by us, not known by new people, overlooked in a hurry, etc. Moreover, editing the input data (dependencies) and the code (rules/recipes) in the same eventually file causes troubles when you want to combine old code with new dependencies or vice versa, when you need to cherry-pick individual file lines instead of just replacing the whole file.

All these problems are addressed in the current system by generating the `.grammars.d` file automatically. It contains dependencies separated from the rule. Dependencies are updated automatically whenever the source file, `optimade.rst`, is changed. If new grammars are added, new dependencies are generated on the fly, and the same rules are applied to generated them (this already happened once in this project when the `Numbers.ebnf` sub-grammar was marked up for testing). So far, it works fine, doen't it? And the GNU Make message about creating `.grammars.d` reminds you of what is going on, and helps new people to figure our where the dependencies come from (provided they are somewhat familiar with GNU Make, but we need to know our tools; we don't complain that RST renders in a strange/unintuitive/unfamiliar way, do we? Instead, we just read the fine docs ;)

> Why can we not just make all grammar testing targets depend on `.grammars.d`?

The `.grammar.d` file is included into the Makefiles using the `include` directive, and so the GNU Make functionality is to (re)build this file each time its prerequisite (`optimade.rst`) changes; this is the correct behaviour, since otherwise we will get incorrect builds (e.g. some grammars will not be updated and thus will not be tested). I do not think there is a straightforward way to suppress this behaviour *without* introducing inconsistencies into makes/tests.

While I can not claim that there is no way to do what you are asking for, I am not inclined on dwelling too much on that; I think the system works well and as intended; it's implemented in a fairly standard way (most projects that use GNU Make autogenerate dependencies in a similar way, that where I looked it up) and we could spend our time for more worthy goals of implementing and testing OPTiMaDe itself.

If the makefile generation is so unwelcome to you, you can always resort to a standard functionality of using one particular Makefile; for instance, if you only want to check the spelling without generating the dependencies, you could do:

```
make spell -f tests/makefiles/Makelocal-spell
```

the shorter version also works:

```
make -f tests/makefiles/Makelocal-spell
```

since `spell` is the default target there; and you can shorten the command in `bash` by:

```
alias optispell='make -f tests/makefiles/Makelocal-spell'
```

or even shorter

```
alias os='make -f tests/makefiles/Makelocal-spell'
```

Would that suite your needs?

sauliusg:
> Are you OK with the solution of adding the SVN code and Grammatica as git subrepositories to the OPTiMaDe one? It is my impression that this feature is precisely indended to address the situation we are having. But, it means that to run the tests one needs an extra step after `git clone` of:
> 
> ```
> git submodule update --init --recursive
> ```

The git submodule handling is awkward, and the need of extra options and extra actions on update will be, IMHO, a source of additional confusion and mistakes. That said, it is still probably better than copying code (even as a read-only copy) to the OPTiMaDe repo itself. If we ant to stay with `git` I do not see a much better solution.

Another solution would be to copy Grammatica and the "SVN code" (called "Grammatiker", btw) into the '.../external/' subdirectory of the OPTiMaDe repo, mark them with specific tag/commit/revision specifiers, and clearly indicate that these are readonly copies intended to be compiled and run bot not to be modified and forked in the OPTiMaDe repo. Unfortunately, I do not know any way in git to endorse read-onlyness other than specifying that in a README, and I'm not sure if people will follow the readonly-use advise since it takes extra work to stick to such r/o policy...

So, basically, I'm equally (un)happy with both suggestions, so let's pick the one other people like best.

> However, I think I can add code to the makefile to detect that this hasn't been run and give the user a helpful hint about the need to run this command.

That would definitely help!
> If you approve this solution, can you please open a new repository under Materials-Consortia and upload your SVN repository tools there? More or less exactly as you did with Grammatica.

I'll push it in a short while and let you know...

sauliusg:
> With the solutions implemented to point 3, I'm happy to back off my request of docker being default. Would you then be ok with me just adding a target `test_in_docker` that builds the docker image and runs the tests inside that, and leave all other testing targets as they are?

Fine with me. The `test_in_docker` sounds awkward a bit, but if you find it OK I'll not suggest anything else.

We could also introduce 'make help' to give a help message and to list all available targets.

sauliusg:
> With the suggested solution to point 5, the "normal" way for running the tests would be on the local machine just as it has so far. Is it then so important if the Makefile helper for docker wastes 30 seconds by re-running the tests from scratch each time it is invoked? A fix to this can certainly be implemented, but it seems unnecessary to require this work for what now will be a helper target for those that have difficulty with getting all the dependencies in place.

Not only that, but you need to install docker on your computer for just running grammar checks – IMHO, it is an overkill if you are not on Mac or some other system that can not do tests without the docker. I would prefer not to carry docker on my laptop for just running OPTiMaDe tests...

rartino:
Thanks for the details on what is going on in the interconnected makefiles, this makes it far easier to read them!

For the general discussion, I see your SPOT/DRY principles and raise you the [KISS](https://en.wikipedia.org/wiki/KISS_principle) principle. 

Every programming problem can be generalized infinitely far; e.g., what if we change the block markup to indicate a grammar? So, let's declare that block markup somewhere using a more general pattern language. Wait, what if we want to change the pattern language?, etc., etc. Good software engineering is a lot about striking the right balance between, on the one hand, being too explicit, meaning future changes require lots of manual work and the software gets unmaintainable over time; and, on the other hand, being too abstract and general, spending overly much time on overgeneralized problems and yielding a system so complex that thev deeper understanding of how it works is lost over time and the software becomes unmaintainable. I think you and I mostly agree on the basic principles of good programming, but we appear to disagree on the right balance for our present application: testing a handfull of grammar blocks in a text. I would have been more than happy to manually maintain a single dependency line per grammar in the specification, and do not see any issue with that. I do, however, see an issue with invoking essentially the full power of GNU make's handling of dynamic dependencies to avoid those few lines.

Now, from a more practical standpoint, the latest changes to this PR works with the dynamic dependency system as it is, which is what you want. We've also agreed to tackle point 3 in a separate PR. So, can you please update your review to the present state of the PR? 

(I gather from what your wrote that you want me to generalize the extract_* targets via a general rule, so I will fix that.)

sauliusg:
> (I gather from what your wrote that you want me to generalize the extract_* targets via a general rule, so I will fix that.)

I would say this is not necessary since it is already solved in the existing system as:

```
make
cat tests/generated/Filter.ebnf
cat tests/generated/numbers.ere 
```

and so on, in a more general way. No need to duplicate the existing commands.

rartino:
@sauliusg thank you for the feedback. Next review, would you mind using github's usual way of marking suggested changes in the files?

> 1. `tests/makefiles/Makelocal-docker` specifies `.PHONY: docker_test`, but has only the `docker_audit:` target. As a result, `make docker_test` does nothing, and `docker_audt` is not phony. Unless there is something which I totally overlooked, I suggest fixing that.

This was indeed a bug. Fixed.

> 2. `make docker_audit` fails if the 'make' was successfully run in the Git working copy. Changing 'make clean' to 'make distclean' in the current Dockerfile fixes that.

Done.

> The deeper reason for that is that the Docker copies symlinks to the container, which are absolute and do not find the required directories. The links are generated correctly anew after 'make distcean', however, the current Docker file does only 'make clean'. Either the docker needs to recreate precise test environment, or fix the links, or we need to switch to relative links.

I'd strongly prefer this to be patched to relative links if possible. It is nice for a cloned repository not to break when moved if one can avoid it.

> 3. The docker_audit target each time downloads the grammatica*zip, even if that file was in the local directory already.

Your bug fix is applied as suggested, and then the workaround of explicitly removing the zip is not needed. However, two observations:
- Typical usage for "docker users" will be to not download the zip in the local directory, and then this zip will anyway keep on getting re-downloaded inside docker.
- However, the upcoming PR to fix the security issue of running code downloaded at make-time, the whole "download zip"-thing is meant to go away anyway. So let's not spend too much time refining this.

> 4. [...] I suggest removing 'tests/makefiles/Makelocal-grammars-extract' for the following reasons: [...] It duplicates _code_ from 'Makelocal-grammars'; [...] It duplicates _functionality_ from 'Makelocal-grammars'

Done. It was just that I didn't know this functionality already existed... which begs a reply to your following comment:

> 4.3 The targets there are completely ad hoc; a person experienced in Unix will most probably find files in tests/generated/ (say after `tree` or `ls -lR`) and then use standard tools to process them, but it is much more difficult to guess that 'make extract_filter_grammar' exists, and what will it do

Everyone has different ways of discovering things. I typically read the top makefile for documentation and targets visible there. So, contrary to your point, it simply did not occur to me that the automagical handling of the dependencies had created separate targets for these files; but I would directly have found the `extract_*` targets from reading the now added documentation at the top of the makefile.

But, so, the solution is simple: I just removed `Makelocal-grammars-extract` as requested and changed the documentation to point at your targets instead. 

> 5. The help in the central GNUMakefile is very useful and much needed, but it is disconnected from the Make* files that actually implement the described features; it is not even clear which files implement which targets from the help. Thus we are very likely to get out of sync soon.

I started to work on this. However, with the simple solution you drafted of just running a `find` and merging comments together, we have no control over the order, nor can we group things originating from different Makelocal files that conceptually belong together under joint headlines (e.g., test targets that are spread out.) Hence, as I was trying to implement this, it became apparent that the resulting documentation would be significantly less helpful than what is at the top of the top makefile now, and therefore I reverted back to the centralized help. 

If you feel strongly about the help system you suggest, please accept the present PR with the centralized documentation, and then you can submit a separate PR that reorganizes the help system. We can then take the discussion there about whether the distributed documentation generates output that is helpful enough, compared to the centralized comment version.

> 6. The 'tests/makefiles/Makelocal-spell' spells out 'optimade.rst' explicitly; so do two other sets of rules. This violates SPOT/DRY and we will probably soon get out of sync.

Ok, (somewhat reluctantly) done in the way you ask for. But I accept this change only on the basis of the 'when in Rome'-rule.

Comment: this generalization for "what if we one day want more than one .rst file" and "let's use an indirect reference to the file name in case we change it", are exactly the type of overgeneralization that I discussed upthread. Code like this adds to the complexity and there is not much motivation for doing this work until it actually is needed. If we later decide to split optimade.rst into several files, it wouldn't be particularly expensive to go back and fix this; search/replace is cheap. Whereas until we need that generalization, explicitly hard-coded file names make the rules easier to read, and make it easier to find all targets that relate to a given file via a search for that file name. In general, "premature generalization" creates code that is tricky to test when the cases it applies to do not yet exist (did you test that your code actually runs correctly for multiple .rst files?) One can easily find that to go bug-hunting in an old overgeneralized implementation is more expensive than to follow the natural implement-test-cycle for generalizing code into a new situation that has appeared.

> 7. Short comments at the beginning of the new Makefiles would be helpful. We can use them as 'make help' text, getting two jobs done with one effort :)

I've added short comments to indicate the purposes. I haven't fully documented all targets here; to move that info from the top makefile is the task for someone implementing the distributed help system.

> 8. Would you mind adding `#--*- Makefile -*--` to the beginning of each new Makefile so that Emacs gets the correct mode? Modes for the 'vi' are welcome as well :)

Done.

> 9. Would you mind having '.PHONY:' declarations before the actual target definitions, not after them? The Makefile is then easier to read, since you know what to expect; this is also the way other Make* files are written in the tests/makefiles/ directory.

Done. (I've seen others argue this, so I'm ok with following that "standard". But personally I somewhat fail to see the point. It is usually very clear to a human which LHS are just labels and which are files, and then it is better to get .PHONY out of the way to not distract from the rule definition.)

> 10: Would you mind if we fold lines that they are not longer than 80 chars? 

Done, I think. Let me know if you see somewhere I've missed. (I don't think it is always possible though, you have a few lines longer than 80 chars.)

Please re-review!

rartino:
@merkys, @fekad, @CasperWA Feel free to ignore the long discussion and just look at the file changes. I think there should be no question that this improves the present state of the repo with some helpful targets and the documentation at the top of the main makefile.

It would be particularly helpful if @fekad as the original author of the docker config can check that `make docker_audit` now works on his mac setup. (Still, if he is not around, I don't think it is a dealbreaker; as soon as someone reports the target non-working on their setup we can handle it as a bug.)

tests/makefiles/Makelocal-audit
```diff
@@ -0,0 +1,31 @@
...
+audit_grammars: 
+	$(MAKE) tools
+	$(MAKE) grammars
+	$(MAKE) tests
+	(exit $$($(MAKE) -s listdiff | head -n 1 | wc -l))
+
+
+.PHONY: audit_authors
+
+audit_authors: test_authors
```
merkys: This seems to duplicate target `test_authors` in `Makelocal-authors`. Is there a reason for it?

tests/makefiles/Makelocal-audit
```diff
@@ -0,0 +1,31 @@
...
+
+audit: audit_grammars audit_authors audit_spelling
+	@echo "=============================================="
+	@echo "Audit of specification completed successfully."
+	@echo "=============================================="
+
+
+.PHONY: audit_grammars
+
+audit_grammars:
```
merkys: Is there a specific reason to have `$(MAKE) <target>` instead of `audit_grammars: <target>`?

tests/makefiles/Makelocal-audit
rartino: They are different. In Makelocal-audit:
```
audit_authors: test_authors
	diff -q AUTHORS tests/generated/AUTHORS.sorted
```
and in Makelocal-authors:
```
test_authors: tests/generated/AUTHORS.sorted
	diff -u AUTHORS tests/generated/AUTHORS.sorted || true
```
The audit target prints nothing and exits with a non-zero return code if the files are different, the test targets prints the difference but always returns 0. The different handling of the return code between audit and test targets is documented at the top of the main GNUmakefile.

tests/makefiles/Makelocal-audit
rartino: I wanted to be absolutely sure to replicate *exactly* the behavior of the previous way to validate the grammar without fully understanding it, so that this PR does not break anything.

If you and @sauliusg say that the testing done by the present `audit_grammars` target is exactly equivalent with:
```
audit_grammars: tools grammars tests
	(exit $$($(MAKE) -s listdiff | head -n 1 | wc -l))
```
then I'm happy to change it. 

But, as you see, we will not get rid of `$(MAKE)` anyway since it is needed to process the output of the listdiff target, as that is how audit of the grammar works before this PR. 

However, IMO this processing of listdiff output isn't a good solution, since (as we've discussed upthread) the 'making of makefiles' feature means targets are not in full control of their output. The only reason the above works reliably is because we have meta-knowledge of the build system and know that four invocations of make in a row without changing source files is unlikely to produce output unrelated to the lisftdiff target. Still, I'd argue that relying on this is inelegant, and a better solution is to duplicate the find command in listdiff also under audit_grammars. If you prefer that, let us change it into this (but maybe in a separate PR?)

tests/makefiles/Makelocal-audit
merkys: OK, the duplicated code is very short. Deduplication can take place later.

tests/makefiles/Makelocal-audit
merkys: OK, let's leave it as is for now, with a prospect to refactor later.

PR #201, opened 2019-11-14
Renaming references.type (attribute) to references.bib_type

### LINKED ISSUE #200: References attribute "type"
For filter querying, it is a problem that we allow reference entries to have an attribute called `type`, which is equal to the BibTeX meaning of `type`.

This should either be changed to `bib_type` (or similar) or not be supported at all, as it clashes with the existing "top-level" `type` field that resources MUST have.

merkys:
Closes #200.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1927,7 +1927,7 @@ The following properties are used to provide the bibliographic details:
 
 - **address**, **annote**, **booktitle**, **chapter**, **crossref**, **edition**, **howpublished**, **institution**, **journal**, **key**, **month**, **note**, **number**, **organization**, **pages**, **publisher**, **school**, **series**, **title**, **volume**, **year**: meanings of these properties match the `BibTeX specification <http://bibtexml.sourceforge.net/btxdoc.pdf>`__, values are strings;
 
-- **bibtype**: type of the reference, corresponding to **type** property in the BibTeX specification;
+- **bibtype**: type of the reference, corresponding to **type** property in the BibTeX specification, value is string;
 
 - **authors** and **editors**: lists of *person objects* which are dictionaries with the following keys:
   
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1925,8 +1925,10 @@ References Entries
 The :entry:`references` entries describe bibliographic references.
 The following properties are used to provide the bibliographic details:
 
-- **address**, **annote**, **booktitle**, **chapter**, **crossref**, **edition**, **howpublished**, **institution**, **journal**, **key**, **month**, **note**, **number**, **organization**, **pages**, **publisher**, **school**, **series**, **title**, **type**, **volume**, **year**: Meanings of these properties match the `BibTeX specification <http://bibtexml.sourceforge.net/btxdoc.pdf>`__, values are strings;
-  
+- **address**, **annote**, **booktitle**, **chapter**, **crossref**, **edition**, **howpublished**, **institution**, **journal**, **key**, **month**, **note**, **number**, **organization**, **pages**, **publisher**, **school**, **series**, **title**, **volume**, **year**: meanings of these properties match the `BibTeX specification <http://bibtexml.sourceforge.net/btxdoc.pdf>`__, values are strings;
+
+- **bibtype**: type of the reference, corresponding to **type** property in the BibTeX specification;
```
CasperWA: It may be prudent to use the same logic as we have done for all query parameters that are not taken directly from JSON v1.0, i.e. adding a `_`.
So in this context it would be calling it `bib_type`, using the added `_` to also make it more readable.
What do you think?

Also, for completion the value type should be added. Suggested change:

```suggestion
- **bib_type**: type of the reference given as a string, corresponding to **type** property in the BibTeX specification;
```

merkys: Done in 260e337 and 87d7c9a. I don't think we have to care about property names having `_`, as these will never be used as query parameters, AFAIK.

CasperWA: > Done in [260e337](https://github.com/Materials-Consortia/OPTiMaDe/commit/260e33717b7f93638431744f7ddda708fa472ecc) and [87d7c9a](https://github.com/Materials-Consortia/OPTiMaDe/commit/87d7c9a44d65c9c9727b2fb793e262cf54853c07). I don't think we have to care about property names having `_`, as these will never be used as query parameters, AFAIK.

For sure, it was only to do a similar thing as was done with the query parameters :)

PR #202, opened 2019-11-15
Update and clarify sort and sortable key

CasperWA:
Expand on the `sort` parameter definition and add the mentioning of the field `sortable` to the definition of what can be in `/info/<entries>` properties' dictionaries.

The most major thing here (in my opinion) is the clarification that if sorting is possible for an entry endpoint, the `sortable` key MUST be present for all fields under `data.properties.<field>` at `/info/<entries>` and _can_ be either `true` or `false`.
Mainly this is because I am unsure how to sort on lists/arrays, and would like to flag these as NOT sortable.

A minor, but important, addition is the mentioning of `sortable` under what can be in the `data.properties.<field>` dictionary.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -630,7 +630,7 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
   If it does, it again MUST conform to the JSON API 1.0 specification.
 
   If an implementation supports sorting for an `entry listing endpoint <Entry Listing Endpoints_>`_, then the :endpoint:`/info/<entries>` endpoint MUST include, for each field name :field:`<fieldname>` in its :field:`data.properties.<fieldname>` response value, the key :field:`sortable` with value :field-val:`true` or :field-val:`false`.
-  The set of :field:`<fieldname>` s, with :field:`sortable` equal to :field-val:`true` make up the "sort fields" set according to its definition in the JSON API 1.0 specification.
+  The set of :field:`<fieldname>` s, with :field:`sortable` equal to :field-val:`true` are allowed to be used in the "sort fields" list according to its definition in the JSON API 1.0 specification.
   The field :field:`sortable` is in addition to each property description (and optional unit).
   An example is shown in section `Entry Listing Info Endpoints`_.
 
@@ -921,33 +921,33 @@ Example:
 
 .. code:: jsonc
 
-     {
-       "data": {
-	 "description": "a structures entry",
-	 "properties": {
-	   "nelements": {
-	     "description": "Number of elements",
-	     "sortable": true
-	   },
-	   "lattice_vectors": {
-	     "description": "Unit cell lattice vectors",
-	     "unit": "Å",
-       "sortable": false
-	   }
-	   // ... <other property descriptions>
-	 },
-	 "formats": ["json", "xml"],
-	 "output_fields_by_format": {
-	   "json": [
-	     "nelements",
-	     "lattice_vectors",
-	     // ...
-	   ],
-	   "xml": ["nelements"]
-	 }
-       }
-       // ...
-     }
+    {
+      "data": {
+        "description": "a structures entry",
+        "properties": {
+          "nelements": {
+            "description": "Number of elements",
+            "sortable": true
+          },
+          "lattice_vectors": {
+            "description": "Unit cell lattice vectors",
+            "unit": "Å",
+            "sortable": false
+          }
+          // ... <other property descriptions>
+        },
+        "formats": ["json", "xml"],
+        "output_fields_by_format": {
+          "json": [
+            "nelements",
+            "lattice_vectors",
+            // ...
+          ],
+          "xml": ["nelements"]
+        }
+      }
+      // ...
+    }
 
 Links Endpoint
 --------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -629,7 +629,8 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
   An implementation MAY support multiple sort fields for a single query.
   If it does, it again MUST conform to the JSON API 1.0 specification.
 
-  If an implementation supports sorting for an `entry listing endpoint <Entry Listing Endpoints_>`_, then the :endpoint:`/info/<entries>` endpoint MUST include, for each field name :field:`<fieldname>` in its :field:`data.properties.<fieldname>` response value, the key :field:`sortable` with value :field-val:`true` or :field-val:`false`.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -627,10 +627,11 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
...
 
   An implementation MAY support multiple sort fields for a single query.
-  If it does, it again MUST conform to the JSON API 1.0 spec.
+  If it does, it again MUST conform to the JSON API 1.0 specification.
 
-  If an implementation supports sorting for an `entry listing endpoint <Entry Listing Endpoints_>`_, then the :endpoint:`/info/<entries>` endpoint MUST include, for each field name :field:`<fieldname>` in its :field:`data.properties.<fieldname>` response value, the key :field:`sortable` with value :field-val:`true`.
-  This is in addition to each property description (and optional unit).
+  If an implementation supports sorting for an `entry listing endpoint <Entry Listing Endpoints_>`_, then the :endpoint:`/info/<entries>` endpoint MUST include, for each field name :field:`<fieldname>` in its :field:`data.properties.<fieldname>` response value, the key :field:`sortable` with value :field-val:`true` or :field-val:`false`.
+  The set of :field:`<fieldname>` s, with :field:`sortable` equal to :field-val:`true` make up the "sort fields" set according to its definition in the JSON API 1.0 specification.
```
merkys: It's the first time I've noticed "sort fields" in JSON API specs (thanks!), but to me "sort fields" seem to refer to a specific value taken from query string. Thus, I'd change `make up the "sort fields" set` to `are allowed to be used in the "source fields" list` (`list` instead `set`, as the order is important). What do you think?

By the way, `<fieldname>s` look strange to me. Could we think of better way to stress the plural here?

optimade.rst
```diff
@@ -930,7 +931,8 @@ Example:
 	   },
 	   "lattice_vectors": {
 	     "description": "Unit cell lattice vectors",
-	     "unit": "Å"
+	     "unit": "Å",
+       "sortable": false
```
merkys: Indentation seem off.

rartino:
It is good that `sortable` is added to the example, but I don't understand the rationale for this:

> The most major thing here (in my opinion) is the clarification that if sorting is possible for an entry endpoint, the sortable key MUST be present for all fields under data.properties.<field> at /info/<entries> and can be either true or false. Mainly this is because I am unsure how to sort on lists/arrays, and would like to flag these as NOT sortable.

I don't get why this is necessary? The description now clearly means that an omitted `sortable` means the same thing as `sortable: false`. If I only support sort on a few fields, it just seems unnecessary to force me to include `sortable: false` on *everything else* in my database. But, with the old formulation, if your implementation *prefers* to do it, you are still allowed.

CasperWA:
> I don't get why this is necessary? The description now clearly means that an omitted `sortable` means the same thing as `sortable: false`. If I only support sort on a few fields, it just seems unnecessary to force me to include `sortable: false` on _everything else_ in my database. But, with the old formulation, if your implementation _prefers_ to do it, you are still allowed.

I see your point. I guess my rationale comes from looking at this through the glasses of a client, where simply having the `sortable` key, implies the entry point is sortable (for some keys), while it is not sortable at all if I cannot find a `sortable` key for _any_ property.

optimade.rst
CasperWA: > It's the first time I've noticed "sort fields" in JSON API specs (thanks!), but to me "sort fields" seem to refer to a specific value taken from query string. Thus, I'd change `make up the "sort fields" set` to `are allowed to be used in the "source fields" list` (`list` instead `set`, as the order is important). What do you think?

Sounds great. I believe you're right that "sort fields" refer to the specified values in a query string. And concerning `list` vs. `set`, I completely concur.

> By the way, `<fieldname>s` look strange to me. Could we think of better way to stress the plural here?

Hmm, I could just leave out the brackets or use `properties` instead, but I don't much like that, since `properties` do not have a `sortable` key. We could just use the defined `Field` and go with that - in plural? But then the previous sentence should be altered as well, to have consistency.

merkys: > Hmm, I could just leave out the brackets or use `properties` instead, but I don't much like that, since `properties` do not have a `sortable` key. We could just use the defined `Field` and go with that - in plural? But then the previous sentence should be altered as well, to have consistency.

Maybe just `field names`? I.e.
```
The set of field names, with :field:`sortable` equal to :field-val:`true` ...
```

CasperWA: Would initially say that this would introduce an inconsistency, however that is not true. Looking more closely at the previous sentence, `field names` is indeed already used.
So agreed :)

rartino:
> simply having the sortable key, implies the entry point is sortable (for some keys), while it is not sortable at all if I cannot find a sortable key for any given property.

If I understand you correctly, you are saying that you want clients to be able to use the presence or absence of the `sortable` key under any property as a way to deduce whether "sort is supported"?

I think this promotes a fragile implementation, which should not be supported. In general, for boolean indicator fields like this, it just is more expected, and robust, to let the absence of a key mean the exact same thing as `false`.

In this specific case, assigning this type of meaning to `sortable: false` would make a tricky distinction between "sort is supported, but no properties the client is aware of can be used to sort the output" (i.e., all properties return `sortable: false`) and "sort is not supported" (no `sortable` key present). I can imagine edge cases with sortable database-specific fields that are temporarily included for testing where it gets non-trivial to know whether `sortable: false` or no `sortable` key should be returned for other properties. 

So, IMO, the algorithm that should be used by clients is: if any one property (that the client is aware of) contains `sortable: true`, then sort is supported. The client should anyway almost always want to tabulate the keys that can be used for sorting at the same time as it determines if sorting is supported.

CasperWA:
@rartino I agree with your assessment of it being a fragile implementation. I will look at what I have and adjust it accordingly. Thanks for taking the time to consider this properly :)

CasperWA:
@rartino and @merkys, I have added a new sentence describing what to do with a field name that cannot be used for sorting under an entry listing endpoint that supports sorting. Please feel free to review again :)

CasperWA:
> (...) /info/ endpoints only support json output (right?) so there is no point in trying to use output-format-agnostic language specifically for that output (right?).

Glancing through the specification, I cannot find a place where we specifically say that `/info` endpoints _only_ support JSON output. However, we may state in more subtle ways.
E.g., in the base info response, we have the field `entry_types_by_format`, which is the only field truly connecting up endpoints and output formats. Since this is _has_ to be entry type endpoints (i.e., not `/info` endpoints), then we are basically saying that a server has no other choice but to convey `/info` endpoints in our "default" format, i.e. JSON.

rartino:
@CasperWA what you are saying matches my conclusions. We *could* say it more explicitly, but as far as I can tell, nothing is in error with what we have. I also don't think it would make sense for anything but entry endpoints to deliver data on other formats.

PR #205, opened 2019-11-22
Relieve the requirement level of query support for 'nsites' from MUST to OPTIONAL

merkys:
In the COD each structure is represented by an asymmetric unit + list of symmetry operators. One has to perform full symmetry reconstruction to arrive to the list of all atoms in cell, and this operation is quite intensive. Therefore both listing and querying on atoms and their positions is infeasible in the COD (more details in #184). Thus I would request relieving the requirement levels for querying on such structures. This PR addresses the `nsites`, the only coordinate-based property on which querying `MUST` be supported as of now.

**Edit**: This PR is also included in PR #210.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1696,7 +1696,7 @@ nsites
 - **Requirements/Conventions**:
     
   - **Response**: REQUIRED in the response unless explicitly excluded.
-  - **Query**: SHOULD be a queryable property with support for all mandatory filter operators.
+  - **Query**: Support for queries on this property is OPTIONAL.
     
 - **Examples**:
   
--- a/optimade.rst
+++ b/optimade.rst
@@ -1696,7 +1696,7 @@ nsites
 - **Requirements/Conventions**:
     
   - **Response**: REQUIRED in the response unless explicitly excluded.
-  - **Query**: MUST be a queryable property with support for all mandatory filter operators.
+  - **Query**: SHOULD be a queryable property with support for all mandatory filter operators.
     
 - **Examples**:
   


rartino:
I think this PR could just as well have been handled in #203. They are closely related, and what we agree on affect both of them, the discussion just gets spread out and more difficult to follow. If you agree, you can close this and move the changes over to #203.

Anyway, I'm opposed to having a SHOULD here. If we agree that a CIF-based representative atoms-representation is "as common" as the filled cell one, I think we should fully relax the requirement for filtering on `nsites`.

In short, to avoid juggling a lot of possible rules and special cases for when things should be included, filterable, etc., I propose there are only two possibilities:
1. A few properties are included "by default" and MUST be filterable.
2. All other properties are not included "by default" and filtering is OPTIONAL.

merkys:
> I think this PR could just as well have been handled in #203. They are closely related, and what we agree on affect both of them, the discussion just gets spread out and more difficult to follow. If you agree, you can close this and move the changes over to #203.

Agree.

> In short, to avoid juggling a lot of possible rules and special cases for when things should be included, filterable, etc., I propose there are only two possibilities:
> 
>     1. A few properties are included "by default" and MUST be filterable.
> 
>     2. All other properties are not included "by default" and filtering is OPTIONAL.

This seems quite a big change to me, and while I myself agree to it, I'd like to hear opinions from the other implementations. I'll probably open a separate PR implementing your suggestion, and ask for comments on it during the following Web meeting.

CasperWA:
> > I think this PR could just as well have been handled in #203. They are closely related, and what we agree on affect both of them, the discussion just gets spread out and more difficult to follow. If you agree, you can close this and move the changes over to #203.
> 
> Agree.

Please do this, as I am indeed having trouble following up on which PRs should be handled before each other.
Since we are not having a single commit per PR / per concept edit, it also doesn't matter if a PR "blows up" a bit, so long as it leaves a self-consistent spec after merging.

merkys:
@CasperWA This PR is now included in #210.

rartino:
@merkys is it ok to close this for now to keep the list of open PRs manageable? If #210 is rejected, we can re-open it again?

merkys:
> @merkys is it ok to close this for now to keep the list of open PRs manageable? If #210 is rejected, we can re-open it again?

Agree. Closing this PR for now.

Issue #204, opened 2019-11-20
Clarify response fields for default properties in spec.


CasperWA:
While I am in full support for the overall changes; after making these changes we are not so helpful on the underlying intent with what properties get included or not. The reader is likely to end up with the feeling that one needs to carefully consider each individual `Response` point in Sec 6 to decipher arbitrary conditions of when things should be included or not. 

In reality, our rules are very straightforward:
- If `response_fields` is not present: return only a small set of default properties.
- If `response_fields` is present, return only those fields requested.

I think the right place to say this is early in section 4.1, with maybe a callback from 4.2 to say that the same thing applies for single entry endpoints. Or possibly a new subsection. I started to formulate a suggestion, but do not have time to finish it now. (I'll get back to it eventually if no one else makes an attempt.)

_Originally posted by @rartino in https://github.com/Materials-Consortia/OPTiMaDe/pull/203#issuecomment-555778396_

PR #206, opened 2019-11-22
Property 'fractional_site_positions'

### LINKED ISSUE #196: Format of 'lattice_vectors' is unclear for 2D, 1D and 0D systems
Specification defines `lattice_vectors` as three lattice vectors in Cartesian coordinates. However, it is not clear how the vectors are defined for nonperiodic dimensions. Currently it seems that such vectors are ignored, and as such can contain any values. But it would probably make sense to have a default, and I would suggest having `null`s for all the coordinates of vectors in nonperiodic dimensions.

An example of a 2D surface on XZ plane:
```
...
  "dimension_types": [1, 0, 1],
  "lattice_vectors": [[4, 0, 0], [null, null, null], [0, 0, 4]],
...
```

merkys:
In #196 @rartino and I discussed the need to have a `structures` property for reduced (a.k.a. fractional) atom coordinates. This PR introduces `fractional_site_positions` among the standardized properties. I have copied the most of the text from `cartesian_site_positions` and slightly adjusted the relationships of the two coordinate-based properties. I have also changed a bit the definition of `lattice_vectors`, thus this PR fixes #196.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1656,7 +1656,7 @@ lattice\_vectors
 - **Type**: list of list of floats.
 - **Requirements/Conventions**:
 
-  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` (in this case it is OPTIONAL).
+  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` and `fractional_site_positions`_ are not provided (in this case it is OPTIONAL).
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
   - MUST be a list of three vectors *a*, *b*, and *c*, where each of the vectors MUST BE a list of the vector's coordinates along the x, y, and z Cartesian coordinates.
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
--- a/optimade.rst
+++ b/optimade.rst
@@ -1703,7 +1703,7 @@ fractional\_site\_positions
   - If a component of the position is unknown, the :val:`null` value should be provided instead (see section `Properties with unknown value`_).
     Otherwise, it should be a float value.
     If at least one of the coordinates is unknown, the correct flag in the list property `structure_features`_ MUST be set.
-  - The order of the list MUST match the order of `cartesian_site_positions`_.
+  - The order of the list MUST match the order of `cartesian_site_positions`_, i.e., the list refers to the same positions as `cartesian_site_positions` *in the exact same order*, only expressed in fractional coordinates instead of cartesian.
 
 - **Examples**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1698,7 +1698,7 @@ fractional\_site\_positions
 
   - **Response**: OPTIONAL in the response.
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
-  - It MUST be a list of length N times 3, where N is the number of sites in the structure.
+  - It MUST be a list of length equal to the number of sites in the structure where every element is a list of the three fractional coordinates of a site.
   - An entry MAY have multiple sites at the same fractional position (for a relevant use of this, see e.g., the property `assemblies`_).
   - If a component of the position is unknown, the :val:`null` value should be provided instead (see section `Properties with unknown value`_).
     Otherwise, it should be a float value.
--- a/optimade.rst
+++ b/optimade.rst
@@ -1710,7 +1710,7 @@ fractional\_site\_positions
 nsites
 ~~~~~~
 
-- **Description**: An integer specifying the length of the :property:`cartesian_site_positions` property.
+- **Description**: An integer specifying the length of :property:`cartesian_site_positions` and  :property:`fractional_site_positions` properties.
 - **Type**: integer  
 - **Requirements/Conventions**:
     
--- a/optimade.rst
+++ b/optimade.rst
@@ -1688,6 +1688,25 @@ cartesian\_site\_positions
   
   - :val:`[[0,0,0],[0,0,2]]` indicates a structure with two sites, one sitting at the origin and one along the (positive) *z*-axis, 2 Å away from the origin.
 
+fractional\_site\_positions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Positions of each site expressed as fractions of the `lattice_vectors`_. Description of a site matches the one given in `cartesian_site_positions`_.
+- **Type**: list of list of floats and/or unknown values
+- **Requirements/Conventions**:
+
+  - **Response**: OPTIONAL in the response.
+  - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
+  - It MUST be a list of length N times 3, where N is the number of sites in the structure.
+  - An entry MAY have multiple sites at the same fractional position (for a relevant use of this, see e.g., the property `assemblies`_).
+  - If a component of the position is unknown, the :val:`null` value should be provided instead (see section `Properties with unknown value`_).
+    Otherwise, it should be a float value.
+    If at least one of the coordinates is unknown, the correct flag in the list property `structure_features`_ MUST be set.
+
+- **Examples**:
+
+  - :val:`[[0,0,0],[0.5,0.5,0.5]]` indicates a structure with two sites, one sitting at the origin and one at the center of the unit cell.
+
 nsites
 ~~~~~~
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1653,7 +1653,7 @@ lattice\_vectors
 ~~~~~~~~~~~~~~~~
 
 - **Description**: The three lattice vectors in Cartesian coordinates, in ångström (Å).
-- **Type**: list of list of floats.
+- **Type**: list of list of floats and/or unknown values
 - **Requirements/Conventions**:
 
   - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` and `fractional_site_positions`_ are not provided (in this case it is OPTIONAL).
@@ -1662,6 +1662,7 @@ lattice\_vectors
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
   - For databases that do not define an absolute Cartesian system (e.g., only defining the length and angles between vectors), the first lattice vector SHOULD be set along *x* and the second on the *xy*-plane.
   - This property MUST be an array of dimensions 3 times 3 regardless of the elements of property `dimension_types`_. The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero. The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
+  - If `fractional_site_positions`_ are provided, the vectors in non-periodic directions MUST NOT be expressed as unknown values. Otherwise they MAY be expressed as `[null, null, null]`.
     
 - **Examples**:
     
--- a/optimade.rst
+++ b/optimade.rst
@@ -1656,7 +1656,7 @@ lattice\_vectors
 - **Type**: list of list of floats and/or unknown values
 - **Requirements/Conventions**:
 
-  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` and `fractional_site_positions`_ are not provided (in this case it is OPTIONAL).
+  - **Response**: REQUIRED in the response unless explicitly excluded.
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
   - MUST be a list of three vectors *a*, *b*, and *c*, where each of the vectors MUST BE a list of the vector's coordinates along the x, y, and z Cartesian coordinates.
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
--- a/optimade.rst
+++ b/optimade.rst
@@ -1662,7 +1662,7 @@ lattice\_vectors
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
   - For databases that do not define an absolute Cartesian system (e.g., only defining the length and angles between vectors), the first lattice vector SHOULD be set along *x* and the second on the *xy*-plane.
   - This property MUST be an array of dimensions 3 times 3 regardless of the elements of property `dimension_types`_. The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero. The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
-  - If `fractional_site_positions`_ are provided, the vectors in non-periodic directions MUST NOT be expressed as unknown values. Otherwise they MAY be expressed as `[null, null, null]`.
+  - :val:`null` values are allowed in the output. A vector of :val:`[null, null, null]` MAY be used for non-periodic directions where a numerical vector would not be meaningful. However, if the implementation also supports the `fractional_site_positions`_ property, the lattice vectors are relevant also for the non-periodic directions, and thus no all-`null` vectors are expected in the output. 
     
 - **Examples**:
     
--- a/optimade.rst
+++ b/optimade.rst
@@ -1703,6 +1703,7 @@ fractional\_site\_positions
   - If a component of the position is unknown, the :val:`null` value should be provided instead (see section `Properties with unknown value`_).
     Otherwise, it should be a float value.
     If at least one of the coordinates is unknown, the correct flag in the list property `structure_features`_ MUST be set.
+  - The order of the list MUST match the order of `cartesian_site_positions`_.
 
 - **Examples**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1931,7 +1931,7 @@ structure\_features
   - **List of strings used to indicate special structure features**:
 
     - :val:`disorder`: This flag MUST be present if any one entry in the :property:`species` list has a :property:`chemical_symbols` list that is longer than 1 element.
-    - :val:`unknown_positions`: This flag MUST be present if at least one component of the :property:`cartesian_site_positions` list of lists has value :val:`null`.
+    - :val:`unknown_positions`: This flag MUST be present if at least one component of the :property:`cartesian_site_positions` and :property:`fractional_site_positions` lists of lists has value :val:`null`.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1653,15 +1653,16 @@ lattice\_vectors
 ~~~~~~~~~~~~~~~~
 
 - **Description**: The three lattice vectors in Cartesian coordinates, in ångström (Å).
-- **Type**: list of list of floats.
+- **Type**: list of list of floats and/or unknown values
 - **Requirements/Conventions**:
 
-  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` (in this case it is OPTIONAL).
+  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` and `fractional_site_positions`_ are not provided (in this case it is OPTIONAL).
```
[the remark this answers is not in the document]
merkys: > But before we get into rewriting it to be sound logically; as we hande #203, #204, maybe we should not have any requirements like this at all? I.e., should we not simply say that `lattice_vectors` is always required?

We need to account for the case of structures in nonperiodic setting, i.e., when only Cartesian coordinates are available, and lattice vectors do not make sense. But then null vectors could be provided, so it shouldn't be a problem.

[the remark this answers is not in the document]
merkys: Made `lattice_vectors` always required in fb7c6f3.

merkys:
> The way you have implemented this clearly requires the `cartisian_site_positions` and `fractional_site_positions` to be in the exact same order - line-by-line referring to the same coordinates. I think that requirement should be written down somewhere; I suggest in the list of requirements on `fractional_site_positions`.

Requirement added in 8eefdd2.

optimade.rst
```diff
@@ -1653,15 +1653,16 @@ lattice\_vectors
...
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
   - MUST be a list of three vectors *a*, *b*, and *c*, where each of the vectors MUST BE a list of the vector's coordinates along the x, y, and z Cartesian coordinates.
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
   - For databases that do not define an absolute Cartesian system (e.g., only defining the length and angles between vectors), the first lattice vector SHOULD be set along *x* and the second on the *xy*-plane.
   - This property MUST be an array of dimensions 3 times 3 regardless of the elements of property `dimension_types`_. The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero. The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
+  - If `fractional_site_positions`_ are provided, the vectors in non-periodic directions MUST NOT be expressed as unknown values. Otherwise they MAY be expressed as `[null, null, null]`.
```
rartino: Is this maybe a bit too heavy-handed in forbidding `null` values? We are generally very allowing for `null` in cases when, e.g., values are not known. Also, perhaps the text can be slightly more helpful in explaining the interference with `fractional_site_positions`. Here is an attempt:
```suggestion
  - :val:`null` values are allowed in the output. A vector of :val:`[null, null, null]` MAY be used for non-periodic directions where a numerical vector would not be meaningful. However, if the implementation also supports the `fractional_site_positions`_ property, the lattice vectors are relevant also for the non-periodic directions, and thus no all-`null` vectors are expected in the output. 
```

optimade.rst
```diff
@@ -1688,10 +1689,30 @@ cartesian\_site\_positions
...
+- **Requirements/Conventions**:
+
+  - **Response**: OPTIONAL in the response.
+  - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
+  - It MUST be a list of length N times 3, where N is the number of sites in the structure.
+  - An entry MAY have multiple sites at the same fractional position (for a relevant use of this, see e.g., the property `assemblies`_).
+  - If a component of the position is unknown, the :val:`null` value should be provided instead (see section `Properties with unknown value`_).
+    Otherwise, it should be a float value.
+    If at least one of the coordinates is unknown, the correct flag in the list property `structure_features`_ MUST be set.
+  - The order of the list MUST match the order of `cartesian_site_positions`_.
```
rartino: Somehow I managed to submit my review without the comment I had written about this line:

I'd also like the order statement to be as clear as possible.
```suggestion
  - The order of the list MUST match the order of `cartesian_site_positions`_, i.e., the list refers to the same positions as `cartesian_site_positions` *in the exact same order*, only expressed in fractional coordinates instead of cartesian.
```

merkys:
> As I've written in other issues/commits, I very much like the idea of standardizing `fractional_site_positions`. But I realize now with this PR in hand that if we do it this way, `fractional_site_positions` and `cartesian_site_positions` become "equal citizens". I.e., it would be OK for an implementation to support only the `fractional_site_positions` part of the specification and refuse to deal with `cartesian_site_positions`. This is a problem from a standards-perspective. Can we solve that somehow?

In this PR `cartesian_site_positions` are defined as REQUIRED, and `fractional_site_positions` as OPTIONAL. However, if PRs #203 or #210 are accepted, both properties will become OPTIONAL with no requirement on their support.

Now I'm starting to understand the importance of `REQUIRED in the response` constraint. In the light of PRs #203 and #210 this constraint should be transformed to `REQUIRED to be supported`, that is, a property has to be required to return meaningful values upon request.

optimade.rst
merkys: Agree.

optimade.rst
merkys: Sounds good, thanks for suggestion.

optimade.rst
```diff
@@ -1688,10 +1689,30 @@ cartesian\_site\_positions
   
   - :val:`[[0,0,0],[0,0,2]]` indicates a structure with two sites, one sitting at the origin and one along the (positive) *z*-axis, 2 Å away from the origin.
 
+fractional\_site\_positions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Positions of each site expressed as fractions of the `lattice_vectors`_. Description of a site matches the one given in `cartesian_site_positions`_.
```
sauliusg: As it is currently formulated, `fractional_site_positions` and `cartesian_site_positions` have different status: `cartesian` is the primary definition, and both `species_at_sites` and `fractional...` refer to `cartesian...`. However, mathematically, `cartesian...` and `fractional...` are equivalent descriptions, and either of them or even both may be absent. It then makes this description ambiguous, since, if `cartesian_site_positions` are absent, it is not clear what the `fractional_site_positions` must match.

It seems therefore that it would be more consistent to describe `species_at_site` as a primary definition of an atomic site, and to let both `fractional_site_positions` and `cartesian_site_positions` refer to the `species_at_site` when the order and number of array elements is defined for each `fractional` and `cartesian`.

optimade.rst
```diff
@@ -1688,10 +1689,30 @@ cartesian\_site\_positions
...
+- **Type**: list of list of floats and/or unknown values
+- **Requirements/Conventions**:
+
+  - **Response**: OPTIONAL in the response.
+  - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
+  - It MUST be a list of length N times 3, where N is the number of sites in the structure.
```
sauliusg: This is a mistake – both `fractional_site_positions` and `cartesian_site_positions` are **not** lists of N*3 elements, but rather lists of N elements where each element is a list of length 3. Andrius is writing an issue/PR for this fix for `cartesian...`, should be fixed here as well.

optimade.rst
```diff
@@ -1710,14 +1731,14 @@ nsites
 species\_at\_sites
 ~~~~~~~~~~~~~~~~~~
 
-- **Description**: Name of the species at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_).
+- **Description**: Name of the species at each site (where values for sites are specified with the same order of the properties `cartesian_site_positions`_ and `fractional_site_positions`_).
   The properties of the species are found in the property `species`_.
```
sauliusg: I suggest that `species_at_sites` are moved to the beginning of the specification, and `fractional_site_positions` and `cartesian_size_positions` should both refer to the `species_at_sites` section.

optimade.rst
```diff
@@ -1710,14 +1731,14 @@ nsites
...
 - **Type**: list of strings.
 - **Requirements/Conventions**:
   
   - **Response**: REQUIRED in the response unless explicitly excluded.
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
-  - MUST have length equal to the number of sites in the structure (first dimension of the list property `cartesian_site_positions`_).
+  - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions`_ and `fractional_site_positions`_).
```
sauliusg: When `species_at_site` are moved to the front, the `fractional_site_positions` should the refer to `species_at_sites`, not vice versa.

optimade.rst
```diff
@@ -1688,10 +1689,30 @@ cartesian\_site\_positions
...
+- **Requirements/Conventions**:
+
+  - **Response**: OPTIONAL in the response.
+  - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
+  - It MUST be a list of length N times 3, where N is the number of sites in the structure.
+  - An entry MAY have multiple sites at the same fractional position (for a relevant use of this, see e.g., the property `assemblies`_).
+  - If a component of the position is unknown, the :val:`null` value should be provided instead (see section `Properties with unknown value`_).
+    Otherwise, it should be a float value.
+    If at least one of the coordinates is unknown, the correct flag in the list property `structure_features`_ MUST be set.
+  - The order of the list MUST match the order of `cartesian_site_positions`_, i.e., the list refers to the same positions as `cartesian_site_positions` *in the exact same order*, only expressed in fractional coordinates instead of cartesian.
```
sauliusg: The order of elements in both `fractional_site_positions` and `cartesian_site_positions` should refer to site descriptions, not to each other; the only suitable property that we specify currently is `species_at_sites`.

optimade.rst
```diff
@@ -1688,10 +1689,30 @@ cartesian\_site\_positions
...
   - :val:`[[0,0,0],[0,0,2]]` indicates a structure with two sites, one sitting at the origin and one along the (positive) *z*-axis, 2 Å away from the origin.
 
+fractional\_site\_positions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Positions of each site expressed as fractions of the `lattice_vectors`_. Description of a site matches the one given in `cartesian_site_positions`_.
+- **Type**: list of list of floats and/or unknown values
+- **Requirements/Conventions**:
+
+  - **Response**: OPTIONAL in the response.
```
sauliusg: I suggest specifying that:

```
`fractional_site_positions` and `cartesian_site_positions` MAY be both present in the same response. 
If they are both present, they SHOULD describe the same points in space, i.e. cartesian coordinates 
of sites should be the same as fractional coordinates of the same sites with orthogonalisation 
convention, as expressed in `lattice_vectors`, applied.

cartesian_site_positions[isite][x,y,z] = 
     fractional_site_positions[isite][x,y,z] * latice_vectors[][];

where:
   -  isite is index of the site (isite = 0..length(species_at_site[])-1);
   -  '*' is understood in matrix multiplcation sense;
   - each lattiec_vectors[i] element (i=0..2) is the 
     corresponding lattice vector, \vec{a}, \vec{b} and \vec{c}.
   - both cartesian_site_positions[isite][x,y,z] and
     fractional_site_positions[isite][x,y,z] are row-vectors
     (not column-vectors).
```

rartino:
> Edit: Actually, I've run through the specification, and it seems that even properties that MUST be included in the response may still be included with nulls for values. Thus we nowhere require any properties to be universally supported. Even a JSON document with all properties having null values is valid according to the specification. Or am I missing something?

I have been concerned about this since we started allowing `null`-valued properties, and I think the question was raised on the last physical meeting if we really mean to allow returning `null` for a REQUIRED property.

However, with `null` being allowed for REQUIRED properties these two alternatives have a rather different result:
- For every type of data OPTiMaDe supports, there is only a single format to represent it. Implementations either provide data in that format, or return `null`. 
- OPTiMaDe allows the same data to be returned in *N* number of ways, e.g., `cartesian_site_positions` and `fractional_site_positions` and implementations can choose to return `null` for either one of them (or both).

In the first case, we still have interoperability for all cases where it makes sense. A database either does not support coordinates at all (in which case interoperability with a client that wants coordinates is not possible) or it supports coordinates via `cartesian_site_coordinates` in which case a client that wants coordinates gets them as expected.

In the second case, we set ourselves up for a possible mismatch. The client wants coordinates from `cartesian_site_positions`, but the database returns `null` for that, and only supports coordinates via `fractional_site_positions`. This breaks interoperability.

Hence, it is my firm belief that if we are to accept the present PR, there needs to be some facility that says that IF you support `fractional_site_positions`, you MUST support `cartesian_site_positions`, but not the other way around. I.e., which makes `cartesian_site_positions` "more standard" than `fractional_site_positions`.

With the realization that we are opening a can of worms by allowing more than one standardized way to represent data, perhaps we need to delay this PR beyond 1.0 and sort out exactly how this should be handled?

optimade.rst
```diff
@@ -1653,15 +1653,16 @@ lattice\_vectors
 ~~~~~~~~~~~~~~~~
 
 - **Description**: The three lattice vectors in Cartesian coordinates, in ångström (Å).
-- **Type**: list of list of floats.
+- **Type**: list of list of floats and/or unknown values
 - **Requirements/Conventions**:
 
-  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` (in this case it is OPTIONAL).
+  - **Response**: REQUIRED in the response unless explicitly excluded.
```
CasperWA: Ah, this is where the line is cut for PR #210. I believe we should attempt to merge this PR before #210 then...

optimade.rst
```diff
@@ -1653,15 +1653,16 @@ lattice\_vectors
...
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
   - For databases that do not define an absolute Cartesian system (e.g., only defining the length and angles between vectors), the first lattice vector SHOULD be set along *x* and the second on the *xy*-plane.
   - This property MUST be an array of dimensions 3 times 3 regardless of the elements of property `dimension_types`_. The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero. The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
+  - :val:`null` values are allowed in the output. A vector of :val:`[null, null, null]` MAY be used for non-periodic directions where a numerical vector would not be meaningful. However, if the implementation also supports the `fractional_site_positions`_ property, the lattice vectors are relevant also for the non-periodic directions, and thus no all-`null` vectors are expected in the output.
```
```suggestion
  - :val:`null` values are allowed in the output. A vector of :val:`[null, null, null]` MAY be used for non-periodic directions where a numerical vector would not be meaningful (including a zero vector). However, if the implementation also supports the `fractional_site_positions`_ property, the lattice vectors are relevant also for the non-periodic directions, and thus no all-`null` vectors are allowed in the output.
```

CasperWA: I guess a related question is, why are `null`-vectors necessary, when one can apply zero vectors? Unless zero vectors have a meaning different from `null`-vectors when using `fractional_site_positions`.

optimade.rst
```diff
@@ -1688,10 +1689,30 @@ cartesian\_site\_positions
...
+
+- **Examples**:
+
+  - :val:`[[0,0,0],[0.5,0.5,0.5]]` indicates a structure with two sites, one sitting at the origin and one at the center of the unit cell.
+
 nsites
 ~~~~~~
 
-- **Description**: An integer specifying the length of the :property:`cartesian_site_positions` property.
+- **Description**: An integer specifying the length of :property:`cartesian_site_positions` and  :property:`fractional_site_positions` properties.
```
CasperWA: As was pointed out by @sauliusg, this should refer instead _only_ to `species_at_sites`.

optimade.rst
merkys: Fixed in 03f408838b44184d1be8f1ad9cca9cb19d8f0667 with the same wording as in PR #216.

optimade.rst
```diff
@@ -1710,14 +1731,14 @@ nsites
 species\_at\_sites
 ~~~~~~~~~~~~~~~~~~
 
-- **Description**: Name of the species at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_).
+- **Description**: Name of the species at each site (where values for sites are specified with the same order of the properties `cartesian_site_positions`_ and `fractional_site_positions`_).
```
```suggestion
- **Description**: Name of the species at each site (where values for sites are specified with the same order of the properties `cartesian_site_positions`_ as well as `fractional_site_positions`_).
```

I discourage `and`, even though it is correct, simply due to the fuzzyness it may present for an implementer in terms of "should it be the sum of these properties then"?

optimade.rst
```diff
@@ -1912,7 +1933,7 @@ structure\_features
   - **List of strings used to indicate special structure features**:
 
     - :val:`disorder`: This flag MUST be present if any one entry in the :property:`species` list has a :property:`chemical_symbols` list that is longer than 1 element.
-    - :val:`unknown_positions`: This flag MUST be present if at least one component of the :property:`cartesian_site_positions` list of lists has value :val:`null`.
+    - :val:`unknown_positions`: This flag MUST be present if at least one component of the :property:`cartesian_site_positions` and :property:`fractional_site_positions` lists of lists has value :val:`null`.
```
```suggestion
    - :val:`unknown_positions`: This flag MUST be present if at least one component of the :property:`cartesian_site_positions` or :property:`fractional_site_positions` lists of lists has value :val:`null`.
```

ml-evs:
Did the consensus evolve on this at either of the last 2 workshops? If this is something we still cannot deal with, I would suggest closing this PR (even if only temporarily).

rartino:
> Did the consensus evolve on this at either of the last 2 workshops? If this is something we still cannot deal with, I would suggest closing this PR (even if only temporarily).

I don't think we yet have a resolution to my comment above:

----

> Hence, it is my firm belief that if we are to accept the present PR, there needs to be some facility that says that IF you support fractional_site_positions, you MUST support cartesian_site_positions, but not the other way around. I.e., which makes cartesian_site_positions "more standard" than fractional_site_positions.

----

But the PR is good otherwise, so isn't it better to mark it with a label that means that there is something fundamental that needs to be worked out, but leaving it open?

merkys:
I prefer keeping the PR open as well. Once closed, a PR becomes virtually invisible and it is easy to forget it, scrapping all the effort.

rartino:
It is my understanding that this is going to be superseded by `asym_fractional_coords` in a PR spawned by #416. (No one has so far suggested that we should include both a "filled cell" and an "asymetric unit" version of fractional coords).

merkys:
Closing in favor of solution proposed in https://github.com/Materials-Consortia/OPTIMADE/issues/416.

PR #210, opened 2019-11-25
Common rules for inclusion and querying

### LINKED ISSUE #181: Extend response_fields to support field exclusion
Currently `response_fields` query parameter allows the user to explicitly enumerate fields to be included in the response. I suggest a mechanism to explicitly exclude particular field(s) using the same `response_fields` query parameter.

Now `response_fields=a,b,c` means that the user requests response with only `a`, `b` and `c` fields. As field names cannot have `-` sign, `response_fields=-a` could be defined to stand for *all fields excluding `a`*, and `response_fields=-a,b,c` - *all fields excluding `a`, `b` and `c`* etc.

An alternative proposition is to introduce `response_fields_exclude` query parameter. However, such introduction would also require description of how `response_fields` and `response_fields_exclude` interact.

### LINKED ISSUE #184: Include properties in response upon requests

### LINKED ISSUE #204: Clarify response fields for default properties in spec.

### LINKED ISSUE #209: Exclusion of 'id' and 'type' properties upon request
The specification allows exclusion of `id` and `type` properties upon request. Although having no "special" properties sounds nice, JSON response without `id` and `type` is invalid according to JSON API specification. Moreover, I cannot think of a use case when the output without `id`and `type` would be beneficial. Shall we define `id` and `type` attributes as mandatory regardless the value of `response_fields`?

merkys:
In discussions with @rartino and @CasperWA over #184 the following ideas over property inclusion have been crystallized:

* There should be a small set of properties that MUST be included in the response by default;
* `response_fields` may be used to include optional properties.

Similarly,

* There should be a small set of properties that MUST support querying;
* Other properties support querying OPTIONALLY.

This PR implements these ideas by adding general notices and cleaning up the 'Entry List' section to explicitly mark only those properties that either included by default or must support querying. (I know, this goes against @CasperWA's idea of having a property reference there, but I have little ideas how to do it otherwise. Machine-readable specs maybe? :confused:). Thus fixes #204.

Also this PR:
* Makes coordinate-based properties included upon request only (fixes #184, closes #181, thus includes #203)
* Makes 'lattice_vectors' mandatory in the response (affects #196)
* Reduces query support requirement for `nsites` from MUST to OPTIONAL (includes #205)
* Forbids exclusion of `id` and `type` (fixes #209)


rartino:
**Thank you** for doing the job with this PR!  This will greatly help to avoid a situation with all implementation having slightly incompatible interpretation of the rules of when to, and when not to, include output fields.

However, with the edits you've made so far in this PR, don't you have to also update the definition of `response_fields` in 4.1.1, i.e.,from
> response_fields: a comma-delimited set of fields to be provided in the output. If provided, only these fields MUST be returned and no others. Example: `http://example.com/optimade/v0.9/structures?response_fields=id,url`

into something like

> response_fields: a comma-delimited set of optional fields to be provided in the output. If provided, the implementation MUST include in its response both the indicated fields and the fields marked as REQUIRED for the endpoint. Fields for which no meaningful value can be given MUST still be included with a `null` value. The implementation MUST NOT return any other fields. Example: `http://example.com/optimade/v0.9/structures?response_fields=immutable_id,chemical_formula_hill`

merkys:
Right, the suggested change to `response_fields` would fix #209 as well. However, I am a little hesitant about bundling that many changes in a single PR, as requirements of other implementations might conflict. Maybe let's wait for today's Web meeting to discuss it with wider audience, OK?

optimade.rst
```diff
@@ -589,6 +589,9 @@ There MAY be multiple entry listing endpoints, depending on how many types of en
 Specific standard entry types are specified in section `Entry list`_.
 The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be prefixed by a database-provider-specific prefix.
 
+By default, only those properties MUST be present in entry descriptions that are marked as REQUIRED in section `Entry list`_.
```
CasperWA: Rearranging for easy readability.
```suggestion
By default, only those properties that are marked as REQUIRED in section `Entry list`_ MUST be present in entry descriptions.
```

CasperWA: By the way, isn't this specific to the `/info/<entry_endpoint>` endpoints? That's the only place we return entry property `description`s.

optimade.rst
```diff
@@ -1076,6 +1081,8 @@ An OPTiMaDe filter expression is passed in the parameter :query-param:`filter` a
 API <https://jsonapi.org/format/1.0/#fetching-filtering>`__.
 The filter expression allows desired properties to be compared against search values; several such comparisons can be combined using the logical conjunctions AND, OR, NOT, and parentheses, with their usual semantics.
 
+Only those properties MUST be queryable that are marked as REQUIRED in section `Entry list`_. Query support for other properties is OPTIONAL.
```
```suggestion
Only those properties marked as REQUIRED in section `Entry list`_ MUST be queryable.
Query support for other properties is OPTIONAL.
```

optimade.rst
```diff
@@ -1656,8 +1656,7 @@ lattice\_vectors
 - **Type**: list of list of floats.
 - **Requirements/Conventions**:
 
-  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` (in this case it is OPTIONAL).
-  - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
+  - **Response**: REQUIRED in the response.
```
CasperWA: I presume it is intentional to leave out the conditional here?

optimade.rst
```diff
@@ -1589,8 +1591,6 @@ chemical\_formula\_hill
 - **Type**: string
 - **Requirements/Conventions**:
   
-  - **Response**: OPTIONAL in the response.
-  - **Query**: Support for queries on these properties are OPTIONAL. If supported, only a subset of filter operators MAY be supported.
```
CasperWA: For all OPTIONAL properties, is the queryability always stated as "If supported, only a subset of filter operators MAY be supported"? If not, I guess we should keep this, if yes, then it's definitely fine to leave it out, writing this sentence as generality elsewhere.

optimade.rst
```diff
@@ -1977,7 +1965,6 @@ Database-provider-specific entry types MUST have all properties described above
 
 - **Requirements/Conventions for properties in database-provider-specific entry types**:
   
-  - **Response**: OPTIONAL in the response.
   - **Query**: Support for queries on these properties are OPTIONAL.
```
CasperWA: You've deleted this for some that are OPTIONAL in the response, but not all. How do you justify this? Or rather, what is the implication?

optimade.rst
merkys: Thanks for spotting this. By *entry description* I meant *response*. Fixed the term in c69d5f0eb137690ceb2d3993ff0f443410d9b2f4, as well as accepted your rearranging.

optimade.rst
merkys: Yes. The intent was to simplify the rules for inclusion/exclusion of properties. Also I had PR #206 in mind while developing the current PR.

optimade.rst
```diff
@@ -589,6 +589,9 @@ There MAY be multiple entry listing endpoints, depending on how many types of en
 Specific standard entry types are specified in section `Entry list`_.
 The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be prefixed by a database-provider-specific prefix.
 
+By default, only those properties that are marked as REQUIRED in section `Entry list`_ MUST be present in the response.
+Query parameter *response_fields* MAY be used to explicitly include or exclude properties.
```
rartino: If you follow my suggestion with editing the text about `response_fields`, this needs to be edited
```suggestion
The query parameter *response_fields* is used by the client to request OPTIONAL properties to be included in the response, see Section `Entry Listing URL Query Parameters`_.
```
And the line above should probably drop "By default,"

optimade.rst
merkys: Good point. Clarified this in 502bd44296a69e799f4abdf55b8305bc621ad223.

optimade.rst
merkys: You are right. I was a little hesitant to cleanup the requirements, but I think with the addition of general clarification in 502bd44296a69e799f4abdf55b8305bc621ad223 the rest of the optional clauses can be removed. Did so in 8a6a4b2c9e826b43505847f47d7544239c2a45d8 and 892f60a55dd615e20ad17ba6a9dc6d2e518049ce.

optimade.rst
```diff
@@ -1589,8 +1592,6 @@ chemical\_formula\_hill
 - **Type**: string
 - **Requirements/Conventions**:
   
-  - **Response**: OPTIONAL in the response.
-  - **Query**: Support for queries on these properties are OPTIONAL. If supported, only a subset of filter operators MAY be supported.
```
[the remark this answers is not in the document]
merkys: Added a general statement about the optionally supported operators in 502bd44296a69e799f4abdf55b8305bc621ad223.

merkys:
I have responded to @rartino and @CasperWA comments, addressed #209 on the way. The PR has grown quite heavy, thus I request reviews of other contributors who, AFAIR, were interested in the related discussions.

optimade.rst
```diff
@@ -589,6 +588,9 @@ There MAY be multiple entry listing endpoints, depending on how many types of en
 Specific standard entry types are specified in section `Entry list`_.
 The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be prefixed by a database-provider-specific prefix.
 
+Only the properties that are marked as REQUIRED in section `Entry list`_ MUST be present in the response.
```
```suggestion
Only the properties marked as REQUIRED in section `Entry list`_ MUST be present in the response.
```

optimade.rst
```diff
@@ -589,6 +588,9 @@ There MAY be multiple entry listing endpoints, depending on how many types of en
...
+The query parameter *response_fields* is used by the client to request OPTIONAL properties to be included in the response, see Section `Entry Listing URL Query Parameters`_.
```
```suggestion
The query parameter :query-param:`response_fields` is used by the client to request OPTIONAL properties to be included in the response, see Section `Entry Listing URL Query Parameters`_.
```

optimade.rst
```diff
@@ -644,8 +646,9 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
   The email SHOULD be that of a person and not an automatic system.
   Example: http://example.com/optimade/v0.9/structures?email_address=user@example.com
 - **response\_fields**: a comma-delimited set of fields to be provided in the output.
-  If provided, only these fields MUST be returned and no others.
-  Example: http://example.com/optimade/v0.9/structures?response_fields=id,url
+  If provided, these fields MUST be returned along the REQUIRED fields.
```
```suggestion
  If provided, these fields MUST be returned along with the REQUIRED fields.
```

optimade.rst
```diff
@@ -644,8 +646,9 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
...
+  No other OPTIONAL fields MUST be returned in such case.
+  Example: http://example.com/optimade/v0.9/structures?response_fields=url,nsites
```
CasperWA: We don't have a `url` field.
```suggestion
  Example: http://example.com/optimade/v0.9/structures?response_fields=last_modified,nsites
```

CasperWA: Also, I guess passing a REQUIRED field to `response_fields` is still fine? Although it is understood that it will not alter the response?

merkys: > We don't have a `url` field.

Indeed, it's better to use a real-world example.

> Also, I guess passing a REQUIRED field to `response_fields` is still fine? Although it is understood that it will not alter the response?

Yes and yes.

merkys:
> I am however confused now. Compared to the current spec, we are now forcing a large data response, since, according to this, we are now unable to remove REQUIRED fields in the response, correct?

True, there will be no possibility to remove REQUIRED fields from the response with this PR. This is done both to solve #209 and simplify the rules: there is a set of mandatory properties, additional optional properties can be added using `response_fields`.

I consider the mandatory response quite compact (we have removed coordinates and related data at least). If you feel that some more fields could be made OPTIONAL - just let me know :)

rartino:
The idea is that the REQUIRED part is so small that it doesn't matter that all of it must be included. 

At this point, the REQUIRED part is essentially just various representations of the chemical formula + `lattice_vectors`. That last one stands out a bit from the others when the coordinates are not in the response, so maybe it can also be made non-mandatory.

Or, with apologies to @merkys for wavering on this topic; do we need to refine the categories one step further? Into:
1. REQUIRED in the response
2. Included in the response by default
3. Only included when requested by `response_fields`

And keep the formulation that with `response_fields` set, it specifies the included fields in addition to the REQUIRED ones. I.e., the default ones are then dropped.

Hence, this can be interpreted as saying that not giving a `response_fields` returns the concatenation of REQUIRED fields and default fields.

merkys:
I think yet another requirement layer for properties might be an overkill. But we can introduce it, if needed later. Would such introduction require a major release of specification? If not, maybe let's leave it for v1.1?

rartino:
I have mulled this over in my head for quite a while. I think a major use case we are missing with the present PR is when one makes a query that will return A LOT of entries and just want to grab their IDs transferring as little data as possible. I've used that sometimes when querying existing databases.

So, I apologize profusely for flip-flopping on this, but I now firmly believe that we need the three level version. These are the categories:
1. Only `type` and `id` are marked as REQUIRED in the response, as they are now.
2. All other properties marked as `REQUIRED` now are changed into "MUST be included in the response unless the `response_fields` query parameter is present and does not include this property."
3. All other properties are handled as they are now. I suppose they can remain unmarked, but even in the present PR I think the clarity would be grater saying: "SHOULD not be included in the response unless the `response_fields` query parameter is present and includes this property."

So, as you notice, this is very close to the present version. It is only 'level 2'-properties that needs a search+replace change of the text.

@merkys what do you think? Do you agree? If yes, are you willing to implement this change in your PR this week? Otherwise, I volunteer to do these changes to get them done for the deadline on Friday.

merkys:
> @merkys what do you think? Do you agree? If yes, are you willing to implement this change in your PR this week? Otherwise, I volunteer to do these changes to get them done for the deadline on Friday.

I completely agree with your latest proposal. However, this week is very busy for me, so I'd be grateful if you could make the suggested changes.You should be able to push to my branch.

CasperWA:
> I have mulled this over in my head for quite a while. I think a major use case we are missing with the present PR is when one makes a query that will return A LOT of entries and just want to grab their IDs transferring as little data as possible. I've used that sometimes when querying existing databases.
> 
> So, I apologize profusely for flip-flopping on this, but I now firmly believe that we need the three level version. These are the categories:
> 
>     1. Only `type` and `id` are marked as REQUIRED in the response, as they are now.
> 
>     2. All other properties marked as `REQUIRED` now are changed into "Included in a default response, i.e., MUST be included in the response unless the `response_fields` query parameter is present and does not include this property."
> 
>     3. All other properties are handled as they are now. I suppose they can remain unmarked, but even in the present PR I think the clarity would be grater saying: "SHOULD not be included in the response unless the `response_fields` query parameter is present and includes this property."

This is exactly what I was expecting from this PR and hoping for, hence my earlier "confused" comment.
Great to see we're all see eye-to-eye on this now :+1: Looking forward to this change.

rartino:
Done. Note that I ran into some trouble (that I recall we discussed somewhere) that these changes would lose all distinction between properties that the implementation could just ignore ("fully optional") and properties we meant needs to be 'supported' but are no longer required in any response. 

So, I took it also upon this PR to add a **supported** key in the entries section that clarifies when a property must be supported in the sense that it cannot act as it is `null` all the time. I was very restrictive with the MUST demand here, and most properties that previous was marked as "REQUIRED in the response unless excluded" are now "SHOULD not be `null`".

Well, read the PR, and let me know if you disagree. 

Also, please approve ASAP given that we are trying to work towards the deadline of the 20:th...

optimade.rst
```diff
@@ -516,16 +516,13 @@ The text in this section describes how the API handles properties with the value
 The use of :val:`null` values inside nested property values (such as, e.g., lists or dictionaries) are described in the definitions of those data structures elsewhere in the specification, see section `Entry List`_.
 For these properties, :val:`null` MAY carry a special meaning.
 
-REQUIRED properties with an unknown value MUST be returned in the response, unless explicitly left out (e.g., by using :query-param:`response_fields`, see section `Entry Listing URL Query Parameters`_).
+Responses that include entries where REQUIRED properties have an unknown value MUST include these properties with the value :val:`null`.
 
-OPTIONAL properties with an unknown value MAY be returned in the response.
-If an OPTIONAL property is *not* returned in a *full* response (i.e., not using :query-param:`response_fields`), the client MUST assume the property has an unknown value, i.e., :val:`null`.
+Responses to requests where OPTIONAL properties have been explicitly requested via the :query-param:`response_fields` query parameter, and which includes entries where those properties have an unknown value, MUST include these properties with the value :val:`null`.
```
merkys: "which includes" -> "which include"

optimade.rst
```diff
@@ -588,7 +585,16 @@ Entry Listing Endpoints
 Entry listing endpoints return a list of resource objects representing entries of a specific type.
 For example, a list of structures, or a list of calculations.
 
-Examples:
+For each entry in the list, a set of properties and their corresponding values are given on the form dictated by the output format.
```
merkys: Shouldn't is be "are given on the form" -> "are given based on the form"?

merkys: I must admit I don't fully understand this sentence. I think the set of properties is independent on the output format, so are the values. Or am I missing something?

optimade.rst
```diff
@@ -1463,9 +1477,9 @@ database-provider-specific properties
 - **Type**: Decided by the API implementation.
 - **Requirements/Conventions**:
   
-  - **Response**: OPTIONAL in the response.
-  - **Query**: Support for queries on these properties are OPTIONAL.
-    If supported, only a subset of filter operators MAY be supported.
+  - **Support**: Support for database-provider-specific properties is fully OPTIONAL.
+  - **Query**: Support for queries on these properties are OPTIONAL. If supported, only a subset of filter operators MAY be supported.
+  - **Response**: API implementations are free to choose whether database-provider-specific properties are only included when requested using the query parameter :query-param:`response_fields`, or if they are regarded as part of the "deault response", i.e., also included when the query parameter :query-param:`response_fields` is not present. Implementations SHOULD NOT include database-provider-specific properties when the query parameter :query-param:`response_fields` is present but does not list them.
```
merkys: "deault response" -> "default response"

optimade.rst
rartino: Thanks! 

I'm doing these fixes and will push them to your repo now, but you do realize this is your PR, so nothing prevents you from doing the edits :-)

optimade.rst
rartino: The "form" part of the sentence was just to clarify that how things are encoded in the response is dependent on the output format. However, I think you are right that this clarification confuses more than it helps, so how about just skipping it:

> Each entry in the list includes a set of properties and their corresponding values.
> The section `Entry list`_ specifies properties as belonging to one of three categories:

rartino:
I had forgotten to merge the current state of develop. And, oh wow, that was a painful merge. 

(I've tried to be careful, and my interpretation is that any bad mistake ought to show up in the diff for the present PR, since it would be reverting changes in the present state of develop. But, I'm honestly not sure on that; I remember a case earlier in the history of optimade.md where we managed to revert some changes in an unclean merge. We couldn't even trace back in *which* merge it had happened, there were 3 or so candidates...)

optimade.rst
merkys: I agree with @rartino's suggestion.

merkys:
> I had forgotten to merge the current state of develop. And, oh wow, that was a painful merge.

I trust you did the merge right :) OK, so we need yet one more approve.

rartino:
Pinging @CasperWA, @sauliusg or @giovannipizzi for a review!

optimade.rst
```diff
@@ -310,16 +310,13 @@ The text in this section describes how the API handles properties with the value
 The use of :val:`null` values inside nested property values (such as, e.g., lists or dictionaries) are described in the definitions of those data structures elsewhere in the specification, see section `Entry List`_.
 For these properties, :val:`null` MAY carry a special meaning.
 
-REQUIRED properties with an unknown value MUST be returned in the response, unless explicitly left out (e.g., by using :query-param:`response_fields`, see section `Entry Listing URL Query Parameters`_).
+Responses that include entries where REQUIRED properties have an unknown value MUST include these properties with the value :val:`null`.
 
-OPTIONAL properties with an unknown value MAY be returned in the response.
-If an OPTIONAL property is *not* returned in a *full* response (i.e., not using :query-param:`response_fields`), the client MUST assume the property has an unknown value, i.e., :val:`null`.
+Responses to requests where OPTIONAL properties have been explicitly requested via the :query-param:`response_fields` query parameter, and which include entries where those properties have an unknown value, MUST include these properties with the value :val:`null`.
```
CasperWA: This may then also be rewritten.
```suggestion
OPTIONAL properties, if requested explicitly via the :query-param:`response_fields` query parameter, MUST be included and returned in the response.
```

optimade.rst
```diff
@@ -310,16 +310,13 @@ The text in this section describes how the API handles properties with the value
...
+(For more info on the :query-param:`response_fields` query parameter, see section `Entry Listing URL Query Parameters`_.)
 
-If a property is explicitly requested in a search query without value range filters, then all entries otherwise satisfying the query SHOULD be returned, including those with :val:`null` values for this property.
-These properties MUST be set to :val:`null` in the response.
-
-Filters with :filter-fragment:`IS UNKNOWN` and :filter-fragment:`IS KNOWN` can be used to match entries with values that are, or are not, unknown for some property, respectively.
-This is discussed in section `The Filter Language Syntax`_.
+The interaction of properties with an unknown value with query filters is described in the section `Filtering on Properties with unknown value`_.
```
CasperWA: Shouldn't the title for this section either have plural `values` or an `a` before unknown?

optimade.rst
```diff
@@ -627,7 +624,16 @@ Entry Listing Endpoints
 Entry listing endpoints return a list of resource objects representing entries of a specific type.
 For example, a list of structures, or a list of calculations.
 
-Examples:
+Each entry in the list includes a set of properties and their corresponding values.
+The section `Entry list`_ specifies properties as belonging to one of three categories:
+
+1. Some properties are marked as REQUIRED in the response. These properties MUST always be present for all entries in the response.
```
```suggestion
1. Some properties are marked as REQUIRED in the response.
   These properties MUST always be present for all entries in the response.
```

optimade.rst
```diff
@@ -627,7 +624,16 @@ Entry Listing Endpoints
...
+
+2. Properties marked as part of the "default response" MUST be included if the query parameter :query-param:`response_fields` is not part of the request, or if they are explicity requested in :query-param:`response_fields`. Otherwise they MUST NOT be included.
```
CasperWA: Is `"default response"` defined somewhere?

```suggestion
2. Properties marked as part of the "default response" MUST be included if the query parameter :query-param:`response_fields` is not part of the request, or if they are explicity requested in :query-param:`response_fields`.
   Otherwise they MUST NOT be included.
```

optimade.rst
```diff
@@ -627,7 +624,16 @@ Entry Listing Endpoints
...
+
+3. Properties that are neither REQUIRED, nor part of the "default response" MUST be included if explicity requested in the :query-param:`response_fields`. Otherwise they SHOULD NOT be included.
```
CasperWA: New line & addition of "query parameter" to the first sentence.
```suggestion
3. Properties that are neither REQUIRED, nor part of the "default response" MUST be included if explicity requested in the query parameter :query-param:`response_fields`.
   Otherwise they SHOULD NOT be included.
```

optimade.rst
```diff
@@ -627,7 +624,16 @@ Entry Listing Endpoints
...
+
+Examples of entry listing endpoints:
```
```suggestion
Examples of valid entry listing endpoint URLs:
```

optimade.rst
```diff
@@ -704,8 +710,9 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
   The email SHOULD be that of a person and not an automatic system.
   Example: http://example.com/optimade/v0.9/structures?email_address=user@example.com
 - **response\_fields**: a comma-delimited set of fields to be provided in the output.
-  If provided, only these fields MUST be returned and no others.
-  Example: http://example.com/optimade/v0.9/structures?response_fields=id,url
+  If provided, these fields MUST be returned along with the REQUIRED fields.
+  No other OPTIONAL fields MUST be returned in such case.
```
```suggestion
  Other OPTIONAL fields MUST NOT be returned in such case.
```

optimade.rst
```diff
@@ -1439,8 +1451,9 @@ id
 - **Type**: string.
 - **Requirements/Conventions**:
   
-  - **Response**: REQUIRED in the response unless explicitly excluded.
-  - **Query**: MUST be a queryable property with support for all mandatory filter operators.
+  - **Support**: MUST be supported by all implementations, MUST NOT be :val:`null` for any entries.
```
```suggestion
  - **Support**: MUST be supported by all implementations, MUST NOT be of type `unknown value`, i.e., :val:`null`.
```

CasperWA: My first suggested change was to simply remove ` for any entries`, which is my strongest suggested change in this sentence. The other part is not as crucial for me.

optimade.rst
```diff
@@ -1458,9 +1471,9 @@ type
 - **Type**: string.
 - **Requirements/Conventions**:
   
-  - **Response**: REQUIRED in the response unless explicitly excluded.
-  - **Query**: Support for queries on this property is OPTIONAL.
-    If supported, only a subset of string comparison operators MAY be supported.
+  - **Support**: MUST be supported by all implementations, MUST NOT be :val:`null` for any entries.
```
```suggestion
  - **Support**: MUST be supported by all implementations, MUST NOT be :val:`null`.
```

CasperWA: This is essentially my original suggested change to this line - see previous suggested change for an additional change.

optimade.rst
```diff
@@ -1472,8 +1485,8 @@ immutable\_id
 - **Type**: string.
 - **Requirements/Conventions**:
   
-  - **Response**: OPTIONAL in the response.
-  - **Query**: If present, MUST be a queryable property with support for all mandatory filter operators.
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null` for all entries.
```
```suggestion
  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
```

optimade.rst
```diff
@@ -1501,10 +1515,10 @@ database-provider-specific properties
 - **Type**: Decided by the API implementation.
 - **Requirements/Conventions**:
   
-  - **Response**: OPTIONAL in the response.
-  - **Query**: Support for queries on these properties are OPTIONAL.
-    If supported, only a subset of filter operators MAY be supported.
-  - These MUST be prefixed by a database-provider-specific prefix (see section `Database-Provider-Specific Namespace Prefixes`_).
+  - **Support**: Support for database-provider-specific properties is fully OPTIONAL.
+  - **Query**: Support for queries on these properties are OPTIONAL. If supported, only a subset of filter operators MAY be supported.
```
```suggestion
  - **Query**: Support for queries on these properties are OPTIONAL.
     If supported, only a subset of filter operators MAY be supported.
```

optimade.rst
```diff
@@ -1501,10 +1515,10 @@ database-provider-specific properties
...
+  - **Response**: API implementations are free to choose whether database-provider-specific properties are only included when requested using the query parameter :query-param:`response_fields`, or if they are regarded as part of the "default response", i.e., also included when the query parameter :query-param:`response_fields` is not present. Implementations SHOULD NOT include database-provider-specific properties when the query parameter :query-param:`response_fields` is present but does not list them.
```
```suggestion
  - **Response**: API implementations are free to choose whether database-provider-specific properties are only included when requested using the query parameter :query-param:`response_fields`, or if they are regarded as part of the "default response", i.e., also included when the query parameter :query-param:`response_fields` is not present.
    Implementations SHOULD NOT include database-provider-specific properties when the query parameter :query-param:`response_fields` is present but does not list them.
```

optimade.rst
```diff
@@ -1647,8 +1660,9 @@ chemical\_formula\_hill
 - **Type**: string
 - **Requirements/Conventions**:
   
-  - **Response**: OPTIONAL in the response.
-  - **Query**: Support for queries on these properties are OPTIONAL. If supported, only a subset of filter operators MAY be supported.
+  - **Support**: OPTIONAL support in implementations, i.e., may be :val:`null` for all entries.
```
```suggestion
  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
```

optimade.rst
```diff
@@ -1714,8 +1728,9 @@ lattice\_vectors
 - **Type**: list of list of floats.
 - **Requirements/Conventions**:
 
-  - **Response**: REQUIRED in the response unless explicitly excluded, except when property `dimension_types`_ is equal to :val:`[0, 0, 0]` (in this case it is OPTIONAL).
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
+
```
```suggestion
```

CasperWA: There is some inconsistency in when there is an empty line here or not.

optimade.rst
```diff
@@ -1852,9 +1868,10 @@ assemblies
     
 - **Requirements/Conventions**:
 
-  - **Response**: OPTIONAL in the response (SHOULD be absent if there are no partial occupancies).
-  - **Query**: Support for queries on this property is OPTIONAL.
-    If supported, filters MAY support only a subset of comparison operators.
+  - **Support**: OPTIONAL support in implementations, i.e., may be :val:`null` for all entries.
```
```suggestion
  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
```

optimade.rst
```diff
@@ -1852,9 +1868,10 @@ assemblies
...
+  - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
```
```suggestion
  - **Query**: Support for queries on this property is OPTIONAL.
    If supported, filters MAY support only a subset of comparison operators.
```

optimade.rst
```diff
@@ -1958,7 +1975,7 @@ structure\_features
 - **Type**: list of strings
 - **Requirements/Conventions**:
   
-  - **Response**: REQUIRED in the response (SHOULD be absent if there are no partial occupancies).
+  - **Support**: MUST be supported by all implementations, MUST NOT be :val:`null` for any entries.
```
```suggestion
  - **Support**: MUST be supported by all implementations, MUST NOT be :val:`null`.
```

optimade.rst
```diff
@@ -1996,9 +2013,9 @@ The following properties are used to provide the bibliographic details:
 
 - **Requirements/Conventions**:
   
-  - **Response**: Every references entry MUST contain at least one of the properties.
-  - **Query**: Support for queries on any of these properties is OPTIONAL.
-    If supported, filters MAY support only a subset of comparison operators.
+  - **Support**: OPTIONAL support in implementations, i.e., any of the properties may be :val:`null` for all entries.
```
```suggestion
  - **Support**: OPTIONAL support in implementations, i.e., any of the properties MAY be :val:`null`.
```

optimade.rst
```diff
@@ -1996,9 +2013,9 @@ The following properties are used to provide the bibliographic details:
...
+  - **Query**: Support for queries on any of these properties is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
```
```suggestion
  - **Query**: Support for queries on any of these properties is OPTIONAL.
    If supported, filters MAY support only a subset of comparison operators.
```

optimade.rst
```diff
@@ -2027,14 +2044,13 @@ Example:
...
 
-- **Requirements/Conventions for properties in database-provider-specific entry types**:
-  
-  - **Response**: OPTIONAL in the response.
-  - **Query**: Support for queries on these properties are OPTIONAL.
-    If supported, only a subset of filter operators MAY be supported.
+* **Requirements/Conventions for properties in database-provider-specific entry types**:
+
+  - **Support**: Support for any properties in database-provider-specific entry types is fully OPTIONAL.
+  - **Query**: Support for queries on these properties are OPTIONAL. If supported, only a subset of filter operators MAY be supported.
```
```suggestion
  - **Query**: Support for queries on these properties are OPTIONAL.
    If supported, only a subset of filter operators MAY be supported.
```

optimade.rst
rartino: Your version misses "with an unknown value". Sure, arguably you can say that the statement applies generally--- unknown value or not, but the point of this section is to clarify that these properties need to be in the response even if the value is unknown.

optimade.rst
rartino: To the best of my language abilities, I think both are valid grammatically. I assume you want us to change it, so I'll do that.

optimade.rst
rartino: It is just meant to quote the human-friendly explanation "label" used in that section to mark these. I'll try to rephrase.

optimade.rst
rartino: I'm opposed to an explicit statement that says that `unknown value` is a type, because in my understanding "unknown" is more complicated than that.

But, so why do you feel strongly for removing "for any entries"? To me, this makes it clear with zero ambiguity that the implementation cannot serve any entries where these are null. You may argue that this should be inferred even without that clarification, but I really think it helps to communicate the intent of this limitation clearly.

optimade.rst
```diff
@@ -310,16 +310,13 @@ The text in this section describes how the API handles properties with the value
 The use of :val:`null` values inside nested property values (such as, e.g., lists or dictionaries) are described in the definitions of those data structures elsewhere in the specification, see section `Entry List`_.
 For these properties, :val:`null` MAY carry a special meaning.
 
-REQUIRED properties with an unknown value MUST be returned in the response, unless explicitly left out (e.g., by using :query-param:`response_fields`, see section `Entry Listing URL Query Parameters`_).
+Responses that include entries where REQUIRED properties have an unknown value MUST include these properties with the value :val:`null`.
```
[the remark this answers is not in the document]
rartino: I've edited into:
> REQUIRED properties with an unknown value MUST be included and returned in the response with the value :val:`null`.

optimade.rst
rartino: I've edited into:
> OPTIONAL properties with an unknown value, if requested explicitly via the :query-param:`response_fields` query parameter, MUST be included and returned in the response with the value :val:`null`.

optimade.rst
rartino: Actually, upon editing, I decide to fold on this. I'll cut the "for any entries", it should be obvious. I'll skip the additional reference to unknown here, unless you can motivate why.

optimade.rst
rartino: Ok, I've edited all text that references "default response", because I completely agree that this moniker was not very clear; I now specify the formal definition first, and then clarify that this can be seen as the default value of the `response_fileds` query parameter. Check the new version and let me know if it is not sufficient.

rartino:
@CasperWA Thanks for the careful review! Hopefully, this version is better.

optimade.rst
CasperWA: Thanks - will make another review now. Note, I was also okay with defining this term under our beginning definitions.

optimade.rst
CasperWA: > it should be obvious.

Thanks :+1: 

> I'll skip the additional reference to unknown here, unless you can motivate why.

The reason was simply that we _have_ defined `unknown value` as a recognized type in our specification already, which MAY have the value `null`, and MUST have the value `null` for a JSON response.
See [this section](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#21data-types) and specifically [this line](https://github.com/Materials-Consortia/OPTiMaDe/blame/develop/optimade.rst#L209). However! In the very next line, we do explicitly write that we will consider this type to always have value `null`, so no reason to underline this fact again in the current part of the spec.
I give :) I think we intentionally wrote [this line](https://github.com/Materials-Consortia/OPTiMaDe/blame/develop/optimade.rst#L210) to be able to do exactly what we are currently doing here, i.e., use `null` as a universal value for a property with one or more unknown values.

optimade.rst
```diff
@@ -627,7 +625,19 @@ Entry Listing Endpoints
...
 
-Examples:
+Each entry in the list includes a set of properties and their corresponding values.
+The section `Entry list`_ specifies properties as belonging to one of three categories:
+
+1. Some properties are marked as REQUIRED in the response.
+   These properties MUST always be present for all entries in the response.
+
+2. Some properties are marked as REQUIRED only if the query parameter :query-param:`response_fields` is not part of the request, or if they are explicity requested in :query-param:`response_fields`.
+   Otherwise they MUST NOT be included. One can think of these properties as consituting a default value for :query-param:`response_fields` when that parameter is omitted.
```
```suggestion
   Otherwise they MUST NOT be included.
   One can think of these properties as consituting a default value for :query-param:`response_fields` when that parameter is omitted.
```

optimade.rst
```diff
@@ -627,7 +625,19 @@ Entry Listing Endpoints
 Entry listing endpoints return a list of resource objects representing entries of a specific type.
 For example, a list of structures, or a list of calculations.
 
-Examples:
+Each entry in the list includes a set of properties and their corresponding values.
+The section `Entry list`_ specifies properties as belonging to one of three categories:
+
+1. Some properties are marked as REQUIRED in the response.
```
```suggestion
1. Properties marked as REQUIRED in the response.
```

optimade.rst
```diff
@@ -627,7 +625,19 @@ Entry Listing Endpoints
...
+   These properties MUST always be present for all entries in the response.
+
+2. Some properties are marked as REQUIRED only if the query parameter :query-param:`response_fields` is not part of the request, or if they are explicity requested in :query-param:`response_fields`.
```
```suggestion
2. Properties marked as REQUIRED only if the query parameter :query-param:`response_fields` is not part of the request, or if they are explicitly requested in :query-param:`response_fields`.
```

optimade.rst
```diff
@@ -627,7 +625,19 @@ Entry Listing Endpoints
...
+   Otherwise they MUST NOT be included. One can think of these properties as consituting a default value for :query-param:`response_fields` when that parameter is omitted.
+
+3. Properties that are not marked as REQUIRED in any case, MUST be included only if explicity requested in the query parameter :query-param:`response_fields`.
```
```suggestion
3. Properties not marked as REQUIRED in any case, MUST be included only if explicitly requested in the query parameter :query-param:`response_fields`.
```

optimade.rst
```diff
@@ -627,7 +625,19 @@ Entry Listing Endpoints
...
+   Otherwise they SHOULD NOT be included.
+
+Examples of valid entry listing endpoints URLs:
```
```suggestion
Examples of valid entry listing endpoint URLs:
```

optimade.rst
```diff
@@ -1134,6 +1147,9 @@ An OPTiMaDe filter expression is passed in the parameter :query-param:`filter` a
 API <https://jsonapi.org/format/1.0/#fetching-filtering>`__.
 The filter expression allows desired properties to be compared against search values; several such comparisons can be combined using the logical conjunctions AND, OR, NOT, and parentheses, with their usual semantics.
 
+All properties marked as REQUIRED in section `Entry list`_ MUST be queryable with all mandatory filter featrues.
```
```suggestion
All properties marked as REQUIRED in section `Entry list`_ MUST be queryable with all mandatory filter features.
```

optimade.rst
```diff
@@ -2027,14 +2054,14 @@ Example:
...
 --------------------------------------
 
-Names of database-provider-specific entry types MUST start with database-provider-specific namespace prefix (see section `Database-Provider-Specific Namespace Prefixes`_).
+Names of database-provider-specific entry types MUST start with database-provider-specific namespace prefix (see appendix `Database-Provider-Specific Namespace Prefixes`_).
 Database-provider-specific entry types MUST have all properties described above in section `Properties Used by Multiple Entry Types`_.
 
-- **Requirements/Conventions for properties in database-provider-specific entry types**:
-  
-  - **Response**: OPTIONAL in the response.
+* **Requirements/Conventions for properties in database-provider-specific entry types**:
```
CasperWA: Is there a reason this bullet point has changed from `-` to `*`? Just curious. If it looks better, that's fine. If it looks the same, then maybe change it back to be consistent within the document.

rartino: No idea how this happened. Good catch!

rartino:
@CasperWA done!

PR #213, opened 2019-11-27
Add CC-BY 4.0 license for the specification

### LINKED ISSUE #111: Add license to specification
as discussed I plan to add the Creative Commons 4.0 BY license,
potentially with separate suggestions pointing out that optimade.org is the official optimade organisation and that name clashes should be avoided.

I'll also try to give a look to what other specification bodies do for such cases.

merkys:
Looks like the discussion in #111 have quieted down a little, and CC-BY seems to be an option suitable for the most of people there.

Sorry for spamming with review requests. Nevertheless, I couldn't ask reviews from everyone due to apparent limit of 15 reviewers.

Fixes #111. (Sorry @ltalirz for stealing your PR)

**Edit**: please **don't merge** this PR until all the parties have agreed with the proposal.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,396 @@
+Attribution 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
+Creative Commons public licenses does not create a lawyer-client or
+other relationship. Creative Commons makes its licenses and related
+information available on an "as-is" basis. Creative Commons gives no
+warranties regarding its licenses, any material licensed under their
+terms and conditions, or any related information. Creative Commons
+disclaims all liability for damages resulting from their use to the
+fullest extent possible.
+
+Using Creative Commons Public Licenses
+
+Creative Commons public licenses provide a standard set of terms and
+conditions that creators and other rights holders may use to share
+original works of authorship and other material subject to copyright
+and certain other rights specified in the public license below. The
+following considerations are for informational purposes only, are not
+exhaustive, and do not form part of our licenses.
+
+     Considerations for licensors: Our public licenses are
+     intended for use by those authorized to give the public
+     permission to use material in ways otherwise restricted by
+     copyright and certain other rights. Our licenses are
+     irrevocable. Licensors should read and understand the terms
+     and conditions of the license they choose before applying it.
+     Licensors should also secure all rights necessary before
+     applying our licenses so that the public can reuse the
+     material as expected. Licensors should clearly mark any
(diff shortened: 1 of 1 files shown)


ltalirz:
Maybe add a short section about the license to the README, mentioning that it applies to the specification document only?

merkys:
> Maybe add a short section about the license to the README, mentioning that it applies to the specification document only?

Is there a reason against covering all files by the same license?

merkys:
> If we ever include foreign files that are under different compatible F/LOSS licenses, we'll have to write a note or readme; I guess Andrius can explain better how this is handled in Debian.

Having a single all-covering license seems to be a common practice in the F/LOSS world, and I believe this could be easily applied to documents such as specifications. Should we need to add 3rd party documents/software, we could place them in a separate directory with their licenses, and place a note in README noticing about them.

Having an all-covering license simplifies things a lot. For example, if only `optimade.rst` is covered by CC-SA, it will have to be distributed without the `AUTHORS` file. Moreover, derived projects wouldn't be able to use the testing system, etc.

CasperWA:
> Maybe add a short section about the license to the README, mentioning that it applies to the specification document only?

This seems like the most careful / secure approach, right?
Providing the license _only_ for the spec allows immediate re-usage of the spec, while allowing us to consider a bit more what license should be on our test files, `providers.json`, etc.

On the other hand, it may just postpone the real decision further, leaving us with another long period on non-decision...

CasperWA:
Tagging @csadorf here to give him a chance to add his thoughts, if any, since he was very involved lately in the issue discussion.

rartino:
From a formal copyright perspective, we still need approvals of @ctoher, @fekad, @coreyoses, @ml-evs, @gmrigna, @fawzi, and @muhrin as those have contributed to the repository but not yet approved this PR. I'm tagging them here in the hope that they'll see this, visit this github page, and do the approval.

rartino:
One more attempt to ping the missing people: Hello @ctoher, @fekad, @coreyoses, and @fawzi: please log into github and approve the new repository license! It'll only take a minute!

coreyoses:
Thanks for the poke @rartino. I finally sat down to read about the license.

rartino:
I've now sent personal emails to the few people remaining in hope that we get the final approvals in time for the Dec 20 deadline.

rartino:
@merkys My interpretation is that everyone who has committed to the repository now has approved the licensing. Please check that, and then feel free to press the big green button.

merkys:
OK. Thus a last call: if no one is against, I will merge this PR on Thursday 19th, 08:00 UTC.

PR #214, opened 2019-11-28
Update mcloud provider index base URL

CasperWA:
NOTE: This is against `master` branch, not `develop`.

### DIFF
--- a/providers.json
+++ b/providers.json
@@ -42,7 +42,7 @@
       "attributes": {
         "name": "Example provider",
         "description": "Provider used for examples, not to be assigned to a real database",
-        "base_url": "http://example.com/optimade/index"
+        "base_url": "http://example.com/optimade"
       }
     },
     {
@@ -51,7 +51,7 @@
       "attributes": {
         "name": "The High-Throughput Toolkit",
         "description": "Prefix for implementation-specific identifiers used in the httk implementation at http://httk.org/",
-	"base_url": null
+        "base_url": null
       }
     },
     {
@@ -105,7 +105,7 @@
       "attributes": {
         "name": "OPTiMaDe implementations and libraries",
         "description": "Prefix for implementation-specific identifiers used in API implementations and libraries provided at https://github.com/Materials-Consortia",
-	"base_url": null
+        "base_url": null
       }
     },
     {
--- a/providers.json
+++ b/providers.json
@@ -60,7 +60,7 @@
       "attributes": {
         "name": "Materials Cloud",
         "description": "Materials Cloud: A platform for Open Science built for seamless sharing of resources in computational materials science (http://www.materialscloud.org)",
-        "base_url": null
+        "base_url": "https://www.materialscloud.org/optimade"
       }
     },
     {


ml-evs:
Do you mind if I hijack this to add our new info too?

CasperWA:
> Do you mind if I hijack this to add our new info too?

Sure, go ahead! But remember the URL has to point to the base URL of an [Index meta-database](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#index-meta-database).

> EDIT: just realised that providers.json is quite different in develop, so will make my changes there instead.

Actually, `providers.json` has not changed between `master` and `develop`.

ml-evs:
> Sure, go ahead! But remember the URL has to point to the base URL of an [Index meta-database](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#index-meta-database).

Now I've seen your PR adding the index meta db, I'll wait until I have my server running with that before adding (can be a separate PR).

> > EDIT: just realised that providers.json is quite different in develop, so will make my changes there instead.
> 
> Actually, `providers.json` has not changed between `master` and `develop`.

Weird, no idea what version I had locally then (must be from an old fork).

ml-evs:
Given how smoothly updating to the routing system in optimade-python-tools has gone, could I get write access to this branch @CasperWA so I can add my implementation?

ml-evs:
Thanks @CasperWA , note I'm currently waiting for our SSL cert to refresh for https, http will work until then.

merkys:
Why this PR is against `master`, but not `develop`?

CasperWA:
> Why this PR is against `master`, but not `develop`?

Because I would like to have them in the `providers.json` ASAP :smiley: Not just when we at some point decide to update the version.

merkys:
> Because I would like to have them in the `providers.json` ASAP smiley Not just when we at some point decide to update the version.

Makes sense :)

CasperWA:
> Thanks @CasperWA , note I'm currently waiting for our SSL cert to refresh for https, http will work until then.

If you want to put in the https URL right away, it should be fine, although broken until available, naturally, but otherwise you would need to create another PR.
Do you also want to keep the description empty?

merkys:
Actually, I'm a little hesitant about modifications to `master`. `README.md` states that `master` presents the *stable* version of the specification, and *stable* means immutable. Thus by defining new database-specific prefixes we're theoretically requiring the adopters of v0.10.0 to update their implementations, especially if #162 gets merged. Or am I being too nitpicky?

CasperWA:
> Actually, I'm a little hesitant about modifications to `master`. `README.md` states that `master` presents the _stable_ version of the specification, and _stable_ means immutable. Thus by defining new database-specific prefixes we're theoretically requiring the adopters of v0.10.0 to update their implementations, especially if #162 gets merged. Or am I being too nitpicky?

Ah, @ml-evs has added a new provider-prefix with `odbx`? Yeah, that could be problematic. I would argue that, as the spec is now (in `master`), the two things are decoupled. Ideally, I think `providers.json` and the spec should _always_ be decoupled, since `providers.json` is simply _our_ (i.e., Materials-Consortia's) database provider-specific prefixes. Other consortia or implementers using our spec may use another `providers.json`.
At least I believe this is the intent going forward from previous discussions on this issue between @rartino, @merkys, @sauliusg, and I, maybe more?
Summa summarum, I think it should be fine to update `providers.json` in `master` and add new database provider prefixes as long as the consortium agrees. It should not affect the spec.

ml-evs:
Happy to delay my changes to another PR if that's preferable, just force push over my commits @CasperWA.

CasperWA:
@merkys (and @ml-evs) I have removed @ml-evs's contribution and instead found that there were some indentation issues in two lines.
Most importantly, I have removed the `/index` for the `exmpl` provider. This is _needed_ since no OPTiMaDe implementations, index or otherwise, is allowed to have such a base URL as was written previously. The `index` and `optimade` may be switched, but I think it's better just to remove `index` all-together, what do you think?

CasperWA:
> Happy to delay my changes to another PR if that's preferable, just force push over my commits @CasperWA.

Done. Sorry.
I would make a new PR if I were you, start by making it towards `master` as well, and then we can discuss whether that is correct or not in said PR.

CasperWA:
@merkys are you okay with updating `providers.json` in master for information under existing providers?

merkys:
> @merkys are you okay with updating `providers.json` in master for information under existing providers?

I'm OK, as long as everyone else is fine with touching the `master`. Probably there's no need to be so nitpicky prior to v1.0 release.

Although I'm leaning towards decoupling `providers.json` from the specification for the same reason, still a centralized registry will be needed to keep track of the reserved prefixes.

CasperWA:
Since we have already 4 approvals on this PR, and it doesn't touch anything other than `provider.json`, I will merge it.

CasperWA:
> Although I'm leaning towards decoupling `providers.json` from the specification for the same reason, still a centralized registry will be needed to keep track of the reserved prefixes.

I agree. And I think the centralized registry should be kept elsewhere.

rartino:
How about we just create another repo under Materials-Consortia named providers. In that repo we replicate the URL structure of an index server, with and without major API version number URLs, and commit providers.json there (i.e., practically it will be under two paths with and without version number). The github raw URL becomes our "master" index server for now.

PR #216, opened 2019-11-29
Fix the definition of cartesian_site_positions length

merkys:
Previously the length of `cartesian_site_positions` was defined as N times 3, where N is the number of sites in the structure. This is clearly incorrect, as the length of `cartesian_site_positions` MUST be equal to the number of sites. (`cartesian_site_positions` is list of lists, not a flat list).

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1676,7 +1676,7 @@ cartesian\_site\_positions
   
   - **Response**: REQUIRED in the response unless explicitly excluded.
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1676,7 +1676,7 @@ cartesian\_site\_positions
   
   - **Response**: REQUIRED in the response unless explicitly excluded.
   - **Query**: Support for queries on this property is OPTIONAL. If supported, filters MAY support only a subset of comparison operators.
-  - It MUST be a list of length N times 3, where N is the number of sites in the structure.
+  - It MUST be a list of length equal to the number of sites in the structure.
```
rartino: It is good that you fix this, but you are losing information from what this was intended to mean
```suggestion
  - It MUST be a list of length equal to the number of sites in the structure where every element is a list of the three cartesian coordinates of a site.
```

PR #219, opened 2019-12-03
Add `include` query parameter and be more explicit about top-level `included`

### LINKED ISSUE #183: Handle query parameter 'include' as mandated by JSON API
AFAIK we're trying to support JSON API as closely as possible. [JSON API specifies the use of `include` parameter](https://jsonapi.org/format/#fetching-includes) to limit the types of resources which go to `included` list. To fulfill this requirement of JSON API, we have to either inherit the specification of `include` from JSON API, or respond with `400 Bad Request` to requests with `include`.

### LINKED ISSUE #212: Update requirements for top-level "included"
From the consortium meeting on 26.11.2019 we have decided the following:

If a related resource is of type `references` it SHOULD be included under the top-level `included` field.
For any other related resource types it MAY be included under the top-level `included` field.

There is as of yet no consensus on limitation on how many resource objects may be returned under `included`.

There is also no consensus on whether we want to use the `include` query parameter.

CasperWA:
Closes #212
Closes #183

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -652,7 +652,7 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
   This means :entry:`references` entries MUST always be included under the top-level field :field:`included` as default, since a server assumes if :query-param:`include` is not specified by a client in the request, it is still specified as :query-string:`include=references`.
   Note, if a client explicitly specifies :query-param:`include` and leaves out :query-val:`references`, :entry:`references` resource objects MUST NOT be included under the top-level field :field:`included`, as per the definition of :field:`included`, see section `JSON Response Schema: Common Fields`_.
 
-    **Note**: If a client wishes to explicitly exclude any related resource objects being returned under the top-level field :field:`included`, they can write :query-string:`?include=`.
+    **Note**: A query with the parameter :query-param:`include` set to the empty string means no related resource objects are to be returned under the top-level field :field:`included`.
 
 Standard OPTIONAL URL query parameters not in the JSON API specification:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -430,7 +430,7 @@ The response MAY also return resources related to the primary data in the field:
 - **included**: a list of `JSON API resource objects <http://jsonapi.org/format/1.0/#document-resource-objects>`__ related to the primary data contained in :field:`data`.
   Responses that contain related resources under :field:`included` are known as `compound documents <https://jsonapi.org/format/1.0/#document-compound-documents>`__ in the JSON API.
   
-  The definition of this field is similar to that found ind the `JSON API specification <http://jsonapi.org/format/1.0/#fetching-includes>`__.
+  The definition of this field is similar to that found in the `JSON API specification <http://jsonapi.org/format/1.0/#fetching-includes>`__.
   Specifically, if the query parameter :query-param:`include` is included in the request, :field:`included` MUST NOT include unrequested resource objects.
   For further information on the parameter :query-param:`include`, see section `Entry Listing URL Query Parameters`_.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -430,7 +430,7 @@ The response MAY also return resources related to the primary data in the field:
 - **included**: a list of `JSON API resource objects <http://jsonapi.org/format/1.0/#document-resource-objects>`__ related to the primary data contained in :field:`data`.
   Responses that contain related resources under :field:`included` are known as `compound documents <https://jsonapi.org/format/1.0/#document-compound-documents>`__ in the JSON API.
   
-  The definition of this field is similar to that found in the `JSON API specification <http://jsonapi.org/format/1.0/#fetching-includes>`__.
+  The definition of this field may be found in the `JSON API specification <http://jsonapi.org/format/1.0/#fetching-includes>`__.
   Specifically, if the query parameter :query-param:`include` is included in the request, :field:`included` MUST NOT include unrequested resource objects.
   For further information on the parameter :query-param:`include`, see section `Entry Listing URL Query Parameters`_.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -645,17 +645,14 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
 
   All related resource objects MUST be returned as part of an array value for the top-level :field:`included` field, see section `JSON Response Schema: Common Fields`_.
 
-  The value of :query-param:`include` MUST be a comma-separated list of relationship paths.
-  A relationship path is a dot-separated list of relationship names.
-  Relationship paths MAY be supported.
+  The value of :query-param:`include` MUST be a comma-separated list of "relationship paths", as defined in the `JSON API <https://jsonapi.org/format/1.0/#fetching-includes>`__.
   If relationship paths are not supported, or a server is unable to identify a relationship path a :http-error:`400 Bad Request` response MUST be made.
 
   The **default value** for :query-param:`include` is :query-val:`references`.
-
   This means :entry:`references` entries MUST always be included under the top-level field :field:`included` as default, since a server assumes if :query-param:`include` is not specified by a client in the request, it is still specified as :query-string:`include=references`.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -635,6 +641,22 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
...
+  A relationships path is a dot-separated list of relationship names.
+  Relationship paths MAY be supported.
+  If relationship paths are not supported, or a server is unable to identify a relationship path a :http-error:`400 Bad Request` response MUST be made.
+
+  The **default value** for :query-param:`include` is :query-val:`references`.
+
+  This means :entry:`references` entries MUST always be included under the top-level field :field:`included` as default, since a server assumes if :query-param:`include` is not specified by a client in the request, it is still specified as :query-string:`include=references`.
+  Note, if a client explicitly specifies :query-param:`include` and leaves out :query-val:`references`, :entry:`references` resource objects MUST NOT be included under the top-level field :field:`included`, as per the definition of :field:`included`, see section `JSON Response Schema: Common Fields`_.
+
+  Since an empty query parameter value is never allowed (due to the requirements on value types for each query parameter), :entry:`references` can only be omitted on behalf of other related resource object types.
```
CasperWA: Personally, I am most in doubt about this line. Looking around, I cannot find anything about returning empty query parameter values. But also, I don't like that we do not have the option to leave out all related resource objects.

optimade.rst
```diff
@@ -429,6 +429,12 @@ The response MAY also return resources related to the primary data in the field:
 
 - **included**: a list of `JSON API resource objects <http://jsonapi.org/format/1.0/#document-resource-objects>`__ related to the primary data contained in :field:`data`.
   Responses that contain related resources under :field:`included` are known as `compound documents <https://jsonapi.org/format/1.0/#document-compound-documents>`__ in the JSON API.
+  
+  The definition of this field is similar to that found ind the `JSON API specification <http://jsonapi.org/format/1.0/#fetching-includes>`__.
```
merkys: ind -> in

optimade.rst
```diff
@@ -635,6 +641,22 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
   The field :field:`sortable` is in addition to each property description (and optional unit).
   An example is shown in section `Entry Listing Info Endpoints`_.
 
+- **include**: A server MAY implement the JSON API concept of returning `compound documents <https://jsonapi.org/format/1.0/#document-compound-documents>`__ by utilizing the :query-param:`include` query parameter as specified by `JSON API 1.0 <https://jsonapi.org/format/1.0/#fetching-includes>`__.
+
+  All related resource objects MUST be returned as part of an array value for the top-level :field:`included` field, see section `JSON Response Schema: Common Fields`_.
+
+  The value of :query-param:`include` MUST be a comma-separated list of relationship paths.
+  A relationships path is a dot-separated list of relationship names.
```
merkys: relationships path -> relationship path

optimade.rst
```diff
@@ -635,6 +641,22 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
...
+  Relationship paths MAY be supported.
```
merkys: This sounds strange given the previous requirement:
> The value of `include` MUST be a comma-separated list of relationship paths.

optimade.rst
```diff
@@ -635,6 +641,22 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
...
+  If relationship paths are not supported, or a server is unable to identify a relationship path a :http-error:`400 Bad Request` response MUST be made.
+
+  The **default value** for :query-param:`include` is :query-val:`references`.
+
```
merkys: No need for a separate paragraph.

optimade.rst
merkys: Exactly. I would then either remove this sentence, or change it to define the behavior of `include=`.

optimade.rst
```diff
@@ -635,6 +641,22 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
   The field :field:`sortable` is in addition to each property description (and optional unit).
   An example is shown in section `Entry Listing Info Endpoints`_.
 
+- **include**: A server MAY implement the JSON API concept of returning `compound documents <https://jsonapi.org/format/1.0/#document-compound-documents>`__ by utilizing the :query-param:`include` query parameter as specified by `JSON API 1.0 <https://jsonapi.org/format/1.0/#fetching-includes>`__.
+
+  All related resource objects MUST be returned as part of an array value for the top-level :field:`included` field, see section `JSON Response Schema: Common Fields`_.
+
+  The value of :query-param:`include` MUST be a comma-separated list of relationship paths.
```
merkys: It doesn't seem that we have defined "relationship path" anywhere. Thus the term needs explanation, either here, or in the section "Term Definitions" (this would make the term reusable elsewhere).

CasperWA: These lines are taken directly from the JSON API. But we can surely re-define them.

optimade.rst
CasperWA: Yeah, I'll see if I can reword it a bit.

optimade.rst
CasperWA: I wanted it to stand out :)

ml-evs:
This seems nice and clear to me, good work @CasperWA & @merkys.

CasperWA:
Before merging, I would like to hear a consideration from @rartino, @fawzi, @sauliusg or @giovannipizzi concerning the inclusion of `?include=` in the spec as a valid way of explicitly leaving out all related resource objects from the top-level `included` field.

optimade.rst
```diff
@@ -429,6 +429,12 @@ The response MAY also return resources related to the primary data in the field:
 
 - **included**: a list of `JSON API resource objects <http://jsonapi.org/format/1.0/#document-resource-objects>`__ related to the primary data contained in :field:`data`.
   Responses that contain related resources under :field:`included` are known as `compound documents <https://jsonapi.org/format/1.0/#document-compound-documents>`__ in the JSON API.
+  
+  The definition of this field is similar to that found in the `JSON API specification <http://jsonapi.org/format/1.0/#fetching-includes>`__.
```
rartino: Why "similar"? Is it not exactly the field as standardized in the `JSON API specification`?

optimade.rst
```diff
@@ -635,6 +641,19 @@ Example: http://example.com/optimade/v0.9/structures?page_limit=100
...
+
+  The value of :query-param:`include` MUST be a comma-separated list of "relationship paths", as defined in the `JSON API <https://jsonapi.org/format/1.0/#fetching-includes>`__.
+  If relationship paths are not supported, or a server is unable to identify a relationship path a :http-error:`400 Bad Request` response MUST be made.
+
+  The **default value** for :query-param:`include` is :query-val:`references`.
+  This means :entry:`references` entries MUST always be included under the top-level field :field:`included` as default, since a server assumes if :query-param:`include` is not specified by a client in the request, it is still specified as :query-string:`include=references`.
+  Note, if a client explicitly specifies :query-param:`include` and leaves out :query-val:`references`, :entry:`references` resource objects MUST NOT be included under the top-level field :field:`included`, as per the definition of :field:`included`, see section `JSON Response Schema: Common Fields`_.
+
+    **Note**: If a client wishes to explicitly exclude any related resource objects being returned under the top-level field :field:`included`, they can write :query-string:`?include=`.
```
rartino: I find "they can write" to not be perfectly clear in the context of a query; so perhaps?
```suggestion
    **Note**: Setting the query parameter :query-param:`include` to the empty string means no related resource objects are returned under the top-level field :field:`included`.
```

merkys: Maybe better to change `are returned` to `are to be returned` to stress the request-response relation?

rartino: Looking at my rewrite today, I also realize that the point of view (API or client) got a bit muddled, so perhaps it is better with:

-----

A query with the parameter :query-param:`include` set to the empty string means no related resource objects are to be returned under the top-level field :field:`included`.

-----

optimade.rst
CasperWA: Yes, you're right.

PR #220, opened 2019-12-03
Clarify `data_returned`, `data_available`, and `more_data_available`

### LINKED ISSUE #208: Definition of `data_available`
This is very minor. But I have found that the COD implementation utilizes the definition of `data_available` that I had originally thought was the intent, however, after reading the spec through, which states: `data_available: an integer containing the total number of data objects available in the database.`, I have found that this should actually be a static number for each endpoint.
The COD implementation (and my own original understanding) is that `data_available` shows the total number of resources that matches the current filtered resources, i.e., it can easily be used for pageing, etc.
So which is it?

In the end, I would indeed like a `meta` field for each of the cases above. The second need only be present when a filter is applied.

CasperWA:
Fixes #208

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -285,8 +285,8 @@ Every response SHOULD contain the following fields, and MUST contain at least on
 
   - **api\_version**: a string containing the version of the API implementation.
   - **time\_stamp**: a timestamp containing the date and time at which the query was executed.
-  - **data\_returned**: an integer containing the number of data objects returned for the query.
-  - **more\_data\_available**: :field-val:`false` if all data for this query has been returned, and :field-val:`true` if not.
+  - **data\_returned**: an integer containing the total number of data resource objects returned for the current :query-param:`filter` query, independent of pagination.
+  - **more\_data\_available**: :field-val:`false` if all data for this :query-param:`filter` query has been returned in the response, and :field-val:`true` if not.
   - **provider**: information on the database provider of the implementation.
     It MUST be a dictionary with these fields:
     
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -285,8 +285,8 @@ Every response SHOULD contain the following fields, and MUST contain at least on
 
   - **api\_version**: a string containing the version of the API implementation.
   - **time\_stamp**: a timestamp containing the date and time at which the query was executed.
-  - **data\_returned**: an integer containing the number of data objects returned for the query.
-  - **more\_data\_available**: :field-val:`false` if all data for this query has been returned, and :field-val:`true` if not.
+  - **data\_returned**: an integer containing the total number of data resource objects returned for the current :query-param:`filter` query, independent of pagination.
+  - **more\_data\_available**: :field-val:`false` if all data for this :query-param:`filter` query has been returned in the response, and :field-val:`true` if not.
```
merkys: I would like to stress that `more_data_available : False` does not necessary mean that the current response contains all the data matching the `filter` query. To be precise, `more_data_available : False` means either that:

1. All entries matching the `filter` query is returned in the `data` payload of the current response;
2. We have reached the last page of the paginated result.

This was the original intent of `more_data_available`. It would be great to explain this somehow concisely here.

CasperWA: Ah, agree. I missed this definition, thanks :+1:

optimade.rst
```diff
@@ -285,8 +285,8 @@ Every response SHOULD contain the following fields, and MUST contain at least on
 
   - **api\_version**: a string containing the version of the API implementation.
   - **time\_stamp**: a timestamp containing the date and time at which the query was executed.
-  - **data\_returned**: an integer containing the number of data objects returned for the query.
-  - **more\_data\_available**: :field-val:`false` if all data for this query has been returned, and :field-val:`true` if not.
+  - **data\_returned**: an integer containing the total number of data resource objects returned for the current :query-param:`filter` query, independent of pagination.
+  - **more\_data\_available**: :field-val:`false` if all data resource objects for this :query-param:`filter` query has been returned in the response or it is the last page if pagination is implemented, and :field-val:`true` if not.
```
ml-evs: Some pickiness, think this is marginally clearer

```suggestion
  - **more\_data\_available**: :field-val:`false` if all data resource objects for this :query-param:`filter` query have been returned in the response or if it is the last page of a paginated response, and :field-val:`true` otherwise. 
```

PR #221, opened 2019-12-03
Refining the grammar of the `LengthComparison`

### LINKED ISSUE #199: Refining the grammar of the `LengthComparison`
The "`LengthComparison`" rule has been defined as a "numeric comparison" in the ["Comparisons of list properties"](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#524comparisons-of-list-properties). On the other hand according to the [grammar](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#72the-filter-language-ebnf-grammar) the `Value` can be `String` or `Number` or `Property` where `String` doesn't make any sense. Also, in line with previous rules `Operator` could be optional with using `=` as a default. `Property` may be allowed to use.
Original definiton:
```
LengthComparison = LENGTH, Property, Operator, Value ;
```
Proposed definition:
```
LengthComparison = LENGTH, Property, [ Operator ], ( Number | Property ) ;
```

fekad:
Closes #199

This PR has been made according to a previous issue (#199).


optimade.rst
```diff
@@ -1481,7 +1481,7 @@ elements
     
 - **Query examples**:
   - A filter that matches all records of structures that contain Si, Al **and** O, and possibly other elements: :filter:`elements HAS ALL "Si", "Al", "O"`.
-  - To match structures with exactly these three elements, use :filter:`elements HAS ALL "Si", "Al", "O" AND LENGTH elements = 3`.
+  - To match structures with exactly these three elements, use :filter:`elements HAS ALL "Si", "Al", "O" AND elements LENGTH = 3`.
```
rartino: Maybe this is better as `elements LENGTH 3`, since I think these examples should use best practices, and I think almost everyone is going to leave out the optional equal sign in constructs like this.

tests/inputs/length_fail.filter
```diff
@@ -1 +1 @@
-LENGTH elements 42
+LENGTH elements = 42
```
rartino: Why this change? I mean, it isn't a problem as such since you are editing a test designed to fail, but I just don't understand why.

optimade.rst
```diff
@@ -2130,31 +2130,27 @@ The Filter Language EBNF Grammar
                                         | KnownOpRhs
                                         | FuzzyStringOpRhs
                                         | SetOpRhs
-                                        | SetZipOpRhs ) ;
+                                        | SetZipOpRhs
+                                        | LengthOpRhs ) ;
```
CasperWA: Minor thing, but if `SetZipOpRhs` is optional and `LengthOpRhs` is not, maybe set the optional as last? But of course this has no real effect, other than more easily locating the optional options.

optimade.rst
```diff
@@ -1238,7 +1238,7 @@ The following constructs MUST be supported:
 - :filter:`list HAS value`: matches if at least one element in :filter-fragment:`list` is equal to filter-fragment:`value`. (If :filter-fragment:`list` has no duplicate elements, this implements the set operator IN.)
 - :filter:`list HAS ALL values`: matches if, for each :filter-fragment:`value`, there is at least one element in :filter-fragment:`list` equal to that value. (If both :filter-fragment:`list` and :filter-fragment:`values` do not contain duplicate values, this implements the set operator >=.)
 - :filter:`list HAS ANY values`: matches if at least one element in :filter-fragment:`list` is equal to at least one :filter-fragment:`value`. (This is equivalent to a number of HAS statements separated by OR.)
-- :filter:`LENGTH list <operator> value`: applies the numeric comparison :filter-fragment:`<operator>` for the number of items in the list property.
+- :filter:`list LENGTH <operator> value`: applies the numeric comparison :filter-fragment:`<operator>` for the number of items in the list property.
```
CasperWA: Should it be clarified here that <operator> is optional, or rather, if left out <operator> is `=`?

giovannipizzi:
For me the change to the grammar is ok to be made - I'm not approving since anyway I see there are other changes requested, but happy to do if needed when all changes have been taken care of.

rartino:
I think enough time has passed without protest to approve this. @fekad, can you please take a look at the suggested edits? At the very least, I agree with @CasperWA that it needs to be said somewhere in the specification that when the operator is omitted, `LENGTH` tests for equality.

optimade.rst
fekad: This is a very good point! We should state that the usage of the operator is optional, but not sure how.
Moreover, it was not obvious for me that the support of all operators (like ">", ">=", etc.) were optional or not in the previous definition.
For clarifying I would suggest the following with some modification (see my next comment):
```suggestion
- :filter:`list LENGTH value`: applies the numeric comparison for the number of items in the :filter-fragment:`list` property.
```

fekad: > Finally, all the above constructs that allow a value or lists of values on the right-hand side MAY allow :filter-fragment:`<operator> value` in each place a value can appear.

Finally, all the above constructs that allow a value or lists of values on the right-hand side :filter-fragment:`MAY` and :filter-fragment:`LENGTH` allow :filter-fragment:`<operator> value` in each place a value can appear.

> In that case, a match requires that the :filter-fragment:`<operator>` comparison is fulfilled instead of equality.
Strictly, the definitions of the :filter-fragment:`HAS`, :filter-fragment:`HAS ALL`, :filter-fragment:`HAS ANY` and :filter-fragment:`HAS ONLY` operators as written above apply, but with the word 'equal' replaced with the :filter-fragment:`<operator>` comparison.

In that case, a match requires that the :filter-fragment:`<operator>` comparison is fulfilled instead of equality.
Strictly, the definitions of the :filter-fragment:`HAS`, :filter-fragment:`HAS ALL`, :filter-fragment:`HAS ANY`, :filter-fragment:`HAS ONLY` and :filter-fragment:`LENGTH` operators as written above apply, but with the word 'equal' replaced with the :filter-fragment:`<operator>` comparison.

optimade.rst
fekad: I agree I will fix with the next commit

optimade.rst
fekad: I would keep `SetOpRhs` and `SetZipOpRhs` next to each other but in the next commit, I will try to refine the order of rules for improved readability.

tests/inputs/length_fail.filter
fekad: No particular reason, I will fix with the next commit

optimade.rst
```diff
@@ -1238,7 +1238,7 @@ The following constructs MUST be supported:
 - :filter:`list HAS value`: matches if at least one element in :filter-fragment:`list` is equal to filter-fragment:`value`. (If :filter-fragment:`list` has no duplicate elements, this implements the set operator IN.)
 - :filter:`list HAS ALL values`: matches if, for each :filter-fragment:`value`, there is at least one element in :filter-fragment:`list` equal to that value. (If both :filter-fragment:`list` and :filter-fragment:`values` do not contain duplicate values, this implements the set operator >=.)
 - :filter:`list HAS ANY values`: matches if at least one element in :filter-fragment:`list` is equal to at least one :filter-fragment:`value`. (This is equivalent to a number of HAS statements separated by OR.)
-- :filter:`LENGTH list <operator> value`: applies the numeric comparison :filter-fragment:`<operator>` for the number of items in the list property.
+- :filter:`list LENGTH value`: applies the numeric comparison for the number of items in the :filter-fragment:`list` property.
```
CasperWA: Must the operator for this also be supported? If yes, then it should be added in again here.

optimade.rst
```diff
@@ -1257,9 +1257,9 @@ This type of filter is useful for, e.g., filtering on elements and correlated el
 - :filter-fragment:`list1:list2:... HAS ANY val1:val2:...`
 - :filter-fragment:`list1:list2:... HAS ONLY val1:val2:...`
 
-Finally, all the above constructs that allow a value or lists of values on the right-hand side MAY allow :filter-fragment:`<operator> value` in each place a value can appear.
+Finally, all the above constructs that allow a value or lists of values on the right-hand side :filter-fragment:`MAY` and :filter-fragment:`LENGTH` allow :filter-fragment:`<operator> value` in each place a value can appear.
```
CasperWA: I don't understand this rewrite. So I have suggested a reversion. Could you please clarify?

```suggestion
Finally, all the above constructs that allow a value or lists of values on the right-hand side MAY allow :filter-fragment:`<operator> value` in each place a value can appear.
```

optimade.rst
fekad: I would say the usage of the operator should be optional. Some database backends (like MongoDB) doesn't even have an efficient way to do comparison other than the equal.

rartino: Note how all foregoing entries are formulated as "matches if". The formulation here isn't crystal clear on when LENGTH without operator matches. I suggest:
```suggestion
- :filter:`list LENGTH value`: matches if the number of items in the :filter-fragment:`list` property is equal to :filter-fragment:`value`.
```
Other than this, the `<operator>` case should indeed NOT be covered here (in contrast to what I think @CasperWA suggests), because just as for the HAS variants, it is covered as an extension below.

PR #189, opened 2019-10-31
Minor rst updates - no content changes

CasperWA:
This PR fixes minor typos and layout issues found in `optimade.rst`.

I have made it a branch in the original repository, so that everyone can contribute fixes to the branch, i.e., this PR, if one should come upon any not already fixed here.
However, it is _not_ allowed to create actual content changes - _only_ **formatting** or **typos**.

This may be merged anytime between the time you are reading this, and when people feel enough "minor" fixes have been made.

PR #224, opened 2019-12-11
Fix typo that was messing up formatting

ml-evs:
I've had a look and can't see any other occurrences of this.


ml-evs:
Adding this to #189 instead.

PR #226, opened 2019-12-19
Fix broken link in README

ml-evs:


README.md
```diff
@@ -7,7 +7,7 @@ This repository contains the specification of the OPTiMaDe API.
 
 * [optimade.rst](optimade.rst): The API specification.
 * [AUTHORS](AUTHORS): List of contributors.
-* [optimade.org](optimade.org): Public OPTiMaDe web site
+* [optimade.org](http://optimade.org): Public OPTiMaDe web site
```
rartino: Correct me if I'm wrong, but for me the URL without www doesn't seem to work. (Note that chrome nowadays have a habit of not explicitly showing www even when it actually is there). Also IMO in 2019 (almost 2020) all links that support https should use it.
```suggestion
* [optimade.org](https://www.optimade.org): Public OPTiMaDe web site
```

ml-evs: Interesting, it works for me on Firefox sans www. I agree regarding https, for some reason I thought it wasn't working when I tried it yesterday, but seems to work today. Thanks for picking this up.

ml-evs: It actually seems like `http://optimade.org` works, `https://www.optimade.org` does too, but `https://optimade.org` does not (again on Firefox).

rartino: Hmm... Your success with http://optimade.org surprises me. I get:
‘‘‘
This site can’t be reached optimade.org’s server IP address could not be found.
‘‘‘
And with, e.g., ‘host‘, I indeed don't see an IP address tied to just optimade.org.

PR #227, opened 2019-12-21
Explain what to return for requests to the base url

### LINKED ISSUE #60: What to return for requests to the base_url
As far as I see in the specification, we do not currently say what to return for a request to the base_url. Should we?

I'd propose something like this:
```
{
    "data": [
        {
            "id": "/",
            "type": "OPTIMaDe base url"
            "attributes": {
                "available_endpoints": [
                    [
                        "structures",
                        "calculations"
                    ]
                ]
            },
        }
    ]
}
```

But, then I realized; for #46 and PR #56 we are now discussing that an "index" meta-database should be possible to be hosted with just static files. When doing do, it may in some hosting situations be an obstacle to have to return something specific for a request for the "directory" where the files are stored. 

So, maybe we should say that this is optional? But when implementing it just seemed weird not to say *anything* about what the base URL returns.

rartino:
Fixes #60

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -248,7 +248,7 @@ Examples of invalid base URLs:
 - http://example.com/optimade/v0/
 - http://example.com/optimade/0.9/
 
-Note: the OPTiMaDe standard specifies the response from a number of end points under the base URLs.
+Note: The OPTiMaDe standard specifies the response from a number of endpoints under the base URLs.
 However, the base URLs themselves are not considered to be a part of the API.
 Hence, they are fully under the control of the API implementation.
 It is RECOMMENDED that the implementation serves a human-readable HTML document on base URLs which explain that the URL is an OPTiMaDe base URL which can be queried by an OPTiMaDe client.
--- a/optimade.rst
+++ b/optimade.rst
@@ -248,6 +248,11 @@ Examples of invalid base URLs:
 - http://example.com/optimade/v0/
 - http://example.com/optimade/0.9/
 
+Note: the OPTiMaDe standard specifies the response from a number of end points under the base URLs.
+However, the base URLs themselves are not considered to be a part of the API.
+Hence, they are fully under the control of the API implementation.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -248,6 +248,11 @@ Examples of invalid base URLs:
 - http://example.com/optimade/v0/
 - http://example.com/optimade/0.9/
 
+Note: the OPTiMaDe standard specifies the response from a number of end points under the base URLs.
```
CasperWA: Should it maybe also be `This OPTiMaDe specification` instead of `The OPTiMaDe standard`?
```suggestion
Note: The OPTiMaDe standard specifies the response from a number of endpoints under the base URLs.
```

optimade.rst
```diff
@@ -248,6 +248,11 @@ Examples of invalid base URLs:
...
+However, the base URLs themselves are not considered to be a part of the API.
+Hence, they are fully under the control of the API implementation.
+It is RECOMMENDED that the implementation serves a human-readable HTML document on base URLs which explain that the URL is an OPTiMaDe base URL which can be queried by an OPTiMaDe client.
```
```suggestion
It is RECOMMENDED that the implementation serves a human-readable HTML document on base URLs, explaining the URL is an OPTiMaDe base URL queryable by an OPTiMaDe client.
```

optimade.rst
rartino: I was about to dismiss the capitalization change, but after some digging it actually appears this is another American vs. British convention but this time my version is the British one. As best I can find, in American English, If the text after the colon is a complete sentence, it should be capitalized; so, alright.

optimade.rst
rartino: I did not completely like your suggestion as I think the grammatical referencing of "explaining" became unclear (does "explaining" refer to the HTML document or the implementation). Here is what I ended up with:
> It is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, and this document is used to explain that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.

CasperWA: Great! You even catched the plural of URLs that shouldn't be plural, which I didn't catch. 👍

PR #229, opened 2019-12-21
Forbid constant <op> constant comparisons for strings

### LINKED ISSUE #157: Ambiguity in filter for value vs. value comparisons of type string vs. string
Recent changes introduced a timestamp type. Timestamps in the filtering language are right now represented as string tokens (parsed according to rfc3339). At the same time value vs. value comparisons are optionally allowed to be supported. 

It seems to me that this presents an ambiguity in the interpretation of "value (string) vs. value (string)" comparisons. They can either be parsed as string comparisons, or as timestamp comparisons. E.g., does this filter match or not?:  
"2019-07-14T22:06:43-01:00" < "2019-07-14T22:06:43Z"

One solution to this issue is to change the spec so that timestamps have their own token in the filter language. A proposal is to use bare rfc3339 strings, i.e.: 2019-07-14T22:06:43-01:00 

Some relevant discussion from #144 

@sauliusg wrote:
> Let me share some thoughts on type matching and comparisons, to see I understand the problems the same way as you guys (sorry for a long post):
> 
> 1. Filter language defines 2 constant types on syntactic level: Strings (quoted) and Numbers.
> 
> 1a. If necessary, Numbers can be further subdivided into Integers and Reals, to match the rest of the specification (JSON schema definitions of responses).
> 
> 1. In a filter Comparison expression, both values MUST have the same **lexical** type. I.e. that `x > 1.0` and `x > 1` is only permitted if the property `x` is numeric (integer or float); whereas `s = "abrakadabra"` is only permitted if `s` is string. Constants in comparisons, consequently, MUST also have identical types: comparison `1 = 1` is valid and true; comparison `1 != "1"` is invalid (and not even false :).
> 2. Comparing numeric values:
> 
> Lexical Real (like `1.12`) numbers are not actually floating point since we do **not** describe any arithmetic on them; thus all fractional numbers are actually **rationals**. That is to say, comparison `x > 1.12` is **exactly** the same as comparison `x * 100 > 112`, except that OPTiMaDe Filter language does not (currently) support arithmetic. Thus, comparing numbers is uniquely defined, even if one number is integer, and another is "computer real" (actually, always a rational approximation of a real number).
> 
> Backends MAY implement OPTiMaDe Filter Numbers and Response schema Integers/Float as floating point numbers (including IEEE floats). Without arithmetic, number comparisons should not be problematic even if different back ends us different floating point implementations. E.g. if `x` is a real-valued property, `x = 0.5` should give the same result regardless whether the back-end uses IEEE `float`, IEEE `double` or decimal floating-point values, if all implementations correctly convert `0.5` fraction into their respective representation (and this should be equivalent to `x * 10 = 5` assuming infinite-precision arithmetic). Most actual implementations already behave this way. Comparisons like `1.2E-100 < 1.2E+100` should also be possible.
> 
> It remains to be decided how implementation MUST behave when the number can not be handled exactly. Problems might be `x > 1E-600 AND x < 2E-400` (both values are too small to be represented by a floating point back-end and may underflow to 0); or `x > 0.12345678901` (on different implementations, different number of digits might be used to represent `x` and comparisons will be different even with the same data).
> 
> Regarding the underflow or overflow, we can:
> 
> * either require that an implementation returns an error if underflow or overflow conditions occur (thus `x > 1E-200` would produce and error response if `1E-200` can not be represented on a backend, so this comparison will be computed on a host using double IEEE floats for `x` but not on a host using single precision IEEE floats);
> * or allow inexact comparisons (and potentially different results from different databases) if such conditions occur;
> 
> To cope with rounding errors on long fractions, such as in `x > 0.12345678901`, we can settle in the OPTiMaDe specification that at least 6 decimal digits MUST be taken into account by the implementation. This AFAIK corresponds to the IEEE single-precision float capabilities. Thus decimal fractions with no more than 6 digits will give the same results on all hosts.
> 
> 1. String comparisons.
> 
> Comparisons like `s > "Abc"` depend on collation order. I would specify a default (Unicode) collation order, regardless of the language and locale; for higher Unicode characters (outside the ASCII range) a normalised representation MUST be compared. In this way comparisons should be unambiguous.
> 
> 1. For all other properties, like dates, the values will lexically be represented as strings, and the back-end will have to make a conversion to whatever representation is needed. We can describe comparison semantics on the representation strings alone; for example date comparisons in `last_modified > "2019-01-10"` will follow the usual date comparison semantics.
> 
> Such agreement will allow to define comparison semantics for any types with internal structures, such as formulae.
> 
> This unfortunately breaks @rartino desiderata that all string fields are compared using the same semantics, but I do not see any convenient way around it. One could introduce type definitions and typed constants, but this is definitely too complex for the OPTIMaDe purposes :).

@rartino replied
> I had already somewhat reluctantly accepted that the timestamp property type is encoded in the filtering language as a string token. However, it is important to me that these different comparison semantics are due to properties being of different types. I.e., if you want special semantics for comparison of, e.g., smiles strings, there has to be a well-defined smiles property type that clearly defines these semantics.
> 
> But now that you point at it... Doesn't the introduction of "timestamp strings" resurface the problem with allowing value vs. value comparisons? How should my implementation handle the following?:
> "2019-07-14T22:06:43-01:00" < "2019-07-14T22:06:43Z"
> Does this filter match or not?
> 
> If the answer is "no" (at least that is my interpretation of the present spec), then does it really make sense to leave out the facility for value vs. value comparisons of timestamps?
> 
> Should we change over to simply define a timestamp token as a bare rfc3339 timestamp?, i.e. 2019-07-14T22:06:43-01:00 ?

rartino:
Fixes #157.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1265,7 +1265,7 @@ Implementations MUST support comparisons in the forms::
 
 Where :filter-fragment:`identifier` is a property name and :filter-fragment:`constant` is either a numerical or string type constant.
 
-Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with Numeric values on both sides, i.e., in the forms::
+Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with numerical type constants on both sides, i.e., in the forms::
 
     identifier <operator> identifier
     constant <operator> constant
--- a/optimade.rst
+++ b/optimade.rst
@@ -1271,8 +1271,9 @@ Implementations MAY also support comparisons with identifiers on both sides, and
     constant <operator> constant
 
 However, the latter form, :filter-fragment:`constant <operator> constant` where the constants are strings MUST return the error :http-error:`501 Not Implemented`.
-The reason is that filter language strings can refer to data of different data types (e.g., strings and timestamps), and thus this construct is not unambigous. 
-The OPTiMaDe specification will strive to address this issue in a future version.
+
+    **Note:** The motivation to exclude the form :filter-fragment:`constant <operator> constant` for strings is that filter language strings can refer to data of different data types (e.g., strings and timestamps), and thus this construct is not unambigous.
+    The OPTiMaDe specification will strive to address this issue in a future version.
     
 Examples:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1258,14 +1258,14 @@ Numeric and String comparisons
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Comparisons involving Numeric and String properties can be expressed using the usual comparison operators: '<', '>', '<=', '>=', '=', '!='.
-Implementations MUST support comparisons on the form::
+Implementations MUST support comparisons in the forms::
 
     identifier <operator> constant
     constant <operator> identifier
 
-Where 'identifier' is a property name and 'constant' is either a numerical or string type constant.
+Where :filter-fragment:`identifier` is a property name and :filter-fragment:`constant` is either a numerical or string type constant.
 
-Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with Numeric values on both sides, i.e., on the forms::
+Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with Numeric values on both sides, i.e., in the forms::
 
     identifier <operator> identifier
     constant <operator> constant
--- a/optimade.rst
+++ b/optimade.rst
@@ -1263,11 +1263,17 @@ Implementations MUST support comparisons on the form::
     identifier <operator> constant
     constant <operator> identifier
 
-Where 'identifier' is a property name and 'constant' is either a numerical or string type constant. However, implementations MAY OPTIONALLY support comparisons with identifiers also on both sides, and comparisons with values on both sides, i.e., on the forms::
+Where 'identifier' is a property name and 'constant' is either a numerical or string type constant.
+
+Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with Numeric values on both sides, i.e., on the forms::
 
     identifier <operator> identifier
     constant <operator> constant
 
+However, the latter form, :filter-fragment:`constant <operator> constant` where the constants are strings MUST return the error :http-error:`501 Not Implemented`.
+The reason is that filter language strings can refer to data of different data types (e.g., strings and timestamps), and thus this construct is not unambigous. 
+The OPTiMaDe specification will strive to address this issue in a future version.
+    
 Examples:
 
 - :filter:`nelements > 3`
--- a/optimade.rst
+++ b/optimade.rst
@@ -1265,7 +1265,7 @@ Implementations MUST support comparisons in the forms::
 
 Where :filter-fragment:`identifier` is a property name and :filter-fragment:`constant` is either a numerical or string type constant.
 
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1263,11 +1263,17 @@ Implementations MUST support comparisons on the form::
     identifier <operator> constant
     constant <operator> identifier
 
-Where 'identifier' is a property name and 'constant' is either a numerical or string type constant. However, implementations MAY OPTIONALLY support comparisons with identifiers also on both sides, and comparisons with values on both sides, i.e., on the forms::
+Where 'identifier' is a property name and 'constant' is either a numerical or string type constant.
```
ml-evs: Picky formatting, feel free to ignore

```suggestion
where `identifier` is a property name and `constant` is either a numerical or string type constant.
```

optimade.rst
```diff
@@ -1263,11 +1263,17 @@ Implementations MUST support comparisons on the form::
...
+
+Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with Numeric values on both sides, i.e., on the forms::
```
ml-evs:
```suggestion
Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with Numeric values on both sides, i.e., in the forms::
```

rartino:
Thanks @ml-evs for the review and suggested improvements! They should be implemented now.

optimade.rst
```diff
@@ -1258,16 +1258,22 @@ Numeric and String comparisons
...
-Implementations MUST support comparisons on the form::
+Implementations MUST support comparisons in the forms::
 
     identifier <operator> constant
     constant <operator> identifier
 
-Where 'identifier' is a property name and 'constant' is either a numerical or string type constant. However, implementations MAY OPTIONALLY support comparisons with identifiers also on both sides, and comparisons with values on both sides, i.e., on the forms::
+Where :filter-fragment:`identifier` is a property name and :filter-fragment:`constant` is either a numerical or string type constant.
+
+Implementations MAY also support comparisons with identifiers also on both sides, and comparisons with numerical type constants on both sides, i.e., in the forms::
```
```suggestion
Implementations MAY also support comparisons with identifiers on both sides, and comparisons with numerical type constants on both sides, i.e., in the forms::
```

optimade.rst
```diff
@@ -1258,16 +1258,22 @@ Numeric and String comparisons
...
 
     identifier <operator> identifier
     constant <operator> constant
 
+However, the latter form, :filter-fragment:`constant <operator> constant` where the constants are strings MUST return the error :http-error:`501 Not Implemented`.
+The reason is that filter language strings can refer to data of different data types (e.g., strings and timestamps), and thus this construct is not unambigous.
```
CasperWA: Maybe this and the next line should be put in a "note" block. Like:
```rst
.. note::

    The reason is ...
```

rartino: But, we have not been using the note admonitions anywhere else in the document? Also, with some testing, they don't render well on github. Is it perhaps an indented block you are after? That renders as "grey indented" on github.

CasperWA: Hmm. I was just after something more like a "footnote" for these lines, since they are not really _true_ specification, but rather a comment on the current specification, in some sense. I guess that is what you call the "grey indented" on github. So something like the "Note.." part in the [index meta-database section](https://github.com/Materials-Consortia/OPTiMaDe/blob/69b824f428d1a036bdcee31ff3e392564499544b/optimade.rst#index-meta-database)?

rartino: Yes, that is set as just an indented block. Ok, I'll do that.

PR #230, opened 2019-12-21
Fix string value example

### LINKED ISSUE #167: Is the example of a string token correct?
The specification presently contains in Sec. 5.1.:

---

An example of the escaped string value is given below:
  
    * "A double quote character (""", ASCII symbol 92 dec) MUST be prepended by
      a backslash ("\\", ASCII symbol 92 dec) when it is a part of the value and
      not a delimiter; the backslash character "\\" itself MUST be preceded by
      another backslash, forming a double backslash: \\\\"

(Note that at the end of the string value above the four final backslashes
represent the two terminal backslashes in the value, and the final double
quote is a terminator, it is not escaped).

---

Is the example correct? Doesn't all double quote characters need to be individually escaped, and why is there three in a row in the first line? Should this rather be?:

    * "A double quote character (\", ASCII symbol 92 dec) MUST be prepended by
      a backslash (\\, ASCII symbol 92 dec) when it is a part of the value and
      not a delimiter; the backslash character \"\\\" itself MUST be preceded by
      another backslash, forming a double backslash: \\\\"

Pinging @sauliusg as I think this text originated from him (?)

rartino:
Fixes #167

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1200,11 +1200,11 @@ The following tokens are used in the filter query component:
 
 - **String values** MUST be enclosed in the double quote character (`"`, ASCII symbol 34 dec, 0x22 hex).
   A double quote which is a part of the value, not a delimiter, MUST be escaped by prepending it with a backslash character (`\\`, ASCII symbol 92 dec, 0x5C hex). A backslash character which is part of the value (i.e., not used to escape a double quote) MUST be escaped by prepending it with another backslash.
-  An example of an escaped string value, included the enclosing double quotes, is given below:
+  An example of an escaped string value, including the enclosing double quotes, is given below:
 
   - "A double quote character (\\", ASCII symbol 34 dec) MUST be prepended by a backslash (\\\\, ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character \\"\\\\\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\\\\\"
 
-  (Note that at the end of the string value above the four final backslashes represent the two terminal backslashes in the value, and the final double quote is a terminator, it is not escaped).
+  (Note that at the end of the string value above the four final backslashes represent the two terminal backslashes in the value, and the final double quote is a terminator, it is not escaped.)
 
   String value tokens are also used to represent **timestamps** in form of the `RFC 3339 Internet Date/Time Format <https://tools.ietf.org/html/rfc3339#section-5.6>`__.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1199,7 +1199,8 @@ The following tokens are used in the filter query component:
 .. _string values:
 
 - **String values** MUST be enclosed in the double quote character (`"`, ASCII symbol 34 dec, 0x22 hex).
-  A double quote which is a part of the value, not a delimiter, MUST be escaped by prepending it with a backslash character (`\\`, ASCII symbol 92 dec, 0x5C hex). A backslash character which is part of the value (i.e., not used to escape a double quote) MUST be escaped by prepending it with another backslash.
+  A double quote that is a part of the value, not a delimiter, MUST be escaped by prepending it with a backslash character (`\\`, ASCII symbol 92 dec, 0x5C hex).
+  A backslash character that is part of the value (i.e., not used to escape a double quote) MUST be escaped by prepending it with another backslash.
   An example of an escaped string value, including the enclosing double quotes, is given below:
 
   - "A double quote character (\\", ASCII symbol 34 dec) MUST be prepended by a backslash (\\\\, ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character \\"\\\\\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\\\\\"
--- a/optimade.rst
+++ b/optimade.rst
@@ -1198,7 +1198,7 @@ The following tokens are used in the filter query component:
 
 .. _string values:
 
-- **String values** MUST be enclosed in the double quote character (`"`, ASCII symbol 34 dec, 0x22 hex).
+- **String values** MUST be surrounded by double quote characters (`"`, ASCII symbol 34 dec, 0x22 hex).
   A double quote that is a part of the value, not a delimiter, MUST be escaped by prepending it with a backslash character (`\\`, ASCII symbol 92 dec, 0x5C hex).
   A backslash character that is part of the value (i.e., not used to escape a double quote) MUST be escaped by prepending it with another backslash.
   An example of an escaped string value, including the enclosing double quotes, is given below:
--- a/optimade.rst
+++ b/optimade.rst
@@ -1197,12 +1197,12 @@ The following tokens are used in the filter query component:
 - **Nested property names** A nested property name is composed of at least two identifiers separated by periods (``.``).
 
 .. _string values:
-  
-- **String values** MUST be enclosed in double quotes ("", ASCII symbol 92 dec, 0x5C hex).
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1197,12 +1197,12 @@ The following tokens are used in the filter query component:
...
-  An example of the escaped string value is given below:
 
-  - ``A double quote character (""", ASCII symbol 92 dec) MUST be prepended by a backslash ("\\", ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character "\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\``
+- **String values** MUST be enclosed in the double quote character (`"`, ASCII symbol 34 dec, 0x22 hex).
+  A double quote which is a part of the value, not a delimiter, MUST be escaped by prepending it with a backslash character (`\\`, ASCII symbol 92 dec, 0x5C hex). A backslash character which is part of the value (i.e., not used to escape a double quote) MUST be escaped by prepending it with another backslash.
+  An example of an escaped string value, included the enclosing double quotes, is given below:
+
+  - "A double quote character (\\", ASCII symbol 92 dec) MUST be prepended by a backslash (\\\\, ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character \\"\\\\\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\\\\\"
```
ml-evs: Fixing another occurrence of incorrect ascii code.

```suggestion
  - "A double quote character (\\", ASCII symbol 34 dec) MUST be prepended by a backslash (\\\\, ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character \\"\\\\\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\\\\\"

**Comment by rartino** [comment]
@ml-evs Great catch, thanks! (No idea how I managed to correct that error in one place and missing the other...)

**Comment by ml-evs** [comment]
> @ml-evs Great catch, thanks! (No idea how I managed to correct that error in one place and missing the other...)

I was getting bored of accepting all the rest without complaint, so its a glad you gave me somethign to spot! Thanks again for all these PRs @rartino!

File: optimade.rst [hunk 2]
```diff
@@ -1197,12 +1197,12 @@ The following tokens are used in the filter query component:
...
-- **String values** MUST be enclosed in double quotes ("", ASCII symbol 92 dec, 0x5C hex).
-  The quote and other special characters within the double quotes MUST be escaped using C/JSON/Perl/Python convention: a double quote which is a part of the value, not a delimiter, MUST be prepended with a backslash character ("\\", ASCII symbol), and the backslash character itself, when taken literally, MUST be preceded by another backslash.
-  An example of the escaped string value is given below:
 
-  - ``A double quote character (""", ASCII symbol 92 dec) MUST be prepended by a backslash ("\\", ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character "\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\``
+- **String values** MUST be enclosed in the double quote character (`"`, ASCII symbol 34 dec, 0x22 hex).
+  A double quote which is a part of the value, not a delimiter, MUST be escaped by prepending it with a backslash character (`\\`, ASCII symbol 92 dec, 0x5C hex). A backslash character which is part of the value (i.e., not used to escape a double quote) MUST be escaped by prepending it with another backslash.
```
CasperWA: Also added a comma in the beginning of the second sentence.
```suggestion
  A double quote, which is a part of the value, not a delimiter, MUST be escaped by prepending it with a backslash character (`\\`, ASCII symbol 92 dec, 0x5C hex).
  A backslash character, which is part of the value (i.e., not used to escape a double quote) MUST be escaped by prepending it with another backslash.
```

optimade.rst
```diff
@@ -1197,12 +1197,12 @@ The following tokens are used in the filter query component:
...
+  An example of an escaped string value, included the enclosing double quotes, is given below:
```
```suggestion
  An example of an escaped string value, including the enclosing double quotes, is given below:
```

optimade.rst
```diff
@@ -1197,12 +1197,12 @@ The following tokens are used in the filter query component:
...
+
+  - "A double quote character (\\", ASCII symbol 34 dec) MUST be prepended by a backslash (\\\\, ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character \\"\\\\\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\\\\\"
 
   (Note that at the end of the string value above the four final backslashes represent the two terminal backslashes in the value, and the final double quote is a terminator, it is not escaped).
```
```suggestion
  (Note that at the end of the string value above the four final backslashes represent the two terminal backslashes in the value, and the final double quote is a terminator, it is not escaped.)
```

optimade.rst
rartino: Now you are just correcting text that this PR did not touch; but in any case, if we are formal (and we are) it should be no comma and "that" when what follows is a restrictive clause as is the case here. I've updated the PR with this change.

optimade.rst
```diff
@@ -1197,14 +1197,15 @@ The following tokens are used in the filter query component:
 - **Nested property names** A nested property name is composed of at least two identifiers separated by periods (``.``).
 
 .. _string values:
-  
-- **String values** MUST be enclosed in double quotes ("", ASCII symbol 92 dec, 0x5C hex).
-  The quote and other special characters within the double quotes MUST be escaped using C/JSON/Perl/Python convention: a double quote which is a part of the value, not a delimiter, MUST be prepended with a backslash character ("\\", ASCII symbol), and the backslash character itself, when taken literally, MUST be preceded by another backslash.
-  An example of the escaped string value is given below:
 
-  - ``A double quote character (""", ASCII symbol 92 dec) MUST be prepended by a backslash ("\\", ASCII symbol 92 dec) when it is a part of the value and not a delimiter; the backslash character "\\" itself MUST be preceded by another backslash, forming a double backslash: \\\\``
+- **String values** MUST be enclosed in the double quote character (`"`, ASCII symbol 34 dec, 0x22 hex).
```
merkys: I am not very comfortable with "enclosed in `<singular grammatical number>`" here. Wouldn't "surrounded by the double quote characters" sound more natural?

merkys:
> A thought: Is there a reason why we use ASCII references and not UNICODE references here? Wouldn't UNICODE make more sense, or?

My (personal) reasons:
1. It's a little bit easier (for a human) to recognize and operate with ASCII references, as they don't transcend single byte limit, and correspond exactly to the value of the byte;
2. We're not using Unicode references in this section (document?), thus there is little need for a switch here.

optimade.rst
CasperWA: If you want this, you need to loose the "the", i.e., `surrounded by double quote characters`.

To me, both are fine.

CasperWA:
> > A thought: Is there a reason why we use ASCII references and not UNICODE references here? Wouldn't UNICODE make more sense, or?
> 
> My (personal) reasons:
> 
> 1. It's a little bit easier (for a human) to recognize and operate with ASCII references, as they don't transcend single byte limit, and correspond exactly to the value of the byte;

Don't know enough about this to say either for or against this claim.
But in the end, it's a machine encoding language, right? So if it's more readable if it says an ASCII code or Unicode code here should be the same for a human reading it, or?

> 2. We're not using Unicode references in this section (document?), thus there is little need for a switch here.

True. Maybe this issue should be raised in a separate issue, where it can be discussed? And have the possibility for a separate future PR.

As far as I know, Unicode is the main basis for most web-based platforms at this point. However, this is not my field of expertise, it is merely what I think to have gleaned from experience and periphery knowledge.

optimade.rst
rartino: Ok --- but I also contemplated keeping "enclosed". But after some minimal research, it appears "surrounded by" is somewhat more common than "enclosed in" in this context, so I'm going with @CasperWA's version.

PR #233, opened 2019-12-21
Standardize all use of RFC2119-type terms

rartino:
This PR does minor text changes that are not meant to affect the specification, but which corrects all use of the rfc2119 terms (e.g., SHOULD, MUST, etc.) to be uppercase when meant in the rfc2119 sense, and reformulated when not meant that way.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1138,7 +1138,7 @@ Custom Extension Endpoints
 --------------------------
 
 API implementations MAY provide custom endpoints under the Extensions endpoint.
-These endpoints are then on the form "<base\_url>/extensions/<custom paths>".
+These endpoints MUST be on the form "<base\_url>/extensions/<custom paths>".
 
 API Filtering Format Specification
 ==================================
--- a/optimade.rst
+++ b/optimade.rst
@@ -1137,8 +1137,8 @@ This will make all OPTiMaDe databases and implementations by the provider discov
 Custom Extension Endpoints
 --------------------------
 
-API implementations can provide custom endpoints under the Extensions endpoint.
-These endpoints SHOULD have the form "<base\_url>/extensions/<custom paths>".
+API implementations MAY provide custom endpoints under the Extensions endpoint.
+These endpoints are then on the form "<base\_url>/extensions/<custom paths>".
 
 API Filtering Format Specification
 ==================================
--- a/optimade.rst
+++ b/optimade.rst
@@ -1598,7 +1598,8 @@ elements\_ratios
     
 - **Query examples**:
   
-  - Note: useful filters can be formulated using the set operator syntax for correlated values. However, since the values are floating point values, the use of equality comparisons is generally inadvisable.
+  - Note: Useful filters can be formulated using the set operator syntax for correlated values.
+    However, since the values are floating point values, the use of equality comparisons is generally inadvisable.
   - A filter that matches structures where approximately 1/3 of the atoms in the structure are the element Al is: :filter:`elements:elements_ratios HAS ALL "Al":>0.3333, "Al":<0.3334`.
 
 chemical\_formula\_descriptive
--- a/optimade.rst
+++ b/optimade.rst
@@ -512,7 +512,7 @@ The response MAY also return resources related to the primary data in the field:
     When the current response is the last page of data, this field MUST be either omitted or :field-val:`null`\ -valued.
 
   An implementation MAY also use the following reserved fields for pagination.
-They represent links in a similar way as for :field:`next`.
+  They represent links in a similar way as for :field:`next`.
 
   - **prev**: the previous page of data. :field-val:`null` or omitted when the current response is the first page of data.
   - **last**: the last page of data.
--- a/optimade.rst
+++ b/optimade.rst
@@ -193,7 +193,7 @@ An API implementation handles data types and their representations in three diff
 
 - In the HTTP URL query filter, see section `API Filtering Format Specification`_.
 - In the HTTP response. The default response format is JSON-based and thus uses JSON data types.
-  However, other response formats may use different data types.
+  However, other response formats can use different data types.
   For more info, see section `Responses`_.
 - The underlying database backend(s) from which the implementation serves data.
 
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -478,7 +478,7 @@ The response MAY also return resources related to the primary data in the field:
 
 - **links**: `JSON API links <http://jsonapi.org/format/1.0/#document-links>`__ is MANDATORY for implementing pagination.
   (see section `Entry Listing URL Query Parameters`_.)
-  Each field of a links object, i.e. a "link", must be either
+  Each field of a links object, i.e. a "link", MUST be one of:
```
```suggestion
  Each field of a links object, i.e., a "link", MUST be one of:
```

optimade.rst
```diff
@@ -511,9 +511,8 @@ The response MAY also return resources related to the primary data in the field:
   - **next**: represents a link to fetch the next set of results.
     When the current response is the last page of data, this field MUST be either omitted or :field-val:`null`\ -valued.
 
-  The following fields are reserved for pagination.
-  Their values are as with :field:`next`, in the sense that they should be a "link".
-  An implementation MAY offer these links:
+  An implementation MAY also use the following reserved fields for pagination.
+  In a similar way as for :field:`next`, they represent links.
```
```suggestion
They represent links in a similar way as for :field:`next`.
```

optimade.rst
```diff
@@ -1139,7 +1138,7 @@ Custom Extension Endpoints
 --------------------------
 
 API implementations can provide custom endpoints under the Extensions endpoint.
-These endpoints should have the form "<base\_url>/extensions/<custom paths>".
+These endpoints SHOULD have the form "<base\_url>/extensions/<custom paths>".
```
CasperWA: MUST they not, or?

optimade.rst
```diff
@@ -1320,7 +1319,7 @@ The following construct MAY be supported:
 - :filter:`list HAS ONLY values`: matches if all elements in :filter-fragment:`list` are equal to at least one :filter-fragment:`value`.
   (If both :filter-fragment:`list` and :filter-fragment:`values` do not contain duplicate values, this implements the <= set operator.)
 
-This construct is OPTIONAL as it may be difficult to realize in some underlying database implementations.
+This construct is OPTIONAL as it can be difficult to realize it in some underlying database implementations.
```
CasperWA: No reason for the extra object inclusion :) But feel free to keep it if you feel it makes is easier to read and clearer.
```suggestion
This construct is OPTIONAL as it can be difficult to realize in some underlying database implementations.
```

optimade.rst
```diff
@@ -1550,9 +1549,9 @@ elements
   
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
-  - The strings are the chemical symbols, written as uppercase letter plus optional lowercase letters.
+  - The strings are the chemical symbols, i.e., either a single uppercase letter or an uppercase letter followed by a number of lowercase letters.
   - The order MUST be alphabetical.
-  - Note: This may not contain the "X" or "vacancy" suggested in chemical_symbols for the :property:`species` property.
+  - Note: This property SHOULD NOT contain the string "X" to indicate non-chemical elements or "vacancy" to indicate vacancies (in contrast to the field :field:`chemical_symbols` for the :property:`species` property.)
```
CasperWA: I really like this rewrite and clarification. Thanks!
```suggestion
  - Note: This property SHOULD NOT contain the string "X" to indicate non-chemical elements or "vacancy" to indicate vacancies (in contrast to the field :field:`chemical_symbols` for the :property:`species` property).
```

optimade.rst
rartino: This PR was not meant to change the specification, but I think you are right. It gets quite weird if one think what it means with a SHOULD here. Are we allowed to provide extensions anywhere we want? Ok, I'll fix it in this PR.

optimade.rst
rartino: Gah, you got me with treating paratheses the same way as quotes again. Will I ever learn?

optimade.rst
rartino: Actually, I realized the sentence before this is using 'can', when it really means the RFC2119 MAY. (Maybe I should also search more broadly for misused 'can'.) With that, it made sense to skip a conformance word in the second sentence, avoiding the issue. Look at the updated text and see if you are happy.

CasperWA: Just out of curiousity: Did you find any other uses of `can` that needed changing? And we should consider in a review?

optimade.rst
```diff
@@ -511,9 +511,8 @@ The response MAY also return resources related to the primary data in the field:
   - **next**: represents a link to fetch the next set of results.
     When the current response is the last page of data, this field MUST be either omitted or :field-val:`null`\ -valued.
 
-  The following fields are reserved for pagination.
-  Their values are as with :field:`next`, in the sense that they should be a "link".
-  An implementation MAY offer these links:
+  An implementation MAY also use the following reserved fields for pagination.
+They represent links in a similar way as for :field:`next`.
```
CasperWA: I think maybe this and the next lines need to be slightly indented?

optimade.rst
```diff
@@ -1599,7 +1598,7 @@ elements\_ratios
     
 - **Query examples**:
   
-  - Note: useful filters can be formulated using the set operator syntax for correlated values. However, since the values are floating point values, the use of equality comparisons is generally not recommended.
+  - Note: useful filters can be formulated using the set operator syntax for correlated values. However, since the values are floating point values, the use of equality comparisons is generally inadvisable.
```
CasperWA: Another case of non-capitalization after colon. A possibility to correct the existing mistake... :)
```suggestion
  - Note: Useful filters can be formulated using the set operator syntax for correlated values.
    However, since the values are floating point values, the use of equality comparisons is generally inadvisable.
```

optimade.rst
```diff
@@ -1138,8 +1137,8 @@ This will make all OPTiMaDe databases and implementations by the provider discov
 Custom Extension Endpoints
 --------------------------
 
-API implementations can provide custom endpoints under the Extensions endpoint.
-These endpoints should have the form "<base\_url>/extensions/<custom paths>".
+API implementations MAY provide custom endpoints under the Extensions endpoint.
+These endpoints are then on the form "<base\_url>/extensions/<custom paths>".
```
CasperWA: Hmm. I think the `MUST` is still advisable in this sentence. Unless you _really_ don't want to put it, it's fine. But I see no problem in having `MUST` here: It doesn't make extension endpoints mandatory, only their "base" URLs ...

rartino: I didn't really think a MUST here adds much that can be evaluated for conformance; the sentence before it already says "under the Extensions endpoint", the second sentence just says what that means. But, alright, it is added.

rartino:
Sigh, had to merge develop, so now I need new approvals...

PR #236, opened 2019-12-22
Relaxed requirements for index meta-databases to support static hosting

### LINKED ISSUE #72: JSON API headers for Index Meta-Database
Lifted from PR #68.

> (..) I think in the final version of the pull request that put this info here, we lost a sentence either here or elsewhere saying that these endpoints on index meta-databases are exempt from the jsonapi header requirements on the `Content-type` and `Accept` headers, and may serve the response with any `Content-type`.
> 
> But maybe this is something we should take in a separate PR, since it isn't only a cosmetic change. It seems very important, though.

_Originally posted by @rartino in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjM2NjMzNDQw/pull_request_reviews/more_threads_

rartino:
This is my concrete suggestion for #72. (Fixes #72.)

It was clear on the last telco that there were some differing opinions on this. My suggestion is to allow v1.0-rc1 to be released without this PR merged, and to sort this out between v1.0-rc1 and v1.0.

Issue #72 has become more severe since we (the Materials-Consortia) are now serving an index meta-database of OPTiMaDe providers on http://www.optimade.org/providers/ from github, and as far as I am aware, it is not possible to serve an index meta-database as static files from github without the relaxations in this PR.

I disagree with releasing v1.0 with our own index meta-database of providers in violation of our own specification. 

Hence, we either need this PR merged --- OR --- an alternative hosting solution set up for https://www.optimade.org/providers with comparable reliability/redundancy that auto-updates from the Materials Consortia providers github repo. 

Edit: a preview of these changes is visible at: https://deploy-preview-15--optimade-providers.netlify.com/

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -286,7 +286,7 @@ Database-Provider-Specific Namespace Prefixes
 
 This standard refers to database-provider-specific prefixes and database providers.
 
-A list of known providers and their assigned prefixes is published in the form of a statically hosted OPTiMaDe Index Meta-Database with base URL `https://www.optimade.org/providers/ <https://www.optimade.org/providers/>`__.
+A list of known providers and their assigned prefixes is published in the form of a statically hosted OPTiMaDe Index Meta-Database with base URL `https://providers.optimade.org <https://providers.optimade.org>`__.
 Visiting this URL in a web browser gives a human-readable description of how to retrieve the information in the form of a JSON file, and specifies the procedure for registration of new prefixes.
 
 API implementations SHOULD NOT make up and use new prefixes without first getting them registered in the official list.
--- a/optimade.rst
+++ b/optimade.rst
@@ -266,7 +266,7 @@ The value for :field:`is_index` MUST be :field-val:`true`.
 
 A few suggestions and mandatory requirements of the OPTIMaDe specification are specifically relaxed **only for index meta-databases** to make it possible to serve them in the form of static files on restricted third-party hosting platforms:
 
-- When serving an index meta-database in the form of static files, it is RECOMMENDED that the responses only contain the `data` field (as described in the section `JSON Response Schema: Common Fields`_.)
+- When serving an index meta-database in the form of static files, it is RECOMMENDED that the responses only contain the :field:`data` field (as described in the section `JSON Response Schema: Common Fields`_.)
   The motivation is that static files cannot keep dynamic fields such as :field:`time_stamp` updated.
 
 - The `JSON API specification <http://jsonapi.org/format/1.0>`__ requirements on content negotiation using the HTTP headers :http-header:`Content-type` and :http-header:`Accept` are NOT mandatory for index meta-databases.
--- a/optimade.rst
+++ b/optimade.rst
@@ -273,8 +273,7 @@ A few suggestions and mandatory requirements of the OPTIMaDe specification are s
   Hence, API Implementations MAY ignore the content of these headers and respond to all requests.
   The motivation is that static file hosting is typically not flexible enough to support these requirements on HTTP headers.
 
-- API implementations SHOULD serve JSON content with either the JSON API mandated HTTP header :http-header:`Content-Type: application/vnd.api+json` or :http-header:`Content-Type: application/json`. However, if the hosting platform does not allow this, JSON content MAY be served with other content types.
-  This relaxed requirement means that clients MUST ignore the value of the :http-header:`Content-Type` HTTP header in responses from index meta-databases and accept as valid any response with an HTTP body that can be parsed as UTF-8 encoded JSON if it matches the expected schema.
+- API implementations SHOULD serve JSON content with either the JSON API mandated HTTP header :http-header:`Content-Type: application/vnd.api+json` or :http-header:`Content-Type: application/json`. However, if the hosting platform does not allow this, JSON content MAY be served with :http-header:`Content-Type: text/plain`.
 
 ..
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -264,6 +264,20 @@ The :field:`index_base_url` field MUST be included in every response in the :fie
 The :field:`is_index` field under :field:`attributes` as well as the :field:`relationships` field, MUST be included in the :endpoint:`info` endpoint for the index meta-database (see section `Base URL Info Endpoint`_).
 The value for :field:`is_index` MUST be :field-val:`true`.
 
+A few suggestions and mandatory requirements of the OPTIMaDe specification are specifically relaxed **only for index meta-databases** to make it possible to serve them in the form of static files on restricted third-party hosting platforms:
+
+- When serving an index meta-database in the form of static files, it is RECOMMENDED that the responses only contain the `data` field (as described in the section `JSON Response Schema: Common Fields`_.)
+  The motivation is that static files cannot keep dynamic fields such as :field:`time_stamp` updated.
+
+- The `JSON API specification <http://jsonapi.org/format/1.0>`__ requirements on content negotiation using the HTTP headers :http-header:`Content-type` and :http-header:`Accept` are NOT mandatory for index meta-databases.
+  Hence, API Implementations MAY ignore the content of these headers and respond to all requests.
+  The motivation is that static file hosting is typically not flexible enough to support these requirements on HTTP headers.
+
+- API implementations SHOULD serve JSON content with either the JSON API mandated HTTP header :http-header:`Content-Type: application/vnd.api+json` or :http-header:`Content-Type: application/json`. However, if the hosting platform does not allow this, JSON content MAY be served with other content types.
+  This relaxed requirement means that clients MUST ignore the value of the :http-header:`Content-Type` HTTP header in responses from index meta-databases and accept as valid any response with an HTTP body that can be parsed as UTF-8 encoded JSON if it matches the expected schema.
+
+..
+
     **Note**: A list of database providers acknowledged by the **Open Databases Integration for Materials Design** consortium is maintained externally from this specification and can be retrieved as described in section `Database-Provider-Specific Namespace Prefixes`_.
     This list is also machine-readable, optimizing the automatic discoverability.
 


rartino:
One thought closely related to this PR is that we probably should change the location of our providers index meta-database from:
  https://www.optimade.org/providers/
to: 
  https://providers.optimade.org/
since the latter gives us more options in the future to change the way we host it. If it is confined under www.optimade.org, it can be tricky to alter its hosting without affecting our web pages.

For this to be possible, we need someone at UCLouvain to add a DNS CNAME for `providers.optimade.org` to `materials-consortia.github.io`.

sauliusg:
> I disagree with releasing v1.0 with our own index meta-database of providers in violation of our own specification.

I fully support this principle!

sauliusg:
> Issue #72 has become more severe since we (the Materials-Consortia) are now serving an index meta-database of OPTiMaDe providers on http://www.optimade.org/providers/ from github, and as far as I am aware, it is not possible to serve an index meta-database as static files from github without the relaxations in this PR.

IMHO, serving a JSON file that is required in the specification to be JSON with a `text/html` Content-Type: is a no-go. If we do that, then, the whole specification of CT makes no sense.

> Hence, we either need this PR merged --- OR --- an alternative hosting solution set up for https://www.optimade.org/providers with comparable reliability/redundancy that auto-updates from the Materials Consortia providers github repo.

Alternative hosting, for sure. If a tool (GitHub) dos not work as required, we should not bend the specification to match an unsuitable tool, but pick a right tool instead.

As a compromise, we could serve a *raw* github file, and redirect to it from a stable URI, e.g.:

Apache2 config:

`RewriteRule ^optimade(/v0.10.0)?(/links) https://raw.githubusercontent.com/Materials-Consortia/providers/master/src/v0.10/providers.json`

example URIs:

`https://www.crystallography.net/cod/optimade/links`

`https://www.crystallography.net/cod/optimade/v0.10.0/links`

The `Content-Type:` is then `text/plain`. It is not `application/json`, but at least it does not "lie" that the CT is `text/html`. One can argue that JSON is kind-of text, so `application/json` is at the same time `text/plain`. We could the not in the specification that the providers list can be served either with `Content-Type: text/plain; charset=utf-8` or with `Content-Type: application/json`; the latter is preferable but the former is permitted.

optimade.rst
```diff
@@ -264,6 +264,20 @@ The :field:`index_base_url` field MUST be included in every response in the :fie
...
+
+- When serving an index meta-database in the form of static files, it is RECOMMENDED that the responses only contain the :field:`data` field (as described in the section `JSON Response Schema: Common Fields`_.)
+  The motivation is that static files cannot keep dynamic fields such as :field:`time_stamp` updated.
+
+- The `JSON API specification <http://jsonapi.org/format/1.0>`__ requirements on content negotiation using the HTTP headers :http-header:`Content-type` and :http-header:`Accept` are NOT mandatory for index meta-databases.
+  Hence, API Implementations MAY ignore the content of these headers and respond to all requests.
+  The motivation is that static file hosting is typically not flexible enough to support these requirements on HTTP headers.
+
+- API implementations SHOULD serve JSON content with either the JSON API mandated HTTP header :http-header:`Content-Type: application/vnd.api+json` or :http-header:`Content-Type: application/json`. However, if the hosting platform does not allow this, JSON content MAY be served with other content types.
```
sauliusg: `served with other content types` is too unspecific. I suggest replacing it by:

```
if the hosting platform does not allow this, JSON content MAY be served with :http-header:`Content-Type: text/plain`
```

optimade.rst
```diff
@@ -264,6 +264,20 @@ The :field:`index_base_url` field MUST be included in every response in the :fie
...
+  This relaxed requirement means that clients MUST ignore the value of the :http-header:`Content-Type` HTTP header in responses from index meta-databases and accept as valid any response with an HTTP body that can be parsed as UTF-8 encoded JSON if it matches the expected schema.
```
sauliusg: I suggest removing this. Ignoring http headers is not a wise thing to do. Better let's set them correctly.

rartino: Rather than suggesting removing this, don't you mean it should be changed to say that clients need to accept responses with either of the three content types, ‘application/vnd.api+json‘, ‘application/json‘ or ‘text/plain‘?

rartino:
@sauliusg thanks for the input.

What would you primarily prefer? To not give any leeway with the content type at all (meaning it must be ’application/vnd.api+json’) or to relax it to ’text/plain’.

HTTP redirects... are those allowed in the middle of interacting with the API? I suppose we haven't explicitly disallowed them. Still, it is unexpected that fetching some resources via the API may suddenly trigger an HTTP redirect. If it is our intent that this is fine, we may want to drop an implementers note somewhere reminding clients about being prepared to receive a redirect at any point.

sauliusg:
> What would you primarily prefer? To not give any leeway with the content type at all (meaning it must be ’application/vnd.api+json’) or to relax it to ’text/plain’.

As a suggestion for a reasonable engineering compromise, I would say:

- an implementation SHOULD use the most specific MIME type that matches the content;

- If the response content matches the [JSON:API ](https://jsonapi.org/), then the `Content-Type:` SHOULD be specified as `application/vnd.api+json`; if the `Content-Type:`  is given as `application/vnd.api+json`, then the response MUST match the specification of [JSON:API ](https://jsonapi.org/)

- If the response has JSON syntax (but not necessarily the [JSON:API ](https://jsonapi.org/) specification or schema), the `Content-Type:` SHOULD be specified as `application/json`, as specified in [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt); if the `Content-Type:` is `application/json`, then the response MUST conform to the JSON syntax;

- The response `Content-Type:` MAY be specified as `text/plain` (even if the JSON:API or generic JSON is returned), for example when technical limitations do not allow specifying a more precise content type; however, some clients in this case may fail on the request, may take longer to process or require additional manual configuration.

- an implementation MAY NOT use incorrect `Content-Type:` header (such as `text/html` for JSON; but this should be obvious from the HTTP specs, shouldn't it?)

> HTTP redirects... are those allowed in the middle of interacting with the API? I suppose we haven't explicitly disallowed them. Still, it is unexpected that fetching some resources via the API may suddenly trigger an HTTP redirect. If it is our intent that this is fine, we may want to drop an implementers note somewhere reminding clients about being prepared to receive a redirect at any point.

I think redirects (`302 Found\nLocation: ...`) are a standard part of HTTP protocol, an old one, and most client support it. It is very handy in some cases (e.g. when you POST to an endpoint to create a new resource on the server, it is very reasonable that you get redirect to that newly created resource in the response, no? In that way a standard Web HTTP browser immediately shows you a new resource). At least we use redirects explicitly in our (other) REST design, assuming it is perfectly valid and totally expected.

rartino:
> when you POST to an endpoint to create a new resource on the server, it is very reasonable that you get redirect to that newly created resource in the response, no?

Sure, but that is very different from redirecting an attempted access of a REST endpoint. I'm skeptical to the reading that our present spec allows endpoint redirection; we say the endpoints should *be* at these URLs. That isn't exactly the same thing as "one can locate the endpoint via a series of standard HTTP requests starting at this URL".

As for other APIs, there is a draft specification from 2014:
- https://tools.ietf.org/id/draft-hunt-http-rest-redirect-00.html

Which says you SHOULD NOT use 301-306 redirects for REST operations, but MAY use 307 and 308 as follows:
- 307 for temporary redirection
- 308 "In the event that a REST resource has been relocated or reassigned" or "in order to permit aliasing of RESTful resources"

I think what you are suggesting may fall under aliasing?

I also find notes that it isn't uncommon for client libraries to have trouble with REST redirects, e.g., see
- https://docs.aws.amazon.com/AmazonS3/latest/dev/RESTRedirect.html

so from a practical standpoint it isn't very nice.

> - an implementation MAY NOT use incorrect Content-Type: header (such as text/html for JSON; but this should be obvious from the HTTP specs, shouldn't it?)

Well, it certainly is against the intent, but technically I haven't found support for the most strict dismissal. RFC7231 says that it is aware that servers sometimes not setting the Content-Type header correctly and that clients sometimes override it. The spec doesn't seem particularly happy about that, but it also never says that clients 'MUST NOT' do it:
> In practice, resource owners do not always properly configure their origin server to provide the correct Content-Type for a given representation, with the result that some clients will examine a payload's content and override the specified type.  Clients that do so risk drawing incorrect conclusions, which might expose additional security risks (e.g., "privilege escalation").  Furthermore, it is impossible to determine the sender's intent by examining the data format: many data formats match multiple media types that differ only in processing semantics.  Implementers are encouraged to provide a means of disabling such "content sniffing" when it is used." 

In the end, for now, I'm leaning towards trying to set up a hosting solution that serves the correct MIME type but doesn't require redirects. And then put in the specification your version of the relaxed requirements for index meta-databases. We can always relax them further later on. Un-relaxing them is more difficult as it is not backward compatible.

shyamd:
Why not just host on JSON Bin : https://jsonbin.io/

rartino:
> Why not just host on JSON Bin : https://jsonbin.io/

We need our JSON files served under a URL structure that looks like an index base URL (i.e., as a response to URLs `.../info`, `.../links`) Can this service do that? Can it serve those URLs under our domain? (i.e., CNAME support.) 

Unless I'm missing something, this service would only give us fixed URLs to separate files served with content-type application/json, and in that case we can just as well use
- https://raw.githubusercontent.com/Materials-Consortia/providers/master/src/v0.10/providers.json

rartino:
I have found that *netlify* apparently is a major well-known hosting service with a significant free tier that publish websites straight out of a github repository much like github pages. It allows customizing headers [using a `_headers` file](https://docs.netlify.com/routing/headers/).

Is anyone opposed to me connecting their service to our providers repository and trying out if it can work for us? It apparently involves giving them some limited write rights to [checks, commit statues, and pull requests.](https://docs.netlify.com/configure-builds/repo-permissions-linking/) I'm not entirely sure what "write rights to pull requests" means, but they claim it is only used for notifications in the GitHub UI. To me they seem well-known enough to be trusted for this.

shyamd:
I forgot about Netlify. It's actually a good service, so should be fine.

giovannipizzi:
@rartino I also think that it is ok to test this out. Please feel free to go ahead.

rartino:
I've been testing Netlify, but I cannot seem to get it to deploy the symbolic links to files that we have in the repository in any sane way.

Netlify just seems to exclude them, whereas regular files named similarly alongside them gets deployed. Weirdly I also do not find any discussion about this issue anywhere online.

Symbolic links to directories seem to work, as long as there is at least one regular file inside the linked directory that is to be deployed.

But with the latest changes to the specification where there is no longer any "default version" served under the base URL without a version number, perhaps I can work around any need for symbolic links to files.

rartino:
Ok, I've done exactly the alterations to this PR that @sauliusg proposed, which is now possible for us to fulfill at https://providers.optimade.org/ hosted via Netifly.

sauliusg:
> Ok, I've done exactly the alterations to this PR that @sauliusg proposed, which is now possible for us to fulfill at https://providers.optimade.org/ hosted via Netifly.
> 
> ~We should just wait until [Materials-Consortia/providers#15](https://github.com/Materials-Consortia/providers/pull/15) gets merged, and then this PR should be mergeable.~ (done)

I would say the links are fine now at [https://providers.optimade.org/](https://providers.optimade.org/). The concrete version number underlying the abstract `<version>` link clarifies well enough what  was meant, at least for me.

rartino:
> I would say the links are fine now at https://providers.optimade.org/. The concrete version number underlying the abstract <version> link clarifies well enough what was meant, at least for me.

Great! @sauliusg, when you are able, please do a full review of #236 to make sure all parts of the relaxed requirements for http headers etc. seems good now. If all is good, please approve it and we can finally put #72 to rest in preparation of v1.0.0.

rartino:
I'm re-pinging @sauliusg for his review and approval of this PR in hope that we can merge it before the teleconference at the end of this week.

optimade.rst
```diff
@@ -272,7 +285,7 @@ Database-Provider-Specific Namespace Prefixes
 
 This standard refers to database-provider-specific prefixes and database providers.
 
-A list of known providers and their assigned prefixes is published in the form of a statically hosted OPTiMaDe Index Meta-Database with base URL `https://www.optimade.org/providers/ <https://www.optimade.org/providers/>`__.
+A list of known providers and their assigned prefixes is published in the form of a statically hosted OPTiMaDe Index Meta-Database with base URL `https://providers.optimade.org <https://providers.optimade.org>`__.
```
sauliusg: I the Optimade specification now says where to find human readable instructions, which is nice to have; but I am more worried about the machine readable description. Wouldn't it be more efficient for interoperability to include lines:

"The list of known providers in machine readable form is served from the URI `https://providers.optimade.org/providers.json`"

I suggest including this sentence into the specification.

Of course this adds extra promise to keep, but I find it more helpful than a references to "human-readable description", and with the introduction of `providers.optimade.org` it is now probably easy to fulfill this promise, isn't it?

optimade.rst
```diff
@@ -264,6 +264,19 @@ The :field:`index_base_url` field MUST be included in every response in the :fie
 The :field:`is_index` field under :field:`attributes` as well as the :field:`relationships` field, MUST be included in the :endpoint:`info` endpoint for the index meta-database (see section `Base URL Info Endpoint`_).
 The value for :field:`is_index` MUST be :field-val:`true`.
 
+A few suggestions and mandatory requirements of the OPTIMaDe specification are specifically relaxed **only for index meta-databases** to make it possible to serve them in the form of static files on restricted third-party hosting platforms:
+
+- When serving an index meta-database in the form of static files, it is RECOMMENDED that the responses only contain the :field:`data` field (as described in the section `JSON Response Schema: Common Fields`_.)
+  The motivation is that static files cannot keep dynamic fields such as :field:`time_stamp` updated.
+
+- The `JSON API specification <http://jsonapi.org/format/1.0>`__ requirements on content negotiation using the HTTP headers :http-header:`Content-type` and :http-header:`Accept` are NOT mandatory for index meta-databases.
```
sauliusg: I suggest dropping reference to JSON API altogether. Now that we use 'providers.optimade.org' server name, assigning proper Content-type should be possible, and the next sentence mandates deviation from JSON API anyway, so no compatibility with be maintained here.

optimade.rst
rartino: Well, formally, when we say *"in the form of a statically hosted OPTiMaDe Index Meta-Database with base URL ..."* this should already be telling you that you can retrieve the machine readable description at \<URL\>/v1/links 

I'm not so opposed to include also a URL to providers.json, but if we do it, we should be sure that we want to keep it at that URL. And you had reservations against `https://providers.optimade.org/providers.json` in the discussion of Materials-Consortia/providers#14. So does it not  make sense to keep the one web-page indirection for finding providers.json until we have sorted this out?

optimade.rst
rartino: I'm not sure I follow. As far as I can see, no sentence *mandates* deviation from JSON API. You are allowed to serve an index meta-database completely following the JSON API specification. But because of the exceptions that this PR adds, you are allowed to somewhat break JSON API conformance when serving index meta-databases.

What more precisely is it we should drop? Is it this bullet point?:

- The `JSON API specification <http://jsonapi.org/format/1.0>`__ requirements on content negotiation using the HTTP headers :http-header:`Content-type` and :http-header:`Accept` are NOT mandatory for index meta-databases.

If we leave out that bullet point, it means an index-meta database MUST, e.g., "respond with a 415 Unsupported Media Type status code if a request specifies the header Content-Type: application/vnd.api+json with any media type parameters." This is very tricky to pull off with static hosting.

optimade.rst
sauliusg: My reservations were directed at the URL of the form `https://www.optimade.org/providers/providers.json`. It duplicated the resource name.

Now that you have introduced the `https://providers.optimade.org' domain, the duplication is not so grave. We can have domain name and resource name overlapping, it ise used and useful sometimes (e.g. [https://en.wikipedia.org/wiki/Wikipedia](https://en.wikipedia.org/wiki/Wikipedia) :)

New endpoints can be introduced later, e.g. `https://providers.optimade.org/v1.0/providers.json` for a specific version or `https://providers.optimade.org/providers.xml` for a different format. So I do not see big danger here.

Issue #147, opened 2019-06-28
Add `homepage_url` to providers.json


giovannipizzi:
As it was already partially suggested in #46 (see e.g. `'db_provider_homepage': 'http://example.com'` in point 4 of the description), it would be good to have a separate field to indicate the URL of the homepage of a provider in the `providers.json` file. 

In this way, when displaying a list of known providers, one can also directly point to the homepage for people interested in knowing better what it is.

For the current structure:
```json
    {
      "type": "provider",
      "id": "exmpl",
      "attributes": {
        "name": "Example provider",
        "description": "Provider used for examples, not to be assigned to a real database",
        "base_url": "http://example.com/optimade/index"
      }
    }
```
I suggest adding a field `homepage_url` to the `attributes` but I'm open to discussions on a different/better name.

PR #238, opened 2019-12-23
Change homepage_url -> homepage in providers.json and add it to examples

rartino:
When making PR #228 to implement issue #147 I missed updating the examples. This PR fixes that (sorry).

One thing, though. I implemented the suggestion from #147 to use the field name `homepage_url`, which probably was suggested because it matches `base_url`. However, in the `provider` field in the `info` endpoint we have already used the field name `homepage` for the same link. Should we live with this inconsistency, or should we change either one of them? I suppose when the field value is a "link object" which can be more that just an url, `homepage` makes more sense. (However `base_url` should still remain `base_url`, because that combination of words is for us a moniker for "the root of an API implementation")

Also, do you agree that listing a homepage in providers.json should be mandatory?

I've added an issue in the providers repository to add `homepage_url` to all entries, Materials-Consortia/providers#6.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1055,7 +1055,7 @@ The resource objects' response dictionaries MUST include the following fields:
     - **href**: a string containing the OPTiMaDe base URL.
     - **meta**: a meta object containing non-standard meta-information about the implementation.
 
-  - **homepage\_url**: `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage URL for this implementation, either directly as a string, or as a links object, which can contain the following fields:
+  - **homepage**: `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage URL for this implementation, either directly as a string, or as a links object, which can contain the following fields:
 
     - **href**: a string containing the implementation homepage URL.
     - **meta**: a meta object containing non-standard meta-information about the homepage.
@@ -1073,7 +1073,7 @@ Example:
 	     "name": "Index",
 	     "description": "Index for example's OPTiMaDe databases",
 	     "base_url": "http://example.com/optimade/index",
-	     "homepage_url": "http://example.com"
+	     "homepage": "http://example.com"
 	   }
 	 },
 	 {
@@ -1088,7 +1088,7 @@ Example:
 		 "_exmpl_catalyst_group": "denox"
 	       }
 	     },
-	     "homepage_url": "http://example.com"
+	     "homepage": "http://example.com"
 	   }
 	 },
 	 {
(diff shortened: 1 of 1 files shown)


CasperWA:
> One thing, though. I implemented the suggestion from #147 to use the field name `homepage_url`, which probably was suggested because it matches `base_url`. However, in the `provider` field in the `info` endpoint we have already used the field name `homepage` for the same link. Should we live with this inconsistency, or should we change either one of them? I suppose when the field value is a "link object" which can be more that just an url, `homepage` makes more sense. (However `base_url` should still remain `base_url`, because that combination of words is for us a moniker for "the root of an API implementation")

Most importantly, I think it's crucial the field names are the same for our `meta` provider as for the resource. So I would definitely go for `homepage`, plain and simple.

> Also, do you agree that listing a homepage in providers.json should be mandatory?

Sure, but it can be `null`.

merkys:
I agree with @CasperWA, `homepage` looks better to me.

rartino:
Ok, `homepage_url` -> `homepage` done.

optimade.rst
```diff
@@ -1072,7 +1072,8 @@ Example:
 	   "attributes": {
 	     "name": "Index",
 	     "description": "Index for example's OPTiMaDe databases",
-	     "base_url": "http://example.com/optimade/index"
+	     "base_url": "http://example.com/optimade/index",
```
```suggestion
	     "base_url": "http://example.com/index/optimade",
```

optimade.rst
```diff
@@ -1095,7 +1097,8 @@ Example:
 	   "attributes": {
 	     "name": "Zeolitic Frameworks",
 	     "description": "",
-	     "base_url": "http://example.com/optimade/zeo_frameworks"
+	     "base_url": "http://example.com/optimade/zeo_frameworks",
```
```suggestion
	     "base_url": "http://example.com/zeo_frameworks/optimade",
```

optimade.rst
```diff
@@ -1104,7 +1107,8 @@ Example:
 	   "attributes": {
 	     "name": "Example provider",
 	     "description": "Provider used for examples, not to be assigned to a real database",
-	     "base_url": "http://example.com/optimade/index"
+	     "base_url": "http://example.com/optimade/index",
```
```suggestion
	     "base_url": "http://example.com/index/optimade",
```

optimade.rst
merkys: Maybe just `http://example.com/optimade`? I guess this is remnant from the old times.

optimade.rst
merkys: Maybe just `http://example.com/optimade` as well.

rartino:
Sorry for getting lazy over the holidays :), but the fix is in now. Let's try to move things forward so we get the v1.0-rc1 released asap. It seems to me not much is left to do.

PR #240, opened 2020-01-08
clarify requirements on the base url

ltalirz:
This PR tries to clarify a couple of issues in the requirements for the base url.

 - The spec seemed to define a way for clients to construct new base
   urls themselves. This bit is removed, since clients can easily discover
   the base urls of supported versions via the info/ endpoint.

 - It was unclear whether the spec *mandated* serving the API both under
   a baseurl *and* under a baseurl + version suffix. This PR makes clear
   that the second part is optional.

 - The spec differentiated between treatment of vMAJOR, vMAJOR.MINOR and
   vMAJOR.MINOR.PATCH version numbers. Since clients use /info to access
   supported versions, this special treatment is not needed.

 - The spec gave examples of invalid base urls. Strictly speaking, all
   base urls are valid - it is just certain combinations of base URLs that
   are invalid.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -231,6 +231,7 @@ Every URL path segment that follows the base URL MUST behave as standardized in 
 All access to the API is provided under the **versioned base URLs**, which append a URL path segment to the base URL that SHOULD be of the form :query-url:`/vMAJOR`, :query-url:`/vMAJOR.MINOR` or :query-url:`/vMAJOR.MINOR.PATCH`. An implementation MUST provide the :query-url:`/vMAJOR` form, whereas the other forms are OPTIONAL.
 Here, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
 For all major versions supported by the provider, the :val:`/vMAJOR` URL MUST serve the *latest* minor/patch version implemented by the provider.
+If the version has a suffix, e.g., `-rc<number>` for release candidate versions, this suffix SHOULD be omitted in the URL path segment.
 
     **For implementers**: Clients are recommended to discover the highest version supported by both the client and the API implementation by trying versioned base URLs in order of priority.
     E.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
--- a/optimade.rst
+++ b/optimade.rst
@@ -228,13 +228,13 @@ For any major version supported by the provider, the :val:`<URL>/vMAJOR` URL MUS
 
     **For implementers**: Clients are recommended to discover the highest commonly supported version by the client and the API implementation by trying versioned base URL in order of priority, e.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
 
-Examples of valid base URLs:
+Examples of valid versioned base URLs:
 
 - http://example.com/optimade/v0/
 - http://example.com/v0.9.1/
 - http://example.com/v1/
 
-Examples of invalid base URLs:
+Examples of invalid versioned base URLs:
 
 - http://example.com/optimade/0.9/
 - http://example.com/optimade/
--- a/optimade.rst
+++ b/optimade.rst
@@ -232,7 +232,8 @@ All access to the API is provided under the **versioned base URLs**, which appen
 Here, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
 For all major versions supported by the provider, the :val:`/vMAJOR` URL MUST serve the *latest* minor/patch version implemented by the provider.
 
-    **For implementers**: Clients are recommended to discover the highest commonly supported version by the client and the API implementation by trying versioned base URLs in order of priority, e.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
+    **For implementers**: Clients are recommended to discover the highest version supported by both the client and the API implementation by trying versioned base URLs in order of priority.
+    E.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
 
 Examples of valid versioned base URLs:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -228,7 +228,7 @@ Base URL
 Each database provider will publish one or more **base URLs** that serve the API, for example: http://example.com/optimade/.
 Every URL path segment that follows the base URL MUST behave as standardized in this API specification.
 
-All access to the API is provided under the **versioned base URLs**, which append a URL path segment to the base URL that SHOULD be on the form :query-url:`/vMAJOR`, :query-url:`/vMAJOR.MINOR` or :query-url:`/vMAJOR.MINOR.PATCH`. An implementation MUST provide the :query-url:`/vMAJOR` form, whereas the other forms are OPTIONAL.
+All access to the API is provided under the **versioned base URLs**, which append a URL path segment to the base URL that SHOULD be of the form :query-url:`/vMAJOR`, :query-url:`/vMAJOR.MINOR` or :query-url:`/vMAJOR.MINOR.PATCH`. An implementation MUST provide the :query-url:`/vMAJOR` form, whereas the other forms are OPTIONAL.
 Here, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
 For all major versions supported by the provider, the :val:`/vMAJOR` URL MUST serve the *latest* minor/patch version implemented by the provider.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -219,14 +219,14 @@ General API Requirements and Conventions
 Base URL
 --------
 
-Each database provider will publish one or more **base URLs** that serve the API, for example: http://example.com/optimade/v1.
+Each database provider will publish one or more **base URLs** that serve the API, for example: http://example.com/optimade/
 Every URL component that follows the base URL MUST behave as standardized in this API specification.
 
-Base URLs SHOULD have the form :val:`<URL>/vMAJOR`, :val:`<URL>/vMAJOR.MINOR` or :val:`<URL>/vMAJOR.MINOR.PATCH`.
-Here, :val:`<URL>` represents a generic URL, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
+All access to the API is provided under **versioned base URLs**, which append a URL component to the base URL that SHOULD be on the form :query-url:`vMAJOR`, :query-url:`/vMAJOR.MINOR` or :query-url:`/vMAJOR.MINOR.PATCH`. An implementation MUST provide the :query-url:`/vMAJOR` form, whereas the other forms are OPTIONAL.
+Here, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
 For any major version supported by the provider, the :val:`<URL>/vMAJOR` URL MUST point to the **latest** minor/patch version implemented by the provider.
 
-    **For implementers**: Clients can discover the latest supported version of the standard by cycling through :val:`<URL>/v0`, :val:`<URL>/v1`, :val:`<URL>/v2` etc.
+    **For implementers**: Clients are recommended to discover the highest commonly supported version by the client and the API implementation by trying versioned base URL in order of priority, e.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
 
 Examples of valid base URLs:
 
@@ -241,8 +241,8 @@ Examples of invalid base URLs:
 
 Database providers SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any major and minor version they support.
 
-Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to a base URL.
-However, it is RECOMMENDED that implementations serve a human-readable HTML document on a base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
+Note: The base URLs and versioned base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to them.
+However, it is RECOMMENDED that implementations serve a human-readable HTML document on the base URLs and versioned base URLs, which explains that the URL is an OPTiMaDe URL meant to be queried by an OPTiMaDe client.
 
 Index Meta-Database
 -------------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -232,7 +232,7 @@ All access to the API is provided under the **versioned base URLs**, which appen
 Here, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
 For all major versions supported by the provider, the :val:`/vMAJOR` URL MUST serve the **latest** minor/patch version implemented by the provider.
 
-    **For implementers**: Clients are recommended to discover the highest commonly supported version by the client and the API implementation by trying versioned base URL in order of priority, e.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
+    **For implementers**: Clients are recommended to discover the highest commonly supported version by the client and the API implementation by trying versioned base URLs in order of priority, e.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
 
 Examples of valid versioned base URLs:
 
@@ -1026,7 +1026,7 @@ Links Endpoint
 --------------
 
 This endpoint exposes information on other OPTiMaDe API implementations that are linked to the current implementation.
-The links endpoint is accessed under the versioned base URL at :endpoint:`/links`.
+The links endpoint MUST be provided under the versioned base URL at :endpoint:`/links`.
 
 It can be considered an introspective endpoint, similar to the Info endpoint, but at a higher level: that is, Info endpoints provide information on the given implementation, while the Links endpoint provides information on the links between immediately related implementations (in particular, an array of none or a single :object:`parent` object and none or more child-type objects, see section `Parent and Child Objects`_).
 
@@ -1494,7 +1494,7 @@ type
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
   - **Response**: REQUIRED in the response.
   - MUST be an existing entry type.
-  - The entry of type `<type>` and ID `<id>` MUST be given as the response to a request for :endpoint:`/<type>/<id>` under the versioned base URL.
+  - The entry of type `<type>` and ID `<id>` MUST be returned in response to a request for :endpoint:`/<type>/<id>` under the versioned base URL.
     
 - **Example**: :val:`"structures"`
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -222,32 +222,34 @@ Base URL
 Each database provider will publish one or more base URLs that serve the API, for example: http://example.com/optimade/.
 Every URL component that follows the base URL MUST behave as standardized in this API specification.
 
-If database providers choose to implement multiple versions of the API standard, they MAY serve them under the same base URL, followed by the version number prefixed with the letter "v".
-Allowed formatting of the version are vMAJOR and vMAJOR.MINOR and vMAJOR.MINOR.PATCH, where MAJOR is the major version number, MINOR is the minor version number and PATCH is the patch version number of the corresponding standard.
+If database providers choose to implement multiple versions of the API standard, they MAY serve them using version-specific base URLs below the same generic base URL.
+In this case, the generic base URL MUST serve the **latest** version of the standard implemented by the provider.
+Version-specific base URLs consist of the generic base URL, followed by the version number prefixed with the letter "v".
+Allowed formatting of the version are vMAJOR and vMAJOR.MINOR, vMAJOR.MINOR.PATCH, where MAJOR is the major version number, MINOR is the minor version number and PATCH is the patch version number of the corresponding standard.
 If the major version number is 0, then the minor version number MUST also be included.
 
 Examples of valid base URLs:
 
-- http://example.com/optimade/, together with
-  http://example.com/optimade/v0.9/ and
-  http://example.com/optimade/v0.9.1/
+- http://example.com/optimade/
 - http://example.com/
 - http://example.com/some/path/
 
-Examples of invalid base URLs:
+Examples of valid version-specific base URLs:
 
-- http://example.com/v0, together with
-  http://example.com/
-- http://example.com/optimade/0.9/, together with 
-  http://example.com/optimade/
+- http://example.com/optimade/v0.9/
+- http://example.com/v0.9.1/
 
-The base URL MUST serve the latest version of the standard that the provider implements.
+Examples of invalid version-specific base URLs:
+
+- http://example.com/v0
+- http://example.com/optimade/0.9/
+
 The database provider SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any supported major and minor version.
 
 Clients can retrieve supported versions using the :field:`available_api_versions` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
 
-Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to the base URL.
-However, it is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
+Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to a base URL.
+However, it is RECOMMENDED that implementations serve a human-readable HTML document on a base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
 
 Index Meta-Database
 -------------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -244,10 +244,10 @@ Examples of invalid base URLs:
 The base URL MUST serve the latest version of the standard that the provider implements.
 The database provider SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any supported major and minor version.
 
-Clients can query for supported versions using the :field:`available_api_versions` field of the :field:`attributes` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
+Clients can retrieve supported versions using the :field:`available_api_versions` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
 
 Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to the base URL.
-It is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
+However, it is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
 
 Index Meta-Database
 -------------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -219,34 +219,27 @@ General API Requirements and Conventions
 Base URL
 --------
 
-Each database provider will publish one or more base URLs that serve the API, for example: http://example.com/optimade/.
+Each database provider will publish one or more **base URLs** that serve the API, for example: http://example.com/optimade/v1.
 Every URL component that follows the base URL MUST behave as standardized in this API specification.
 
-If database providers choose to implement multiple versions of the API standard, they MAY serve them using version-specific base URLs below the same generic base URL.
-In this case, the generic base URL MUST serve the **latest** version of the standard implemented by the provider.
-Version-specific base URLs consist of the generic base URL, followed by the version number prefixed with the letter "v".
-Allowed formatting of the version are vMAJOR and vMAJOR.MINOR, vMAJOR.MINOR.PATCH, where MAJOR is the major version number, MINOR is the minor version number and PATCH is the patch version number of the corresponding standard.
-If the major version number is 0, then the minor version number MUST also be included.
+Base URLs SHOULD have the form :val:`<URL>/vMAJOR`, :val:`<URL>/vMAJOR.MINOR` or :val:`<URL>/vMAJOR.MINOR.PATCH`.
+Here, :val:`<URL>` represents a generic URL, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
+For any major version supported by the provider, the :val:`<URL>/vMAJOR` URL MUST point to the **latest** minor/patch version implemented by the provider.
+
+    **For implementers**: Clients can discover the latest supported version of the standard by cycling through :val:`<URL>/v0`, :val:`<URL>/v1`, :val:`<URL>/v2` etc.
 
 Examples of valid base URLs:
 
-- http://example.com/optimade/
-- http://example.com/
-- http://example.com/some/path/
-
-Examples of valid version-specific base URLs:
-
-- http://example.com/optimade/v0.9/
+- http://example.com/optimade/v0/
 - http://example.com/v0.9.1/
+- http://example.com/v1/
 
-Examples of invalid version-specific base URLs:
+Examples of invalid base URLs:
 
-- http://example.com/v0
 - http://example.com/optimade/0.9/
+- http://example.com/optimade/
 
-The database provider SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any supported major and minor version.
-
-Clients can retrieve supported versions using the :field:`available_api_versions` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
+Database providers SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any major and minor version they support.
 
 Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to a base URL.
 However, it is RECOMMENDED that implementations serve a human-readable HTML document on a base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -219,39 +219,35 @@ General API Requirements and Conventions
...
 - http://example.com/
 - http://example.com/some/path/
 
 Examples of invalid base URLs:
 
-- http://example.com/optimade/v0/
-- http://example.com/optimade/0.9/
+- http://example.com/v0, together with
+  http://example.com/
+- http://example.com/optimade/0.9/, together with
```
CasperWA: I kind of get where you want to go with the `together with` parts, but it seems very confusing and not well explained.
I would probably rather go for clean examples, and just list them, one by one, as was previously done. The base URLs don't pair up, they are each separate and their own.

optimade.rst
```diff
@@ -219,39 +219,35 @@ General API Requirements and Conventions
...
 Examples of invalid base URLs:
 
-- http://example.com/optimade/v0/
-- http://example.com/optimade/0.9/
+- http://example.com/v0, together with
+  http://example.com/
+- http://example.com/optimade/0.9/, together with 
+  http://example.com/optimade/
+
+The base URL MUST serve the latest version of the standard that the provider implements.
```
CasperWA: What is the base URL in this case? I would consider `.../optimade` as well as the version-specific URLs to all be base URLs.

optimade.rst
```diff
@@ -219,39 +219,35 @@ General API Requirements and Conventions
...
+The database provider SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any supported major and minor version.
+
+Clients can query for supported versions using the :field:`available_api_versions` field of the :field:`attributes` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
```
```suggestion
Clients can retrieve supported versions using the :field:`available_api_versions` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
```

optimade.rst
```diff
@@ -219,39 +219,35 @@ General API Requirements and Conventions
...
 
-Note: The OPTiMaDe standard specifies the response from a number of endpoints under the base URLs.
-However, the base URLs themselves are not considered to be a part of the API.
-Hence, they are fully under the control of the API implementation.
-It is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, and this document is used to explain that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
+Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to the base URL.
+It is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
```
```suggestion
However, it is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
```

optimade.rst
ltalirz: Happy for better suggestions!
However, I find the original examples confusing since they are not correct.

ltalirz:
I agree that the use of the "base URL" term in this paragraph is confusing - I did not intend to change its meaning.
I will make an update to try to address this.

One issue, however, is the fact that we allow this "nesting" of base URLs in the first place, which (to my understanding) is not common practice in REST APIs.

optimade.rst
ltalirz: Agreed. I will try to make this clearer.

optimade.rst
ltalirz: One could e.g. just list them with commas

ltalirz: I've introduced a distinction for a version-specific base URL, which also solves the issue with the examples, I believe.

optimade.rst
```diff
@@ -219,39 +219,37 @@ General API Requirements and Conventions
...
 Every URL component that follows the base URL MUST behave as standardized in this API specification.
 
-The client MAY include a version number in the base URL, prefixed with the letter "v", where the version number indicates the version of the API standard that the client requests.
-The format is either vMAJOR or vMAJOR.MINOR where MAJOR is the major version number, and MINOR is the minor version number of the standard being referenced.
-If the major version is 0, the minor version MUST also be included.
-The database provider MAY support further levels of versioning separated from the major and minor version by a decimal point, e.g., patch version on the format vMAJOR.MINOR.PATCH. However, the client MUST NOT assume levels beyond the minor version are supported.
-
-If the client does not include a version number in the base URL, the request is for the latest version of this standard that the database provider implements.
-A query that includes a major and/or minor version is for the latest subversion of that major and/or minor version that the database provider implements.
```
rartino: As far as I can see, this PR removes the requirement that `optimade/vMAJOR` MUST refer to the latest MINOR.PATCH version supported by the implementation, and that `optimade/vMAJOR.MINOR` MUST refer to the latest PATCH version supported by the implementation. Should these requirements be dropped? Why?

(The top PR text mentions that there is no need for a client to perform URL discovery, but even then it makes a difference if an implementation is allowed to serve a prior minor version under `optimade/vMAJOR`)

ltalirz:
Thanks @rartino for giving this a read!

> However, my reading of the present text is that accessing the API via the major version number is REQUIRED to be supported by implementations.

I agree (I also read the current spec that way) - and I strongly suggest to drop this requirement.

> Because the URL structure of /info and the format of /info may change in future versions, so IMO it would be a design fault for a client to assume it will find something it understands under optimade/info.

A change of the /info endpoint would require a change in spec - the same argument applies to the specification of the version-specific base URLs.

> The safest and recommended way for a client to access an OPTiMaDe endpoint is to always go via optimade/vN/ where N is the latest major version the client supports.

I don't think that is true. The client can only access optimade/vN if *both* the client *and* the server support it.
In order for the client to figure out which base URLs are supported it will anyhow need to go to the /info endpoint. 
At that point, it can just as well take the base URLs that the server reports for the particular version.

rartino:
> A change of the /info endpoint would require a change in the spec - the same argument applies to changing the specification of the version-specific base URLs.

I suppose this point is formally correct; future versions of the standard *could*, in theory, chose to abandon the version-specific URLs so that `v1/`, `v2/`, etc. no longer mean the same thing. However, in practice, that is not something one does in a design like this. I suppose we could explicitly put down that commitment in the specification, but I would argue that it is implicit in the design (as with many other APIs that use the same system.)

However, If I understand your main point, you want this PR to change the standard to discourage all forms of versioning negotiation based on URLs and instead push for version negotiation to go through the "raw" base URL `/info` endpoint? You thus mean that just as well as the commitment to the fixed URLs above, we can make the same type of commitment for upholding the location and response format of the `/info` endpoint for ALL future versions of OPTiMaDe? 

I don't think that is a good idea. We may, in the future, need to change our URL structure so that the meaning of `/info` changes, e.g., to adhere to future API standards such as a future version of JSONAPI. We may need to change the format of the response (we've already been through two formats for our JSON representation for the URL to version mapping...). We may even, at some point, want to abandon JSONAPI (or even JSON) for other standards. 

>> The safest and recommended way for a client to access an OPTiMaDe endpoint is to always go via optimade/vN/ where N is the latest major version the client supports.

> The client can only access optimade/vN if both the client and the server support it.
> In order for the client to figure out which versions are supported by the server it will anyhow need to go to the /info endpoint.

In URL-based version negotiation, if a client supports multiple main versions, it tries `optimade/vN` for all values of N it supports in order of priority. *If* we end up having to release so many main versions of OPTiMaDe that this scheme starts to become a nuisance (and I *really* hope we won't) we can at that point design a solution that applies to all versions onwards.

> At that point, it can just as well take the base URL that the server reports for the particular version (and I suggest it should).

To reiterate my point, the client cannot do that if it cannot safely assume it will understand the response it gets from `/info`.

I'll also add that URL-based versioning of the kind we have in the specification now is more or less *the* standard way of doing this. You see the same design if you look at e.g., the APIs of Facebook, Twitter, etc. So, to change to a different way of doing it would have to be thoroughly motivated.

ltalirz:
You make some fair points - I think you're right that the response from /info is more likely to change than the API schema (although the actual versions supported by the server might change).

This means it is probably better if clients don't use the /info endpoint to figure out which URL to go to (perhaps we should state this in the spec).

I will change that back accordingly.

> I'll also add that URL-based versioning of the kind we have in the specification now is more or less the standard way of doing this. You see the same design if you look at e.g., the APIs of Facebook, Twitter, etc.

I had a look at the facebook API and - as you said - they are indeed serving the API *both* under graph.facebook.com (still serving v2.11) *and* under graph.facebook.com/v2.11, ..., /v5.0 - which I wouldn't have thought - thanks for pointing this out.

Thinking back to the issue with the optimade-python-tools server that originally drove me here, there is one main remaining question that I have:
The fact that the /info endpoint is supposed to inform about available API versions seems to suggest that all the APIs (say, those served under /optimade, /optimade/v1, /optimade/v2) would need to be run from the same instance of an application (otherwise, how could those three APIs know about the availability of the other ones?).
I'm wondering whether that is a good idea?

If that's variant (A), then alternative (B) would of course be to simply have self-contained instances of APIs that know just about themselves and nothing about the rest of the world (they would still be served under base URLs according to the spec, but it wouldn't be "baked into the application").

Going with route (B) makes the implementation of the server a bit easier, since you don't have to add functionality to support multiple API versions in the same code base (e.g. optimade-python-tools currently does not have this).
When a new version of the API comes out, you can update your codebase and shoot up a new instance of the server (while keeping the old ones running).

There may be no need to force people into either route, but perhaps it would be worth to *allow* rout (B) as well, and perhaps give a little bit of guidance on this in the "base URL" section.

rartino:
> The fact that the /info endpoint is supposed to inform about available API versions seems to suggest that all the APIs (say, those served under /optimade, /optimade/v1, /optimade/v2) would need to be run from the same instance of an application (otherwise, how could those three APIs know about the availability of the other ones?).
> I'm wondering whether that is a good idea?

The design doesn't necessarily force all versions to be served from one implementation. It just requires `available_api_versions` to be manually configurable to point at all your versions.

But, you have a point in that this puts work on "route (B)" implementations that is susceptible to likely human error. Hence, I'd be fine with the requirement on /info being relaxed to only mandating reporting subversions of the same main version, and optional beyond that. However, reading the specification, I don't actually see a compliance statement about to what degree `available_api_versions` must be complete, so in a strict reading, this may already be allowed (but we should probably clarify that).

That said, a "route (B)" implementation still requires quite a bit of cooperation from the webserver, because versioned URLs must be re-routed to the different implementations, even though they are UNDER an URL that is served by the implementation for the latest version.

Would it be better if we completely remove the serving of the API on non-versioned URLs and REQUIRE clients to indicate their desired main version? I honestly would see that as a win, because the non-versioned "latest version" base URL will just lead to lazy client implementators cursing every time we release a new major version that is not backwards-compatible. I guess this is the reason Facebook still serves 2.11 under graph.facebook.com.

ltalirz:
> reading the specification, I don't actually see a compliance statement about to what degree available_api_versions must be complete, so in a strict reading, this may already be allowed (but we should probably clarify that).

Yes - I think a comment / some guidance on this question in the spec would be helpful.

> That said, a "route (B)" implementation still requires quite a bit of cooperation from the webserver, because versioned URLs must be re-routed to the different implementations, even though they are UNDER an URL that is served by the implementation for the latest version.

It is indeed an invasion of the namespace of the API hosted at the top-level base URL, and that's why I was a bit surprised to see facebook doing this.
However, the fact that it works for them suggests that this is not a problem in practice, and technically it is of course very straightforward to configure at the apache/nginx level, outside the application.
So perhaps there is no need to be too worried about this.

> Would it be better if we completely remove the serving of the API on non-versioned URLs and REQUIRE clients to indicate their desired main version? I honestly would see that as a win, because the non-versioned "latest version" base URL will just lead to lazy client implementators cursing every time we release a new major version that is not backwards-compatible. I guess this is the reason Facebook still serves 2.11 under graph.facebook.com.

Just to be sure I understand correctly: you would disallow serving an API under `example.com/optimade`?

I suspect (?, this was before my time) that one reason for the original spec was to allow database providers the "simple" solution of just having one API under one URL, and to just keep updating the implementation of this one API as they go along.
It's not the cleanest solution but it would seem like a simple one.

However, it's true that allowing unversioned URIs raises the question: how *should* clients check the version of the API they are connected to?
I guess that would have to be the /info endpoint after all, which would bring us back to the beginning of our argument... hm.

rartino:
> Just to be sure I understand correctly: you would disallow serving an API under example.com/optimade

Yes, all API access would have to go through `example.com/optimade/v1`, etc. That is also how people write clients for other APIs today (e.g., facebook, twitter).

> However, it's true that allowing unversioned URIs raises the question: how should clients check the version of the API they are connected to?

I think all use of `example.com/optimade` by automated clients is, more or less, a design error. As far as I can see, it is only for a human that manually (or semi-manually via a client library) explores the API that the unversioned URL makes some sense. The human would actively look at the responses and understand that the interactions work as expected. If the implementation suddenly updates the version served on the non-versioned URL (which could happen in the middle of a session!), there could be errors, but the human would see them, understand what is happening, and react appropriately.

I'm coming to the realization from this discussion that, at the very least, we should include a text segment that heavily discourages non-human-supervised access of the non-versioned URL. But it honestly seem as a major mistake to put in the standard that `example.com/optimade` serves an API version that can at any point suddenly change in a way that isn't backward compatible.

rartino:
@ltalirz Will you update the PR according to the discussion? I had the impression you wanted this through before we released v1.0-rc1? (It would be good to be ready to release at the teleconference on Friday.)

ltalirz:
Very sorry about the delay on this - I've had a lot of urgent work that could not be postponed.

Just to be clear, I should modify the section to require (or recommend?) to *only* use base URLs with the `/v...` suffix , correct?
 
> all API access would have to go through example.com/optimade/v1, etc. That is also how people write clients for other APIs today

If that's the case, how does version discovery work in practice? Do they simply try `<URL>/v1`, `<URL>/v2`, `<URL>/v3`, ... ?
That already becomes a bit more difficult with our scheme, since we don't have integer versioning.

I think I should write a sentence in the spec on how clients are suggested to discover versions, since they shouldn't use the `/info` endpoint.

rartino:
> If that's the case, how does version discovery work in practice? Do they simply try <URL>/v1, <URL>/v2, <URL>/v3, ... ?

Yes, but usually rather in the order <URL>/v3, <URL>/v2, <URL>/v1. 

> That already becomes a bit more difficult with our scheme, since we don't have a single integer as the version number.

The present spec mandates that regardless of what other sub-versions you support as versioned endpoints, the single integer endpoint must be supported (and point to the latest minor version that the API supports.) As long as there is no backward-incompatible change between minor versions, you can simply read, e.g., <URL>/v2/info and learn what minor and patch version you got and what other endpoints exist. I'm just asking for you to re-introduce this requirement in your text.

> I think I should write a sentence in the spec on how clients are suggested to discover versions, since they shouldn't use the /info endpoint.

I think this is fine as an implementors note. But please then set it in the same format as other implementor notes.

ltalirz:
> The present spec mandates that regardless of what other sub-versions you support as versioned endpoints, the single integer endpoint must be supported (and point to the latest minor version that the API supports.)

What I meant is: the spec currently asks for `/v0.10` and then, at some point,  `/v1`. So, how do you search here? `/v3`, `/v2`, `/v1`, `/v0.99`?

merkys:
> What I meant is: the spec currently asks for `/v0.10` and then, at some point, `/v1`. So, how do you search here? `/v3`, `/v2`, `/v1`, `/v0.99`?

As far as I understand, @rartino's comment extends consistently also for `/v0`, thus the search sequence is `/v3`, `/v2`, `/v1`, `/v0`. In our case, `/v0` will point to `/v0.10.0` as per the latest non-stable release.

ltalirz:
Ok, makes sense. It's just not what the spec currently says:

> If the major version is 0, the minor version MUST also be included.

But I agree that if we want this URL-based version discovery, we should probably allow `<URL>/v0`

rartino:
We are just about to release v1.0-rc1 of the specification, and all implementations we are aware about will implement that. So I expect the question about v0 to not be very relevant in practice.

That said, I think the point of the previous formulations was that for pre 1.0 versions, the minor version number takes the same role as the major number post 1.0. I.e., when doing larger, not backward compatible things, we are supposed to update the minor version number. Hence, it makes sense that to access pre v1.0 versions, you have to include also the minor number. 

So, what I formally mean is: /v3, /v2, /v1, /v0.10, /v0.9, /v0.8, ...

I prefer that to remain because I think it is a good practice. But I don't feel very strongly about it (i.e., I'm ok with v0 if you want to have it), since, in practice, this won't matter.

merkys:
> Ok, makes sense. It's just not what the spec currently says:
> 
> > If the major version is 0, the minor version MUST also be included.

Indeed, I forgot about this. Then `/v0` is not a good idea after all.

As pre-releases are sparse, iterating down from `/v1` does not make a lot of sense. Base URL Info Endpoint is supposed to contain `available_api_versions`, which could be sorted in descending order using version ordering. Then the implementation of the latest pre-release spec could be chosen from there.

ltalirz:
@rartino Finally ready for re-review. Here the link to the rendered text:
https://github.com/ltalirz/OPTiMaDe/blob/clarify-baseurl/optimade.rst#base-url

I've gone with `/v0` for the moment, since it makes the specification more compact.
If you feel it's too compact, let me know and we can add special cases / more explicitness back in.

optimade.rst
```diff
@@ -219,39 +219,30 @@ General API Requirements and Conventions
...
-Each database provider will publish one or more base URL that serves the API.
-An example could be: http://example.com/optimade/.
+Each database provider will publish one or more **base URLs** that serve the API, for example: http://example.com/optimade/v1.
 Every URL component that follows the base URL MUST behave as standardized in this API specification.
 
-The client MAY include a version number in the base URL, prefixed with the letter "v", where the version number indicates the version of the API standard that the client requests.
-The format is either vMAJOR or vMAJOR.MINOR where MAJOR is the major version number, and MINOR is the minor version number of the standard being referenced.
-If the major version is 0, the minor version MUST also be included.
-The database provider MAY support further levels of versioning separated from the major and minor version by a decimal point, e.g., patch version on the format vMAJOR.MINOR.PATCH. However, the client MUST NOT assume levels beyond the minor version are supported.
+Base URLs SHOULD have the form :val:`<URL>/vMAJOR`, :val:`<URL>/vMAJOR.MINOR` or :val:`<URL>/vMAJOR.MINOR.PATCH`.
```
```suggestion
All access to the API is provided under **versioned base URLs**, which append a URL component to the base URL that SHOULD be on the form :query-url:`vMAJOR`, :query-url:`/vMAJOR.MINOR` or :query-url:`/vMAJOR.MINOR.PATCH`. An implementation MUST provide the :query-url:`/vMAJOR` form, whereas the other forms are OPTIONAL.
```

.

optimade.rst
```diff
@@ -219,39 +219,30 @@ General API Requirements and Conventions
...
+Here, :val:`<URL>` represents a generic URL, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
```
```suggestion
Here, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
```

optimade.rst
```diff
@@ -219,39 +219,30 @@ General API Requirements and Conventions
...
+For any major version supported by the provider, the :val:`<URL>/vMAJOR` URL MUST point to the **latest** minor/patch version implemented by the provider.
 
-If the client does not include a version number in the base URL, the request is for the latest version of this standard that the database provider implements.
-A query that includes a major and/or minor version is for the latest subversion of that major and/or minor version that the database provider implements.
-
-A database provider MAY choose to only support a subset of possible versions.
-The client can find out which versions are supported using the :field:`available_api_versions` field of the :field:`attributes` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
-The database provider SHOULD strive to implement the latest subversion of any major and minor version supported.
-Specifically, the latest version of this standard SHOULD be supported.
+    **For implementers**: Clients can discover the latest supported version of the standard by cycling through :val:`<URL>/v0`, :val:`<URL>/v1`, :val:`<URL>/v2` etc.
```
rartino: I don't think "cycling" is the right word here (it would mean re-trying them in a loop), and I still think the order should be by priority of the client.
```suggestion
    **For implementers**: Clients are recommended to discover the highest commonly supported version by the client and the API implementation by trying versioned base URL in order of priority, e.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
```

optimade.rst
```diff
@@ -219,39 +219,30 @@ General API Requirements and Conventions
...
 
 Examples of valid base URLs:
```
```suggestion
Examples of valid versioned base URLs:
```

optimade.rst
```diff
@@ -219,39 +219,30 @@ General API Requirements and Conventions
...
 
-- http://example.com/optimade/
-- http://example.com/optimade/v0.9/
-- http://example.com/
-- http://example.com/some/path/
+- http://example.com/optimade/v0/
+- http://example.com/v0.9.1/
+- http://example.com/v1/
 
 Examples of invalid base URLs:
```
```suggestion
Examples of invalid versioned base URLs:
```

optimade.rst
```diff
@@ -219,39 +219,30 @@ General API Requirements and Conventions
...
+- http://example.com/optimade/
+
+Database providers SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any major and minor version they support.
 
-Note: The OPTiMaDe standard specifies the response from a number of endpoints under the base URLs.
-However, the base URLs themselves are not considered to be a part of the API.
-Hence, they are fully under the control of the API implementation.
-It is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, and this document is used to explain that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
+Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to a base URL.
+However, it is RECOMMENDED that implementations serve a human-readable HTML document on a base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
```
rartino: I don't understand why this word was changed, so I revert it because "each" is more precise
```suggestion
However, it is RECOMMENDED that implementations serve a human-readable HTML document on each base URL, which explains that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
```

optimade.rst
```diff
@@ -219,39 +219,30 @@ General API Requirements and Conventions
...
 - http://example.com/optimade/0.9/
+- http://example.com/optimade/
+
+Database providers SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any major and minor version they support.
 
-Note: The OPTiMaDe standard specifies the response from a number of endpoints under the base URLs.
-However, the base URLs themselves are not considered to be a part of the API.
-Hence, they are fully under the control of the API implementation.
-It is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, and this document is used to explain that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
+Note: The base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to a base URL.
```
```suggestion
Note: The base URLs and versioned base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to them.
```

ltalirz:
Thanks for the suggestions @rartino !

While starting to look through uses of "base URL" in the rest of the document, I noticed this:
https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#base-url-info-endpoint
> The Info endpoint on the base URL or directly after the version number (e.g. http://example.com/optimade/v0.9/info) returns information relating to the API implementation.

I.e. there is already an explicit distinction between the base URL and the version-specific baser URL here, but if I understand correctly, the unspecific one will disappear from here, since we will no longer support it (correct?).

Before I go about replacing things in the rest of the document: In general, is there any place where you suggest to use the term "base URL" meaning the `<URL>` part, and not `<URL>/v1`?
Because if not, it would seem easier to just talk about base URLs.

rartino:
>> The Info endpoint on the base URL or directly after the version number (e.g. http://example.com/optimade/v0.9/info) returns information relating to the API implementation.

> I.e. there is already an explicit distinction between the base URL and the version-specific base URL here, but if I understand correctly, the unspecific one will disappear from here, since we will no longer support it (correct?).

Yes. e.g., "The Info endpoint is located directly under a versioned base URL (e.g. http://example.com/optimade/v0.9/info) and returns information relating to the API implementation."

> Before I go about replacing things in the rest of the document: In general, is there any place where you suggest to use the term "base URL" meaning the <URL> part, and not <URL>/v1?
Because if not, it would seem easier to just talk about base URLs instead of version-specific ones.

I've looked through all mentions of "Base URL". The ones I see that refer to the **un**-versioned URL are:
- Section 3.2 Index Meta-Database in several places.
- Section 3.3 Database-Provider-Specific Namespace Prefixes: reference to the providers index-meta database.
- Section 4.2 JSON Response Schema: Common Fields: definition of `index_base_url` field under `meta->provider`.
- Section 5.4.1 Links Endpoint JSON Response Schema: `attributes->base_url`.

Then also for Section 4.2 JSON Response Schema: Common Fields we need to decide if the field `representation` should include the version number path segment or not (it is presently unclear, which is a weakness in the specification...)

And the thing that prevents renaming what "Base URL" refers to is in particular the definition of the `base_url` field in the `links` endpoint. It absolutely should refer to the un-versioned URL and I'm strongly opposed to renaming that field now at this late point since it directly affects all implementations. On the other hand "versioned base URL" could be given a simpler name, if you have any good ideas.

ltalirz:
> I've looked through all mentions of "Base URL". The ones I see that refer to the un-versioned URL are:

> * Section 3.2 Index Meta-Database in several places.

Right - but taking your first [example](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#index-meta-database), following your suggestion this will then need to be replaced by "versioned base URL".
After all, the meta index DB is a special optimade api (with index and links endpoints)  and should be versioned just as the regular api.

I'll now go through the document and have a stab at replacing these mentions.

ltalirz:
Ok, I see now that the index meta database is already served under an "unversioned" URL https://www.optimade.org/providers/ and I guess you don't want to change that?

I personally think simplicity is one of the most important aspects driving adoption and reliability of a standard, and if possible we should avoid having multiple notions of base URLs + exceptions for index meta databases etc.

But it's also late, and I don't have a good feeling for what is acceptable to change at this stage and what isn't.
 @rartino You seem to have a clear idea of what changes you want to see and what you want the standard to look like.
In order not to waste your time with this back-and-forth, can I suggest you close this PR and either start from it (or from scratch) to make the changes you have in mind?

Happy to help reviewing for logic issues; I'll hold back concerning complexity etc. for this.

rartino:
I think you may have misunderstood me.

The URL https://www.optimade.org/providers/ is the (un-versioned) base URL of the providers index meta-database. It should be expected to work as a usual (non-versioned) base URL, i.e., a client interacting with it MUST append a version number. That works right now. (But we should *remove* the ability to access /info and /links without appending a version number.)

However, the term "base URL" appears three times in section 3.2. Two of those refer to un-versioned base URLs, and one (which is part of a section title) refers to the versioned base URL.

> In order not to continue wasting your time with this back-and-forth, can I suggest you close this PR and either start from it (or from scratch) to make the changes you have in mind?

That is fair-- the idea was really that we were supposed to release v1.0-rc1 this weekend, and this PR is now delaying this. I think I will be able to push to your branch (unless you deliberately unchecked that when you submitted the PR), so I will keep the PR open and do that. I should be able to get to this in the evening.

ltalirz:
Sounds good!

I've made the PR from my fork, so I think you won't be able to push to it by default, but I've invited you as a collaborator to the fork, so you'll be able to.

rartino:
Ok, I think I am done now. Please review.

Note that this now actually affects some response examples, and thus may lead to us needing some adjustments in our implementations. But I see no option to doing that when we had fields named `base_url` that in some examples were shown adding the version path segment, and sometimes not.

(Please make sure to ASAP approve #244 as well so we can fix the travis error here.)

optimade.rst
```diff
@@ -219,39 +225,30 @@ General API Requirements and Conventions
...
-The database provider MAY support further levels of versioning separated from the major and minor version by a decimal point, e.g., patch version on the format vMAJOR.MINOR.PATCH. However, the client MUST NOT assume levels beyond the minor version are supported.
+Each database provider will publish one or more **base URLs** that serve the API, for example: http://example.com/optimade/.
+Every URL path segment that follows the base URL MUST behave as standardized in this API specification.
 
-If the client does not include a version number in the base URL, the request is for the latest version of this standard that the database provider implements.
-A query that includes a major and/or minor version is for the latest subversion of that major and/or minor version that the database provider implements.
+All access to the API is provided under the **versioned base URLs**, which append a URL path segment to the base URL that SHOULD be on the form :query-url:`/vMAJOR`, :query-url:`/vMAJOR.MINOR` or :query-url:`/vMAJOR.MINOR.PATCH`. An implementation MUST provide the :query-url:`/vMAJOR` form, whereas the other forms are OPTIONAL.
```
ltalirz:
```suggestion
All access to the API is provided under the **versioned base URLs**, which append a URL path segment to the base URL that SHOULD be of the form :query-url:`/vMAJOR`, :query-url:`/vMAJOR.MINOR` or :query-url:`/vMAJOR.MINOR.PATCH`. An implementation MUST provide the :query-url:`/vMAJOR` form, whereas the other forms are OPTIONAL.
```

optimade.rst
```diff
@@ -219,39 +225,30 @@ General API Requirements and Conventions
...
+Here, :val:`MAJOR` is the major version number, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the standard implemented by the provider.
+For all major versions supported by the provider, the :val:`/vMAJOR` URL MUST serve the *latest* minor/patch version implemented by the provider.
 
-A database provider MAY choose to only support a subset of possible versions.
-The client can find out which versions are supported using the :field:`available_api_versions` field of the :field:`attributes` field from a query to the base URL :endpoint:`info` endpoint (see section `Base URL Info Endpoint`_).
-The database provider SHOULD strive to implement the latest subversion of any major and minor version supported.
-Specifically, the latest version of this standard SHOULD be supported.
+    **For implementers**: Clients are recommended to discover the highest commonly supported version by the client and the API implementation by trying versioned base URLs in order of priority, e.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
```
ltalirz:
```suggestion
    **For implementers**: Clients are recommended to discover the highest version supported by both the client and the API implementation by trying versioned base URLs in order of priority.
    E.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
```

optimade.rst
ltalirz: One could interpret "all access to the API" to mean that this sentence should also apply to index meta databases.
Could you please add a sentence (or half a sentence) to section 3.2 that explicitly states that meta index databases do *not* need to be served under versioned base URLs?

optimade.rst
```diff
@@ -134,6 +134,12 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 **Identifier**
     Names that MUST start with a lowercase letter ([a-z]) or an underscore ("\_") followed by any number of lowercase alphanumerics ([a-z0-9]) and underscores ("\_").
 
+**Base URL**
+    The topmost URL under which the API is served. See section `Base URL`_.
```
ltalirz: This description can be a bit misleading, since (if I understand correctly), there isn't necessarily any API served under the base URL directly.

At the same time, I struggle to find a compact description that explains the actual meaning better.
One way to make things a bit more clear is to add an example here as well.
```suggestion
    The topmost URL under which the API is served, e.g. http://example.com/optimade/. See section `Base URL`_.
```

optimade.rst
```diff
@@ -134,6 +134,12 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
...
+
+**Versioned base URL**
+   A URL formed by the base URL plus a path segment indicating a version of the API. See section `Base URL`_.
```
ltalirz:
```suggestion
   A URL formed by the base URL plus a path segment indicating a version of the API, e.g. http://example.com/optimade/v1. See section `Base URL`_.
```

optimade.rst
ltalirz: Ah, I just fully read you previous comment - so, we *should* be using versioned base URLs for the index meta database (which is good).

Perhaps, could you then clarify in the opposite direction, i.e. although the section 3.2 only mentions "base URL", that access should occur via *version* base URLs?
I think this is not obvious from reading section 3.2

rartino: I don't understand the need to clarify this since there is no distinction between index meta-databases and regular databases in this respect.

I suspect any confusion about this mostly stems from having been exposed to our earlier (admittedly confusing) lack of distinction between various types of "base URL". But now "base URL" means one and only one single thing: the URL to which one adds a version number path segment to access API functions. I honestly think the type of clarification you ask for rather risk adding confusion since it sounds as if this behavior is somehow different for index meta-databases, which it isn't.

We anyway need someone else to approve this PR before we can merge. I ask them to weigh in on this before I make any change here.

optimade.rst
rartino: I'm not strongly opposed, but I also rather avoid bloating the text unnecessarily. Does adding an example here with no additional explanation really provide a benefit? Literally *any* valid URL can be a base URL, so how does the example help? The text is also directly followed by a link to the section with the full discussion about what a base URL is, where that same example appears.

optimade.rst
ltalirz: Ok, if others think this is clear, happy to withdraw this comment.

optimade.rst
ltalirz: Yeah... I agree that the examples don't really solve the problem...
it's difficult to explain the "base URL" without referring to the versioned base URL first.

I'd just argue that it is the *versioned* base URL under which the API is served.
The base URL is simply the versioned base URL minus the version suffix.

Anyhow, I don't mind much going ahead with the current version.

ltalirz:
@rartino As this is my PR, you'll need to approve ;-)

optimade.rst
```diff
@@ -219,39 +225,31 @@ General API Requirements and Conventions
...
+- http://example.com/optimade/
+
+Database providers SHOULD strive to implement the latest released version of this standard, as well as the latest patch version of any major and minor version they support.
 
-Note: The OPTiMaDe standard specifies the response from a number of endpoints under the base URLs.
-However, the base URLs themselves are not considered to be a part of the API.
-Hence, they are fully under the control of the API implementation.
-It is RECOMMENDED that the implementation serves a human-readable HTML document on each base URL, and this document is used to explain that the URL is an OPTiMaDe base URL meant to be queried by an OPTiMaDe client.
+Note: The base URLs and versioned base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to them.
+However, it is RECOMMENDED that implementations serve a human-readable HTML document on base URLs and versioned base URLs, which explains that the URL is an OPTiMaDe URL meant to be queried by an OPTiMaDe client.
```
CasperWA: Since we are doing release candidates and similar, we should probably mention in a note here that release candidates are versioned as its named version without the appendix, i.e., `1.0-rc` -> `1.0`, right?

rartino: You definitely have a point that this should be clarified.

ltalirz:
you guys can go ahead and merge

rartino:
Let's say this is good enough for the release candidate, and if there is a desire to further clarify, PRs for that can be handled before the proper release.

In any case, thanks @ltalirz for the first version; this really resolves a confusing inconsistency in the spec.

PR #241, opened 2020-01-23
Fix available_api_versions example issue found via validation

rartino:
This PR fixes an error in the example for the info endpoint response for the index meta-database.

(It was discovered thanks to @ml-evs great work with adding a validator to the providers repository, see Materials-Consortia/providers#12)


rartino:
Anyone has any idea why the travis check failed? It looks as some oddity with the perl environment. I will close and re-open this PR to trigger a rebuild (I saw it claimed that this is how to trigger a rebuild without having to push a commit)

PR #242, opened 2020-01-24
Add OpenAPI schema JSON

CasperWA:
From optimade-python-tools [v0.3.3](https://github.com/Materials-Consortia/optimade-python-tools/tree/v0.3.3).


openapi_schema.json
```diff
@@ -0,0 +1,4515 @@
+{
+  "openapi": "3.0.2",
+  "info": {
+    "title": "OPTiMaDe API",
+    "description": "The [Open Databases Integration for Materials Design (OPTiMaDe) consortium](http://http://www.optimade.org/) aims to make materials databases interoperational by developing a common REST API.",
+    "version": "0.10.1"
```
merkys: We should make a big notice to whoever does the release to remember to bump this version number here. Should we add something like release manual?

openapi_schema.json
```diff
@@ -0,0 +1,4515 @@
+{
+  "openapi": "3.0.2",
+  "info": {
+    "title": "OPTiMaDe API",
+    "description": "The [Open Databases Integration for Materials Design (OPTiMaDe) consortium](http://http://www.optimade.org/) aims to make materials databases interoperational by developing a common REST API.",
```
merkys: As this API description is autogenerated (right?) by specific software (I hope by a tagged release), it would be nice to see some metadata telling the reader where this document originated from.

shyamd:
The intent of this is just to have a schema for the first release candidate, such that there is some sort of schema for the optimade API. In the future, we plan on moving the `pydantic` models over to this repo and have the JSON schema built based on those, but didn't want to make that a blocking pull request for the release candidate.

CasperWA:
> I wonder why the CI checks have failed. IMHO this should be fixed before merging.

This seems to be an issue with the CI implementation, not this PR.

openapi_schema.json
CasperWA: Good idea! [This](https://github.com/Materials-Consortia/OPTiMaDe/wiki/Versioning) wiki page was started to give a step-by-step on what to do when upping the version of the spec, maybe it could be added here (and given a better page title)?

openapi_schema.json
CasperWA: Good idea. I will make a PR for this to the `optimade-python-tools` and subsequently add the results here.

CasperWA: See the PR Materials-Consortia/optimade-python-tools#151

openapi_schema.json
merkys: Great! Thanks for the link, I wasn't aware of it (or was, but forgot). Surely after acceptance of this PR the note should be added to the wiki page to update the version in the OpenAPI schema.

openapi_schema.json
merkys: Thanks - I've approved the https://github.com/Materials-Consortia/optimade-python-tools/pull/151.

merkys:
Once https://github.com/Materials-Consortia/optimade-python-tools/pull/151 is accepted and the schema is updated here, I'll happily approve. Update should trigger the CI checks too, thus we'll likely get the green tick too.

rartino:
- I updated the schema file the way I think @merkys asked for. 
- I also copied the index meta-database schema, since I don't see why we shouldn't included that as well?
- I put the schemas into a subdirectory to not clutter the root directory.
- I pushed the changes to @CasperWA's repository since he did not disallow this, I hope I have not overstepped any ethical line here, but it should be easy to revert.

Please re-review, especially @merkys and @CasperWA.

ml-evs:
Could we add a note in the README as to what these are, and the caveats between them and the human-readable spec? Something like:

> OpenAPI schema as implemented by [Materials-Consortia/optimade-python-tools](https://github.com/Materials-Consortia/optimade-python-tools) for the [main OPTiMaDe API](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/schemas/openapi_schema.json) and [index meta-database](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/schemas/index_openapi_schema.json)  can be found in `schemas/`. These schema are an approximation of the full human-readable specification and may be missing certain additional constraints, or may fall out of date.

I missed the discussion about the plan to move the models into this repo too, so maybe my understanding of these generated schema is incorrect.

rartino:
It is a good idea with a note in the readme. If they are out of date in the master branch, that is a bug; but maybe if I slightly rephrase your sentence to refer to the develop branch.

CasperWA:
Indeed a good idea with the note. I have slightly adapted the text in the README, no actual content change.
I think this is ready to go now.

README.md
```diff
@@ -10,6 +10,10 @@ This repository contains the specification of the OPTiMaDe API.
 * [optimade.org](https://www.optimade.org): Public OPTiMaDe web site
 * [OPTiMaDe wiki](https://github.com/Materials-Consortia/OPTiMaDe/wiki): Information for developers
 
+The subdirectory `schemas/` contains OpenAPI schemas for the main OPTiMaDe API and index meta-database as implemented by the [optimade-python-tools repository](https://github.com/Materials-Consortia/optimade-python-tools).
+_Note_: These schemas are an approximation of the full human-readable specification and may be missing certain constraints.
+Furthermore, they may not be up to date in the develop branch of this repository.
```
merkys: We could have also said that when in doubt over conflicting requirements between the human-readable document and the OpenAPI, one should hold the human-readable document authoritative.

merkys:
All good!

PR #246, opened 2020-02-01
Update README.md recommendations for implementations beyond v1.0-rc1

rartino:
Beyond v1.0-rc1 we should not recommend external developers to implement a moving development version of the specification, or chaos will ensue.

\- "My server implements v1.1-develop"  
\- "My client too"  
\- "Then why are they incompatible?!"

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,5 +17,5 @@ A commit tagged with the suffix `-rc<number>` indicates a release candidate vers
 
 The [develop branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) contains the present in-development version of the specification.
 
-API and client implementations are RECOMMENDED to implement the lastest named release.
+API and client implementations are RECOMMENDED to implement the latest named release.
 If the latest named release is a release candidate, it is also RECOMMENDED that they implement the latest stable release.
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ This repository contains the specification of the OPTiMaDe API.
 
 ## For developers
 
-The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named release of the specification.
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named version of the specification.
 A commit tagged with a version number without a suffix (alpha, beta, release candidates and similar) indicates a stable release.
 
 The [develop branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) contains the present in-development version of the specification.
 
-API and client implementations are encouraged to support the latest named release.
-If the latest named release is a release candidate, implementations are also encouraged to support the latest stable release.
+API and client implementations are encouraged to support the latest version of the release.
+If the latest version is a release candidate, implementations are also encouraged to support the latest stable release.
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ This repository contains the specification of the OPTiMaDe API.
 
 ## For developers
 
-The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is supposed to be at the latest named release of the specification.
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named release of the specification.
 A commit tagged with a version number without a suffix indicates a stable release.
 A commit tagged with the suffix `-rc<number>` indicates a release candidate version. 
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This repository contains the specification of the OPTiMaDe API.
 ## For developers
 
 The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest release or pre-release version of the specification.
-Versions without a version number suffix (alpha, beta, release candidates and similar) indicates a stable release.
+Versions without a version number suffix (alpha, beta, release candidates and similar) indicate a stable release.
 
 The [develop branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) contains the present in-development version of the specification.
 
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ This repository contains the specification of the OPTiMaDe API.
 
 ## For developers
 
-The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named version of the specification.
-A commit tagged with a version number without a suffix (alpha, beta, release candidates and similar) indicates a stable release.
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest release or pre-release version of the specification.
+Versions without a version number suffix (alpha, beta, release candidates and similar) indicates a stable release.
 
 The [develop branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) contains the present in-development version of the specification.
 
-API and client implementations are encouraged to support the latest version of the release.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,11 +11,11 @@ This repository contains the specification of the OPTiMaDe API.
 
 ## For developers
 
-The latest "stable" version of the specification is found in the [master](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) branch.  
-The latest "in development" version is found in the [develop](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) branch.
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is supposed to be at the latest named release of the specification.
+A commit tagged with a version number without a suffix indicates a stable release.
+A commit tagged with the suffix `-rc<number>` indicates a release candidate version.
```
CasperWA: I think this may be too rigorous a text for a README? E.g., I would expect developers to know that `-rc<number>` indicates a release candidate version.

README.md
```diff
@@ -12,11 +11,11 @@ This repository contains the specification of the OPTiMaDe API.
 
 ## For developers
 
-The latest "stable" version of the specification is found in the [master](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) branch.  
-The latest "in development" version is found in the [develop](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) branch.
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is supposed to be at the latest named release of the specification.
```
CasperWA:
```suggestion
The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named release of the specification.
```

README.md
```diff
@@ -12,11 +11,11 @@ This repository contains the specification of the OPTiMaDe API.
...
+A commit tagged with a version number without a suffix indicates a stable release.
+A commit tagged with the suffix `-rc<number>` indicates a release candidate version. 
 
-If you are a **server** developer, it is _recommended_ to always implement the latest stable version. It is also _recommended_ to implement the latest version in development.
-You may want to use the same _master_ and _develop_ branch model employed here.
+The [develop branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) contains the present in-development version of the specification.
 
-If you are a **client** developer, you are _encouraged_ to support at least the latest stable version, while it is _recommended_ to support the latest version in development.
-You may want to use the same _master_ and _develop_ branch model employed here.
+API and client implementations are RECOMMENDED to implement the lastest named release.
```
CasperWA:
```suggestion
API and client implementations are RECOMMENDED to implement the latest named release.
```

README.md
```diff
@@ -12,11 +11,11 @@ This repository contains the specification of the OPTiMaDe API.
...
+If the latest named release is a release candidate, it is also RECOMMENDED that they implement the latest stable release.
```
CasperWA: Latest `stable` implementations should probably be a `SHOULD`, while everything else is `RECOMMENDED`?

README.md
rartino: Well, *during* the time we do a new release, it may not be for a short while; which is the reason for the weasel words. But is no one else thinks that caveat matters, I guess I can remove them.

README.md
rartino: Please then use the suggest change option to suggest a less rigorous phrasing that explains that these are the two possibilities.

README.md
rartino: Much like your comment above, I'm not sure a readme is the right place to put SHOULD level requirements on anyone. But I brought this on myself by capitalizing RECOMMENDED. How about we just revert to the word "encouraged" as used in the prior version, and not view this readme a formal part of the standard.

README.md
CasperWA: Sure. For this I didn't have a concrete suggested change though. Mostly, I was thinking of the use of RECOMMENDED and so on - the sentiment I was expressing in the other comments again. Anyway, maybe for this one, one could just remove this line (about the `rc`)? And then added an inserted sentence in the prior one (see suggested change).

README.md
```diff
@@ -12,11 +11,11 @@ This repository contains the specification of the OPTiMaDe API.
 
 ## For developers
 
-The latest "stable" version of the specification is found in the [master](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) branch.  
-The latest "in development" version is found in the [develop](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) branch.
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named release of the specification.
+A commit tagged with a version number without a suffix indicates a stable release.
```
CasperWA:
```suggestion
A commit tagged with a version number without a suffix (alpha, beta, release candidates and similar) indicates a stable release.
```

README.md
CasperWA:
```suggestion
```

README.md
```diff
@@ -12,11 +11,10 @@ This repository contains the specification of the OPTiMaDe API.
...
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named release of the specification.
+A commit tagged with a version number without a suffix (alpha, beta, release candidates and similar) indicates a stable release.
 
-If you are a **server** developer, it is _recommended_ to always implement the latest stable version. It is also _recommended_ to implement the latest version in development.
-You may want to use the same _master_ and _develop_ branch model employed here.
+The [develop branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) contains the present in-development version of the specification.
 
-If you are a **client** developer, you are _encouraged_ to support at least the latest stable version, while it is _recommended_ to support the latest version in development.
-You may want to use the same _master_ and _develop_ branch model employed here.
+API and client implementations are encouraged to support the latest named release.
```
merkys: Is *named release* an established term? (In other words: could there be releases other than named?) To me the addition of *named* introduces more confusion when compared to just *the latest release*, I'd feel more comfortable with *named* excluded from the paragraph.

rartino: Hmm... I tried follow your suggestion, but had some trouble with

> If the latest named release is a release candidate, implementations are also encouraged to support the latest stable release.

Which looks wonky if one removes "named". Because a release cannot simultaneously be a release candidate.

I'll try with replacing "named release" with version. See new version of the PR (I'll push the commit in a moment.)

rartino: Please check now; note that I keep one "named version" which perhaps is more ok than "named release"?

(I actually think "named release" IS an established term, though. It has 80,000 hits on google.)

README.md
```diff
@@ -11,10 +11,10 @@ This repository contains the specification of the OPTiMaDe API.
 
 ## For developers
 
-The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named release of the specification.
+The [master branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/master) is at the latest named version of the specification.
```
merkys: I suggest removing *named* from here. We will not have versions other than named, I guess.

README.md
```diff
@@ -11,10 +11,10 @@ This repository contains the specification of the OPTiMaDe API.
...
 A commit tagged with a version number without a suffix (alpha, beta, release candidates and similar) indicates a stable release.
 
 The [develop branch of the repository](https://github.com/Materials-Consortia/OPTiMaDe/tree/develop) contains the present in-development version of the specification.
 
-API and client implementations are encouraged to support the latest named release.
-If the latest named release is a release candidate, implementations are also encouraged to support the latest stable release.
+API and client implementations are encouraged to support the latest version of the release.
```
merkys: release -> specification

rartino:
Ok, @merkys makes good points about version terminology. I've now updated the PR to only use SemVer terminology.

PR #248, opened 2020-02-03
Update OpenAPI schemas to 1.0.0-rc1 (1.0.0)

CasperWA:
As per the comment by @merkys [here](https://github.com/Materials-Consortia/OPTiMaDe/pull/247#issuecomment-581402624), this PR updates the OpenAPI schemas to have the versioned base URL `/optimade/v1` (which technically is not the full base URL, since it's missing the domain, but for OpenAPI and this schema, this is irrelevant).

I have also removed the repetitions for `/optimade/v1.0` and `/optimade/v1.0.0`.

Furthermore, the version has been upped to `1.0.0` - should it be `1.0.0-rc1` instead?


rartino:
Good question, I think 1.0.0 is good. The point with a release candidate is that we shouldn't need to edit much if it is good.

CasperWA:
> Is editing schemas this way directly rather than via the Python implementation in optimade-python-tools a problem? It seems these changes will now need to be back-propagated, or we will have to re-do these changes next time the Schemas generated from Python are updated.

The schemas can be edited directly, that's fine.
These were "edited" via my local `optimade-python-tools` though, and then I copied them over. In other words, these were generated from `optimade-python-tools`, if that was not clear.
I already have a local branch of `optimade-python-tools` that I will create a PR for to update the OpenAPI generation to be similar to these, i.e., without the extra versioned endpoints included.

I am not sure exactly what you mean by the last sentence, but I think the idea is to have the schema generation move to this repo (or a separate one), where the constant updating in two places will not be necessary?

PR #252, opened 2020-02-05
Update master for pre-release v1.0.0-rc1

rartino:
Feel free to approve, but do not merge this PR. It needs to be done manually on the command line as per [instructions in the wiki](https://github.com/Materials-Consortia/OPTiMaDe/wiki/New-release-%28How-to%29).

**Clarification:** we actually need 3 approvals on this to go ahead.


rartino:
Sure, I volunteer. 

(Unless someone feels very strongly about it, I suggest that the normal procedure is that the one posting the PR also does the merging.)

Issue #255, opened 2020-02-11
What version number should we set on our develop branch now?


rartino:
Our wiki [new release page](https://github.com/Materials-Consortia/OPTiMaDe/wiki/New-release-%28How-to%29) states that after merging develop into master as fast forward, we shall create a PR to change the version number in the specification document in the develop branch before merging more PRs. 

>  Increase the patch version number to reflect the next release target in the title of the optimade.rst 

However, we are now at v1.0.0-rc.1, next release may be v1.0.0-rc.2 or v1.0.0. It most certainly will not be v1.0.1. 

The key issue seems to be: can we set a version string for the developer branch that is SemVer-sorted later than the just released version, but before any actual release we may do in the future? It turns out, this isn't easily solved, and others have noted the same. I found a discussion among the SemVer github issues: semver/semver#124.

They land in recommending to update the version number continuously based on what changes are merged into develop. I.e., starting from v1.0.0, as soon as a patch is merged, *the PR with that patch* should also change the version string to v1.0.1-develop. As soon as new functionality is merged, *the PR that adds functionality* should change the version string to v1.1.0-develop. As soon as something backward breaking is merged: v2.0.0-develop. 

If we follow this, we should go with v1.0.0-rc.2-develop now. There will be no big issues with this scheme right now, since anything can go in the release candidates that turn v0.x.x into v1.0.0.

But, I'm not super enthusiastic, because:
- In the future, this will make many PRs mess around with the version number, causing unnecessary merge conflicts.
- For *every PR we consider* we will have to look at the development history and ask "has the version number already been bumped sufficiently for this PR?"

As an alternative proposal, we could instead in our development branch use a version string that adds develop in a way that purposely breaks (!) SemVer as `<last release>~develop`. E.g, v1.0.0-rc.1~develop. 

Why? Because development versions are not properly SemVer ordered anyway; between releases the version string is just a placeholder to clarify that you are looking at a file in which active development has been added on top of the last release. Unless we actually bump up the patch number *for every PR we merge* we are anyway breaking the spirit of SemVer by there being multiple versions with the same version string. To me it seems to be a feature, rather than an issue, that any code that tries to treat these version strings according to SemVer explicitly breaks. 

Then, as soon as we are ready to release, we remove the non-SemVer `~develop` suffix, and the version strings we use for releases are all SemVer-compliant.

We also *could* attempt to petition SemVer to add support for ~develop :-)

merkys:
[...]

PR #256, opened 2020-02-18
Update development version string

rartino:
This updates the version string at the top of optimade.rst in the development branch according to the discussion in #255.

(However, there is another small but awkward issue. The previous release has been referred to as `v1.0.0-rc1`, but to fully follow SemVer praxis it should be `v1.0.0-rc.1` (note the extra dot). Our repository release and git tag use the correct SemVer praxis so the present pre-release will sort correctly compared to upcoming pre-releases. Hence, the only issue is the version string at the top of the present optimade.rst, so I suggest we view that as a bug, which this PR fixes.)


rartino:
@CasperWA Indeed, the key point is that the lack of dot in the present master version is a bug, which is fixed from now on.

PR #257, opened 2020-02-19
Clarify when lattice_vectors MAY be null and how

### LINKED ISSUE #253: Clarify when `lattice_vectors` MAY be `null` and how
During the creation of a model validator for the Python tools (Materials-Consortia/optimade-python-tools#171) we find that the current description of `lattice_vectors` and how it MAY contain `null` vectors are a bit ambiguous.

It states that:

>- This property MUST be an array of dimensions 3 times 3 regardless of the elements of `dimension_types`. The vectors SHOULD by convention be chosen so the determinant of the `lattice_vectors` matrix is different from zero. The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
>- All three elements of the inner lists of floats MAY be `null` for non-periodic dimensions, i.e., those dimensions for which `dimension_types` is `0`.

Does this mean a vector within `lattice_vectors` MUST be either all `null` values or all number values?
And for calculating the determinant, one must then use `0` for `null` or?

Tagging the contributors to the references PR, @shyamd @ml-evs

CasperWA:
Fixes #253 

This may be a bit verbose, but should clarify what is truly meant by the requirements and conventions listed for `lattice_vectors` under the `structures` entry according to the discussion in #253.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1766,10 +1766,10 @@ lattice\_vectors
   - MUST be a list of three vectors *a*, *b*, and *c*, where each of the vectors MUST BE a list of the vector's coordinates along the x, y, and z Cartesian coordinates.
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
   - For databases that do not define an absolute Cartesian system (e.g., only defining the length and angles between vectors), the first lattice vector SHOULD be set along *x* and the second on the *xy*-plane.
-  - This property MUST be an array of dimensions 3 times 3 regardless of the elements of :property:`dimension_types`.
+  - MUST always contain three vectors of three coordinates each, independently of the elements of :property:`dimension_types`.
     The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero.
     The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
-  - In the case of non-periodic dimensions, i.e., those dimensions for which :property:`dimension_types` is :val:`0`, implementers MAY choose to replace the values with a list of :val:`null` values.
+  - The coordinates of the lattice vectors of non-periodic dimensions (i.e., those dimensions for which :property:`dimension_types` is :val:`0`) MAY be given as a list of :val:`null` values.
 
 - **Examples**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1757,7 +1757,7 @@ lattice\_vectors
 ~~~~~~~~~~~~~~~~
 
 - **Description**: The three lattice vectors in Cartesian coordinates, in ångström (Å).
-- **Type**: list of list of floats.
+- **Type**: list of list of floats and/or unknown values.
 - **Requirements/Conventions**:
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
@@ -1770,6 +1770,7 @@ lattice\_vectors
     The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero.
     The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
   - All three elements of the inner lists of floats MAY be :val:`null` for non-periodic dimensions, i.e., those dimensions for which :property:`dimension\_types` is :val:`0`.
+    To clarify, this means a vector MUST he either all floats or all :val:`null` for non-periodic dimensions; a single :val:`null` value MUST never be found.
 
 - **Examples**:
 
@@ -1779,7 +1780,7 @@ cartesian\_site\_positions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 - **Description**: Cartesian positions of each site. A site is an atom, a site potentially occupied by an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual crystal approximation).
-- **Type**: list of list of floats and/or unknown values
+- **Type**: list of list of floats and/or unknown values.
 - **Requirements/Conventions**:
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
--- a/optimade.rst
+++ b/optimade.rst
@@ -1770,7 +1770,7 @@ lattice\_vectors
     The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero.
     The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
   - All three elements of the inner lists of floats MAY be :val:`null` for non-periodic dimensions, i.e., those dimensions for which :property:`dimension\_types` is :val:`0`.
-    To clarify, this means a vector MUST he either all floats or all :val:`null` for non-periodic dimensions; a single :val:`null` value MUST never be found.
+    To clarify, this means a vector MUST be either all floats or all :val:`null` for non-periodic dimensions; a single :val:`null` value MUST never be found.
 
 - **Examples**:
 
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1770,6 +1770,7 @@ lattice\_vectors
     The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero.
     The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
   - All three elements of the inner lists of floats MAY be :val:`null` for non-periodic dimensions, i.e., those dimensions for which :property:`dimension\_types` is :val:`0`.
+    To clarify, this means a vector MUST he either all floats or all :val:`null` for non-periodic dimensions; a single :val:`null` value MUST never be found.
```
ml-evs: Typo:

```suggestion
    To clarify, this means a vector MUST be either all floats or all :val:`null` for non-periodic dimensions; a single :val:`null` value MUST never be found.
```

optimade.rst
```diff
@@ -1770,6 +1770,7 @@ lattice\_vectors
     The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero.
     The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
   - All three elements of the inner lists of floats MAY be :val:`null` for non-periodic dimensions, i.e., those dimensions for which :property:`dimension\_types` is :val:`0`.
```
ml-evs: I think I have more of an issue with this line, and perhaps we could avoid the clarification if this part is rephrased, e.g.

> In the case of non-periodic dimensions (i.e. those dimensions for which :property:`dimension\_types` is :val:`0`, implementers MAY choose to replace the values for that dimension with a list of :val:`null` values.

optimade.rst
```diff
@@ -1757,7 +1757,7 @@ lattice\_vectors
 ~~~~~~~~~~~~~~~~
 
 - **Description**: The three lattice vectors in Cartesian coordinates, in ångström (Å).
-- **Type**: list of list of floats.
+- **Type**: list of list of floats and/or unknown values.
```
ml-evs: Is "unknown values" unambiguous here (and below) in a "type" field? Perhaps we should be more verbose here too, e.g.

> a list of 3 lists, where each sublist also has length 3 and consists of all floats or all :val:`null` values.

CasperWA: Yes and no. I think for the `Type` entry there shouldn't be any restraints or requirements. These should be clearly listed underneath. Here it's only a matter of _what_ it is, i.e., a list of lists containing either floats or unknown values (the `and` should be removed).

ml-evs: Works for me. I guess `cartesian_site_positions` is different as individual values can be `null` to signify uncertainty.

optimade.rst
```diff
@@ -1766,10 +1766,10 @@ lattice\_vectors
   - MUST be a list of three vectors *a*, *b*, and *c*, where each of the vectors MUST BE a list of the vector's coordinates along the x, y, and z Cartesian coordinates.
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
   - For databases that do not define an absolute Cartesian system (e.g., only defining the length and angles between vectors), the first lattice vector SHOULD be set along *x* and the second on the *xy*-plane.
-  - This property MUST be an array of dimensions 3 times 3 regardless of the elements of :property:`dimension\_types`.
+  - This property MUST be an array of dimensions 3 times 3 regardless of the elements of :property:`dimension_types`.
```
rartino: I see issues with this point (even in the state before this PR):
- OPTiMaDe formally doesn't have an array type
- `lattice_vectors` isn't a 3x3 array, it is a list of three vectors, each of 3 elements.
- The requirement that this point adds is quite redundant with the first point.

If we think it is valuable (it was added to aid some confusion), I suggest the following:
```suggestion
  - MUST always contain three vectors of three coordinates each, independently of the elements of :property:`dimension_types`.
```
But I actually think the point can be dropped entirely, because the first point is clear enough on the 3x3 format being mandatory.

optimade.rst
```diff
@@ -1766,10 +1766,10 @@ lattice\_vectors
...
     The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero.
     The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
-  - All three elements of the inner lists of floats MAY be :val:`null` for non-periodic dimensions, i.e., those dimensions for which :property:`dimension\_types` is :val:`0`.
+  - In the case of non-periodic dimensions, i.e., those dimensions for which :property:`dimension_types` is :val:`0`, implementers MAY choose to replace the values with a list of :val:`null` values.
```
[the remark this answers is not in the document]
CasperWA: I'm fine with this.

optimade.rst
CasperWA: For clarification, I think it's good to keep it in. And indeed, we do not have an array type, but define a `list` instead.

CasperWA:
> I'm looking at our pair of definitions for `dimension_types` and `lattice_vectors` and now I'm confused...
> 
>     * The definition of `dimension_types` says that it is a list where each element is 1 if the structure is periodic along the direction of the corresponding vector in `lattice_vectors` and 0 if it is non-periodic along that direction.
> 
>     * The definition of `lattice_vectors` says that if `dimension_types` is 0, it may be `[null, null, null]`. But if so, what is the direction in which the structure is non-periodic?...
> 
> 
> If we want to give implementations the possibility of not giving concrete vectors for non-periodic directions, we need to revisit also the `dimension_types` definition.
> 
> I think I understand the underlying issue: the 3D space divides into a periodic and non-periodic subspace. It is sufficient to give explicit vectors to define the periodic subspace, everything orthogonal to that subspace is in the non-periodic subspace. And we don't care about which precise lattice vectors are used to span the non-periodic subspace since we use cartesian coordinates for the atoms.
> 
> But I'm not perfectly sure yet how to redefine `dimension_types` to express this in a way that is clear and makes sense.

I think this is clear enough considering the note in the description of `dimension_types`:
- Note: the elements in this list each refer to the direction of the corresponding entry in `lattice_vectors` and *not* the Cartesian x, y, z directions.

This means `lattice_vectors` defines the directions, while `dimension_types` defines the periodicity.

optimade.rst
```diff
@@ -1766,10 +1766,10 @@ lattice\_vectors
...
     (Therefore, the first index runs over the three lattice vectors and the second index runs over the x, y, z Cartesian coordinates).
   - For databases that do not define an absolute Cartesian system (e.g., only defining the length and angles between vectors), the first lattice vector SHOULD be set along *x* and the second on the *xy*-plane.
-  - This property MUST be an array of dimensions 3 times 3 regardless of the elements of :property:`dimension\_types`.
+  - MUST always contain three vectors of three coordinates each, independently of the elements of property `dimension\_types`_.
     The vectors SHOULD by convention be chosen so the determinant of the :property:`lattice_vectors` matrix is different from zero.
     The vectors in the non-periodic directions have no significance beyond fulfilling these requirements.
-  - All three elements of the inner lists of floats MAY be :val:`null` for non-periodic dimensions, i.e., those dimensions for which :property:`dimension\_types` is :val:`0`.
+  - The coordinates of the lattice vectors of non-periodic dimensions (i.e., those dimensions for which `dimension\_types`_ is :val:`0`) MAY be given as a list of :val:`null` values.
```
giovannipizzi: I think this means that even just one value can be `null`. From the issue I gather you want that if one is null, all coordinates of that vector are null? In that case, you might want to add an explicit sentence about this.

CasperWA: This was indeed the initial intent of this PR.
I'll attempt further clarification.

PR #263, opened 2020-03-06
Tag more filter examples as optional

### LINKED ISSUE #262: Tag optional filters in a parseable way
We've already gone to some effort to make sure examples are tagged correctly as `:filter:`, which is really helpful. My suggestion is to also be clearer when marking optional features of the filter language. Currently we are very clear in the text, but it would be really helpful when generating examples for validation to either tag optional filters as e.g. `:filter-optional`, or to just ensure that lines that contain optional filter examples are prefixed with "OPTIONAL:" in the text, as we do in many places.

As a concrete example, `HAS ONLY` is introduced [in the spec](https://github.com/Materials-Consortia/OPTiMaDe/blame/develop/optimade.rst#L1342) as:

```
The following construct MAY be supported:

- :filter:`list HAS ONLY values`: matches if all elements in :filter-fragment:`list` are equal to at least one :filter-fragment:`value`.
```

which could be reworded as:

```
The :filter-fragment:`HAS ONLY` construct MAY be supported:

- OPTIONAL: :filter:`list HAS ONLY values`: matches if all elements in :filter-fragment:`list` are equal to at least one :filter-fragment:`value`.
```

ml-evs:
Closes #262, if desired...

- [x] tag more bullet-pointed filter examples as OPTIONAL on the same line where it has already been indicated as optional in the text.
- [ ] ~re-tagged one `:filter:` as a `:filter-fragment:`~
- [x] fixed a spelling typo

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1624,7 +1624,7 @@ elements\_ratios
 
   - Note: Useful filters can be formulated using the set operator syntax for correlated values.
     However, since the values are floating point values, the use of equality comparisons is generally inadvisable.
-  - An OPTIONAL filter that matches structures where approximately 1/3 of the atoms in the structure are the element Al is: :filter:`elements:elements_ratios HAS ALL "Al":>0.3333, "Al":<0.3334`.
+  - OPTIONAL: a filter that matches structures where approximately 1/3 of the atoms in the structure are the element Al is: :filter:`elements:elements_ratios HAS ALL "Al":>0.3333, "Al":<0.3334`.
 
 chemical\_formula\_descriptive
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/optimade.rst
+++ b/optimade.rst
@@ -1387,7 +1387,7 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
 - :filter-fragment:`identifier1` references a list of dictionaries that contain as an identifier :filter-fragment:`identifier2` and the filter matches for a flat list containing only the contents of :filter-fragment:`identifier2` for every dictionary in the list.
   E.g., if :filter-fragment:`identifier1` is the list :filter-fragment:`[{"identifier2":42, "identifier3":36}, {"identifier2":96, "identifier3":66}]`, then :filter-fragment:`identifier1.identifier2` is understood in the filter as the list :filter-fragment:`[42, 96]`.
 
-The API implementation MAY allow this notation to generalize to arbitary depth.
+The API implementation MAY allow this notation to generalize to arbitrary depth.
 A nested property name that combines more than one list MUST, if accepted, be interpreted as a completely flattened list.
 
 Filtering on relationships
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1337,13 +1337,13 @@ The following constructs MUST be supported:
...
+The :filter-fragment:`HAS ONLY` construct MAY be supported:
 
-- :filter:`list HAS ONLY values`: matches if all elements in :filter-fragment:`list` are equal to at least one :filter-fragment:`value`.
+- OPTIONAL: :filter:`list HAS ONLY values`: matches if all elements in :filter-fragment:`list` are equal to at least one :filter-fragment:`value`.
   (If both :filter-fragment:`list` and :filter-fragment:`values` do not contain duplicate values, this implements the <= set operator.)
 
 This construct is OPTIONAL as it can be difficult to realize in some underlying database implementations.
-However, if the desired search is over a property that can only take on a finite set of values (e.g., chemical elements) a client can formulate an equivalent search by inverting the list of values into :filter-fragment:`inverse` and express the filter as :filter:`NOT list HAS inverse`.
+However, if the desired search is over a property that can only take on a finite set of values (e.g., chemical elements) a client can formulate an equivalent search by inverting the list of values into :filter-fragment:`inverse` and express the filter as :filter-fragment:`NOT list HAS inverse`.
```
rartino: Why `:filter:` -> `:filter-fragment:` here? Doesn't `NOT list HAS inverse` parse as a filter?

At least this was my intention when defining `:filter:` and `:filter-fragment:`: everything that is meant to parse as a valid filter should be marked `:filter:`, which means that it can (and should) be grammar-validated in our automatic tests, whereas `:filter-fragment` mark things that should be typeset the same way as `:filter:`, but cannot be validated as a full filter string.

optimade.rst
```diff
@@ -1624,7 +1624,7 @@ elements\_ratios
 
   - Note: Useful filters can be formulated using the set operator syntax for correlated values.
     However, since the values are floating point values, the use of equality comparisons is generally inadvisable.
-  - A filter that matches structures where approximately 1/3 of the atoms in the structure are the element Al is: :filter:`elements:elements_ratios HAS ALL "Al":>0.3333, "Al":<0.3334`.
+  - An OPTIONAL filter that matches structures where approximately 1/3 of the atoms in the structure are the element Al is: :filter:`elements:elements_ratios HAS ALL "Al":>0.3333, "Al":<0.3334`.
```
rartino: I think this minor rearrangement is more consistent with the above uses of OPTIONAL for these filter constructs that can optionally be supported:
```suggestion
  - OPTIONAL: a filter that matches structures where approximately 1/3 of the atoms in the structure are the element Al is: :filter:`elements:elements_ratios HAS ALL "Al":>0.3333, "Al":<0.3334`.
```

optimade.rst
ml-evs: > Why `:filter:` -> `:filter-fragment:` here? Doesn't `NOT list HAS inverse` parse as a filter?

I see now that I included this because I had replaced all the dummy `values` with some concrete examples, but I had not replaced `inverse` so this filter was erroring; I will undo this change.

optimade.rst
ml-evs: Agreed, I've just pushed this

ml-evs:
I haven't been able to spot any more filters that are optional, but are not tagged as such, so I think this is ready for review. We are now validating this list as of Materials-Consortia/optimade-python-tools#213.

ml-evs:
> Looks good.
> 
> Great work with the validator! It seems sensible to push for including it in the main repository to travis-validate changes to the specification, right?

To clarify, we are validating our default server implementation against all the mandatory `:filter:` examples in the spec (and hopefully soon will be doing the same with optional filters). I'm not sure we should go as far as validating the spec with the same tool, but we could discuss this in the next meeting!

PR #264, opened 2020-03-10
Custom endpoints should be under `/extensions`

CasperWA:
A minor clarification/change on how to implement custom endpoints to align with the sentiment under the `/extensions` endpoint section.

I didn't vacuum the spec to see if there are other places in the spec this should be updated/changed.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -661,7 +661,7 @@ Examples of valid entry listing endpoint URLs:
 
 There MAY be multiple entry listing endpoints, depending on how many types of entries an implementation provides.
 Specific standard entry types are specified in section `Entry list`_.
-The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be prefixed by a database-provider-specific prefix.
+The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be situated under the :endpoint:`extensions` endpoint.
 
 Entry Listing URL Query Parameters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/optimade.rst
+++ b/optimade.rst
@@ -662,7 +662,12 @@ Examples of valid entry listing endpoint URLs:
 There MAY be multiple entry listing endpoints, depending on how many types of entries an implementation provides.
 Specific standard entry types are specified in section `Entry list`_.
 
-The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be situated either under the :endpoint:`extensions` endpoint or be prefixed by a database-provider-specific prefix if situated at the root of the base URL.
+The API implementation MAY provide other entry types than the ones standardized in this specification.
+Such entry types MUST be prefixed by a database-provider-specific prefix (i.e., the resource objects' :property:`type` value should start with the database-provider-specific prefix, e.g., :property:`type` = :val:`_exmpl_workflows`).
+Each custom entry type SHOULD be served at a corresponding entry listing endpoint under the base URL with the same name (i.e., equal to the resource objects' :property:`type` value, e.g., :endpoint:`/_exmpl_workflows`).
+It is RECOMMENDED to align with the OPTIMADE API specification practice of using a plural for entry resource types and entry type endpoints.
+
+For more on custom endpoints, see `Custom Extension Endpoints`_.
 
 Entry Listing URL Query Parameters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/optimade.rst
+++ b/optimade.rst
@@ -661,7 +661,8 @@ Examples of valid entry listing endpoint URLs:
 
 There MAY be multiple entry listing endpoints, depending on how many types of entries an implementation provides.
 Specific standard entry types are specified in section `Entry list`_.
-The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be situated under the :endpoint:`extensions` endpoint.
+
+The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be situated either under the :endpoint:`extensions` endpoint or be prefixed by a database-provider-specific prefix if situated at the root of the base URL.
 
 Entry Listing URL Query Parameters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/optimade.rst
+++ b/optimade.rst
@@ -666,6 +666,7 @@ The API implementation MAY provide other entry types than the ones standardized 
 Such entry types MUST be prefixed by a database-provider-specific prefix (i.e., the resource objects' :property:`type` value should start with the database-provider-specific prefix, e.g., :property:`type` = :val:`_exmpl_workflows`).
 Each custom entry type SHOULD be served at a corresponding entry listing endpoint under the base URL with the same name (i.e., equal to the resource objects' :property:`type` value, e.g., :endpoint:`/_exmpl_workflows`).
 It is RECOMMENDED to align with the OPTIMADE API specification practice of using a plural for entry resource types and entry type endpoints.
+Any custom entry listing endpoint MUST also be added to the :property:`available\_endpoints` and :property:`entry\_types\_by\_format` attributes of the `Base Info Endpoint`_.
 
 For more on custom endpoints, see `Custom Extension Endpoints`_.
 


rartino:
Well..., my reading of the present specification is that it is *indented* that a database is allowed to make up its own entry types and serve them under, e.g. `/v1/_exmpl_compound`, rather than`/v1/extensions/_exmpl_compound`. According to this interpretation, `extensions` is for other extensions that are not merely database-specific entries. I like that extendability!

I dug through the history of this line in the specification, and it goes back to the initial version, so there is no originator we can ask.

Do you see a reason to disallow entry types on the form `/v1/_exmpl_compound`? The database-specific prefix should prevent any risk of collision with future standardized endpoints. And if you consider that a database may serve relationships between these entries and other entries, it just seems more elegant to allow them to live in the top URL path.

CasperWA:
> Well..., my reading of the present specification is that it is _indented_ that a database is allowed to make up its own entry types and serve them under, e.g. `/v1/_exmpl_compound`, rather than`/v1/extensions/_exmpl_compound`. According to this interpretation, `extensions` is for other extensions that are not merely database-specific entries. I like that extendability!

It would indeed seem so. And if this is the intention I suggest is it laid out better in the spec.
I just always understood it as the introduction of the `/extensions` endpoint was to put all custom endpoints under it, but if we explicitly want two kinds that's definitely also fine with me.

> I dug through the history of this line in the specification, and it goes back to the initial version, so there is no originator we can ask.

Sure - I also thought this was more a relic that should have been removed after the introduction of the `/extensions` endpoint.

> Do you see a reason to disallow entry types on the form `/v1/_exmpl_compound`? The database-specific prefix should prevent any risk of collision with future standardized endpoints. And if you consider that a database may serve relationships between these entries and other entries, it just seems more elegant to allow them to live in the top URL path.

I see no reason to disallow it, other than it should then be made more explicit throughout the spec.

rartino:
> Sure - I also thought this was more a relic that should have been removed after the introduction of the /extensions endpoint.

But I think it is not, I think both the `/extensions` endpoint and these custom entities were there from the start.

All things said, I prefer the simplicity of always being able to find a type of entry under the endpoint `v1/<entry name>`, rather than having to keep track of what is custom types. If you are convinced of this argument, perhaps you can change this PR to add the clarifications you think should be included for that to be clear?

merkys:
@rartino is right, `/extensions` are here since the beginning. Their purpose is somewhat vague to me, but they are definitely different from endpoints and properties: Supposedly a provider implements custom extension `/extensions/_cod_compare_structures` which, given two structure IDs, would calculate their similarity coefficient. This wouldn't be an endpoint, as it does not represent a logical instance. Whether it should be in OPTiMaDe is another question :)

CasperWA:
Okay, @merkys and @rartino have convinced me of the useability of both having `/extensions/<custom_endpoint>` and `/_exmpl_<custom_endpoint>`.
I'll try and translate their argument into spec-lines through my understanding of it all.

CasperWA:
@rartino and @merkys, I have updated this PR according to our discussion here. Please re-review.

rartino:
Hmm... should this be phrased to say that "entry types" are situated somewhere in the URL tree, or endpoints to access those entries? 

So, I find the overall formulation still a bit confusing. I guess I would go more for ~ "implementations can provide their own entry types, but they must be prefixed with their own database prefix, and endpoints to access them provided under the base URL, with that database prefix. Any other implementation-specific endpoints should be placed under extensions."

CasperWA:
> So, I find the overall formulation still a bit confusing. I guess I would go more for ~ "implementations can provide their own entry types, but they must be prefixed with their own database prefix, and endpoints to access them provided under the base URL, with that database prefix. Any other implementation-specific endpoints should be placed under extensions."

Do you mean the value of a custom entry resource's `type` should also have the provider-specific prefix?
Do you also mean that no matter where the endpoint may be, either under `/extensions` or on the root (alongside `/structures`, `/info`, etc.), it MUST always be prefixed with the provider-specific prefix?

optimade.rst
```diff
@@ -661,7 +661,8 @@ Examples of valid entry listing endpoint URLs:
 
 There MAY be multiple entry listing endpoints, depending on how many types of entries an implementation provides.
 Specific standard entry types are specified in section `Entry list`_.
-The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be prefixed by a database-provider-specific prefix.
+
+The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be situated either under the :endpoint:`extensions` endpoint or be prefixed by a database-provider-specific prefix if situated at the root of the base URL.
```
rartino, line 665:
```suggestion
The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be prefixed by a database-provider-specific prefix. Each such custom entry type SHOULD have a corresponding entry listing endpoint under the base URL with the same name (including the database-provider-specific prefix). Other custom endpoints, which are not entry listing endpoints for database-provider-specific entry types, should be situated under the :endpoint:`extensions`.
```

CasperWA: Right, so indeed you want the provider-specific-prefix as part of the entry resource `type` value.
That makes sense, since we're matching the endpoint names with the resource `type`s.

However, it seems you're also adding a restriction that one cannot serve entry listing endpoints under the `/extensions` endpoint?

rartino: I mean, not for anything you use as "normal" entries--why would you want to do that? But IMO we really could leave out the whole last sentence.

CasperWA: I wouldn't want to try and guess the reasons behind use cases, but instead simply not restrict unnecessarily.

In other words, I don't know :) But adding a restriction seems wrong.
Removing the sentence will leave some ambiguitiy, hence my initial PR.
But I think instead we can simply summarize by stating that _any_ custom endpoint, whatever it may hold, and no matter whether it's placed at the root or under `/extensions`, it should be prefixed with the provider-specific prefix? Or is that also too much of a restriction?

For entry endpoints, the `type`s of the served resources should also be prefixed, and otherwise follow the naming conventions laid out by the specification (not detailing this more, but hoping that developers read and understand the implications of the spec), e.g., endpoint name and resource types are the same.

rartino: Is it not clear from what I suggested to write that `type='...'` and endpoint name need to match?

We can never ensure all developers read and understand the specifications, but isn't that what validators is for?

So maybe change the last sentence to?: "Other custom endpoints should be situated under the :endpoint:`extensions`".

But I have to admit I no longer see what you are seeing here, so maybe I am missing something. There is a whole other section that explains that you can put whatever you like under `/extensions`. What is it that needs clarification about the `/extensions` endpoint here after we have carefully directed people to put their entry listing endpoints directly under the root?

CasperWA: > But I have to admit I no longer see what you are seeing here, so maybe I am missing something. There is a whole other section that explains that you can put whatever you like under `/extensions`. What is it that needs clarification about the `/extensions` endpoint here after we have carefully directed people to put their entry listing endpoints directly under the root?

That was my initial (mis)understanding of this. That custom endpoints could _only_ be situated under `/extensions`, since this is what is most explicitly written in the specification.
That one could add custom entry listing endpoints at the "root" was rather new to me.
I initially wanted to clarify my understanding of it, and it was argued by you and @merkys that it's fine to add "root" entry endpoints as well, as long as the database-provider-specific prefix is added.
So this PR now tries to clarify this extra place to place very specific custom endpoints.

optimade.rst
```diff
@@ -661,7 +661,13 @@ Examples of valid entry listing endpoint URLs:
 
 There MAY be multiple entry listing endpoints, depending on how many types of entries an implementation provides.
 Specific standard entry types are specified in section `Entry list`_.
-The API implementation MAY provide other entry types than the ones standardized in this specification, but such entry types MUST be prefixed by a database-provider-specific prefix.
+
+The API implementation MAY provide other entry types than the ones standardized in this specification.
+Such entry types MUST be prefixed by a database-provider-specific prefix (i.e., the resource objects' :property:`type` value should start with the database-provider-specific prefix, e.g., :property:`type` = :val:`_exmpl_workflows`).
+Each custom entry type SHOULD be served at a corresponding entry listing endpoint under the base URL with the same name (i.e., equal to the resource objects' :property:`type` value, e.g., :endpoint:`/_exmpl_workflows`).
+It is RECOMMENDED to align with the OPTIMADE API specification practice of using a plural for entry resource types and entry type endpoints.
```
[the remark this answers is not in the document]
CasperWA: This is a good point. I think this is sort of already in the spec, but it's never expressed explicitly. And definitely never with a MUST requirement.

Do you mind if we create a new PR for this additional explicit requirement and merge this, or do you feel it is important enough to get it in here as well now? If @giovannipizzi or @rartino is ready to re-approve right away with this addition, it's fine for me to keep add it here.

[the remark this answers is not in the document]
CasperWA: I take @rartino's thumbs up as a "go" for the addition - 2 sec.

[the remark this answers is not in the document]
rartino: Well, it was meant as "I'm here". But if @giovannipizzi isn't, we won't approve this tonight. On the other hand, we have other release-blocking PRs still waiting, so it is hardly critical.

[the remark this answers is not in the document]
CasperWA: So close.. oh well. Maybe I should add something saying like "add entry listing info endpoints for custom entry listing endpoints" as well? Although I feel were starting to reiterate the obvious, but I don't know?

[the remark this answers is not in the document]
rartino: Go ahead and merge while you can :D.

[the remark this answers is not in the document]
ml-evs: I agree that its obvious, my thinking was that a provider probably won't start adding custom endpoints until they have a working server, so by the time they come back and look at how to do that, they've forgotten all about the rest of the spec.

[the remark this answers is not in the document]
CasperWA: Maybe for another PR if relevant.

merkys:
I missed my chance to comment/approve this PR, but I fully agree with the changes.

optimade.rst
```diff
@@ -661,7 +661,14 @@ Examples of valid entry listing endpoint URLs:
...
+Any custom entry listing endpoint MUST also be added to the :property:`available\_endpoints` and :property:`entry\_types\_by\_format` attributes of the `Base Info Endpoint`_.
```
merkys, line 669: ... unless it is a hidden entry listing endpoint :)

CasperWA: Sure, but what would be the point of this? Then maybe just don't publish the data in that endpoint, or put it behind authentication?

CasperWA:
> I missed my chance to comment/approve this PR, but I fully agree with the changes.

Yeah, sorry. I was trying to get it in so that rc.2 can be released ASAP.
Thanks for commenting anyway, and really happy you agree with the changes :)

optimade.rst
merkys: It's just a hypothetical example of the simplest security-through-obscurity solution. I'm not suggesting the specs should do anything about it, it's just that not listing a custom entry listing endpoint in those attributes effectively renders it hidden.

CasperWA: Sure. However, I think this is bad practice (security-through-obscurity). And the spec "agrees" with the wording here and usage of `MUST`.
But you're saying you want to have this practice be available for providers?

merkys: Of course this is a bad practice and I'm not advocating for that – I meant this merely as a joke.

CasperWA: Ah, sorry for the misunderstanding 😅

PR #266, opened 2020-04-29
Handle grammatiker build chain dependency as git submodule

rartino:
This PR attempts to demonstrate one suggestion of how we could improve the present handling of the dependency our build chain has on the grammatiker software. This addresses a problem brought up in #195 (where this solution was also discussed). **Note: there is yet not a consensus on the best way to do this.** I prepared this PR to demonstrate how (un)complicated this solution is in practice and it is meant to be rejected if we decide for another way.

*The present way the dependency on grammatiker is handled:*

The makefile, during build-time, uses svn to download a specific revision (129) of grammatiker from an svn server via a URL specified in the makefile using the svn:// protocol (host: saulius-grazulis.lt). The download is presently not checked for integrity (but this could of course be added in some way).

*With this PR:*

Our github Materials-Consortia/grammatiker repo becomes a git submodule under tests/tools/grammatiker. The binding is to a specific commit in that git repo (presently the one corresponding to SVN revision 129). 

The handling of the submodule with git is mostly transparent for end-users as it is automatically initialized and updated by the makefile when the dependency is needed.

There is also a makefile target: `make grammatiker_svn_pull` that fetches new revisions from the upstream svn server and places them as git commits in the local master branch of the submodule repo. The local repo can then be used to update the Materials-Consortia/grammatiker github repo.

Some properties I like with this solution:

- The way git ties the submodule to a hash guarantees that what is downloaded is the same thing as can be browsed in the github Materials-Consortia/grammatiker repo, where one can look at the changes as individual commits.
 
- When we want to upgrade the revision of grammatiker to use, this will be done via a PR against OPTIMADE for the submodule to point at a new commit. It is then easy to review the changes between revisions in the Materials-Consortia/grammatiker github repo using the same UI we use to reivew the usual OPTIMADE PRs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tests/tools/grammatiker"]
+	path = tests/tools/grammatiker
+	url = https://github.com/Materials-Consortia/grammatiker
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -66,6 +66,29 @@
 # - test_authors: runs tests on the AUTHORS file (checks order)
 #
 #
+# Targets for build chain dependencies
+######################################
+#
+# - grammatiker_svn_update: connects the grammatiker submodule
+#   in tests/tools/grammatiker to the main svn repository
+#   and pulls the latest revisions via svn.
+#
+#   Upon running the above, one can update the git repository
+#   that hosts the submodule:
+#     cd tests/tools/grammatiker
+#     git remote add gitrepo git@github.com:Materials-Consortia/grammatiker.git
+#     git checkout -b new gitrepo/master
+#     git cherry-pick <ranges of hashes not yet in submodule repo>
+#     git push gitrepo master
+#
+#   The grammatiker commit used by OPTIMADE can be updated using:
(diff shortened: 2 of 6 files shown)


merkys:
Slightly related: 'grammatica' is already present in APT repositories (as `libgrammatica-java`) for Ubuntu 20.04 a.k.a. Focal Fossa. Thus if we bump the Dockerfile from `xenial` to `focal`, we may loose the build chain for 'grammatica', also saving some wallclock time. However, 'grammatiker' needs some patching to use the system-provided JAR for 'grammatica', thus not sure if it is worth the effort.

ml-evs:
Hi @rartino, @merkys and I no longer have a clear picture of whether this is needed, which perhaps we can discuss in person! :upside_down_face:

rartino:
This was a proposal to fix something in the "build" that apparently no one else seems concerned about - and the encapsulation of the build into docker addresses it to some degree for myself - I think I will close the PR for now just to clean up our PR list. We can get back to this at a later point.

PR #270, opened 2020-05-19
Add OPTIONAL key `type` to "schema" property

### LINKED ISSUE #269: Allow to specify `type` in `/info/<entry-endpoint>`
To be able to dynamically and effectively handle (especially) provider-specific resource properties, e.g., `_exmpl_some_energy`, it would be useful to know the value type expected.

It would make most sense to add this under the entry-endpoint's specific informational endpoint, e.g., `/info/structures` for the `/structures` entry-endpoint.
It could then be a key (`type`) alongside `description`, `unit`, and `sortable`.
It should be OPTIONAL.

I don't think many would be against adding this, the only tricky thing I see, which could spark a bit of discussion, is the allowed/recognized values.

I would propose to use JSON-typed values, since our spec leans heavily on the JSON API v1.0 already. However, I wouldn't mind custom types either, as long as they have clear and unique links to major programming language types. In this, the JSON types may not even be suitable, and we should instead look towards OpenAPI's extended JSON types to better distinguish between an `array`, an `array` with unique entries, an ordered `array`, etc.

CasperWA:
Closes #269 

This PR first cosmetically updates the list of available keys to describe a property under `/info/<entry-endpoint>`, by splitting it more visually up into the two lists of REQUIRED keys and OPTIONAL keys.
Note, only `description` is REQUIRED.

Then I have added `type` according to #269.
**Please keep any discussion relevant to what the types should be in the issue: #269**.
I will update this PR according to the discussion in #269.

See new section in GitHub formatted RST [here](https://github.com/CasperWA/OPTiMaDe/blob/close_269_add_type/optimade.rst#entry-listing-info-endpoints).

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1004,7 +1004,67 @@ The response for these endpoints MUST include the following information in the :
 
   - :field:`unit`: String. The physical unit symbol in which the property's value is given.
   - :field:`sortable`: Boolean. Whether the property can be used for sorting (see `Entry Listing URL Query Parameters`_ for more information on this field).
-  - :field:`type`: String. The type of the property's value. This can be any of the types defined in `Data types`_: :val:`string`, :val:`integer`, :val:`float`, :val:`boolean`, :val:`timestamp`, :val:`list`, :val:`dictionary`, and :val:`unknown`.
+  - :field:`type`: Dictionary or string.
+    The type of the property's value.
+    This can be any of the types defined in `Data types`_
+
+    :field:`type` SHOULD be given as a string if the property's type is either: :val:`string`, :val:`integer`, :val:`float`, :val:`boolean`, :val:`timestamp`, or :val:`unknown`.
+
+    If the property's type is either: :val:`list` or :val:`dictionary`, i.e., a collection type, the value of :field:`type` SHOULD be given as a dictionary with the keys:
+
+    - :field:`type`: String. Either :val:`list` or :val:`dictionary`.
+    - :field:`content`: Dictionary or string. Similar to the :field:`type` one level up.
+
+    :field:`content` reflects the content of the collection type.
+    This can again be a collection type, which will result in the value of :field:`content` being another dictionary with the keys :field:`type` and :field:`content`.
+
+    While there is no limit to the number of layers of collection types in this specification, a provider should take care to consider the practical limitations of clients and the practical usability and transparency of heavily layered properties.
+
+    If the :field:`type` is :val:`dictionary`, then :field:`content` MUST be a dictionary, where the keys are the keys of the property's dictionary value, and the values are again either a string with the OPTIMADE data type of the dictionary key's value type or a dictionary with the keys :field:`type` and :field:`content` if the dictionary key's value type is a collection type. The values for :field:`type` and :field:`content` follow the rules outlined above.
+
+    Example with the properties :property:`_exmpl_list_prop` with the human-readable type of "list of list of floats or unknown values", :property:`_exmpl_dict_prop` with the human-readable type of "list of dictionary with keys: :field:`name`: string; :field:`occupation`: list of float", and :property:`_exmpl_entry_version` with the human-readable type of "int":
+
+    .. code:: jsonc
+
+        {
+          "data": {
+            // ...
+            "properties": {
+              "_exmpl_entry_version": {
+                // ...
+                "type": "integer"
+              },
+              "_exmpl_list_prop": {
+                // ...
+                "type": {
+                  "type": "list",
+                  "content": {
+                    "type": "list",
+                    "content": "float"
+                  }
+                }
+              },
+              "_exmpl_dict_prop": {
+                // ...
+                "type": {
+                  "type": "list",
+                  "content": {
+                    "type": "dictionary",
+                    "content": {
+                      "name": "string",
+                      "occupation": {
+                        "type": "list",
+                        "content": "float"
+                      }
+                    }
+                  }
+                }
+              }
+              // ... <more properties>
+            }
+          }
+          // ...
+        }
 
 - **formats**: List of output formats available for this type of entry.
 - **output\_fields\_by\_format**: Dictionary of available output fields for this entry type, where the keys are the values of the :field:`formats` list and the values are the keys of the :field:`properties` dictionary.
@@ -1019,12 +1079,20 @@ Example:
         "properties": {
           "nelements": {
             "description": "Number of elements",
-            "sortable": true
+            "sortable": true,
+            "type": "integer"
           },
           "lattice_vectors": {
             "description": "Unit cell lattice vectors",
             "unit": "Å",
-            "sortable": false
+            "sortable": false,
+            "type": {
+              "type": "list",
+              "content": {
+                "type": "list",
+                "content": "float"
+              }
+            }
           }
           // ... <other property descriptions>
         },
--- a/optimade.rst
+++ b/optimade.rst
@@ -428,13 +428,13 @@ Every response SHOULD contain the following fields, and MUST contain at least on
 
     .. code:: jsonc
 
-       {
-	 "id": "dep_chemical_formula_01",
-	 "type": "warning",
-	 "code": "_exmpl_dep_chemical_formula",
-	 "title": "Deprecation Warning",
-	 "detail": "chemical_formula is deprecated, use instead chemical_formula_hill"
-       }
+        {
+          "id": "dep_chemical_formula_01",
+          "type": "warning",
+          "code": "_exmpl_dep_chemical_formula",
+          "title": "Deprecation Warning",
+          "detail": "chemical_formula is deprecated, use instead chemical_formula_hill"
+        }
 
     **Note**: warning :field:`id`\ s MUST NOT be trusted to identify the exceptional situations (i.e., they are not error codes, use instead the field :field:`code` for this.
     Warning :field:`id`\ s can *only* be trusted to be unique in the list of warning resource objects, i.e., together with the :field:`type`.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -994,7 +994,78 @@ The response for these endpoints MUST include the following information in the :
...
+
+  *OPTIONAL keys*:
+
+  - :field:`unit`: String. The physical unit symbol in which the property's value is given.
+  - :field:`sortable`: Boolean. Whether the property can be used for sorting (see `Entry Listing URL Query Parameters`_ for more information on this field).
+  - :field:`type`: Dictionary or string.
+    The type of the property's value.
+    This can be any of the types defined in `Data types`_
+
+    :field:`type` SHOULD be given as a string, if the property's type is either: :val:`string`, :val:`integer`, :val:`float`, :val:`boolean`, :val:`timestamp`, or :val:`unknown`.
```
ml-evs, line 1011:
```suggestion
    :field:`type` SHOULD be given as a string, if the property's type is either: :val:`string`, :val:`integer`, :val:`float`, :val:`boolean`, :val:`timestamp`, or :val:`unknown`, i.e. a primitive type.
```

I think primitive type is a general nomenclature? Just so this matches the collection type section.

optimade.rst
```diff
@@ -994,7 +994,78 @@ The response for these endpoints MUST include the following information in the :
...
+
+    If the property's type is either: :val:`list` or :val:`dictionary`, i.e., a collection type, the value of :field:`type` SHOULD be given as a dictionary with the keys:
+
+    - :field:`type`: String. Either :val:`list` or :val:`dictionary`.
+    - :field:`content`: Dictionary or string. Similar to the :field:`type` one level up.
+
+    :field:`content` reflects the content of the collection type.
+    This can again be a collection type, which will result in the value of :field:`content` being another dictionary with the keys :field:`type` and :field:`content`.
+
+    While there is no limit to the number of layers of collection types in this specification, a provider should take care to consider the practical limitations of clients and the practical usability and transparency of heavily layered properties.
```
ml-evs, line 1021:
```suggestion
    While there is no limit to the number of layers of collection types in this specification, it is RECOMMENDED that a provider takes care to consider the practical limitations of clients and the practical usability and transparency of heavily layered properties.
```

Potentially upgrade this to RECOMMENDED?

optimade.rst
```diff
@@ -994,7 +994,78 @@ The response for these endpoints MUST include the following information in the :
...
-  Each value is a dictionary, with the REQUIRED key :field:`description` and OPTIONAL keys :field:`unit` and :field:`sortable` (see `Entry Listing URL Query Parameters`_ for more information on :field:`sortable`).
+  Each value is a dictionary, with the
+
+  *REQUIRED keys*:
+
+  - :field:`description`: String. Description of the property.
+
+  *OPTIONAL keys*:
+
+  - :field:`unit`: String. The physical unit symbol in which the property's value is given.
```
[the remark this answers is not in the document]
CasperWA: Nice addition.
I think we should consider underlying if a particular set of units (e.g., SI) is preferred as well, e.g., using J (joules) instead of cal (calories) or something weird? Even though one could totally use calories anyway, I guess. A sentence about it would just be nice?

optimade.rst
CasperWA: If that's so, we should add it to the section about data types as well.

CasperWA:
> (... <code examples> ...)
> Thoughts? If we don't implement something like this, I imagine I'll just flatten my entire API, which may be preferable anyway.

If you flatten it, you will lose the containerized information, unless you encode this information in the property name, I guess? E.g., `species.name`.
But then there will be redundant type information ...

Considering the other alternative (putting the "contained" properties' description under the "top-level" `type`), I am not liking the mixing of this.
I'd rather then "rip out" the "contained" properties' description and put it in another "top-level" value, e.g., under `properties` again or something.

Example of this (using the example of @ml-evs above):

```json
{
    "properties": {
        "_odbx_dft_parameters": {
            "description": "container for DFT parameters",
            "properties": {
                "cutoff_energy": {
                    "description": "Plane wave cutoff",
                    "unit": "eV"
                },
                "kpoint_spacing": {
                    "description": "Monkhorst-Pack grid spacing",
                    "unit": "eV/A"
                }
            },
            "type": {
                "type": "dictionary",
                "content": {
                    "cutoff_energy": "float",
                    "kpoint_spacing": "float"
                }
            }
        }
    }
}
```

If there are further dictionaries within, the descriptions+units would go a layer further down into a new `properties` and the `type` would still be kept under the "top-level" `type` key.

The reasoning behind keeping `type` separate, is to reserve the sub-`properties` key only for `dictionary` properties and not have to go several places to retrieve type information. Further sub-descriptions (and units) will also only be relevant for a "top-level" `dictionary` type.

---

So, using this as a first idea, let's see what happens for, e.g., `species`.
First, adding `"type"`.

```json
{
    "properties": {
        "species": {
            "description": "Species of the sites of this structure",
            "type": {
                "type": "list",
                "content": {
                    "type": "dictionary",
                    "content": {
                        "name": "string",
                        "chemical_symbols": {
                            "type": "list",
                            "content": "string"
                        },
                        "concentration": {
                            "type": "list",
                            "content": "float"
                        },
                        "mass": "float",
                        "original_name": "string"
                    }
                }
            }
        }
    }
}
```

So how do we now add descriptions for the contained dictionary within the list?
I think for this, we will have to flatten and simply do:

```json
{
    "properties": {
        "species": {
            "description": "Species of the sites of this structure.",
            "type": {
                "type": "list",
                "content": {
                    "type": "dictionary",
                    "content": {
                        "name": "string",
                        "chemical_symbols": {
                            "type": "list",
                            "content": "string"
                        },
                        "concentration": {
                            "type": "list",
                            "content": "float"
                        },
                        "mass": "float",
                        "original_name": "string"
                    }
                }
            },
            "properties": {
                "name": {
                    "description": "The name of the species, it is unique in the list of species."
                },
                "chemical_symbols": {
                    "description": "Chemical elements composing this species. Can be either a valid chemical element, the value \"X\", representing a non-chemical element or \"vacancy\", representing a non-zero probability of having a vacancy."
                },
                "concentration": {
                    "description": "The relative concentration of the corresponding chemical symbol in this species."
                },
                "mass": {
                    "description": "The species' mass.",
                    "unit": "a.m.u."
                },
                "original_name": {
                    "description": "Name of the species used in the Example database."
                }
            }
        }
    }
}
```

However, this will inherently disallow sub-dictionaries to have the same keys, but with different meaning.
To mitigate this, the `properties` dictionary keys, could simply define the exact location of the key in question, i.e., all of the keys would get `species.` prepended. Even though `species` is a list, this should still uniquely define the key, under the requirements that a sub-container (in this case the dictionary) MUST not change the meaning of its keys within the same "level", i.e., `"name"` is not allowed to mean something different between two dictionaries in the `species` list. While this is clear and implicitly understood in this example, it may not always be as clear for other situations, I think?

---

So what do you think about this? It's a bit of a "hybrid" between the suggestion and ideas by @ml-evs, but more clean, in terms of keeping the "top-level" `type`'s value clear of any other information that `type` information.

CasperWA:
Another idea would be to completely flatten it and not have the nested `type` values. While the property keys would then reflect its "level".

Then it would become:

```json
{
    "properties": {
        "species": {
            "description": "Species of the sites of this structure.",
            "type": {
                "type": "list",
                "content": "dictionary"
            }
        },
        "species.name": {
            "description": "The name of the species, it is unique in the list of species.",
            "type": "string"
        },
        "species.chemical_symbols": {
            "description": "Chemical elements composing this species. Can be either a valid chemical element, the value \"X\", representing a non-chemical element or \"vacancy\", representing a non-zero probability of having a vacancy.",
            "type": {
                "type": "list",
                "content": "string"
            }
        },
        "species.concentration": {
            "description": "The relative concentration of the corresponding chemical symbol in this species.",
            "type": {
                "type": "list",
                "content": "float"
            }
        },
        "species.mass": {
            "description": "The species' mass.",
            "unit": "a.m.u.",
            "type": "float"
        },
        "species.original_name": {
            "description": "Name of the species used in the Example database.",
            "type": "string"
        }
    }
}
```

Then _only_ the type `list` would need to be specified using an "object" value for `type` (i.e., `{"type": {"type": "list", "content": "<whatever type here>"}}`), and the `dictionary` type should then be "fleshed out" using the system above, with sub-named properties.

optimade.rst
CasperWA: They are called "basic types" in the `Data types` section. I will use this term instead of primitive.

optimade.rst
```diff
@@ -994,7 +994,78 @@ The response for these endpoints MUST include the following information in the :
 
 - **description**: Description of the entry.
 - **properties**: A dictionary describing queryable properties for this entry type, where each key is a property name.
-  Each value is a dictionary, with the REQUIRED key :field:`description` and OPTIONAL keys :field:`unit` and :field:`sortable` (see `Entry Listing URL Query Parameters`_ for more information on :field:`sortable`).
+  Each value is a dictionary, with the
+
+  *REQUIRED keys*:
+
+  - :field:`description`: String. Description of the property.
```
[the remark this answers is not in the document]
CasperWA: Concerning the centralised set of descriptions, I don't think it belongs in this repo? But we could consider adding a sort of collection of suggested default values somewhere none the less? Maybe also in this repo. This would include these `/info/<entry-endpoint>` property descriptions, but maybe also some of the suggested standards for some `meta` response values? I don't know what this would look like. Maybe create an issue for this?

ml-evs:
I agree that describing nested properties under `type` is not desirable. I think I prefer the nested version rather than flat (what happens if those nested properties are dicts too?)

I think the limitation you suggest on the keys of list of dicts is a natural one, though it does complicate the explanation slightly...

CasperWA:
> I agree that describing nested properties under `type` is not desirable. I think I prefer the nested version rather than flat (what happens if those nested properties are dicts too?)

In my head the properties would just respect the nesting, e.g., `top.first.second`.
In the end, since we don't mix types, there can be as many in-between lists as we'd want, so `top`, `first`, and `second` need not be "neighbours", but could be on very different nested levels, but with the requirement that the levels always be more and more nested.

> I think the limitation you suggest on the keys of list of dicts is a natural one, though it does complicate the explanation slightly...

Yeah. I guess it's all about writing "unique" a lot? 😅

fekad:
Defining the properties as a nested structure and and storing as a list instead of a dictionary:
```
{
  "properties": [
    {
      "name": "species",
      "description": "Species of the sites of this structure.",
      "type": "dict", // element of the list are dictionaries
      "shape": "[null]", // but it is a list OR ["nelements"] would be even nicer
      "content":[
        {
          "name": "name",
          "description": "The name of the species, it is unique in the list of species.",
          "type": "string"
        },
        {
          "name": "chemical_symbols",
          "description": "Chemical elements composing this species. Can be either a valid chemical element, the value \"X\", representing a non-chemical element or \"vacancy\", representing a non-zero probability of having a vacancy.",
          "type": "string",
          "shape": "[null]" //because it is a list
        },
        {
          "name": "concentration",
          "description": "The relative concentration of the corresponding chemical symbol in this species.",
          "type": "float",
          "shape": "[null]"
        },
        {
          "name": "mass",
          "description": "The species' mass.",
          "unit": "a.m.u.",
          "type": "float"
        },
        {
          "name": "original_name",
          "description": "Name of the species used in the Example database.",
          "type": "string"
        }
      ]
    },
    {
      "name": "_odbx_dft_parameters",
      "description": "container for DFT parameters",
      "type": "dict", // no shape because it is a single dictionary
      "content": [
        { 
          "name": "cutoff_energy",
          "description": "Plane wave cutoff",
          "type":  "float",
          "unit": "eV"
        },
        { 
          "name": "kpoint_spacing",
          "description": "Monkhorst-Pack grid spacing",
          "type":  "float",
          "unit": "eV/A"
        }
      ],
    },
  ]
}
```

rartino:
According to discussions at OPTIMADE workshop 2020 there will be a simplified version of this PR for v1.0, and this will be marked for v1.1.

rartino:
Note: at the end of the 2021 workshop a format for the type specification was decided and there will be a PR forthcoming on that. We can keep this PR open until that PR is posted, but it probably should not be merged.

CasperWA:
> Note: at the end of the 2021 workshop a format for the type specification was decided and there will be a PR forthcoming on that. We can keep this PR open until that PR is posted, but it probably should not be merged.

Indeed, the new PR should supersede this one.

ml-evs:
Superseded by #282 and #376

PR #273, opened 2020-06-11
Extend (and change) the definition of links endpoints

### LINKED ISSUE #22: Discovery mechanism

### LINKED ISSUE #217: Quirks of the /links endpoint
I have found some quirks in the spec concerning the `/links` endpoint and the Index meta-database, while in the process of adding these to the test server in [`optimade-python-tools`](https://github.com/Materials-Consortia/optimade-python-tools).

Quirky things the specification states about `/links` (either ex- or implicitly):
- It is _not_ considered an entry listing endpoint, but still has a list of resource objects.
- The resource objects' types are not plural (like all other resource objects).
- Only allowed query parameters are those equal to a single entry endpoint. This means that there is no filtering in the `/links` endpoint, nor is there a way to handle pagination. (This actually makes for some weird work-arounds in the implementation).
- Should the providers from `providers.json` be supplied here for all OPTiMaDe implementations/databases? Or do we perhaps recommend to only show them in the Index meta-database?

For the Index meta-database:
- Under `/info` one MAY provide a default database using `relationships`. However, this is done under the non-existing `default` relationship-type:

  ```json
  // ...
  "relationships": {
    "default": {
      "data": {"type": "child", "id": "test_server"}
    }
  }
  // ...
  ```

  This should probably be renamed to reflect the endpoint where this related resource may be found, i.e., `links`, or it should be renamed to `child` to reflect the related type. I am not completely sure which is the most appropriate according to JSON API. However, I would suggest `links`.

giovannipizzi:
With this PR:
- the `type` of links is always set to `links`
- we define four `link_type`s for internal links (`child` and `root`), and allow for `external` links, as well to links to provider lists via `providers` link types

(see discussion in [this comment](https://github.com/Materials-Consortia/OPTIMADE/issues/217#issuecomment-642084755) of #217)

This PR fixes #217 
This also fixes #22 providing a discovery mechanism via `external` link types

Co-Authored-By: "Casper W. Andersen <casper.andersen@epfl.ch>"
(@CasperWA)

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1185,7 +1185,7 @@ In practice, this forms a tree structure for the OPTIMADE implementations of a p
 List of Providers Links
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Resource objects with :property:`link_type`=:val:`providers` links to an `Index Meta-Database`_ that supplies a list of OPTIMADE database providers.
+Resource objects with :property:`link_type`=:val:`providers` links MUST point to an `Index Meta-Database`_ that supplies a list of OPTIMADE database providers.
 The intention is to be able to auto-discover all providers of OPTIMADE implementations.
 
 A list of known providers can be retrieved as described in section `Database-Provider-Specific Namespace Prefixes`_.
--- a/optimade.rst
+++ b/optimade.rst
@@ -910,7 +910,7 @@ If this is an index meta-database base URL (see section `Index Meta-Database`_),
     - **data**: `JSON API resource linkage <http://jsonapi.org/format/1.0/#document-links>`__.
       It MUST be either :field-val:`null` or contain a single child identifier object with the fields:
 
-      - **type**: :field-val:`child`
+      - **type**: :field-val:`links`
       - **id**: ID of the provider's chosen default OPTIMADE API database.
         MUST be equal to a valid child object's :field:`id` under the :field:`links` endpoint.
 
@@ -992,7 +992,7 @@ Example for an index meta-database:
 	 },
 	 "relationships": {
 	   "default": {
-	     "data": { "type": "child", "id": "perovskites" }
+	     "data": { "type": "link", "id": "perovskites" }
 	   }
 	 }
        }
@@ -1046,10 +1046,23 @@ Example:
 Links Endpoint
 --------------
 
-This endpoint exposes information on other OPTIMADE API implementations that are linked to the current implementation.
+This endpoint exposes information on other OPTIMADE API implementations that are in some way related to the current implementation.
 The links endpoint MUST be provided under the versioned base URL at :endpoint:`/links`.
 
-It can be considered an introspective endpoint, similar to the Info endpoint, but at a higher level: that is, Info endpoints provide information on the given implementation, while the Links endpoint provides information on the links between immediately related implementations (in particular, an array of none or a single :object:`parent` object and none or more child-type objects, see section `Parent and Child Objects`_).
+Link Types
+~~~~~~~~~~
+
+Each link has a :property:`link_type` attribute that specifies the type of the linked relation.
+
+The :property:`link_type` MUST be one of the following values:
+
+- :val:`child`: a link to another OPTIMADE implementation that MUST be within the same provider. This allows to create a tree-like structure of databases, by pointing to children sub-databases.
+- :val:`root`: a link to the root implementation, within the same provider. This MAY be an `Index Meta-Database`_. There MUST be only one root implementation per provider, and any child (or child of a child, at any depth) MUST have a link to the root implementation.
+- :val:`external`: a link to an external OPTIMADE implementation. This MAY be used to point to any other implementation, also in a different provider.
+- :val:`providers`: a link to a `List of Providers`_ implementation that includes the current implementation, e.g. `https://providers.optimade.org/ <https://providers.optimade.org/>`__. 
+
+
+Thanks to the :val:`root` and :val:`child` link types, links can be used as an introspective endpoint, similar to the Info endpoint, but at a higher level: that is, Info endpoints provide information on the given implementation, while the Links endpoint provides information on the links between immediately related implementations (in particular, an array of none or a single object with link type :val:`root` and none or more objects with link type :val:`child`, see section `Child Links`_).
 
 For Links endpoints, the API implementation MAY ignore any provided query parameters.
 Alternatively, it MAY handle the parameters specified in section `Single Entry URL Query Parameters`_ for single entry endpoints.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1046,21 +1047,36 @@ Example:
...
+Each link has a :property:`link_type` attribute that specifies the type of the linked relation.
+
+The :property:`link_type` MUST be one of the following values:
 
-For Links endpoints, the API implementation MAY ignore any provided query parameters.
+- :field-val:`child`: a link to another OPTIMADE implementation that MUST be within the same provider.
+  This allows the creation of a tree-like structure of databases by pointing to children sub-databases.
+- :field-val:`root`: a link to the root implementation within the same provider.
+  This is RECOMMENDED to be an `Index Meta-Database`_.
+  There MUST be only one :val:`root` implementation per provider and any :val:`child` (or :val:`child` of a :val:`child`, at any depth) MUST have a link to the :val:`root` implementation.
```
rartino, line 1064: Do I read this correctly that the root itself is formally exempt from having a root link? I think it is an odd quirk to not just say that the 'root' link is always mandatory (and thus the root links to itself). So:
```suggestion
  There MUST be only one :val:`root` implementation per provider and all implementations MUST have a link to this :val:`root` implementation. If the provider only supply a single implementation, the :val:`root` link links to the implementation itself.
```

optimade.rst
```diff
@@ -1076,79 +1092,99 @@ The resource objects' response dictionaries MUST include the following fields:
...
+Internal Links: Root and Child Links
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Resource objects that MAY be present under the Links endpoint.
+Any number of resource objects with :property:`link_type`=:val:`child` MAY be present as part of the :field:`data` list.
+A :val:`child` object represents a "link" to an OPTIMADE implementation within the same provider exactly one layer **below** the current implementation's layer.
 
-Either none or a single :object:`parent` object MAY be present as part of the :field:`data` array.
-The :object:`parent` object represents a "link" to the OPTIMADE implementation exactly one layer **above** the current implementation's layer.
+Either none or a single resource object with :property:`link_type`=:val:`root` MAY be present as part of the :field:`data` list.
```
rartino, line 1175:
```suggestion
Exactly one resource object with :property:`link_type`=:val:`root` MUST be present as part of the :field:`data` list.
```
So, see above, I think exactly one root link should always be present. That is easier to validate, and a client can always rely on finding the root from any starting database by following the root link once.

optimade.rst
```diff
@@ -1076,79 +1092,99 @@ The resource objects' response dictionaries MUST include the following fields:
...
+In practice, this forms a tree structure for the OPTIMADE implementations of a provider. 
+**Note**: The RECOMMENDED number of layers is two.
 
-Provider Objects
-~~~~~~~~~~~~~~~~
+List of Providers Links
+~~~~~~~~~~~~~~~~~~~~~~~
 
-The :object:`provider` objects are meant to indicate links to an "Index meta-database" hosted by database providers.
+Resource objects with :property:`link_type`=:val:`providers` are meant to indicate links to an `Index Meta-Database`_ hosted by organisations that want to point to various OPTIMADE database providers.
```
rartino, line 1187: Am I not allowed to link to an index meta-database of providers unless it is maintained by an organization?:
```suggestion
Resource objects with :property:`link_type`=:val:`providers` links to an `Index Meta-Database`_ that supplies a list of OPTIMADE database providers.
```
Also, should there be a MUST somewhere here?

optimade.rst
CasperWA: I see this is an unintended meaning of this sentence. The `root` implementation can of course have a `root` links resource (inevitably pointing to itself).

optimade.rst
CasperWA: It's a bit more stringent, but fine by my accounts. I would like more thumbs up than just my own on this though.

optimade.rst
```diff
@@ -1076,79 +1092,99 @@ The resource objects' response dictionaries MUST include the following fields:
...
+In practice, this forms a tree structure for the OPTIMADE implementations of a provider. 
+**Note**: The RECOMMENDED number of layers is two.
 
-Provider Objects
-~~~~~~~~~~~~~~~~
+List of Providers Links
+~~~~~~~~~~~~~~~~~~~~~~~
 
-The :object:`provider` objects are meant to indicate links to an "Index meta-database" hosted by database providers.
+Resource objects with :property:`link_type`=:val:`providers` links to an `Index Meta-Database`_ that supplies a list of OPTIMADE database providers.
```
CasperWA, line 1187: MUSTy addition
```suggestion
Resource objects with :property:`link_type`=:val:`providers` links MUST point to an `Index Meta-Database`_ that supplies a list of OPTIMADE database providers.
```

optimade.rst
```diff
@@ -1046,21 +1047,36 @@ Example:
...
+Each link has a :property:`link_type` attribute that specifies the type of the linked relation.
+
+The :property:`link_type` MUST be one of the following values:
 
-For Links endpoints, the API implementation MAY ignore any provided query parameters.
+- :field-val:`child`: a link to another OPTIMADE implementation that MUST be within the same provider.
+  This allows the creation of a tree-like structure of databases by pointing to children sub-databases.
+- :field-val:`root`: a link to the root implementation within the same provider.
+  This is RECOMMENDED to be an `Index Meta-Database`_.
+  There MUST be only one :val:`root` implementation per provider and all implementations MUST have a link to this :val:`root` implementation. If the provider only supply a single implementation, the :val:`root` link links to the implementation itself.
```
CasperWA, line 1064: Splitting to one sentence per one line and fixing a minor grammatical mistake.
```suggestion
  There MUST be only one :val:`root` implementation per provider and all implementations MUST have a link to this :val:`root` implementation.
  If the provider only supplies a single implementation, the :val:`root` link links to the implementation itself.
```

optimade.rst
giovannipizzi: Ok with me!

optimade.rst
giovannipizzi: Ok with me

giovannipizzi:
All changes applied

giovannipizzi:
@CasperWA I merged your suggestions, but GitHub still mentions that you want changes, if this is not the case anymore can you dismiss you review?

Issue #275, opened 2020-06-11
Filtering on 'dimension_types'; suggestion for new 'periodic_dimensionality' field


ml-evs:
The discussion in #243 has opened up a more specific issue about the current definition of `dimension_types`.

As it stands, it is very difficult to query for 1D or 2D structures specifically, without checking for equality with all the possible orderings of `dimension_types` (e.g. `[1, 0, 0] OR [0, 1, 0] OR [0, 0, 1]` for 1D). This could be fixed by the suggestions in #243 for more direct filtering on lists, though it is not clear whether this solution will be ready for 1.0.

The suggestion is to create an auxiliary field, say `periodic_dimensionality` (name suggested by @rartino), that simply holds the sum of the current dimension types field for more useable querying. This will become especially useful when aggregating results between multiple databases, some of which may be focused entirely on low-dimensionality systems.

Some other suggestions for names (from discussions with @CasperWA  and @muhrin):
- `periodic_dimensionality` (@rartino's initial suggestion)
- `dimensionality` (so that it is next to `dimension_types` in the alphabetized response...)
- `ndimensions` (to follow the example of `nsites` and `nelements`)
- `periodicity`

This is obviously quite an easy field to add to the specification (pending the name), but there may be objections from providers about putting this field in before the 1.0 release, so please voice your concerns!

mkhorton:
I like the simplicity of `sum(dimension_types)` by default. There are also good examples in the literature of continuous dimensionality measures given the ambiguity sometimes in assigning specific dimensionality (e.g. see doi 10.1103/PhysRevMaterials.3.034003). I could imagine some databases might want to query on this field, which practically would mean allowing for a floating point `dimensionality` in the spec.

muhrin:
Just a quick question that related to @mkhorton 's point.  Does `dimension_types` only refer to the periodicity of the unit cell or some judgement on the periodicity of the thing contained within it.  For example, a cluster (which would be 0D) could still be in a fully periodic unit cell and, therefore, would have a `periodic_dimensionality` or 3.  You can think of any example you like for this (nano ribbons, molecular crystals, etc).

mkhorton:
My view on this is that one is a periodicity imposed on the system, and the other is a property of the system. In this sense, perhaps a floating-point dimensionality is better included under a property API rather than the core API. My intention in raising this though is that if it _is_ part of the core API, it might be important to say that it may be floating point, and it is up to the individual database how they determine periodicity.

[...]

PR #278, opened 2020-06-12
Added nperiodic_dimensions field

ml-evs:
Following the discussions in #275, this PR adds the field `nperiodic_dimensions` to the specification.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1802,6 +1802,27 @@ dimension\_types
   - For a 2D surface/slab, periodic on the plane defined by the first and third lattice vectors: :val:`[1, 0, 1]`
   - For a bulk 3D system: :val:`[1, 1, 1]`
 
+nperiodic\_dimensions
+~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: An integer specifying the number of periodic dimensions in the structure, equivalent to the number of non-zero entries in :property:`dimension\_types`.
+- **Type**: integer
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: MUST be a queryable property with support for all mandatory filter features.
+  - The integer value MUST be between 0 and 3 inclusive.
+  - This property only reflects the treatment of the lattice vectors provided for the structure, and not any physical interpretation of the dimensionality of its contents.
+
+- **Examples**:
+
+  - :val:`2` should be indicated in cases where :property:`dimension\_types` is any of :val:`[1, 1, 0]`, :val:`[1, 0, 1]`, :val:`[0, 1, 1]`.
+
+- **Query examples**:
+
+  - Match only structures with exactly 3 periodic dimensions: :filter:`nperiodic\_dimensions=3`
+  - Match all structures with 2 or fewer periodic dimensions: :filter:`nperiodic\_dimensions<=2`
+
 lattice\_vectors
 ~~~~~~~~~~~~~~~~
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1811,7 +1811,7 @@ nperiodic\_dimensions
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
-  - The integer value MUST be between 0 and 3 inclusive.
+  - The integer value MUST be between 0 and 3 inclusive and MUST be equal to the sum of the items in the `dimension_types`_ property.
   - This property only reflects the treatment of the lattice vectors provided for the structure, and not any physical interpretation of the dimensionality of its contents.
 
 - **Examples**:


optimade.rst
```diff
@@ -1802,6 +1802,27 @@ dimension\_types
...
+nperiodic\_dimensions
+~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: An integer specifying the number of periodic dimensions in the structure, equivalent to the number of non-zero entries in :property:`dimension\_types`.
+- **Type**: integer
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: MUST be a queryable property with support for all mandatory filter features.
+  - The integer value MUST be between 0 and 3 inclusive.
```
rartino, line 1814: Suggestion for clarity:
```suggestion
  - **Query**: MUST be a queryable property with support for all mandatory filter features.
  - The integer value MUST be between 0 and 3 inclusive and MUST be equal to the sum of the items in the `dimension_types`_ property.
```

optimade.rst
```diff
@@ -1802,6 +1802,27 @@ dimension\_types
...
+nperiodic\_dimensions
+~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: An integer specifying the number of periodic dimensions in the structure, equivalent to the number of non-zero entries in :property:`dimension\_types`.
+- **Type**: integer
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: MUST be a queryable property with support for all mandatory filter features.
+  - The integer value MUST be between 0 and 3 inclusive and MUST be equal to the sum of the items in the `dimension_types`_ property.
```
CasperWA, line 1814: Just a question: The way of stating that both values (0 and 3) are inclusive here that is the standard English way of writing this? Or are there some "fluff" missing around it?

optimade.rst
```diff
@@ -1802,6 +1802,27 @@ dimension\_types
   - For a 2D surface/slab, periodic on the plane defined by the first and third lattice vectors: :val:`[1, 0, 1]`
   - For a bulk 3D system: :val:`[1, 1, 1]`
 
+nperiodic\_dimensions
+~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: An integer specifying the number of periodic dimensions in the structure, equivalent to the number of non-zero entries in :property:`dimension\_types`.
```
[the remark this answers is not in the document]
rartino: Do we really need another 'integer', when the property is defined to be integer?

Also, there is a subtle difference between "sum of entries" and "number of non-zero entries". Do we have a preference, and if so, why?

(I have a preference *against* sum of non-zero entries, because that is just unnecessarily convoluted way to express this, since the zero entries doesn't contribute to the sum...)

optimade.rst
ml-evs:
```suggestion
  - This property MUST take an integer value from 0 to 3 and MUST be equal to the sum of the items in the `dimension_types`_ property.
```

I think I prefer my original wording, but is this clearer for you?

optimade.rst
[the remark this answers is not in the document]
giovannipizzi: I agree with @rartino

optimade.rst
giovannipizzi: If useful, both would be fine with me

optimade.rst
```diff
@@ -1802,6 +1802,27 @@ dimension\_types
...
+- **Type**: integer
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
```
giovannipizzi, line 1812: I guess it's ok since also `dimension_types` is SHOULD, but I would rather make both `MUST`? Or there was a reason for which we said SHOULD? Anyway, this can be treated in a different discussion thread maybe?

ml-evs: I was actually quite surprised at how few of our properties are `MUST`s, (`id`, `type`, `structure_features` only), especially when support from querying on almost all properties is a `MUST`...

optimade.rst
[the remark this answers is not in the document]
CasperWA: Consensus has been reached!

optimade.rst
CasperWA: The `inclusive` part is integral, it was more that I didn't immediately recognize the sentence structure when using `inclusive`. To be better for non-native English speakers it could simply be a more verbose structure, e.g.:
```suggestion
  - The integer value MUST be between 0 and 3 including both extrema and MUST be equal to the sum of the items in the `dimension_types`_ property.
```
Or something similar, I'm also not loving this structure, so in the end, if it's proper English the first time around, then that's also fine with me. And I then apologize for my lack of mastering the English language :) But happy to learn.

optimade.rst
CasperWA: This comes from a rewrite by @merkys in discussion with @rartino and myself at some earlier point, re-evaluating "required" properties in a response.

optimade.rst
rartino: I did not have trouble reading the original formulation and find "including both extrema" a very unusual way of expressing this. But if we really must reformulate it, we'd save characters on: "MUST be 0, 1, 2 or 3". There is also "MUST be a value from 0 to 3", which I think generally is viewed as less ambiguous than "between" and then may not need the 'inclusive' clarification.

rartino: We now have two approvals for this PR, but perhaps we should wait for @CasperWA to say if he is happy as it is, or if he wants the "From ... to..." solution, or something else.

giovannipizzi: I like @ml-evs latest suggestion. @CasperWA ?

optimade.rst
giovannipizzi: Still, I'm surprised that `dimension_types` is optional. A Si_2 molecule and a Si crystal might only differ on the `dimension_types`, and they are *very different* things. I would vote to reintroduce this back as a MUST even if I understand this might require further discussion... do you have a link to the issue/PR where this was discussed?

rartino: Why are you singling out `dimension_types`? All of the structural properties, `cartesian_site_positions`, `speices_at_sites`, and `species`, (and so on) are SHOULD. 

I suppose I was thinking that we put them as SHOULD now and see how it works out in implementations. If everyone implements them and no one sees any issues with them, we can increase the requirement to MUST.

(It is easier to change properties to mandatory than the other way around, since making mandatory fields non-mandatory means that you are technically breaking backwards compatibility.)

giovannipizzi: I see, fine. But I'll open an issue about this.

optimade.rst
CasperWA: I withdraw my suggested corrections for this, as there is consensus on it :)

PR #279, opened 2020-06-12
New representation of atoms at unknown positions

### LINKED ISSUE #258: Query support of structure_features properties
COD will not be able to handle queries for `structure_features` property, at least not until the release of OPTiMaDe, and even some time afterwards. We do not store the symmetry-reconstructed structures yet, although we plan to do so in future. Would it be possible to reduce the query support requirement level from MUST to RECOMMENDED? If not, the COD will not be fully supporting OPTiMaDe specification for quite some time.

### LINKED ISSUE #260: Definition of an unknown position
Currently, structure feature `unknown_positions` is defined as the following:

> `unknown_positions`: This flag MUST be present if at least one component of the `cartesian_site_positions` list of lists has value `null`.

Taking crystallographic structure descriptions in CIF format into consideration, this seemingly means "*explicit* sites with unknown coordinates". However, there might be *implicit* sites such as:

* Attached hydrogens. Sometimes hydrogen coordinates are left undefined. Instead, their number is listed as a property of the atom they are bound to.
* Solvent molecules. These are usually left out, and this fact is not necessary documented.
* Other non-modeled sites. Existence of them in a crystal could be inferred from mismatch of the given and the calculated formulae (not as a rule, though).

How should we deal with these implicit sites? Are they important enough to be listed in `species_at_sites`, and, in turn, affect the `unknown_positions`?

rartino:
This PR follows discussions at the OPTIMADE 2020 workshop.

As per issues #260 and #258 (and also see #50, #93) there has been quite some discussion about how to handle unknown positions. It was previously decided that we would allow `null` in the `cartesian_site_positions` property in the `structure` entry for this. However, since then it has come up that: (i) it is not desirable for databases to on-the-fly have to insert these null entries, (ii) it can be difficult for databases to provide reasonably efficient queries on whether such `null`s will be present or not.

Hence, after some discussion, this PR implements:

- The old notion of `unknown_atoms` is completely removed. `nulls` are no longer allowed in the `cartesian_site_position` property.
- A species definition may now include attached atoms, e.g., one can define 'CH4' to be a Carbon with four attached Hydrogens. This feature is provided by keys `attached` and `nattached` in the `sites` list. There is a flag `site_attachments` in `structure_features` to indicate when this feature is used so queries easily can exclude/include these structures.
- An implementation may serve structures with `implicit atoms`. This means that not all atoms are actually included in the structure definition provided in `cartisian_site_positions` and `species_at_sites`. This usually means there is a mismatch between the properties indicating the chemical formula and what is found in this structural definition. There is a flag `implicit_atoms` in `structure_features` to indicate when this feature is used so queries can exclude/include these structures.

(Other people prominently being part of the discussions shaping this PR: @sauliusg, @giovannipizzi, and @merkys).

Closes #260 
Closes #258

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1828,20 +1828,15 @@ lattice\_vectors
 cartesian\_site\_positions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- **Description**: Cartesian positions of each site. A site is an atom, a site potentially occupied by an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual crystal approximation).
-- **Type**: list of list of floats and/or unknown values.
+- **Description**: The cartesian positions of each site in the structure. A site is the position of a species (cf. the :property:`species` property.) as indicated by the :property:`species_at_sites` property.
+- **Type**: list of list of floats
 - **Requirements/Conventions**:
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: Support for queries on this property is OPTIONAL.
     If supported, filters MAY support only a subset of comparison operators.
-  - It MUST be a list of length equal to the number of sites in the structure where every element is a list of the three Cartesian coordinates of a site.
-  - An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see e.g., the property `assemblies`_).
-  - If a component of the position is unknown, the :val:`null` value SHOULD be provided instead (see section `Properties with an unknown value`_).
-    Otherwise, it SHOULD be a float value, expressed in angstrom (Å).
-    If at least one of the coordinates is unknown, the correct flag in the list property `structure_features`_ MUST be set.
-  - **Notes**: (for implementers) While this is unrelated to this OPTIMADE specification: If you decide to store internally the :property:`cartesian_site_positions` as a float array, you might want to represent :val:`null` values with :field-val:`NaN` values.
-    The latter being valid float numbers in the IEEE 754 standard in `IEEE 754-1985 <https://doi.org/10.1109/IEEESTD.1985.82928>`__ and in the updated version `IEEE 754-2008 <https://doi.org/10.1109/IEEESTD.2008.4610935>`__.
+  - It MUST be a list of length equal to the number of sites in the structure where every element is a list of the three Cartesian coordinates of a site expressed as float values, expressed in the unit angstrom (Å).
+  - An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see e.g., the property `assemblies`_). 
 
 - **Examples**:
 
@@ -1869,7 +1864,7 @@ nsites
 species\_at\_sites
 ~~~~~~~~~~~~~~~~~~
 
-- **Description**: Name of the species at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_).
+- **Description**: Name of the species at each site specified by the property `cartesian_site_positions`_ (in the same order.)
   The properties of the species are found in the property `species`_.
 - **Type**: list of strings.
 - **Requirements/Conventions**:
@@ -1891,7 +1886,9 @@ species\_at\_sites
 species
 ~~~~~~~
 
-- **Description**: A list describing the species of the sites of this structure. Species can be pure chemical elements, or virtual-crystal atoms representing a statistical occupation of a given site by multiple chemical elements.
+- **Description**: A list describing the species of the sites of this structure.
+  Species can represent pure chemical elements, virtual-crystal atoms representing a statistical occupation of a given site by multiple chemical elements, and/or a location to which there are attached atoms, i.e., atoms whose precise location are unknown beyond that they are attached to that position (more or less exclusively used to indicate hydrogen atoms attached to other elements.) 
+  
 - **Type**: list of dictionary with keys:
 
   - :property:`name`: string (REQUIRED)
@@ -1925,6 +1922,12 @@ species
 
       Note that concentrations are uncorrelated between different site (even of the same species).
 
+    - **attached**: OPTIONAL; if provided MUST be a list of length 1 or more of strings of chemical symbols for the elements attached to this site, or "X" for a non-chemical element.
+    - **nattached**: OPTIONAL; if provided MUST be a list of length 1 or more of integers indicating the number of attached atoms of the kind specified in the value of the :field:`attached` key.
+
+      The implementation MUST include either both or none of the :field:`attached` and :field:`nattached` keys, and if they are provided, they MUST be of the same length.
+      Furthermore, if they are provided, the `structure_features`_ property MUST include the string :val:`site_attachments`.
+      
     - **mass**: OPTIONAL. If present MUST be a float expressed in a.m.u.
     - **original\_name**: OPTIONAL. Can be any valid Unicode string, and SHOULD contain (if specified) the name of the species that is used internally in the source database.
 
@@ -2070,11 +2073,12 @@ structure\_features
   - If a special feature listed below is not used, the list MUST NOT contain the corresponding string.
   - **List of strings used to indicate special structure features**:
 
-    - :val:`disorder`: This flag MUST be present if any one entry in the :property:`species` list has a :property:`chemical_symbols` list that is longer than 1 element.
-    - :val:`unknown_positions`: This flag MUST be present if at least one component of the :property:`cartesian_site_positions` list of lists has value :val:`null`.
-    - :val:`assemblies`: This flag MUST be present if the property `assemblies`_ is present.
+    - :val:`disorder`: this flag MUST be present if any one entry in the `species`_ list has a :field:`chemical_symbols` list that is longer than 1 element.
+    - :val:`implicit_atoms`: this flag MUST be present if the structure contains atoms that are not assigned to sites via the property `species_at_sites`_ (e.g., because their positions are unknown). When this flag is present, the properties related to the chemical formula will likely not match the type and count of atoms represented by the `species_at_sites`_, `species`_, and `assemblies`_ properties. 
+    - :val:`site_attachments`: this flag MUST be present if any one entry in the `species`_ list includes :field:`attached` and :field:`nattached`.
+    - :val:`assemblies`: this flag MUST be present if the property `assemblies`_ is present.
 
--  **Examples**: A structure having unknown positions and using assemblies: :val:`["assemblies", "unknown_positions"]`
+-  **Examples**: A structure having implcit atoms and using assemblies: :val:`["assemblies", "implicit_atoms"]`
 
 Calculations Entries
 --------------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -1829,7 +1829,7 @@ cartesian\_site\_positions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 - **Description**: Cartesian positions of each site in the structure.
-   A site is usually used to describe positions of atoms; what atoms can be encountered at a given site is conveyed by the :property:`species_at_sites` property, and the species themselves are described in the :property:`species` property.
+  A site is usually used to describe positions of atoms; what atoms can be encountered at a given site is conveyed by the :property:`species_at_sites` property, and the species themselves are described in the :property:`species` property.
 - **Type**: list of list of floats
 - **Requirements/Conventions**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1828,7 +1828,8 @@ lattice\_vectors
 cartesian\_site\_positions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- **Description**: The cartesian positions of each site in the structure. A site is the position of a species (cf. the :property:`species` property.) as indicated by the :property:`species_at_sites` property.
+- **Description**: Cartesian positions of each site in the structure. 
+  A site is the location of a species (cf. the `species`_ property) as assigned by the `species_at_sites`_ property.
 - **Type**: list of list of floats
 - **Requirements/Conventions**:
 
@@ -1864,7 +1865,7 @@ nsites
 species\_at\_sites
 ~~~~~~~~~~~~~~~~~~
 
-- **Description**: Name of the species at each site specified by the property `cartesian_site_positions`_ (in the same order.)
+- **Description**: Name of the species at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_).
   The properties of the species are found in the property `species`_.
 - **Type**: list of strings.
 - **Requirements/Conventions**:
@@ -1887,7 +1888,7 @@ species
 ~~~~~~~
 
 - **Description**: A list describing the species of the sites of this structure.
-  Species can represent pure chemical elements, virtual-crystal atoms representing a statistical occupation of a given site by multiple chemical elements, and/or a location to which there are attached atoms, i.e., atoms whose precise location are unknown beyond that they are attached to that position (more or less exclusively used to indicate hydrogen atoms attached to other elements.) 
+  Species can represent pure chemical elements, virtual-crystal atoms representing a statistical occupation of a given site by multiple chemical elements, and/or a location to which there are attached atoms, i.e., atoms whose precise location are unknown beyond that they are attached to that position (frequently used to indicate hydrogen atoms attached to other elements). 
   
 - **Type**: list of dictionary with keys:
 
@@ -2078,7 +2079,7 @@ structure\_features
     - :val:`site_attachments`: this flag MUST be present if any one entry in the `species`_ list includes :field:`attached` and :field:`nattached`.
     - :val:`assemblies`: this flag MUST be present if the property `assemblies`_ is present.
 
--  **Examples**: A structure having implcit atoms and using assemblies: :val:`["assemblies", "implicit_atoms"]`
+-  **Examples**: A structure having implicit atoms and using assemblies: :val:`["assemblies", "implicit_atoms"]`
 
 Calculations Entries
 --------------------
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1828,20 +1828,15 @@ lattice\_vectors
 cartesian\_site\_positions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- **Description**: Cartesian positions of each site. A site is an atom, a site potentially occupied by an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual crystal approximation).
-- **Type**: list of list of floats and/or unknown values.
+- **Description**: The cartesian positions of each site in the structure. A site is the position of a species (cf. the :property:`species` property.) as indicated by the :property:`species_at_sites` property.
```
sauliusg: The formulation njow is very obscure, and refers to the syntactic structure of the file (which should be evident from the schema, eventually), and not to the intended meaning (semantics) of the atomic site. This makes specification too formal and very difficult to read.

I suggest leaving the old text with an additional clarification: "**Description**: Cartesian positions of each site. A site is an atom, a site potentially occupied by an atom, or a placeholder for a virtual mixture of atoms (e.g., in a virtual crystal approximation, or an atom with additional atoms attached to it, like in the case of CH3 group tree hydrogen are attached to a carbon)."

We then would specify, as suggested, that coordinates are a list of floats, and no `null` values should be present there.

rartino: Ok, sure I can revert the old clarificatio. But as I was editing these definition I really felt that there was unnecessary redundancy here that could be cleaned up, since essentially the same sentence appears also in the definition of a species. Should we really copy+paste the same rather long "explanation" of what a species is at multiple places?

rartino: I should add that the present version of this PR also renders, in this description of `cartisian_site_positions`, a link to the species definition. And, a reason why I cut the longer description away was that if I want to be formal, I now also need to add the possibility of the site having attached atoms. (Which I did add to the `species` definition).

sauliusg: > Ok, sure I can revert the old clarification. But as I was editing these definitions, I really felt that there was unnecessary redundancy here that could be cleaned up, since essentially the same sentence appears also in the definition of a species. Should we really copy+paste the same rather long "explanation" of what a species is at multiple places?

Duplication of definitions if of course a bad thing, but my point is that it is much clearer when the spec refers to indented meaning/use, or at least clearly gives an example for such.

So while copy+past of the main definition is not good, I agree, the same wording could be used in all places. IMHO in the end we refer to atoms, which we all thing about (and thing we understand about them ;), and not to abstractions of the interface like "species", "sites" and so on. 'species_at_sites" is the way we describe atomic structure of matter, isn't it?

optimade.rst
```diff
@@ -1891,7 +1886,9 @@ species\_at\_sites
 species
 ~~~~~~~
 
-- **Description**: A list describing the species of the sites of this structure. Species can be pure chemical elements, or virtual-crystal atoms representing a statistical occupation of a given site by multiple chemical elements.
+- **Description**: A list describing the species of the sites of this structure.
```
sauliusg, line 1890: I think it would be helpful to give an example here, so for instance a carbon with three attached hydrogens might represent a methyl group, -CH3

optimade.rst
sauliusg: Trying a different wording:

> cartesian\_site\_positions
> - **Description**: The Cartesian positions of each site in the structure. A site is usually used to describe positions of atoms; what atoms can be encountered at a given site is conveyed by the :property:`species_at_sites` property, and the species themselves are described in the :property:`species` property.
**Type**: list of list of floats

What I want to achieve:

- let the reader understand the intended use of the `cartesian_site_positions` is to describe atomic positions without traversing all indirections in the specification (cartesian_...→species_at_sites→spiecies);
- give the reference to all relevant descriptions without repeating too much of details.

I admit that I do not understand the meaning of the sentence "A site is the position of a species (cf. the :property:`species` property.) as indicated by the :property:`species_at_sites` property" even on the fifth reading...

rartino:
Thanks for the review @sauliusg, I've made the modifications suggested.

rartino:
Oh, I think I misunderstood you. I added the "example" of a "-CH3" to the description of species
> (frequently used to indicate hydrogen atoms attached to another element, e.g., a carbon with three attached hydrogens might represent a methyl group, -CH3).
But you mean to add an example to the list of species examples, right? Lets do that in this PR. Is this ok?:
```
[ {"name": "-CH3", "chemical_symbols": ["C"], "concentration": [1.0], "attached":"H", "nattached":3} ]
```

sauliusg:
> [ {"name": "-CH3", "chemical_symbols": ["C"], "concentration": [1.0], "attached":"H", "nattached":3} ]

Probably just "CH3" would be sufficient to give a computer-name like string, but of course "-CH3" would be more explicit that this is a methyl attached to something. I'm find with both ways the example would go (and assume a server can assign any string that is convenient).

rartino:
Alright, @sauliusg, @giovannipizzi, @CasperWA please re-review so we can get this merged. Thanks.

optimade.rst
```diff
@@ -1828,20 +1828,16 @@ lattice\_vectors
...
-  - If a component of the position is unknown, the :val:`null` value SHOULD be provided instead (see section `Properties with an unknown value`_).
-    Otherwise, it SHOULD be a float value, expressed in angstrom (Å).
-    If at least one of the coordinates is unknown, the correct flag in the list property `structure_features`_ MUST be set.
-  - **Notes**: (for implementers) While this is unrelated to this OPTIMADE specification: If you decide to store internally the :property:`cartesian_site_positions` as a float array, you might want to represent :val:`null` values with :field-val:`NaN` values.
-    The latter being valid float numbers in the IEEE 754 standard in `IEEE 754-1985 <https://doi.org/10.1109/IEEESTD.1985.82928>`__ and in the updated version `IEEE 754-2008 <https://doi.org/10.1109/IEEESTD.2008.4610935>`__.
+  - It MUST be a list of length equal to the number of sites in the structure where every element is a list of the three Cartesian coordinates of a site expressed as float values, expressed in the unit angstrom (Å).
+  - An entry MAY have multiple sites at the same Cartesian position (for a relevant use of this, see e.g., the property `assemblies`_).
```
CasperWA, line 1840: Would this also be equivalent to multiple `chemical_symbols` in a species?
I am slightly unaligned with the exact definition of `site` from the property description, and what it entails for this sentence.
Note, I don't see a need to change the description, but rather its implication on this sentence, i.e., extending this sentence with a comment on what it means for multiple `chemical_symbols` in a species; whether multiple sites at the same Cartesian position can also be set by adding multiple values to `chemical_symbols` of a `species` or if not (then one could add something like `not to be confused with adding multiple values for "chemical_symbols" of the related "species"`, or something?)

optimade.rst
```diff
@@ -1943,6 +1947,7 @@ species
   - :val:`[ {"name": "BaCa", "chemical_symbols": ["vacancy", "Ba", "Ca"], "concentration": [0.05, 0.45, 0.5], "mass": 88.5} ]`: any site with this species is occupied by a Ba atom with 45 % probability, a Ca atom with 50 % probability, and by a vacancy with 5 % probability. The mass of this site is (on average) 88.5 a.m.u.
   - :val:`[ {"name": "C12", "chemical_symbols": ["C"], "concentration": [1.0], "mass": 12.0} ]`: any site with this species is occupied by a carbon isotope with mass 12.
   - :val:`[ {"name": "C13", "chemical_symbols": ["C"], "concentration": [1.0], "mass": 13.0} ]`: any site with this species is occupied by a carbon isotope with mass 13.
+  - :val:`[ {"name": "CH3", "chemical_symbols": ["C"], "concentration": [1.0], "attached":"H", "nattached":3} ]` : any site with this species is occupied by a methyl group, -CH3, which is represented without specifying precise positions of the hydrogen atoms.
```
CasperWA, line 1950: The values of `attached` and `nattached` should be _lists_ of strings/integers right?
```suggestion
  - :val:`[ {"name": "CH3", "chemical_symbols": ["C"], "concentration": [1.0], "attached": ["H"], "nattached": [3]} ]` : any site with this species is occupied by a methyl group, -CH3, which is represented without specifying precise positions of the hydrogen atoms.
```

optimade.rst
```diff
@@ -2070,11 +2075,12 @@ structure\_features
   - If a special feature listed below is not used, the list MUST NOT contain the corresponding string.
   - **List of strings used to indicate special structure features**:
 
-    - :val:`disorder`: This flag MUST be present if any one entry in the :property:`species` list has a :property:`chemical_symbols` list that is longer than 1 element.
-    - :val:`unknown_positions`: This flag MUST be present if at least one component of the :property:`cartesian_site_positions` list of lists has value :val:`null`.
-    - :val:`assemblies`: This flag MUST be present if the property `assemblies`_ is present.
+    - :val:`disorder`: this flag MUST be present if any one entry in the `species`_ list has a :field:`chemical_symbols` list that is longer than 1 element.
+    - :val:`implicit_atoms`: this flag MUST be present if the structure contains atoms that are not assigned to sites via the property `species_at_sites`_ (e.g., because their positions are unknown). When this flag is present, the properties related to the chemical formula will likely not match the type and count of atoms represented by the `species_at_sites`_, `species`_, and `assemblies`_ properties.
```
CasperWA, line 2079:
```suggestion
    - :val:`implicit_atoms`: this flag MUST be present if the structure contains atoms that are not assigned to sites via the property `species_at_sites`_ (e.g., because their positions are unknown).
      When this flag is present, the properties related to the chemical formula will likely not match the type and count of atoms represented by the `species_at_sites`_, `species`_, and `assemblies`_ properties. 
```

CasperWA, line 2079: When would this be present without also having `site_attachments` present?
Or phrasing the question in another way: How could I add implicit atoms without mentioning them somewhere else? Or is that the whole point of implicit atoms? That it should be present if the structure contains "species/atoms" which are not accounted for by any property, e.g., crystalline/bound water or other solvents or the like?

optimade.rst
sauliusg: > Would this also be equivalent to multiple `chemical_symbols` in a species?

I would say physically yes, but logically no.

If two sites have the same Cartesian coords, then I would understand that physically either one atom or another can occupy the same location, with probabilities described in `group_probabilities`. Of course then the sites MUST belong to different groups. Bu the fact that their coordinates are the same is accidental; you can run a positional refinement and the sites will diverge.

If we have just one site with a species consisting of two atoms, then I understand that either one or another atom can be encountered at each site where these species is present (but there are no correlations between sites, right?); the probabilities are given in `concentration`. `vacancy` can be specified for a missing atom, with its own concentration. In this case atoms will (should?) never ever diverge from the given site.

> I am slightly unaligned with the exact definition of `site` from the property description, and what it entails for this sentence.

Hmmm... I wold understand that a `site` is a position in space (given by Cartesian coordinates) potentially occupied by an atom of one or another chemical kind; what atoms, in what proportions and in what correlations is described by `species`, `species_at_sites`, `sites_in_groups`. Actually, the current spec gives an example of describing the same situation using mixed-element species and alternatively using assemblies.

optimade.rst
sauliusg: > How could I add implicit atoms without mentioning them somewhere else? 

In OPTIMADE, you can not.

In CIFs (experimental crystallography), you can add a summary formula that species more atoms than are listed in the coordinates table; you can also add special _sqeeze data items that in my understanding describe electron density occupied by molecules that could not be resolved and modelled by discrete atoms because they are disordered. These molecules (e.g. solvent) will still be present in the formula, but absent from the atom list, and not attached to any particular atom. In that case, you would return for OPTIMADE `attached_atoms == false && implicit_atoms == true`

> Or is that the whole point of implicit atoms?

Yes!

optimade.rst
CasperWA: Thanks @sauliusg, I think your understanding aligns perfectly with my understanding as well.

For the purpose of the spec, I would then add a mention of `species`, but link to the comparison of `assemblies` and `species` to describe the exact meaning of `multiple sites at the same Cartesian position`, or rather examples of this.

optimade.rst
CasperWA: So for my two questions you've answered them with opposing answers, as I understand it.
While we in OPTIMADE do not allow the flag `"implicit_atoms"` to be `true` without also adding the atoms somewhere else (i.e., in another property), then that is still the flags purpose?

sauliusg: > So for my two questions you've answered them with opposing answers, as I understand it.
> While we in OPTIMADE do not allow the flag `"implicit_atoms"` to be `true` without also adding the atoms somewhere else (i.e., in another property), then that is still the flags purpose?

I probably misunderstood your (last?) question.

Let me try to clarify (for myself and for discussion with all of us):

- As intended in the current OPTIMADE PR, as I understand it, the flags `implicit_atoms` and attached_atoms` are orthogonal and independent. All four combinations are permissible:

1. `implicit_atoms == false && attached_atoms == false`: This is a structure where every atom that has been modelled has a corresponding `species_at_sites` entry and the Cartesian coordinates; these are structures that computational material scientists are mostly after, but not the only structures that experimentalists produce :) ;

2. `implicit_atoms == true && attached_atoms == false`: we *know* that the material described contains more atoms than specified in the coordinates list, but we do not know where these atoms are attached. In experimental databases, this will manifest through the fact that a formula computed from atomic coordinates and the chemical formula declared by the authors do not match. We do not know, however, where these atoms are in the cell.

3. `implicit_atoms == false && attached_atoms == true`: there are atoms to which other atoms attached, and we know *all* such attachments. Most often the attached atoms will be hydrogens (but can probably be deuterium; although deuterium is usually determined by neutron diffraction and its position is know precisely). The calculated formula matches the declared formula if you take attached atoms into account; no other atoms in the unit cell are known/modelled/claimed.

For such structures you can reasonably well reconstruct the hydrogens; for example in a benzene ring a hydrogen atom attached to a ring carbon will sit in a plane of the ring on a bissectrice at a known distance. If you place all such atoms (you can do this automatically most often), then you will know coordinates of all atoms and you can proceed with the calculation. Sometimes the position of hydrogens will be ambiguous, e.g. a methyl group (-CH3) can be rotated around its bond; but hopefully the differences are small for the coordinate relaxation to converge (I wold be interested to try such calculation, actually...). 

While placing hydrogens in this case is possible, the database will *not* do it for you ;)

4. Finally, `implicit_atoms == true && attached_atoms == true`: it means that both situations of from cases 2 and 3 are present. Some atoms are attached and you can roughly reconstruct them; but some will still be missing. It's up to you what you do with such structure; most probably computationalists will discard it (as well as the case 2), but the model can be uselful in many different cases. Also, a big question for computational material scientists: is it possible to *guess* an approximate positions of the missing atoms and then *compute* them using DFT?

CasperWA: So again, I believe I understand all of these scenarios and can come up with examples of my own.
My concern is more of clarification of whether the OPTIMADE `structures` data structure needs to contain the information of the implicit atoms, i.e., one could add `species` which do not have a site, which could be the implicit atoms.
One may or may not even know the content (but not position) of such species, so how would this be specified in the `concentration` field?

My concern here is simply that the `implicit_atoms` is not well defined in terms of what it requires otherwise in the OPTIMADE `structures` data structure.
We may say that this is fine, and providers can do as they wish, however, it may be valuable to understand which species are considered implicit if possible, and hence recommend a certain way of supplying this information.

A concrete example may be crystalline water, where it's easier to provide a 3D density mapping of the probability of where to find the water. It may be catalytic compounds that shift around so much that they cannot be pinpointed to a specific location, by could then (surely) be attached (i.e., one sets the `attached_atoms` flag and fill in the appropriate fields in `species`), however, if it's not attached anywhere, or more realistically, it is unknown where part of it is attached, then where should one put this information?

CasperWA: Now, just re-reading your latest post @sauliusg, I think my main point is this:

We should come up with a recommendation on how to supply information on implicit atoms/species.
And our data structure / specification should be flexible enough to accommodate this. Otherwise, we're losing information using the OPTIMADE schema compared to the original database schema (relating to the implicit atoms), while at the same time knowingly stating that we are missing this information (through the `implicit_atoms` flag).

optimade.rst
rartino: The precise point on how to interpret repeated coordinates, and how to formulate the text around it, was a tricky issue to agree on, so we should not edit this without multiple other people taking a look. If what you ask for is to add a further clarification on text that this PR does not modify (?), can you please post a separate PR with the suggested clarification so we can discuss that separately? I don't think this has anything to do with atoms at unknown coordinates?

On the actual point, my view is aligned with @sauliusg. Repeated coordinates represent *different* sites that happen to have the same numerical values as coordinates, but those coordinates could diverge with higher accuracy. There are very few reasons to ever use this in practice, aside assemblies. Multiple `chemical_symbols` + `concentrations` on one site represent correlated disordered atoms that should sit on exactly the *same* site, correlated in such a way that two atoms never sit on that site at the same time. If you tried to represent the same thing with two sites with repeated coordinates in OPTIMADE with 50% atom + 50% vacancy, they would individually, and in an uncorrelated way, disorder at respective site as 50% atom/50% vacancy; meaning that sometimes both atoms would be sitting there at the same coordinates at the same time (which probably is not what you want).

 It was an important point that there never should be a need on either client or server to "detect" repeated coordinates and prescribe a special meaning to that situation, compared to just treating them as two different sites.

optimade.rst
rartino: Great save; thank you.

optimade.rst
rartino: At this time there is no standardized formal way of specifying which atoms are implicit, except that they MAY be indicated in the formulas (but the specification presently does not make this *mandatory*). It is my impression from the lengthy discussions around this that we should not add any mandatory requirements to further specify this, as in the case of many cif files, it will be impossible to do so without inserting someones interpretation of the data.

I don't have anything against additional optional specifications of what the missing atoms are. Please post as a separate PR.

rartino:
I've commited @CasperWA's edits. For his larger points, I'd like to see them as separate PRs as I write in the comments. I don't think they are blocking v1.0.0-rc.2 (which we are already late for...)

optimade.rst
CasperWA: I didn't mean to add further mandatory requirements, but rather specification recommendations, i.e., RECOMMENDED instead of MUST or SHOULD.
However, to not delay this any longer, I will direct any suggested changes I may have into another issue/PR.

rartino: Right, so what could be useful is some optional field for specifying the implicit atoms, which can go in another PR.

(I just note that as far as I read [rfc2119](https://www.ietf.org/rfc/rfc2119.txt), there is no difference between SHOULD and RECOMMENDED.)

PR #282, opened 2020-06-12
Add `type` key to properties (simple)

CasperWA:
This addresses #269, but does not close it. That will be up to #270.

The `type` key will hold the value of the outermost type of the property it describes, i.e., for `species`, which is described as a "list of dictionaries", the `type` will be `"list"`.

The value of `type` MUST be one of the OPTIMADE data types described [here](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#data-types) in the specification.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -23,6 +23,8 @@ OPTIMADE API specification v1.0.0-rc.1~develop
      property : data item that pertains to an entry.
      val : value examples that properties can be.
            :val: is ONLY used when referencing values of actual properties, i.e., information that pertains to the database.
+     type : data type of values.
+            MUST equal a valid OPTIMADE data type as listed and defined under `Data types`_.
 
      # URL queries
 
@@ -66,6 +68,8 @@ OPTIMADE API specification v1.0.0-rc.1~develop
 .. role:: property(literal)
 
 .. role:: val(literal)
+
+.. role:: type(literal)
 
 .. role:: property-fail(literal)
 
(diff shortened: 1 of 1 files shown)


CasperWA:
> I'm happy with this as is, but do you think there should be a discussion of the `type` field in the section about provider-specific types? i.e. "Providers MUST/SHOULD provide the type of custom fields at the info/entry endpoint", or some such? Otherwise I don't see what value is gained from this PR alone for clients trying to display custom fields (as probably no-one will implement this field).

That would be great, but considering so many of the properties are already not a MUST, I think it's hard to justify this point.

However, we could always add a RECOMMENDED?

rartino:
When we later upgrade this field to be a more complete type description; will we have to consider this a backwards-breaking change (i.e., new major version number) or a minor version? 

How about we write in this text to that if the value isn't a simple OPTIMADE Datatype, the client should ignore the field?

optimade.rst
```diff
@@ -1008,7 +1008,27 @@ The response for these endpoints MUST include the following information in the :
...
+  *OPTIONAL keys*:
+
+  - :field:`unit`: String.
+    The physical unit symbol in which the property's value is given.
+    It is RECOMMENDED that non-standard (non-SI) units are described in the description for the property.
+  - :field:`sortable`: Boolean.
+    Whether the property can be used for sorting (see `Entry Listing URL Query Parameters`_ for more information on this field).
+  - :field:`type`: String.
+    The type of the property's value.
+    This can be any of the types defined in `Data types`_.
```
rartino, line 1027:
```suggestion
  - :field:`type`: String.
    The type of the property's value.
    This can be any of the types defined in `Data types`_.
    For the purpose of compatibility with future versions of this specification, a client must accept values that is not a String specifying a recognized type, and if so MUST disregard the `type` field.
```

giovannipizzi: Maybe a simplification: `a client must accept values that is not a String specifying a recognized type, and if so MUST disregard the `type` field` -> `a client MUST disregard the ``type`` field if its value is unknown.`

rartino: @giovannipizzi Is your edit meant to be synonymous with what I wrote, or is it meant to change the meaning into something you think is better? 

I see several possible interpretations of "the value is unknown" for a json field, so this does not seems as precise as I tried to be. Especially with the possible confusion (?) with our own definition in the spec that "unknown values" are ~ `nulls`... ?

giovannipizzi: Wanted to be synonymous. I see your point. What about `a client MUST disregard the ``type`` field if its value is not recognized.`

rartino: Seems slightly less precise than my formulation, but I'm ok with it. I guess I'll let @CasperWA decide since this is his PR.

CasperWA: I've added something now using some of @rartino's words, adding my own, to try and have something I also understand myself :)

CasperWA:
In the latest commit I've added the following:

- A database-provider-specific property MUST be listed in its respective entry listing `/info` endpoint.
- Future-proofing the `type` field.
- New `:type:` rst role to be used when writing OPTIMADE data types.

PR #284, opened 2020-06-12
Adding the aggregate option for links

giovannipizzi:
This was suggested during the OPTIMADE 2020 workshop.

It adds the `aggregate` option (with four possible values: `ok`, `no` (general catch-all), `test` (test DBs) and `staging` (almost-production-ready DBs). 
The specs are future-proofs, and it will be possible to add more reasons in the future (any new string will mean "don't follow unless you know better").

Also adding an optional `no_aggregate_reason` string field for those DBs that want to say why `aggregate != "ok"` (note that it's not required to provide it).

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1096,6 +1096,18 @@ The resource objects' response dictionaries MUST include the following fields:
   - **link\_type**: a string containing the link type.
     It MUST be one of the values listed above in section `Link Types`_.
 
+  - **aggregate**: a string indicating whether a client that is following links to aggregate results from different OPTIMADE implementations should follow this link or not. This flag SHOULD NOT be indicated for links where :property:`link_type` is not :val:`child`.
+
+    If not specified, clients MAY assume that the value is :val:`ok`.
+    If specified, and the value is anything different than :val:`ok`, the client MUST assume that the server is suggesting not to follow the link during aggregation by default (also if the value is not among the known ones, in case a future specification adds new accepted values).
+
+    Specific values indicate the reason why the server is providing the suggestion.
+    A client MAY follow the link anyway if it has reason to do so (e.g., if the client is looking for all test databases, it MAY follow the links marked with :property:`aggregate`=:val:`test`).
+  
+    If specified, it MUST be one of the values listed in section `Link Aggregate Options`_.
+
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1198,6 +1235,16 @@ If the provider implements an `Index Meta-Database`_, it is RECOMMENDED to adopt
 
 This will make all OPTIMADE databases and implementations by the provider discoverable as links with :val:`child` link type, under the Links endpoint of the `Index Meta-Database`_.
 
+Link Aggregate Options
+~~~~~~~~~~~~~~~~~~~~~~
+
+If specified, the :property:`aggregate` attributed MUST have one of the following values:
+
+- :val:`ok` (default value, if unspecified): it is ok to follow this link when aggregating OPTIMADE results.
```
rartino, line 1243: I'm not sure I am super-happy about `ok` instead of the `yes` we agreed on. I understand that you want to further emphasis that aggregating results is always up to the client. Still, when a parameter is "essentially a boolean", 'yes' is more expected than 'ok' as the truth value.

I'm not going to block merge based on that, though. Next reviewer can comment on their agreement or disagreement.

giovannipizzi: I see, but it's not really a boolean, in the sense that e.g. `test` does not *really* mean don't aggregate, but rather "this is a test db, aggregate only if this type of DB is of interest to you".

Anyway, I can change to `yes` if we prefer (I had forgotten, I have to admit). Happy to hear someone else out!

Or I can change `no -> rather_not` ;-)

rartino: If you prefer "ok", keep it; I don't feel strongly about this at all. But we anyway need one more approval, so maybe with luck we get a third vote on this.

sauliusg: > I see, but it's not really a boolean, in the sense that e.g. `test` does not _really_ mean don't aggregate, but rather "this is a test db, aggregate only if this type of DB is of interest to you".

I agree with @giovannipizzi , this is not a Boolean value and so "yes" would not be a Boolean truth value, but rather a hit that it is "ok" to aggregate.

rartino: ok 

:)

PR #288, opened 2020-06-14
Adjust query requirements on dimension_types

### LINKED ISSUE #276: We cannot filter for equality on list properties
In the list of properties, for some properties with list type values, the spec specifies a requirement of equality querying. This is however not possible according with the current filter grammar.

An example of this can be seen for [`dimension_types`](https://github.com/Materials-Consortia/OPTIMADE/blob/7157cc0ca42d4747fe8466f342903685441118ea/optimade.rst#dimension-types).

These requirements should be relaxed to having the `property HAS value` required, which is equality of a value within the list.

rartino:
Fixes #276, I think. 

The only property I could find to reference queries on equality on a list is `dimension_types`. As per other discussions, this PR tunes it down so all queries on this field are OPTIONAL, since it is now mandatory to support queries instead on `nperiodic_dimensions` which should support all the relevant use cases for queries on this, but in a much simpler way.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1874,8 +1874,7 @@ dimension\_types
 - **Requirements/Conventions**:
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
-  - **Query**: MUST be a queryable property.
-    Support for equality comparison is REQUIRED, support for other comparison operators are OPTIONAL.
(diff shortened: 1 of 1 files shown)


rartino:
Lets do it the present way for now, and if you want to add this as a small clarification, put it in a new PR we can merge between -rc.2 and proper.

PR #289, opened 2020-06-16
Formatting fixes

ml-evs:
Some minor, and one not so minor, changes to formatting in the specification.

I'm leaving this as a draft for now until I've finished going through all the models for the python-tools, which is why I spotted these.

Most of these arise in the rendering of the rst only: often things like ```":property:`aggregate`=:val:`value`"``` were attempted, which do not render nicely at all. These sections were reworded slightly to avoid this construct.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -272,7 +272,7 @@ The :field:`index_base_url` field MUST be included in every response in the :fie
 The :field:`is_index` field under :field:`attributes` as well as the :field:`relationships` field, MUST be included in the :endpoint:`info` endpoint for the index meta-database (see section `Base Info Endpoint`_).
 The value for :field:`is_index` MUST be :field-val:`true`.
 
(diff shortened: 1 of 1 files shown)


CasperWA:
@ml-evs Can I add to this PR as well? :)

ml-evs:
> @ml-evs Can I add to this PR as well? :)

Yes! You should have access to my branch by the time you read this

rartino:
I added a small fix of changing a "MANDATORY" to "REQUIRED" in one place, since MANDATORY isn't actually one of the words defined in rfc2119.

ml-evs:
I haven't been able to spot anything else to fix arising from our model updates, it would be good to lock this in before we update our models, in case we spot something else...

ml-evs:
@CasperWA did you still have stuff to add here?

CasperWA:
> @CasperWA did you still have stuff to add here?

I forget, there's probably something, but I'll open another PR if it's the case as to not hold this up.

ml-evs:
Found another couple of required fixes from @CasperWA's review in python-tools. Will re-request reviews once we're done with it.

PR #290, opened 2020-06-17
Serving the API on the unversioned base URL

### LINKED ISSUE #277: Standardizing serving the API on the unversioned URL
This was brought up at the end of the OPTIMADE 2020 workshop:

Right now the specification does not forbid, but also does not in any way endorse, serving the the API on the unversioned base URL.

Do we want to standardize how the API is served on the unversioned base url URL?

If we cannot agree, then the "default" is that we release as v1.0.0 the specification that neither forbids, nor endorses, serving something at the unversioned URL.

This question also ties into what the proper procedure for version negotiation between a client and server is. Presently the specification does prescribe the following procedure:
> For implementers: Clients are recommended to discover the highest version supported by both the client and the API implementation by trying versioned base URLs in order of priority. E.g., if major version 2 and lower are supported by the client, it would try: /v2, /v1, and then /v0.

rartino:
This PR is an attempt at accommodating the various views in #277 for how version negotiation between server and client should be done, and what should be allowed and not with regards to serving the API on the unversioned base URL.

Closes #277

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -285,7 +285,7 @@ Implementations serving the API on the unversioned base URL have a few alternati
 Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
 
 The client MAY provide a query parameter :query-param:`api_hint` to hint the server about a preferred API version.
-When this parameter is provided, the request is to be handled as described in section `Version Negotiation`_, which allows a "best sutiable" version of the API to be selected to serve the request (or forward the request to).
+When this parameter is provided, the request is to be handled as described in section `Version Negotiation`_, which allows a "best suitable" version of the API to be selected to serve the request (or forward the request to).
 However, if :query-param:`api_hint` is not provided, the implementation SHOULD serve (or redirect to) its preferred version of the API (i.e., the lastest, most mature, and stable version).
 In this case, that version MUST also be the first version in the response of the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -236,7 +236,7 @@ Versioned base URLs
 ~~~~~~~~~~~~~~~~~~~
 
 Access to the API is primarily provided under **versioned base URLs**.
-An implementation MUST provide the form where a URL path segment :query-url:`/vMAJOR` is appended to the base URL, where :val:`MAJOR` is one of the major version numbers of the API that the implementation supports.
+An implementation MUST provide access to the API under a URL where the first path segment appended to the base URL is :query-url:`/vMAJOR`, where :val:`MAJOR` is one of the major version numbers of the API that the implementation supports.
 This URL MUST serve the *latest* minor/patch version supported by the implementation.
 For example, the latest minor and patch version of major version 1 of the API is served under :query-url:`/v1`.
 
@@ -304,7 +304,7 @@ The OPTIMADE API provides three concurrent mechanisms for version negotiation be
    This parameter is described in more detail below.
 
 The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
-However, for endpoints under a versioned base URL this parameter SHOULD be ignored, and the request served as usual according to the version specified in the URL path segment.
+However, for endpoints under a versioned base URL this parameter MUST be ignored, and the request served as usual according to the version specified in the URL path segment.
 If the client provides the parameter, the value SHOULD have the format :val:`vMAJOR` or :val:`vMAJOR.MINOR`, where MAJOR is a major version and MINOR is a minor version of the API.
 For example, if a client appends :query-string:`api_hint=v1.0` to the query string, the hint provided is for major version 1 and minor version 0.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -305,7 +305,7 @@ The OPTIMADE API provides three concurrent mechanisms for version negotiation be
 
 The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
 However, for endpoints under a versioned base URL this parameter SHOULD be ignored, and the request served as usual according to the version specified in the URL path segment.
-If the client provides the parameter, the value SHOULD be on the format :val:`vMAJOR` where MAJOR is a major version of the API.
+If the client provides the parameter, the value SHOULD have the format :val:`vMAJOR` where MAJOR is a major version of the API.
 For example, if a client appends :query-string:`api_hint=v1` to the query string, the hint provided is for major version 1.
 
 If the server supports the version indicated by the :query-param:`api_hint` parameter, it SHOULD serve the request using this version.
--- a/optimade.rst
+++ b/optimade.rst
@@ -278,9 +278,9 @@ Clients that perform automated processing of responses SHOULD access the API via
 
 Implementations serving the API on the unversioned base URL have a few alternative options:
 
-1. Direct access may be provided to the full API.
-2. Requests to endpoints under the unversioned base URL may be redirected using an HTTP 307 temporary redirect to the corresponding endpoints under a versioned base URL.
-3. Direct access may be limited to only single entry endpoints (see section `Single Entry Endpoints`_), i.e., so that this form of access is only available for permanent links to resource objects.
+1. Direct access MAY be provided to the full API.
+2. Requests to endpoints under the unversioned base URL MAY be redirected using an HTTP 307 temporary redirect to the corresponding endpoints under a versioned base URL.
+3. Direct access MAY be limited to only single entry endpoints (see section `Single Entry Endpoints`_), i.e., so that this form of access is only available for permanent links to resource objects.
 
 Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
 
@@ -289,7 +289,7 @@ If this parameter is provided, the server MAY serve the request using (or redire
 However, if :query-param:`api_hint` is not provided, the implementation SHOULD serve (or redirect to) its preferred version of the API (i.e., the lastest, most mature, and stable version).
 In this case, that version MUST also be the first version in the response of the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
 
-    **For implementers**: Before enabling access to the API on unversioned base URLs, implementers are advised to consider that an upgrade of the major version of the API served this way may change the behaviors of associated endpoints in ways that are not backward compatible.
+    **For implementers**: Before enabling access to the API on unversioned base URLs, implementers are advised to consider that an upgrade of the major version of the API served this way can change the behaviors of associated endpoints in ways that are not backward compatible.
 
 Version Negotiation
 -------------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -285,7 +285,7 @@ Implementations serving the API on the unversioned base URL have a few alternati
 Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
 
 The client MAY provide a query parameter :query-param:`api_hint` to hint the server about a preferred API version.
-If this parameter is provided, the server MAY serve the request using (or redirect the request to) any version of the API that it finds appropriate, see section `Version Negotiation`_ for more details.
+When this parameter is provided, the request is to be handled as described in section `Version Negotiation`_, which allows a "best sutiable" version of the API to be selected to serve the request (or forward the request to).
 However, if :query-param:`api_hint` is not provided, the implementation SHOULD serve (or redirect to) its preferred version of the API (i.e., the lastest, most mature, and stable version).
 In this case, that version MUST also be the first version in the response of the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
 
@@ -305,7 +305,7 @@ The OPTIMADE API provides three concurrent mechanisms for version negotiation be
 
 The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
 However, for endpoints under a versioned base URL this parameter SHOULD be ignored, and the request served as usual according to the version specified in the URL path segment.
-The client SHOULD set the value on the format :val:`vMAJOR` where MAJOR is a major version of the API.
+When a client provides the parameter, the value SHOULD be on the format :val:`vMAJOR` where MAJOR is a major version of the API.
 For example, if a client appends to the query string :query-string:`api_hint=v1` the hint provided is for major version 1.
 
 The intent of the :query-param:`api_hint` parameter is that if a client hints a version that is not supported by the server, the server can use the value to make a best-effort attempt at still serving the request, e.g., by invoking the closest supported version of the API.
--- a/optimade.rst
+++ b/optimade.rst
@@ -713,7 +713,7 @@ Versions Endpoint
 The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
 This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
 
-The response to a query to this endpoint is in the :RFC:`4180` CSV (`text/csv; header=present`) format (and does therefore not comply with the JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
+The response to a query to this endpoint is in the :RFC:`4180` CSV (`text/csv; header=present`) format.
 In the present version of the API, the response contains only a single field that is used to list the major versions of the API that the implementation supports.
 However, clients MUST accept responses that include other fields that follow the version.
 The CSV format header line MUST be provided, and the header for the first field MUST be :val:`version`.
--- a/optimade.rst
+++ b/optimade.rst
@@ -684,10 +684,10 @@ Versions Endpoint
 The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
 This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
 
-The response to a query to this endpoint is in the :RFC:`4180` CSV format (and does therefore not comply with JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
-In the present version of the API the response contains only a single field, which is used to list the major versions of the API that the implementation supports.
+The response to a query to this endpoint is in the :RFC:`4180` `text/csv; header=present` format (and does therefore not comply with JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
+In the present version of the API, the response contains only a single field that is used to list the major versions of the API that the implementation supports.
 However, clients MUST accept responses that include other fields that follow the version.
-If the CSV response includes a header line, the name of the first field MUST be "#version", and the client can therefore safely disregard the first line if it starts with the '#' character.
+The csv format header line MUST be provided, and the header for the first field MUST be `#version`.
 
 The major API versions in the response are to be ordered according to the preference of the API implementation.
 If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the top line of the response (after the optional CSV header).
--- a/optimade.rst
+++ b/optimade.rst
@@ -282,7 +282,7 @@ Implementations serving the API on the unversioned base URL have a few alternati
 2. Requests to endpoints under the unversioned base URL MAY be redirected using an HTTP 307 temporary redirect to the corresponding endpoints under a versioned base URL.
 3. Direct access MAY be limited to only single entry endpoints (see section `Single Entry Endpoints`_), i.e., so that this form of access is only available for permanent links to resource objects.
 
-Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
+Implementations MAY combine direct access to single entry endpoints with redirects for other API queries.
 
 The client MAY provide a query parameter :query-param:`api_hint` to hint the server about a preferred API version.
 When this parameter is provided, the request is to be handled as described in section `Version Negotiation`_, which allows a "best suitable" version of the API to be selected to serve the request (or forward the request to).
@@ -304,8 +304,8 @@ The OPTIMADE API provides three concurrent mechanisms for version negotiation be
    This parameter is described in more detail below.
 
 The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
-However, for endpoints under a versioned base URL the request MUST be served as usual according to the version specified in the URL path segment regardless of the value of `api_hint`.
-In this case, the server MAY issue a warning if the value of `api_hint` suggests that the query may not be properly supported.
+However, for endpoints under a versioned base URL the request MUST be served as usual according to the version specified in the URL path segment regardless of the value of :query-param:`api_hint`.
+In this case, the server MAY issue a warning if the value of :query-param:`api_hint` suggests that the query may not be properly supported.
 If the client provides the parameter, the value SHOULD have the format :val:`vMAJOR` or :val:`vMAJOR.MINOR`, where MAJOR is a major version and MINOR is a minor version of the API.
 For example, if a client appends :query-string:`api_hint=v1.0` to the query string, the hint provided is for major version 1 and minor version 0.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -249,11 +249,11 @@ A URL on the form  :query-url:`/vMAJOR.MINOR` MUST serve the *latest* patch vers
 
 API versions that are published with a suffix, e.g., `-rc<number>` to indicate a release candidate version, SHOULD be served on versioned base URLs without this suffix.
 
-If a request is made to a versioned base URL that begins with `/v` and an integer followed by any other characters, indicating a version that the implementation does not recognize or support, it SHOULD respond with the custom HTTP server error status code :http-error:`553 API Version Not Supported`, perferably along with a user-friendly error message that directs the client to adapt the request to a version it provides.
+If a request is made to a versioned base URL that begins with `/v` and an integer followed by any other characters, indicating a version that the implementation does not recognize or support, the implementation SHOULD respond with the custom HTTP server error status code :http-error:`553 API Version Not Supported`, perferably along with a user-friendly error message that directs the client to adapt the request to a version it provides.
 
 It is the intent that future versions of this standard will not assign different meanings to URLs that begin with :query-url:`/v` and an integer followed by other characters.
 Hence, a client can safely attempt to access a specific version of the API via the corresponding versioned base URL. 
-Other forms of version negotiation is provided by the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
+Other forms of version negotiation are provided by the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
 
 Examples of valid versioned base URLs:
 
@@ -276,23 +276,23 @@ Unversioned base URLs
 
 Implementations MAY also provide access to the API on the **unversioned base URL** as described in this subsection.
 
-Access via the unversioned URL is primarily intended for (i) convinience when manually interacting with the API, and (ii) to provide version agnostic permanent links to resource objects.
+Access via the unversioned URL is primarily intended for (i) convenience when manually interacting with the API, and (ii) to provide version agnostic permanent links to resource objects.
 Clients that perform automated processing of responses SHOULD access the API via versioned base URLs.
 
 Implementations MAY provide direct access to the full API under the unversioned base URL. 
 In this case, the implementation SHOULD serve its preferred version of the API (i.e., the lastest, most mature, and stable version.) 
 
-Implementations MAY instead issue a HTTP 307 temporary redirect for queries issued to endpoints under the unversioned base URL.  
-In that case, the redirect SHOULD be to the corresponding versioned URL for the implementations preferred version of the API.
+Implementations MAY instead issue an HTTP 307 temporary redirect for queries issued to endpoints under the unversioned base URL.  
+In that case, the redirect SHOULD be to the corresponding versioned URL for the preferred version of the API supported by the implementation.
 
 As a third alternative, implementations MAY choose to provide direct access only to Single Entry Endpoints (see section `Single Entry Endpoints`_) under the unversioned base URL, i.e., so that this form of access is only available for permanent links to resource objects.
-In that case, the response format SHOULD be that of the implementations preferred version of the API.
+In that case, the response format SHOULD be that of the preferred version of the API supported by the implementation.
 Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
 
 When an implementation serves (or redirects to) a preferred major version of the API on the unversioned base URL, that version MUST also be the version at the first line in the response of the :endpoint:`versions` endpoint (see `Versions Endpoint`_).
 
-    **For implementers**: Before enabling access to the API on unversioned base URLs, implementations are advised to consider that this feature may encourage careless automated access from clients that, without verification, assumes a specific version of the API to be served on the unversioned base URL.
-    The implementation needs to be prepared to break such clients on major version upgrades, or will otherwise get stuck at serving the *oldest* API version on the unversioned URL.
+    **For implementers**: Before enabling access to the API on unversioned base URLs, implementers are advised to consider that this feature may encourage careless automated access from clients that, without verification, assumes a specific version of the API to be served on the unversioned base URL.
+    The implementer needs to be prepared to break such clients on major version upgrades, or will otherwise get stuck at serving the *oldest* API version on the unversioned URL.
 
 Index Meta-Database
 -------------------
@@ -666,7 +666,7 @@ For implementation-specific warnings, they MUST start with ``_`` and the databas
 API Endpoints
 =============
 
-Access to API endpoints as desribed in the subsections below are to be provided under the versioned and/or the unversioned base URL as desribed in the section `Base URL`_.
+Access to API endpoints as described in the subsections below are to be provided under the versioned and/or the unversioned base URL as explained in the section `Base URL`_.
 
 The endpoints are:
 
@@ -686,7 +686,7 @@ Versions Endpoint
 The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
 This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
 
-The response to a query to this endpoint is in the :RFC:`4180` `text/csv; header=present` format (and does therefore not comply with JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
+The response to a query to this endpoint is in the :RFC:`4180` `text/csv; header=present` format (and does therefore not comply with the JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
 In the present version of the API, the response contains only a single field that is used to list the major versions of the API that the implementation supports.
 However, clients MUST accept responses that include other fields that follow the version.
 The csv format header line MUST be provided, and the header for the first field MUST be `#version`.
@@ -704,8 +704,8 @@ Example response:
   1
   0
 
-The above response means that the API versions 1 and 0 are served under the versioned base URLs `/v1` and `/v0` respectively.
-The order of the versions indicate that the API implementation regards version 1 as preferred over version 0.
+The above response means that the API versions 1 and 0 are served under the versioned base URLs `/v1` and `/v0`, respectively.
+The order of the versions indicates that the API implementation regards version 1 as preferred over version 0.
 If the API implementation allows access to the API on the unversioned base URL, this access has to be to version 1, since the number 1 appears in the first (non-header) line.
 
 Entry Listing Endpoints
--- a/optimade.rst
+++ b/optimade.rst
@@ -242,7 +242,7 @@ Versioned base URLs
 Access to the API is primarily provided under **versioned base URLs**.
 An implementation MUST provide the form where a URL path segment :query-url:`/vMAJOR` is appended to the base URL, where :val:`MAJOR` is one of the major version numbers of the API that the implementation supports.
 This URL MUST serve the *latest* minor/patch version supported by the implementation.
-For example, the lastest minor and patch version of major version 1 of the API is served under :query-url:`/v1`.
+For example, the latest minor and patch version of major version 1 of the API is served under :query-url:`/v1`.
 
 An implementation MAY also provide versioned base URLs on the forms :query-url:`/vMAJOR.MINOR` and :query-url:`/vMAJOR.MINOR.PATCH`.
 Here, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the API.
@@ -692,7 +692,7 @@ However, clients MUST accept responses that include other fields that follow the
 The CSV format header line MUST be provided, and the header for the first field MUST be `version`.
 
 The major API versions in the response are to be ordered according to the preference of the API implementation.
-If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the top line of the response (after the optional CSV header).
+If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the top line of the response (after the CSV header).
 
 It is the intent that all future versions of this specification retain this endpoint and the meaning of the first field of the response.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -249,7 +249,9 @@ A URL on the form  :query-url:`/vMAJOR.MINOR` MUST serve the *latest* patch vers
 
 API versions that are published with a suffix, e.g., `-rc<number>` to indicate a release candidate version, SHOULD be served on versioned base URLs without this suffix.
 
-It is the intent that future versions of this standard will not assign different meanings to URLs that begin with :query-url:`/v` and an integer.
+If a request is made to a versioned base URL that begins with `/v` and an integer followed by any other characters, indicating a version that the implementation does not recognize or support, it SHOULD respond with the custom HTTP server error status code :http-error:`553 API Version Not Supported`, perferably along with a user-friendly error message that directs the client to adapt the request to a version it provides.
+
+It is the intent that future versions of this standard will not assign different meanings to URLs that begin with :query-url:`/v` and an integer followed by other characters.
 Hence, a client can safely attempt to access a specific version of the API via the corresponding versioned base URL. 
 Other forms of version negotiation is provided by the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -692,7 +692,7 @@ However, clients MUST accept responses that include other fields that follow the
 The CSV format header line MUST be provided, and the header for the first field MUST be `version`.
 
 The major API versions in the response are to be ordered according to the preference of the API implementation.
-If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the top line of the response (after the CSV header).
+If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the first value in the response (i.e., it MUST be on the second line of the response directly following the required CSV header).
 
 It is the intent that all future versions of this specification retain this endpoint and the meaning of the first field of the response.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -232,10 +232,6 @@ Base URL
 Each database provider will publish one or more **base URLs** that serve the API, for example: http://example.com/optimade/.
 Every URL path segment that follows the base URL MUST behave as standardized in this API specification.
 
-Implementations MUST serve a :endpoint:`versions` endpoint directly under the **base URLs**.
-This endpoint allows a client to discover the major versions of the API that the implementation provides.
-For the specification of this endpoint see section `Versions Endpoint`_.
-
 Versioned base URLs
 ~~~~~~~~~~~~~~~~~~~
 
@@ -248,13 +244,13 @@ An implementation MAY also provide versioned base URLs on the forms :query-url:`
 Here, :val:`MINOR` is the minor version number and :val:`PATCH` is the patch version number of the API.
 A URL on the form  :query-url:`/vMAJOR.MINOR` MUST serve the *latest* patch version supported by the implementation of this minor version.
 
-API versions that are published with a suffix, e.g., `-rc<number>` to indicate a release candidate version, SHOULD be served on versioned base URLs without this suffix.
+API versions that are published with a suffix, e.g., :val:`-rc<number>` to indicate a release candidate version, SHOULD be served on versioned base URLs without this suffix.
 
-If a request is made to a versioned base URL that begins with `/v` and an integer followed by any other characters, indicating a version that the implementation does not recognize or support, the implementation SHOULD respond with the custom HTTP server error status code :http-error:`553 API Version Not Supported`, perferably along with a user-friendly error message that directs the client to adapt the request to a version it provides.
+If a request is made to a versioned base URL that begins with :query-url:`/v` and an integer followed by any other characters, indicating a version that the implementation does not recognize or support, the implementation SHOULD respond with the custom HTTP server error status code :http-error:`553 Version Not Supported`, preferably along with a user-friendly error message that directs the client to adapt the request to a version it provides.
 
 It is the intent that future versions of this standard will not assign different meanings to URLs that begin with :query-url:`/v` and an integer followed by other characters.
 Hence, a client can safely attempt to access a specific version of the API via the corresponding versioned base URL. 
-Other forms of version negotiation are provided by the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
+For other forms of version negotiation, see section `Version Negotiation`_.
 
 Examples of valid versioned base URLs:
 
@@ -272,27 +268,48 @@ Database providers SHOULD strive to implement the latest released version of thi
 Note: The base URLs and versioned base URLs themselves are not considered part of the API, and the standard does not specify the response for a request to them.
 However, it is RECOMMENDED that implementations serve a human-readable HTML document on base URLs and versioned base URLs, which explains that the URL is an OPTIMADE URL meant to be queried by an OPTIMADE client.
   
-Unversioned base URLs
-~~~~~~~~~~~~~~~~~~~~~
+Unversioned base URL
+~~~~~~~~~~~~~~~~~~~~
 
 Implementations MAY also provide access to the API on the **unversioned base URL** as described in this subsection.
 
 Access via the unversioned URL is primarily intended for (i) convenience when manually interacting with the API, and (ii) to provide version agnostic permanent links to resource objects.
 Clients that perform automated processing of responses SHOULD access the API via versioned base URLs.
 
-Implementations MAY provide direct access to the full API under the unversioned base URL. 
-In this case, the implementation SHOULD serve its preferred version of the API (i.e., the lastest, most mature, and stable version.) 
+Implementations serving the API on the unversioned base URL have a few alternative options:
 
-Implementations MAY instead issue an HTTP 307 temporary redirect for queries issued to endpoints under the unversioned base URL.  
-In that case, the redirect SHOULD be to the corresponding versioned URL for the preferred version of the API supported by the implementation.
+1. Direct access may be provided to the full API.
+2. Requests to endpoints under the unversioned base URL may be redirected using an HTTP 307 temporary redirect to the corresponding endpoints under a versioned base URL.
+3. Direct access may be limited to only single entry endpoints (see section `Single Entry Endpoints`_), i.e., so that this form of access is only available for permanent links to resource objects.
 
-As a third alternative, implementations MAY choose to provide direct access only to Single Entry Endpoints (see section `Single Entry Endpoints`_) under the unversioned base URL, i.e., so that this form of access is only available for permanent links to resource objects.
-In that case, the response format SHOULD be that of the preferred version of the API supported by the implementation.
 Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
 
-When an implementation serves (or redirects to) a preferred major version of the API on the unversioned base URL, that version MUST also be the version at the first line in the response of the :endpoint:`versions` endpoint (see `Versions Endpoint`_).
+The client MAY provide a query parameter :query-param:`api_hint` to hint the server about a preferred API version.
+If this parameter is provided, the server MAY serve the request using (or redirect the request to) any version of the API that it finds appropriate, see section `Version Negotiation`_ for more details.
+However, if :query-param:`api_hint` is not provided, the implementation SHOULD serve (or redirect to) its preferred version of the API (i.e., the lastest, most mature, and stable version).
+In this case, that version MUST also be the first version in the response of the :endpoint:`versions` endpoint (see section `Versions Endpoint`_).
 
-    **For implementers**: Before enabling access to the API on unversioned base URLs, implementers are advised to consider that an upgrade of the major version of the API served this way may change the behaviors of associated endpoints in ways that are not backwards compatible.
+    **For implementers**: Before enabling access to the API on unversioned base URLs, implementers are advised to consider that an upgrade of the major version of the API served this way may change the behaviors of associated endpoints in ways that are not backward compatible.
+
+Version Negotiation
+-------------------
+The OPTIMADE API provides three concurrent mechanisms for version negotiation between client and server.
+
+1. The :endpoint:`versions` endpoint served directly under the unversioned base URL allows a client to discover all major API versions supported by a server in the order of preference (see section `Versions Endpoint`_).
+
+2. A client can access the API under versioned base URLs.
+   In this case, the server MUST respond according to the specified version or return an error if the version is not supported (see section `Versioned Base URLs`_).
+  
+3. When accessing the API under the unversioned base URL, a client MAY append the query parameter :query-param:`api_hint` to hint the server about a preferred API version.
+   This parameter is described in more detail below.
+
+The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
+However, for endpoints under a versioned base URL this parameter SHOULD be ignored, and the request served as usual according to the version specified in the URL path segment.
+The client SHOULD set the value on the format :val:`vMAJOR` where MAJOR is a major version of the API.
+For example, if a client appends to the query string :query-string:`api_hint=v1` the hint provided is for major version 1.
+
+The intent of the :query-param:`api_hint` parameter is that if a client hints a version that is not supported by the server, the server can use the value to make a best-effort attempt at still serving the request, e.g., by invoking the closest supported version of the API.
+However, the use of this parameter, if any, is entirely at the discretion of the implementation, and clients MUST NOT expect the response to be served according to the version that is hinted.
 
 Index Meta-Database
 -------------------
@@ -680,6 +697,11 @@ The endpoints are:
 
 These endpoints are documented below.
 
+Query parameters
+----------------
+Query parameters to the endpoints are documented in the respective subsections below.
+However, in addition, all API endpoints MUST accept the :query-param:`api_hint` parameter described under `Version Negotiation`_.
+
 Versions Endpoint
 -----------------
 
@@ -689,7 +711,7 @@ This endpoint is special in that it MUST be provided directly on the unversioned
 The response to a query to this endpoint is in the :RFC:`4180` CSV (`text/csv; header=present`) format (and does therefore not comply with the JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
 In the present version of the API, the response contains only a single field that is used to list the major versions of the API that the implementation supports.
 However, clients MUST accept responses that include other fields that follow the version.
-The CSV format header line MUST be provided, and the header for the first field MUST be `version`.
+The CSV format header line MUST be provided, and the header for the first field MUST be :val:`version`.
 
 The major API versions in the response are to be ordered according to the preference of the API implementation.
 If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the first value in the response (i.e., it MUST be on the second line of the response directly following the required CSV header).
@@ -704,7 +726,7 @@ Example response:
   1
   0
 
-The above response means that the API versions 1 and 0 are served under the versioned base URLs `/v1` and `/v0`, respectively.
+The above response means that the API versions 1 and 0 are served under the versioned base URLs :query-url:`/v1` and :query-url:`/v0`, respectively.
 The order of the versions indicates that the API implementation regards version 1 as preferred over version 0.
 If the API implementation allows access to the API on the unversioned base URL, this access has to be to version 1, since the number 1 appears in the first (non-header) line.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -713,15 +713,18 @@ Versions Endpoint
 The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
 This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
 
-The response to a query to this endpoint is in the :RFC:`4180` CSV (`text/csv; header=present`) format.
+The response to a query to this endpoint is in a restricted subset of the :RFC:`4180` CSV (`text/csv; header=present`) format.
+The restrictions are: (i) field values and header names MUST NOT contain commas, newlines, or double quote characters; (ii) Field values and header names MUST NOT be enclosed by double quotes; (iii) The first line MUST be a header line.
+These restrictions allow clients to parse the file line-by-line, where each line can be split on all occurences of the comma ',' character to obtain the head names and field values.
+
 In the present version of the API, the response contains only a single field that is used to list the major versions of the API that the implementation supports.
+The CSV format header line MUST specify :val:`version` as the name for this field.
 However, clients MUST accept responses that include other fields that follow the version.
-The CSV format header line MUST be provided, and the header for the first field MUST be :val:`version`.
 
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -639,6 +678,34 @@ The URL path segment that follows the versioned base URL MUST represent one of t
...
+Versions Endpoint
+-----------------
+
+The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
+This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
+
+The response to a query to this endpoint is in the :RFC:`4180` CSV format (and does therefore not comply with JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
+In the present version of the API the response contains only a single field, which is used to list the major versions of the API that the implementation supports.
+However, clients MUST accept responses that include other fields that follow the version.
+If the CSV response includes a header line, the name of the first field MUST be "#version", and the client can therefore safely disregard the first line if it starts with the '#' character.
```
sauliusg, line 690: I would make the header line mandatory in all OPTIMADE APIs starting from v1.0, even if it contains just a single line 'version'. In this way simple clients that simply skip the first line and get the first number on the second line will be future-proof.

rartino: Ok, agreed.

rartino:
New version pushed that makes the CSV header mandatory, and which adds a SHOULD requirement for HTTP error handling of missing versioned endpoints; which from the discussions in #277 appear to be very useful for future-proofing of the API.

optimade.rst
```diff
@@ -255,6 +270,29 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+
+As a third alternative, implementations MAY choose to provide direct access only to Single Entry Endpoints (see section `Single Entry Endpoints`_) under the unversioned base URL, i.e., so that this form of access is only available for permanent links to resource objects.
+In that case, the response format SHOULD be that of the preferred version of the API supported by the implementation.
+Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
+
+When an implementation serves (or redirects to) a preferred major version of the API on the unversioned base URL, that version MUST also be the version at the first line in the response of the :endpoint:`versions` endpoint (see `Versions Endpoint`_).
+
+    **For implementers**: Before enabling access to the API on unversioned base URLs, implementers are advised to consider that this feature may encourage careless automated access from clients that, without verification, assumes a specific version of the API to be served on the unversioned base URL.
```
sauliusg, line 294: I would not make assumption that this "feature may encourage careless automated access from clients that, without verification, assumes a specific version of the API". Let's just write that "implementers are advised that the API version served on an unversioned endpoint may change without notice, even to a different *major* API version that may break responsibility"

optimade.rst
```diff
@@ -255,6 +270,29 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+    The implementer needs to be prepared to break such clients on major version upgrades, or will otherwise get stuck at serving the *oldest* API version on the unversioned URL.
```
sauliusg, line 295: Of you serve the *oldest* API on your unversioned URI you will break *new* clients that do not support the oldest version. No way around it.

I would suggest leaving this discussion out of the spec.

optimade.rst
```diff
@@ -232,13 +232,28 @@ Base URL
...
-    **For implementers**: Clients are recommended to discover the highest version supported by both the client and the API implementation by trying versioned base URLs in order of priority.
-    E.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
+Versioned base URLs
+~~~~~~~~~~~~~~~~~~~
+
+Access to the API is primarily provided under **versioned base URLs**.
+An implementation MUST provide the form where a URL path segment :query-url:`/vMAJOR` is appended to the base URL, where :val:`MAJOR` is one of the major version numbers of the API that the implementation supports.
+This URL MUST serve the *latest* minor/patch version supported by the implementation.
+
+An implementation MAY also provide versioned base URLs on the forms :query-url:`/vMAJOR.MINOR` and :query-url:`/vMAJOR.MINOR.PATCH`.
```
sauliusg, line 246: I like these extensions, these are handy for us, may be helpful and are actually implemented in out current server :)

optimade.rst
```diff
@@ -232,13 +232,28 @@ Base URL
...
+This endpoint allows a client to discover the major versions of the API that the implementation provides.
+For the specification of this endpoint see section `Versions Endpoint`_.
 
-    **For implementers**: Clients are recommended to discover the highest version supported by both the client and the API implementation by trying versioned base URLs in order of priority.
-    E.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
+Versioned base URLs
+~~~~~~~~~~~~~~~~~~~
+
+Access to the API is primarily provided under **versioned base URLs**.
+An implementation MUST provide the form where a URL path segment :query-url:`/vMAJOR` is appended to the base URL, where :val:`MAJOR` is one of the major version numbers of the API that the implementation supports.
```
sauliusg, line 243: I suggest putting example of the versioned endpoint to avoid the "metasymbol problem" (in the text above, `/vMAJOR`, `/v` are symbols while `MAJOR` are metasymbols to be replaced by a version number, and this is not self-evident for someone who ix not into the specs). I suggest adding the text:

> An implementation MUST provide the form where a URL path segment :query-url:`/vMAJOR` is appended to the base URL, where :val:`MAJOR` is one of the major version numbers of the API that the implementation supports.

> For example, :query-url:`/v12` if 12 is the served major version of the API.

And similar examples for the `/v1.0` and `/v1.2.3` URLs.

BTW, do we call them URLs or URIs?

optimade.rst
```diff
@@ -255,6 +270,29 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+Access via the unversioned URL is primarily intended for (i) convenience when manually interacting with the API, and (ii) to provide version agnostic permanent links to resource objects.
+Clients that perform automated processing of responses SHOULD access the API via versioned base URLs.
+
+Implementations MAY provide direct access to the full API under the unversioned base URL. 
+In this case, the implementation SHOULD serve its preferred version of the API (i.e., the lastest, most mature, and stable version.) 
+
+Implementations MAY instead issue an HTTP 307 temporary redirect for queries issued to endpoints under the unversioned base URL.  
+In that case, the redirect SHOULD be to the corresponding versioned URL for the preferred version of the API supported by the implementation.
+
+As a third alternative, implementations MAY choose to provide direct access only to Single Entry Endpoints (see section `Single Entry Endpoints`_) under the unversioned base URL, i.e., so that this form of access is only available for permanent links to resource objects.
```
sauliusg, line 288: If this is what you are going to do in your API then let's leave it, but I think this optionality makes the OPTIMADE standard even more complicated and does not bring anything. I would suggest – if you provide unversioned endpoints, then it the same interface as on the versioned ones. All or nothing.

optimade.rst
```diff
@@ -639,6 +680,34 @@ The URL path segment that follows the versioned base URL MUST represent one of t
...
+Versions Endpoint
+-----------------
+
+The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
+This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
+
+The response to a query to this endpoint is in the :RFC:`4180` `text/csv; header=present` format (and does therefore not comply with the JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
+In the present version of the API, the response contains only a single field that is used to list the major versions of the API that the implementation supports.
+However, clients MUST accept responses that include other fields that follow the version.
+The csv format header line MUST be provided, and the header for the first field MUST be `#version`.
```
sauliusg, line 692: We could skip the comment sign '#' but we can also leave it; for CSV it makes no difference IMHO since comments are not supported anyway...

optimade.rst
rartino: Right now we are consistently referring to them as URLs rather than URIs, we have even named fields according to this, e.g., `base_url`. If you can point at a use case where saying URL limits something someone may want to do, we can consider changing this. But it would have to be done consistently throughout the specification, which is not for this PR.

optimade.rst
rartino: Yes, I will implement static resource objects and redirects for other API requests; after quite some consideration this matches very well our needs for publishing and query use. So, I appreciate if it can stay.

optimade.rst
sauliusg: > Right now we are consistently referring to them as URLs rather than URIs, we have even named fields according to this, e.g., `base_url`. If you can point at a use case where saying URL limits something someone may want to do, we can consider changing this. But it would have to be done consistently throughout the specification, which is not for this PR.

OK, let's the leave the URLs if this is so throughout the document.

I tend to think (and to use the terms accordingly) that URL is a _locator_, i.e. names a location (server, page) where certain resource can be currently found; whereas URI is _identifier_, i.e. a (location independent) name of the resource. When the resource is moved from one location to another, URLs may (and probably will) change, but URIs should not. In principle, URIs are resolved to URLs, preferably automatically. So, for the COD we call 'https://www.crystallography.net/cod/2000000.cif' and URI, and a corresponding URL might be 'http://cod.ibt.lt/cod/2000000.cif'. The former is kept stable and used for referencing, the latter may change at any time. URIs are "cool" in a [well-known sense](https://www.w3.org/Provider/Style/URI.html), URLs are not :).

So I would tend to see OPTIMADE endpoints as URIs. But I admit that my use of URIs and URLs is not the "official" one in any sense, the IETF seems undecided what terminology should be used, and also there are URNs which I do not consider at all. In my use, the difference between URI and URL is in the *intention* of their use, not in the technical implementation or syntax.

optimade.rst
merkys, line 692: csv -> CSV (abbreviation; also, "CSV" is used later in the text)

optimade.rst
rartino: Ok, it is fair to tone this "warning" down. I've pushed what I hope is a more agreeable formulation that still says what I think needs to be said to the API implemeter. Lets see if you agree with the new version.

optimade.rst
rartino: Well, [rfc 3986](https://tools.ietf.org/html/rfc3986#section-1.1.3) as well as other sources (e.g., [Wikipedia](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier)) seem pretty clear on that URLs are a subset of URIs. Hence, I indeed find it a non-standard use of the terms to suggest that something can be a URL but not a URI, based on its intended use. 

My interpretation of the discussion and examples in [rfc 3986](https://tools.ietf.org/html/rfc3986#section-1.1.2) is that those URIs that implicitly tell you *where to find something* (locally, over the network, or in some other way) in addition to just being *identifiers* are URLs. On the other hand, if you have an identifier that follows the URI scheme, but really just *names* the resource without telling you where and how to retrieve it, that is a URI but not a URL (but probably a URN). The one clear URN example in rfc3986 is: `urn:oasis:names:specification:docbook:dtd:xml:4.1.2`. 

So, the way I see this is: do we want "those things" to be allowed when we right now in the specification say "URL"? If so, we should change URL to URI. But, probably not?

rartino:
Ok, the PR is updated according to all feedback. Please re-review.

optimade.rst
```diff
@@ -255,6 +271,28 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+In that case, the redirect SHOULD be to the corresponding versioned URL for the preferred version of the API supported by the implementation.
+
+As a third alternative, implementations MAY choose to provide direct access only to Single Entry Endpoints (see section `Single Entry Endpoints`_) under the unversioned base URL, i.e., so that this form of access is only available for permanent links to resource objects.
+In that case, the response format SHOULD be that of the preferred version of the API supported by the implementation.
+Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
+
+When an implementation serves (or redirects to) a preferred major version of the API on the unversioned base URL, that version MUST also be the version at the first line in the response of the :endpoint:`versions` endpoint (see `Versions Endpoint`_).
+
+    **For implementers**: Before enabling access to the API on unversioned base URLs, implementers are advised to consider that an upgrade of the major version of the API served this way may change the behaviors of associated endpoints in ways that are not backwards compatible.
```
sauliusg, line 295: Perfect!

optimade.rst
sauliusg: > Yes, I will implement static resource objects and redirects for other API requests; 

But if you will use redirects for queries then you are basically having 1+2 alternatives (which is just fine if we view redirection as a suggested implementation possibility). So do you still need the "third alternative: under this scenario?

If you do, them let's leave it, but I would be hesitant to endorse non-orthogonality of the interface. It is a pain in the neck to work with existing non-orthogonal REST APIs (will not provide links, but there are many of them :), I think OPTIMADE can do it better!

optimade.rst
```diff
@@ -639,6 +680,34 @@ The URL path segment that follows the versioned base URL MUST represent one of t
...
+
+.. code:: CSV
+
+  version
+  1
+  0
+
+The above response means that the API versions 1 and 0 are served under the versioned base URLs `/v1` and `/v0`, respectively.
+The order of the versions indicates that the API implementation regards version 1 as preferred over version 0.
+If the API implementation allows access to the API on the unversioned base URL, this access has to be to version 1, since the number 1 appears in the first (non-header) line.
```
sauliusg, line 709: Looks good!

optimade.rst
sauliusg: > Hence, I indeed find it a non-standard use of the terms to suggest that something can be a URL but not a URI, based on its intended use.

Well, I do not want to imply this, and I try to make my use is compatible with this definition, although I do not depend on it.

> My interpretation of the discussion and examples in [rfc 3986](https://tools.ietf.org/html/rfc3986#section-1.1.2) is that those URIs that implicitly tell you _where to find something_ (locally, over the network, or in some other way) in addition to just being _identifiers_ are URLs.

True, I agree.

> On the other hand, if you have an identifier that follows the URI scheme, but really just _names_ the resource without telling you where and how to retrieve it, that is a URI but not a URL

But without the ability to find the resource (that is accessible) names are pretty useless, so in practice URI always maps to URL. That's exactly what I meant. One can make a redirection URI->URL.

And, since URI identifies a *resource* regardless to its *location*, URIs will persist even if the resource location changes (unless one rename a resource, which one shouldn't ;) ). So your intepretation with some implication applied from real life situation yields the same situation as I have described.

> (but probably a URN).

Indeed. And since I have not seen any software that reasonably well handles URNs, I tend not to think about them, they are pretty useless in the current situation, aren't they?

sauliusg: > So, the way I see this is: do we want "those things" to be allowed when we right now in the specification say "URL"? If so, we should change URL to URI. But, probably not?

In practice all this URI/URL/URN distinction is somewhat moot so I am fine with leaving URLs in the specs (too much work to change it for too little benefit...) If I were writing it from scratch I would probably use URI as a term but for now lets leave it as is.

I think this resolves this particular question.

optimade.rst
sauliusg: For some reason GitHub still shows the old text here, but I've finally found the new version and it sounds perfectly OK there. So if that other text is adopted (https://github.com/Materials-Consortia/OPTIMADE/pull/290/commits/4d1811ce8a15eb9166d8a8eb64babbbb2d0123e0) I find this issue resolved.

optimade.rst
```diff
@@ -232,13 +232,29 @@ Base URL
...
 
-    **For implementers**: Clients are recommended to discover the highest version supported by both the client and the API implementation by trying versioned base URLs in order of priority.
-    E.g., if major version 2 and lower are supported by the client, it would try: :query-url:`/v2`, :query-url:`/v1`, and then :query-url:`/v0`.
+Versioned base URLs
+~~~~~~~~~~~~~~~~~~~
+
+Access to the API is primarily provided under **versioned base URLs**.
+An implementation MUST provide the form where a URL path segment :query-url:`/vMAJOR` is appended to the base URL, where :val:`MAJOR` is one of the major version numbers of the API that the implementation supports.
+This URL MUST serve the *latest* minor/patch version supported by the implementation.
+For example, the lastest minor and patch version of major version 1 of the API is served under :query-url:`/v1`.
```
giovannipizzi, line 245:
```suggestion
For example, the latest minor and patch version of major version 1 of the API is served under :query-url:`/v1`.
```

optimade.rst
```diff
@@ -639,6 +680,34 @@ The URL path segment that follows the versioned base URL MUST represent one of t
 
 These endpoints are documented below.
 
+Versions Endpoint
+-----------------
+
+The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
+This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
```
giovannipizzi, line 687: Would it be ok to say SHOULD NOT in the second one, for those implementers that have issues in removing it?

optimade.rst
```diff
@@ -639,6 +680,34 @@ The URL path segment that follows the versioned base URL MUST represent one of t
...
+
+The response to a query to this endpoint is in the :RFC:`4180` CSV (`text/csv; header=present`) format (and does therefore not comply with the JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
+In the present version of the API, the response contains only a single field that is used to list the major versions of the API that the implementation supports.
+However, clients MUST accept responses that include other fields that follow the version.
+The CSV format header line MUST be provided, and the header for the first field MUST be `version`.
+
+The major API versions in the response are to be ordered according to the preference of the API implementation.
+If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the top line of the response (after the optional CSV header).
```
giovannipizzi, line 695:
```suggestion
If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the top line of the response (after the CSV header).
```
(it's not optional anymore, right?)

optimade.rst
sauliusg: IMHO 'SHOULD NOT' would be fine, we are not breaking anything if we still have the same versions mapped under the versioned route.

But I would appreciate @rartino comment on this, maybe he will see some dangerous interaction.

optimade.rst
sauliusg: Right, I suggest committing it, or even strengthening to 'CSV header required by OPTIMADE', since in the CSV proper the header is indeed optional.

optimade.rst
rartino: My motivation to put MUST here was that the existence or not of `/versions` then becomes an indication of whether a URL is an unversioned or versioned OPTIMADE base URL, which can otherwise be tricky to determine if you are just given a URL. This would simplify scripts that try to do some simple validation on that they are given the right kind of URL, preventing the easy user mistake of giving a versioned URL instead of the unversioned one. It also becomes very easy to clearly communicate what URL is expected, e.g., "This should be the URL under which you serve `/versions`". 

Do you really think there will be implementations that have difficulty to not serve `/versions` under versioned URLs?

optimade.rst
rartino: Oops, good catch.

optimade.rst
giovannipizzi: I see your point. The only concern is on how to allow for a change that is not backward-incompatible.

Changing later to `SHOULD NOT` might break clients that assume that `/versions` doesn't exist. Changing later from `SHOULD NOT` to `MUST` is less risky (we just need to check that all existing servers can do it)?

Anyway, in order not to block the release: I'll accept this as is, but I suggest that, in the 2-week span, we emphasise to implementers the new features, and ask explicitly if some of the latest changes could be problematic for them.

sauliusg: > Changing later to `SHOULD NOT` might break clients that assume that `/versions` doesn't exist. Changing later from `SHOULD NOT` to `MUST` is less risky (we just need to check that all existing servers can do it)?

Isn't it more likely that the clients will break that assumed the endpoint *exists* when it stops to do so?

When an endpoint is not supposed to exist, you do not depend on it, you simple do not query it. When it appears, you do not notice.

When an endpoint exists, you query it and assume an error if it does not. In this case removing existing endpoint is more breaking than adding one.

In this sense, as it is currently written, "MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs" is safer than permitting it everywhere.

rartino: I agree with @giovannipizzi that, formally in this case, SHOULD NOT -> MUST NOT is not a backward breaking change, whereas MUST NOT -> SHOULD NOT, is. Just as I intended to, clients may rely on that the endpoint is NOT there to identify a versioned base URL, and this breaks when the mandate is changed to SHOULD NOT.

That said, if we don't put MUST NOT now, we can just as well forget about the whole thing and stick with MAY. This is useless as an easy detection mechanism for versioned vs unversioned URLs if it doesn't work on servers that remain on v1.0.

rartino:
@sauliusg, @merkys, @giovannipizzi Quite large changes pushed to also introduce the `api_hint` parameter we've in #277 concluded is useful if, and only if, introduced now in v1.0.

optimade.rst
sauliusg: I have overlooked previously you intention to detect versioned vs. unversioned base URLs, which is an important functionality of we want clients to navigate fully automatically.

By now you have convince me to leave the original formulation "MUST be provided directly on the unversioned base URL at :query-url:/versions and MUST NOT be provided under the versioned base URLs" is necessary for this to work. Since `/versions` can be served form a static file, hopefully this feature will not be too difficult to implement by everybody, and since @giovannipizzi has approved, as I understand, this version of the text, he does not see any obstacles for implementation either. So we go with "MUST"+"MUST NOT"? I'm fine with the text.

optimade.rst
```diff
@@ -272,27 +268,48 @@ Database providers SHOULD strive to implement the latest released version of thi
...
 
-As a third alternative, implementations MAY choose to provide direct access only to Single Entry Endpoints (see section `Single Entry Endpoints`_) under the unversioned base URL, i.e., so that this form of access is only available for permanent links to resource objects.
-In that case, the response format SHOULD be that of the preferred version of the API supported by the implementation.
 Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
 
-When an implementation serves (or redirects to) a preferred major version of the API on the unversioned base URL, that version MUST also be the version at the first line in the response of the :endpoint:`versions` endpoint (see `Versions Endpoint`_).
+The client MAY provide a query parameter :query-param:`api_hint` to hint the server about a preferred API version.
+If this parameter is provided, the server MAY serve the request using (or redirect the request to) any version of the API that it finds appropriate, see section `Version Negotiation`_ for more details.
```
sauliusg, line 288: After reading `Version Negotiation` I see that the policy is explained there, but reading the current text the policy and the purpose is not evident. Also, the MAY here and the SHOULD level in `Version Negotiation` seem confusing (if not contradictory), and extra indirection makes it more difficult to perceive the text.

Maybe we state the requirement as follows:

> If this parameter is provided, the server SHOULD serve the request using (or redirect the request to) the "best suitable" version of the API that is selected as described in the section `Version Negotiation`.

In my understanding, MAY implies full freedom for implementer's fantasy, while SHOULD with a link to the section tells that you should read this section before taking decisions on your implementation. The insert "the "best suitable"" hints to the purpose of this provision without to much risk of becoming inconsistent when `Version Negotiation` is amended.

sauliusg: From my perspective, the PR is ready for green light when we resolve this point.

rartino: But, isn't processing of the `api_hit` meant to be completely optional for the implementation? Your SHOULD formulation suggests it isn't, and all well-working implementations should process it unless they have good reasons not to. Do you mean to say that?

I'm not sure what SHOULD-level requirement you refer to in Version Negotiation? The only relevant SHOULD I see there was really only meant to mandate the format of the parameter when it is set. I think I should clarify that with an additional "**When provided**, the client SHOULD set the value on the format ..."

rartino: Ok, I've just pushed a reformulation that I think side-steps the issue by using your formulation but without any rfc2119-terms:

------

The client MAY provide a query parameter `api_hint` to hint the server about a preferred API version. When this parameter is provided, the request is to be handled as described in section `Version Negotiation`, which allows a "best suitable" version of the API to be selected to serve the request (or forward the request to). However, if `api_hint` is not provided, the implementation SHOULD serve (or redirect to) its preferred version of the API (i.e., the lastest, most mature, and stable version). In this case, that version MUST also be the first version in the response of the `versions` endpoint (see section `Versions Endpoint`).

------

Does this seem ok?

sauliusg: > But, isn't processing of the `api_hit` meant to be completely optional for the implementation? Your SHOULD formulation suggests it isn't, and all well-working implementations should process it unless they have good reasons not to. Do you mean to say that?

My understanding of SHOULD level in RFC 2119 is that if you do not do not follow a SHOULD recommendation, you are still compliant with the specification but you should carefully weight the consequences of this design. In other words, the default, if you do not have special reasons not to do what SHOULD mandates, you implement this.

For the `api_hint` to be maximally useful, I think the right level of specs is that you SHOULD provide the requested version (if you can); and you SHOULD provide the *closest* matching version if you do not serve the same version as requested. In other words, by default one would implement the server as provided in the two SHOULD clauses above, *unless* there is a serious reason to deviate from the spec (lack of manpower to implement this part of the spec is of course a valid reason :) ).

This is different from MAY where every implemented decides essentially what they wish; for example they may choose to always server v1, or allways the latest version (essentially ignoring `api_hint`). While such behaviour is compliant with the spec both in case of 'MAY' or 'SHOULD' level, I thing 'SHOULD' matches closer the intended purpose of the `api_hint`, namely – to maintain maximum possible compatibility.

sauliusg: > The client MAY provide a query parameter `api_hint` to hint the server about a preferred API version. When this parameter is provided, the request is to be handled as described in section `Version Negotiation`, which allows a "best suitable" version of the API to be selected to serve the request (or forward the request to). However, if `api_hint` is not provided, the implementation SHOULD serve (or redirect to) its preferred version of the API (i.e., the lastest, most mature, and stable version). In this case, that version MUST also be the first version in the response of the `versions` endpoint (see section `Versions Endpoint`).

I must admit that I am not happy with the combined text of the current paragraph + 'Version Negotiation' section. There are a lot of discussions how `api_hint` can be ignored and what to do if it is not supported, but does not describe the main purpose of the flag, namely:

IF you *can* support version requested by `api_hint`, THEN you DO serve this version; ELSE serve the closest matching one END IF;

Saying that the server selects the "best suitable" version must be met by the explanation of what is "best suitable", and stress that it is "best suitable" for the *client* to process, not for the server to serve or some other purposes.

I am not sure when the 'Version Negotiation' text was introduced, but apparently it is being changed in this PR as well, so I thing we should add clarification with a SHOULD complicance level.

sauliusg: > I'm not sure what SHOULD-level requirement you refer to in Version Negotiation? The only relevant SHOULD I see there was really only meant to mandate the format of the parameter when it is set. I think I should clarify that with an additional "**When provided**, the client SHOULD set the value on the format ..."

Yes, I probably misread the text. I was assuming something along the lines I proposed in the previous comment is there, but apparently it is not.

optimade.rst
```diff
@@ -305,7 +305,7 @@ The OPTIMADE API provides three concurrent mechanisms for version negotiation be
 
 The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
 However, for endpoints under a versioned base URL this parameter SHOULD be ignored, and the request served as usual according to the version specified in the URL path segment.
-The client SHOULD set the value on the format :val:`vMAJOR` where MAJOR is a major version of the API.
+When a client provides the parameter, the value SHOULD be on the format :val:`vMAJOR` where MAJOR is a major version of the API.
```
sauliusg, line 308: Maybe better '... the value SHOULD have the format ...'? The phrase 'SHOULD be on the format' sounds very unusual to me...

optimade.rst
sauliusg: > I must admit that I am not happy with the combined text of the current paragraph + 'Version Negotiation' section. There are a lot of discussions how `api_hint` can be ignored and what to do if it is not supported, but does not describe the main purpose of the flag, namely:
> 
> IF the server _can_ support version requested by `api_hint`, THEN it DOES serve this version; ELSE it serves the closest matching version (not necessary the default!) END IF;

@rartino : if you are not against, I'll try to insert the suggested wording into the 'Version Negotiation' section.

rartino: Ok, please take a look at the latest version and see if you are ok with how this is handled now I am perfectly happy iterating a few times more on this if we are not yet on the right formulations.

I adopted your "algorithm" for version selection as follows: if the hinted version is supported, that version SHOULD be used. Beyond that, however, I put the requirement on MAY-level since it essentially represents an ad-hoc algorithm, and IMO it is confusing to mandate ad-hoc algorithms to be implemented beyond the MAY level.

Since you ask, the `Version Negotiation` section is new now with `api_hint`, since with its introduction it became clear that we need a place to clarify the interaction between the multiple version-related features.

rartino:
@sauliusg This is ready to be re-reviewed.

Other things drew my attention away from this for a few days. However, one thing I started to think about is: we have now very carefully considered backward-breaking upgrades of the major version. But, should we also take a quick look at upgrades of the *minor* version to make sure we haven't missed anything there? 

I see no issues for a single specific database. It will only get "more features" at point release upgrades. So all types of perma-links etc. should be fine. (As long as they are against an endpoint under the versioned base URL for the major version, e.g. `/v1/...` or under the unversioned base URL. People should, however, generally avoid storing links under versioned URLs for specific minor versions, e.g. `/v1.1/...`, since those should be expected to be dropped, and the specification then mandates an *error* in response.)

However, it gets a bit more tricky for a client sending one query to multiple databases, when we have a new `v1.1`, which, e.g., standardizes a few new `/structure` fields, adds a `/properties` endpoint, and maybe adds another feature to the filtering language? Let's say, exmpl1 only supports `v1.0` and exmpl2 supports `v1.1`.

According to our negotiation protocol, a careful client trying to send one query to both these databases would do this, right?:

1. Deduce the required subversion of the query; is it a `v1.1`-type query, or can it be represented under `v1.0`?
2. Go to each database base URL and request `v1/info`. 
3. Compare the version of the query with the reported version of the implementation.
4. Back out, if the supported minor version < the required minor version for the query.
5. Otherwise submit the query and fetch the results.

Alternatively, I suppose we could try to devise some scheme that makes it possible for a `v1.0` server to know, without any version hint from the client, that a given query would be servable by a newer version than is supported (as opposed to, e.g., a syntax error.) so that it can respond with a detectable error for this. This seems difficult.

As a third alternative, would maybe allowing also minor versions in `api_hint` be a good idea? Then the server can do the version comparison and decide if it can serve the query. But it overloads that parameter a bit.

optimade.rst
```diff
@@ -639,6 +702,39 @@ The URL path segment that follows the versioned base URL MUST represent one of t
...
+Query parameters to the endpoints are documented in the respective subsections below.
+However, in addition, all API endpoints MUST accept the :query-param:`api_hint` parameter described under `Version Negotiation`_.
+
+Versions Endpoint
+-----------------
+
+The :endpoint:`versions` endpoint aims at providing a stable and future-proof way for a client to discover the major versions of the API that the implementation provides. 
+This endpoint is special in that it MUST be provided directly on the unversioned base URL at :query-url:`/versions` and MUST NOT be provided under the versioned base URLs.
+
+The response to a query to this endpoint is in the :RFC:`4180` CSV (`text/csv; header=present`) format (and does therefore not comply with the JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification).
```
sauliusg, line 716: I suggest dropping the braced statement '(and does therefore not comply with the JSON:API data format of the `JSON API v1.0 <http://jsonapi.org/format/1.0>`__ specification)' :

- It is evident for those who read the JSON API 1.0 (the response is not JSON anyway), so coveys 0 bits of information;

- we are already not following JSON API 1.0 exactly, and there is no such thing as "90% compatibility". So I suggest not pretending that we somehow are bound by JSON API 1.0.

- For the clients to be able to reuse existing JSON API 1.0 libraries, we could indicate that "the specified OPTIMADE response format for this **particular** endpoint follows JSON API 1.0 response specification (ref...). Thus clients should be able to reuse libraries intended for the JSON API 1.0 processing". Unless we strictly specify in the OPTIMADE "the OPTIMADE endpoint response MUST follow the syntax specified in the JSON API 1.0 response specification (ref...).", and *skip* the description of the response altogether, the reference to JSON API are no more than implementation hints, IMHO.

optimade.rst
sauliusg: > I adopted your "algorithm" for version selection as follows: if the hinted version is supported, that version SHOULD be used. Beyond that, however, I put the requirement on MAY-level since it essentially represents an ad-hoc algorithm, and IMO it is confusing to mandate ad-hoc algorithms to be implemented beyond the MAY level.
> 
> Since you ask, the `Version Negotiation` section is new now with `api_hint`, since with its introduction it became clear that we need a place to clarify the interaction between the multiple version-related features.

To me, the Version Negotiation section now reads great, and I agree with the MAY level in the second sentence. Sound perfect!

sauliusg: > Alternatively, I suppose we could try to devise some scheme that makes it possible for a `v1.0` server to know, without any version hint from the client, that a given query would be servable by a newer version than is supported (as opposed to, e.g., a syntax error.) so that it can respond with a detectable error for this. This seems difficult.

I agree, this is difficult do make in general and probably not worth the trouble. At he version 'v1.0' you do not know what can be added as an extension for 'v1.1', so you can only say that the request is not compatible with your current version...

> As a third alternative, would maybe allowing also minor versions in `api_hint` be a good idea? Then the server can do the version comparison and decide if it can serve the query. But it overloads that parameter a bit.

I general, I tend to think about the version number specs as follows: if I specify that I want 'v1', then any minor and any patch-level (p.l.) of the 'v1' should do; if I specify 'v1.1', then any minor equal to or *higher* than the specified is acceptable, e.g. server can use 'v1.1', 'v1.2' or 'v1.110', with any patch-level, to process the query, and it should work. If Y specify a patch-level ('v1.1.20'), them any *higher* or *equal* p.l. or any higher minor with any p.l. should work, e.g. 'v1.1.21', 'v1.2.0' and so on.

If this is the viewpoint then allowing `api_hint=v1.1` and even `api_hint=v1.1.11` seems like a natural completion of what we have specified right now. Hopefully this is a closed set of functions and no "creeping featurism" will happen :).

However, I am thinking about the different use scenarios for the `version_hint=v...`.

For the "cited published URL" scenario, since there the `api_hint=v1` would presumably be provided for an existing v1 implementation (say v1.2 at the moment of publishing), and the minor version of the API in the given implementation is only expected to increase (indeed, why should we switch from the v1.2 back to v1.0?), the unversioned endpoint should remain compatible with all previously issued historic requests that specify `api_hind=v1`, provided the major version v1 is supported.

However, if we copy the query and try to send it to another OPTIMADE database (which we might want to do some time after the publication, why not?), then it can happen that the supported minor OPTIMADE version on the other database will be lower than the version used in the query (and in the original cited URL), and in that case the query can break.

But in that case:

- If the server only supports the lower minor version (v1.0), and the request needs a higher one (v1.1), then the query will result in an error anyway;

- It is hard, in advance, to say if a query needs v1.1 or it can be served with v1.0; the easiest way to find out is just to try the query. 

So the minor version specification of the `apo_hint` QS parameter is only necessary if:

1. we expect the server to *downgrade*   the served minor version for a long time (not temporarily!). Is this an often/realistic scenario?

2. we expect the same unversioned endpoint query to be sent to other OPTIMADE servers that can happen to serve a lower minor revision. This will only help if the server a) supports that higher minor version (e.g. v.1.1) and b) does *not* serve it on the unversioned endpoint with `api_hint=v1` by default. Is this scenario likely/useful?

Of course having `api_hint=v1.1` would provide clients, servers and users with additional information which might help to solve compatibility problems. Thus, the `api_hind=v1` query, even if serviced by 'v1' endpoint, may break if the query tacitly implies 'v1.1' and the server only serves 'v1.0', and there is no direct way to find this out except for a knowledgeable intelligence to look at the query and recognize that it will not work under `v1.0` (and then there may be yet other reasons to fail, in addition to the minor version incomaptibility). Adding explicit `api_hind=v1.1` would make such things easier to decide.

But the people/clients who will use and quote queries for 'v1.1' will probably not examine if each query they make can be served by a lower version ('v1.0', and with time there will be more lower versions, e.g. a query for 'v1.10' might be OK for 'v1.9', 'v1.8' and 'v1.7', but not for 'v1.6' and lower); that is too much work and to complicated to do, with too little benefit. So at the end we will have a lot of queries tagged 'v1.1' or 'v1.10' just because they happened to work with these versions, and this will be overcautious – the same queries might be perfectly OK for 'v1.0', just nobody cared to check...

So in the end we need to consider how important the "server downgrade" scenario is?

sauliusg:
> However, it gets a bit more tricky for a client sending one query to multiple databases, when we have a new `v1.1`, which, e.g., standardizes a few new `/structure` fields, adds a `/properties` endpoint, and maybe adds another feature to the filtering language? Let's say, exmpl1 only supports `v1.0` and exmpl2 supports `v1.1`.
> 
> According to our negotiation protocol, a careful client trying to send one query to both these databases would do this, right?:

Indeed, you algorithm is extra-cautious and should work, but I expect implementers to take some shortcuts (I would...):

>     1. Deduce the required subversion of the query; is it a `v1.1`-type query, or can it be represented under `v1.0`?

This is tricky to decide in general; the only sure way is to try to service the query under `v1.0` and see if it works (correctly...). For a client to do this it will have to contain at least partial server implementation, which seems to be an overkill. So most probably the client will just send the query to the server and (hopefully!) correctly process any error responses it gets.

>     2. Go to each database base URL and request `v1/info`.

True, except that this adds extra query for every DB. Of course the response can (should?) be cached.

>     3. Compare the version of the query with the reported version of the implementation.
> 
>     4. Back out, if the supported minor version < the required minor version for the query.

Since we have troubles in the client to determine if the query can be serviced by a lower minor (see comment on p.1), and since in general supporting old versions is a tedious work that does not pay well, I suspect that the more realistic scenario would be the following:

4'. IF the supported minor version < the required minor _of_ _the_ _client_, THEN issue a warning that the minor version is not what the client in general needs;

>     5. Otherwise submit the query and fetch the results.

5'. Submit the query regardless of the 4' check (the user has been warned!). If the query succeeds, return the result (caveat emptor); if the server returns an error, we report that error (and we can do no better anyway). The client can add additional explanation that the error may be caused by the minor version incompatibility.

optimade.rst
```diff
@@ -255,7 +267,55 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+
+If the server supports the version indicated by the :query-param:`api_hint` parameter, it SHOULD serve the request using this version.
+If the server does not support the version hinted, it MAY use the provided value to make a best-effort attempt at still serving the request, e.g., by invoking the closest supported version of the API.
+If the hinted version is not supported by the server and the request is not served using an alternative version, the server SHOULD respond with the custom HTTP server error status code :http-error:`553 Version Not Supported`.
+Note that the above protocol means that clients MUST NOT expect a returned response is served according to the version that is hinted.
+
+    **For end users**: Users are strongly encouraged to include the :query-param:`api_hint` query parameter for URLs in, e.g., journal publications for queries on endpoints under the unversioned base URL.
+    The version hint will make it possible to serve such queries in a reasonable way even after the server changes the major API version used for requests without version hints.
+
```
sauliusg, line 318: I am now happy with the text an would approve the PR.

If we decide not to include minor version into the `api_hint=v1` string, I think we are done for v1.0 in this section.

If we want to allow specifying minor versions, we should discuss and update the text accordingly.

I think we can do this extention for later 'v1.x' update, since it will be backwards compatible, no?

sauliusg:
> I see no issues for a single specific database. It will only get "more features" at point release upgrades. So all types of perma-links etc. should be fine. (As long as they are against an endpoint under the versioned base URL for the major version, e.g. `/v1/...` or under the unversioned base URL. People should, however, generally avoid storing links under versioned URLs for specific minor versions, e.g. `/v1.1/...`, since those should be expected to be dropped, and the specification then mandates an _error_ in response.)

True, I think we should be very explicit that minor version endpoints are mostly for development, testing and migration, and once the new minor is fully implemented and tested it should migrate to the `vMAJOR` endpoint.

> However, it gets a bit more tricky for a client sending one query to multiple databases, when we have a new `v1.1`, which, e.g., standardizes a few new `/structure` fields, adds a `/properties` endpoint, and maybe adds another feature to the filtering language? Let's say, exmpl1 only supports `v1.0` and exmpl2 supports `v1.1`.

Hmmm... indeed, this is a problem. With the current spec. we will end up in:

- different servers serving different minors (`v1.0`, `v1.1`, `v1.2`);

- some queries, that require `v1.1` features, will be processed with an error if sent to a `v1.0` server...

One can mitigate these spurious errors by specifying a desired API  version in the client; e.g. my hypothetical CLI client would have and option

```
optimade-query --filter='nelements=2 AND elements HAS "Si" AND featureX = "kva" --api=v1.1'
```

and the client would only send this query to servers supporting OPTIMADE v1.1 or higher. This is completely client specific and does not need to be specified in the OPTIMADE spec (maybe only mentioned as an implementation hint).

Are we OK with such scenario? That is, a query may fail because of minor version incompatibility, in addition to other reasons, and it is up to the client to select appropriate servers using `/versions` and `/info`.

If yes, then I think the current spec is OK.

optimade.rst
rartino: OK, let's drop the braced note here. 

However, my primary motivation to not have a problem with this is that in a recent re-read of the JSON API spec I noticed that the MUST requirement is actually formulated as follows (emphasis mine): *"Servers MUST send **all JSON:API data** in response documents with the header Content-Type: application/vnd.api+json without any media type parameters."* 

Output in CSV format isn't  "JSON:API data". Nor is, e.g., HDF5 if that is selected with `response_format`, so we could drop the small warning we now have about that as well.

With this interpretation, I think normal OPTIMADE implementations are, in fact, 100% compatible with JSON API v1.0. Index meta-dbs are allowed not to be, because of the dropped requirement on handling "Accept"-headers and the relaxation on "Content-Type" (the latter could **possibly** be argued to not break the spec, but I see no way around the former.)

sauliusg: > OK, let's drop the braced note here.

Agreed. I'll remove it on GitHub and approve the PR.

What remains is to see if we really want minor revisions to be specified for `api_hint` and if we do, do we need it for OPTIMADE v1.0? Your ideas on my comments are highly appreciated.

sauliusg: TL;DR on my part but still adding a short comment...

> However, my primary motivation to not have a problem with this is that in a recent re-read of the JSON API spec I noticed that the MUST requirement is actually formulated as follows (emphasis mine): _"Servers MUST send **all JSON:API data** in response documents with the header Content-Type: application/vnd.api+json without any media type parameters."_
> 
> Output in CSV format isn't "JSON:API data". Nor is, e.g., HDF5 if that is selected with `response_format`, so we could drop the small warning we now have about that as well.

Sure.

> With this interpretation, I think normal OPTIMADE implementations are, in fact, 100% compatible with JSON API v1.0. Index meta-dbs are allowed not to be, because of the dropped requirement on handling "Accept"-headers and the relaxation on "Content-Type" (the latter could **possibly** be argued to not break the spec, but I see no way around the former.)

I am not sure what the guys meant writing "all JSON:API data", but in any case `/versions` does not return JSON, and you can not reuse any JSON (API) tools for dealing with it. So what?

My problem with JSON API is that they conflate API and JSON, which I want to see strictly separated, since they are orthogonal concerns. API, in my view, should specify endpoints (for a REST), QS parameters and *interpretation*  of data items in an abstract sense (i.e. what does the data name "lattice_vectors" mean, what is it physical interpretation, units of measurement, etc.). JSON is a *carrier* format, one of many; API can and should be able to switch freely between JSON, XML, ASN.1, CSV, HDF5, CIF ... – you name it. We should be able to change the carrier format without changing the API specs (or with a minor addition that "the response MAY be served in (XML|CSV|CIF|etc.)"). OPTIMADE can do this, which in my view is great; I am not sure of the "JSON API" is flexible enough for this or of they insist on JSON only. If that is the case, I have little appreciation for the JSON API.

optimade.rst
```diff
@@ -692,7 +692,7 @@ However, clients MUST accept responses that include other fields that follow the
 The CSV format header line MUST be provided, and the header for the first field MUST be `version`.
 
 The major API versions in the response are to be ordered according to the preference of the API implementation.
-If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the top line of the response (after the CSV header).
+If a version of the API is served on the unversioned base URL as described in the section `Base URL`_, that version MUST be the first value in the response (i.e., it MUST be on the second line of the response directly following the required CSV header).
```
sauliusg, line 722: One last comment: CSV, in general, is somewhat more complicated than a comma-separated-tokens – the values may be in quotes, and quoted values may contain new line characters. While this makes CSV flexible enough to accommodate arbitrary data, it also makes it more difficult to parse than simply `split(",", $csv_line)`.

Currently, OPTIMADE does not need a full power of CSV. So the response of `/versions` *can* be parsed with the aforementioned `split(",",$_)`. But in future we may want to extend the `/versions` content, and we do not know what we will need to put in there.

We have two options:

1. Promise that OPTIMADE will not include new-lines into CSV fields, allowing `split(",",$_)` processing and not requiring any specialised CSV library;

2. Stating that OPTIMADE is using CSV, implying that compliant clients MUST use full-featured CSV parser just for the sake of compatibility with hypothetical future extensions.

(1) is simpler, but (2) is more future-proof. Which one we take?

Probably (2) is safer in a long run and more standard.

rartino: Very good point! I had not considered all the possibilities within RFC4180.

My vote is for limiting the CSV format so that an ad-hoc parser can be trivially implemented. The point of this was that the format should be simple. If we get to the point of having to invoke a parser library, I much rather go for JSON, since our implementations need such a parser already.

So, I would go with:
- No newlines, nor any commas, in fields
- No double quotes will appear in the file at all (neither within fields, nor to quote fields)

With that, `split(",",$_)` should be safe.

sauliusg: > Very good point! I had not considered all the possibilities within RFC4180.
> 
> My vote is for limiting the CSV format so that an ad-hoc parser can be trivially implemented. The point of this was that the format should be simple. If we get to the point of having to invoke a parser library, I much rather go for JSON, since our implementations need such a parser already.
> 
> So, I would go with:
> 
>     * No newlines, nor any commas, in fields
> 
>     * No double quotes will appear in the file at all (neither within fields, nor to quote fields)
> 
> 
> With that, `split(",",$_)` should be safe.

I am fine with both decisions (both are compromises).

One last thought: if we now mandate full CSV and restrict servers in the future major release, this will be still compatible with older clients.

If, however, we now restrict and will have to expand to full CSV in the future, this will potentially break older clients that do not expect this extension, and make compatibility over major version changes more tricky.

Also, even in we allow full CSV, getting the *main* supported version is possible AFAIU with a "trivial" parser, viz.:

`curl https://example.com/optimade/versions | head -n 2 | tail -n 1 | awk -F, '{print $1}'`

will always yield the **main** supported major version, no matter how complicated further CSV is.

So maybe full CSV is the safest version, still? (My colleagues in COD have recently persuaded me that sticking to actual CSV standard with no further embellishments is a good thing...)

rartino: > One last thought: if we now mandate full CSV and restrict servers in the future major release, this will be still compatible with older clients.

I don't like this solution, because it means that I have to put a complete parser of the full CSV specification for my `v1.0` client implementation right now, and that goes against the whole idea of a *trivially simple* format for `/versions`. (If I don't, future servers may break its version resolution, which was the whole point to avoid with this `/versions` setup.)

Your suggestion for a "trivial" parser seem to break on valid CSV files: e.g., if a quoted header field contains a newline (which is valid according to the ABNF grammar in RFC4180), or if the version number is double-quoted. I think this shows why ad-hoc parsing of the full CSV format isn't a great idea.

And why do we want this? Can anyone present a scenario where not allowing for the full CSV format in this "meant-to-be-trivial" versions format can give issues later? I *really* cannot see this. In the *extremely unlikely* event that we find ourselves regretting the decision of not allowing newlines, commas and quotation marks in the fields, there is always the possibility to add a post-parser unquoting mechanism that works with both standard and ad-hoc parsers: e.g. \44 for comma, \10 for newline, \34 for a double quote, and \92 for backslash. Sure, it isn't a super-elegant solution to add a home-brewed quoting mechanism to standard CSV, but it would only be needed in the *super unlikely event* that we feel urged to add support for such values, and I fail to see a possible rationale for that.

My order of preference is:
1. Standardize on super-simplified CSV: clients are guaranteed no commas, no newlines, no double quotes within fields. Fields are never double-quoted. Header can be assumed to always be present ⇒ split-comma-parsing works without issues. **← strongly preferred by me**
2. Standardize on a JSON list format, where the first element in each list is the integer version number. Everything else in the list should be ignored. This allows for the same freedom as the full CSV standard and does not require us to add new parsers to our implementations.
3. Full CSV format.

sauliusg: > My order of preference is:
> 
>     1. Standardize on super-simplified CSV: clients are guaranteed no commas, no newlines, no double quotes within fields. Fields are never double-quoted. Header can be assumed to always be present ⇒ split-comma-parsing works without issues. **← strongly preferred by me**
> 
>     2. Standardize on a JSON list format, where the first element in each list is the integer version number. Everything else in the list should be ignored. This allows for the same freedom as the full CSV standard and does not require us to add new parsers to our implementations.
> 
>     3. Full CSV format.

OK, let's go for 1!

My order of preference would be 3, 1, 2, but 3 $\approx$ 1, so I easily agree on 1 :)

sauliusg: > And why do we want this? Can anyone present a scenario where not allowing for the full CSV format in this "meant-to-be-trivial" versions format can give issues later? I _really_ cannot see this. In the _extremely unlikely_ event that we find ourselves regretting the decision of not allowing newlines, commas and quotation marks in the fields, there is always the possibility to add a post-parser unquoting mechanism that works with both standard and ad-hoc parsers: e.g. \44 for comma, \10 for newline, \34 for a double quote, and \92 for backslash. Sure, it isn't a super-elegant solution to add a home-brewed quoting mechanism to standard CSV, but it would only be needed in the _super unlikely event_ that we feel urged to add support for such values, and I fail to see a possible rationale for that.

I agree. The need of extension at the moment seems extremely unlikely.

rartino: > Will you insert the text into the "versions endpoint"?

Yes, I mean do the edits for this later this evening, along with allowing subversions in `api_hint`.

rartino:
> With the current spec. we will end up in:
> 
> * different servers serving different minors (`v1.0`, `v1.1`, `v1.2`);
> * some queries, that require `v1.1` features, will be processed with an error if sent to a `v1.0` server...
> 
> One can mitigate these spurious errors by specifying a desired API version in the client; e.g. my hypothetical CLI client would have and option
> 
> ```
> optimade-query --filter='nelements=2 AND elements HAS "Si" AND featureX="kva" --api=v1.1'
> ```
> 
> and the client would only send this query to servers supporting OPTIMADE v1.1 or higher, the ones I know support the `featureX`. This is completely client specific and does not need to be specified in the OPTIMADE spec (maybe only mentioned as an implementation hint).
> 
> Are we OK with such scenario? That is, a query may fail because of minor version incompatibility, in addition to other reasons, and it is up to the client to select appropriate servers using `/versions` and `/info`.
> 
> If yes, then I think the current spec is OK.

Based on this discussion, I think allowing minor versions in `api_hint` makes a lot of sense. But if we want that, it needs to go in the specification now. Otherwise a `v1.0` server will be confused by future clients specifying, e.g., `api_hint=v1.1`.

So, I updated this PR to allow an optional minor version in `api_hint`.

This means that there is another way for you to implement your CLI client:
> ```
> optimade-query --filter='nelements=2 AND elements HAS "Si" AND featureX="kva" --api=v1.1'
> ```

It can now translate that into requests on the following form:
```
https://example.com/optimade/structures?filter=nelements=2 AND elements HAS "Si" AND featureX="kva"&api_hint=v1.1
```
This makes client behavior more streamlined, as it does not need to crawl through `/info` endpoints to decide who gets the request and who does not *or* try to intelligently parse errors to figure out which ones are caused by obsolete versions. Instead the responsibility ends up on the database to "do the best thing possible" for this `v1.1`-hinted request. I think this is better.

I am quite happy about the solution in the PR right now. The only thing that irks me is that since servers are allowed a lot of freedom in interpreting `api_hint`, there is no way for a client to be more firm: "This is a v1.1 query. Either serve it with v1.X with X>=1 or return 551 Version not supported".

sauliusg:
> Based on this discussion, I think allowing minor versions in `api_hint` makes a lot of sense. But if we want that, it needs to go in the specification now. Otherwise a `v1.0` server will be confused by future clients specifying, e.g., `api_hint=v1.1`.

I agree.

sauliusg:
> This means that there is another way for you to implement your CLI client:
> 
> > ```
> > optimade-query --filter='nelements=2 AND elements HAS "Si" AND featureX="kva" --api=v1.1'
> > ```
> 
> It can now translate that into requests on the following form:
> 
> ```
> https://example.com/optimade/structures?filter=nelements=2 AND elements HAS "Si" AND featureX="kva"&api_hint=v1.1
> ```

This is a possible scenario, but it is not an equivalent transformation. The 'client --api=v1.1' can choose not to send a query to a server that does not for sure support v1.1 or a higher minor version. This will eliminate error messages from servers that can not serve a request, but may miss some data.

The `..&api_hint=v1.1` variant will presumably send the query to *unversioned* endpoints of all servers, and may get extra error messages, but it may get some more useful data (if, for instance, a v2.0 supports the `featureX` property in its queries, and the filter syntax is not radically changed :).

Both scenarios have merits and drawbacks in real life, IMHO.

> This makes client behavior more streamlined, as it does not need to crawl through `/info` endpoints to decide who gets the request and who does not _or_ try to intelligently parse errors to figure out which ones are caused by obsolete versions. Instead the responsibility ends up on the database to "do the best thing possible" for this `v1.1`-hinted request. I think this is better.

I though crawling `/versions` + `/info` was the standard intended way to ensure compatibility, and the only way to proceed prior to standardization of unversioned endpoints. So its probably not a drawback? In any case, client has a choice...

sauliusg:
> I am quite happy about the solution in the PR right now. The only thing that irks me is that since servers are allowed a lot of freedom in interpreting `api_hint`, there is no way for a client to be more firm: "This is a v1.1 query. Either serve it with v1.X with X>=1 or return 551 Version not supported".

I would not include this feature in a server since it can be fully implemented in a client, and the Unix principle which I follow tells us to split functionality and not to include extra function into a program it if can be fully implemented in a separate program :).

> To do such a firm request, the client still has to crawl `/info`,

Yes, but, as discussed, crawling `/info` was a standard procedure, wasn't it? And, since it can be done in a client, and the response can be cached, I would leave it to clients, and not overload the spec and the server. In any case, if it would be a MUST feature, OPTIMADE servers would become even more complicated, and a SHOULD/MAY feature will probably be left unimplemented by most servers...

> or rely on the assumption that servers return version info in their responses (which right now is a SHOULD and not a MUST.)

Opps... What was the reason it was set to SHOULD? I would argue for MUST. Does any implementation have problems in including API support level info?

optimade.rst
```diff
@@ -639,6 +702,42 @@ The URL path segment that follows the versioned base URL MUST represent one of t
 
 These endpoints are documented below.
 
+Query parameters
+----------------
+Query parameters to the endpoints are documented in the respective subsections below.
+However, in addition, all API endpoints MUST accept the :query-param:`api_hint` parameter described under `Version Negotiation`_.
```
sauliusg, line 708: This sentence duplicates the sentence in 'Version Negotiation' par. 3, so we have a mild SPOT violation. I'm not sure how important it is; it is definitely good to have a reminder in the 'Query parameters' section, but when one place is changed, the other will have to be adjusted as well, otherwise we risk a contradiction...

rartino:
> > or rely on the assumption that servers return version info in their responses (which right now is a SHOULD and not a MUST.)

> Opps... What was the reason it was set to SHOULD? I would argue for MUST. Does any implementation have problems in including API support level info?

I think it may just have followed along from the corresponding requirement in JSON API. Our JSON response format mandates that there SHOULD be 'meta' and 'data' fields in the outermost response and at least one MUST be provided. We can consider changing that SHOULD to MUST. But then we probably need to soften up what MUST be present in the meta field. Because, e.g., `time_stamp` isn't trivial for endpoints served from static files. The solution to that problem, for now, have been to exclude the `meta` field.

optimade.rst
```diff
@@ -255,7 +267,55 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+1. The :endpoint:`versions` endpoint served directly under the unversioned base URL allows a client to discover all major API versions supported by a server in the order of preference (see section `Versions Endpoint`_).
+
+2. A client can access the API under versioned base URLs.
+   In this case, the server MUST respond according to the specified version or return an error if the version is not supported (see section `Versioned Base URLs`_).
+  
+3. When accessing the API under the unversioned base URL, clients are encouraged to append the OPTIONAL query parameter :query-param:`api_hint` to hint the server about a preferred API version for the request.
+   This parameter is described in more detail below.
+
+The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
+However, for endpoints under a versioned base URL this parameter SHOULD be ignored, and the request served as usual according to the version specified in the URL path segment.
```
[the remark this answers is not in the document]
rartino: I agree, and have made the edits.

> The last thing: what should happen is a request with api_hint=3 is sent to, say, the /optimade/v1/structures endpoint, i.e. to the different (lower) major version? Either it shold be an error, or a best effort service with a warning. I would be in favour of the last behaviour (give the response you can + warning). This would also amount to essentially ignoring the api_hint on the versioned endpoint.

I'm not sure I completely follow (is the best effort attempt you mention here based on `api_hint` or not?) but I agree with what you wrote above that `api_hint` MUST be ignored on versioned endpoints. Hence, `/optimade/v1/structures?api_hint=v3` just follows the usual procedure for a `/v1/` URL, i.e., it is served with `v1.X`, where X is the highest point release supported. The `api_hint` is completely ignored.

Personally, I really like the fact that the versioned URLs mean "serve exactly this or give me an error". Those who want a softer behavior should use the unversioned endpoint and `api_hint`.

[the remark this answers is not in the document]
sauliusg: > I'm not sure I completely follow (is the best effort attempt you mention here based on `api_hint` or not?) but I agree with what you wrote above that `api_hint` MUST be ignored on versioned endpoints. Hence, `/optimade/v1/structures?api_hint=v3` just follows the usual procedure for a `/v1/` URL, i.e., it is served with `v1.X`, where X is the highest point release supported. The `api_hint` is completely ignored.

Yes, this is my understanding as well.

It seems possible, however, to use `api_hint` even in this case to form a *warning*, e.g. "Your requested API version v3 can not be served on the endpoint `/optimade/v1/structures` and therefore can fail since major versions are incompatible". Or something more fluent...

Such warning would make it (hopefully) easier for the client to find out what is the reason why the request failed.

In such case, *technically* the `api_hint` is not completely ignored but used for analysing the situation and producing warnings. I think it would be good to allow such use and explicitly say that "even on a *versioned* endpoint an implementation MAY use `api_hint` for producing warning messages', but, as you have written now, "an implementation MUST ignore `api_hints` on a versioned endpoint when selecting which version of API to use for serving the request".

Does that make sense? 

> Personally, I really like the fact that the versioned URLs mean "serve exactly this or give me an error". Those who want a softer behaviour should use the unversioned endpoint and `api_hint`.

I agree.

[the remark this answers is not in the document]
rartino: I could change the sentence to, e.g.,

-----

However, for endpoints under a versioned base URL the request MUST be served as usual according to the version specified in the URL path segment regardless of the value of `api_hint`. In this case, the server MAY issue a warning if the value of `api_hint` suggests that the query may not be properly supported.

----

However, I'm somewhat against this. Because:

1. It makes the rules around this a bit less straightforward.

2. I don't think it is going to be a very common problem that users gets confused about providing both `api_hint` and a versioned URL. If they try it, they'll soon figure out based on the responses that `api_hint` does nothing in this case.

3. Let's say there is a system somewhere that, perhaps with help of `api_hint`, figures out the precise version that can serve the request and handles this via a redirect to the versioned URL. In this case, it is more convenient if one can be sure `api_hint` will be completely ignored, because then there is no need to edit the QS, it can just be passed on verbatim. If there is a risk that irrelevant warnings will be issued based on `api_hint`, then the redirecting system needs to edit the QS to remove `api_hint`. 

Hence, I somewhat prefer to know that `api_hint` will always be completely ignored by versioned endpoints. But I don't feel very strongly about it, if you want to argue the other side.

rartino:
@sauliusg, I just responded to your final comment in the thread above. Please consider my argument; and if you still think it would be better to adjust the text to allow warnings based on `api_hint` on versioned endpoints, I'll make the change.

optimade.rst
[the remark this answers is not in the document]
sauliusg: > However, for endpoints under a versioned base URL the request MUST be served as usual according to the version specified in the URL path segment regardless of the value of `api_hint`. In this case, the server MAY issue a warning if the value of `api_hint` suggests that the query may not be properly supported.

IMHO this is a very good text and I would be happy to see it included. I think it clarifies both the intended use of `api_hint` and makes clear what behaviour is compliant, without extra restrictions or complications for implementer's

If, as you say, you do not feel strong about it we include the text you suggest?

> However, I'm somewhat against this. Because:
> 
>     1. It makes the rules around this a bit less straightforward.

I agree; but on the other hand it's a small expansion of the docs, and it makes docs more explicit on what you can/may/are supposed to do on the server side.

>     2. I don't think it is going to be a very common problem that users gets confused about providing both `api_hint` and a versioned URL. If they try it, they'll soon figure out based on the responses that `api_hint` does nothing in this case.

Probably not, but we never know. A typical scenario might be – someone tales a query published in a paper a while ago (unversioned endpoint, with `api_hint`), passes it to an active client, and the client finds out the latest version and passes the whole query (with the api hint) to a *versioned* endpoint that does not understand it. The query fails, and it will be very confusing to see `api_hint` set and the query still failing. 

A good warning from the server would make the situation clear immediately – your `api_hint` sais you need v1.1, but you used the versioned endpoint `/v3/`... People will tend to forget what is the priority of URL versions (whether the path has precedence over the QS or the other way round), so it would be very nice if the server reminds them in a warning.

>     3. Let's say there is a system somewhere that, perhaps with help of `api_hint`, figures out the precise version that can serve the request and handles this via a redirect to the versioned URL. In this case, it is more convenient if one can be sure `api_hint` will be completely ignored, because then there is no need to edit the QS, it can just be passed on verbatim. If there is a risk that irrelevant warnings will be issued based on `api_hint`, then the redirecting system needs to edit the QS to remove `api_hint`.

My intention is that for the purpose of generating the returned data, the `api_hint` is indeed completely ignored on a versioned endpoint. The server is, however, *allowed* to make use of the `api_hint` to formulated warning messages or other metadata.

> Hence, I somewhat prefer to know that `api_hint` will always be completely ignored by versioned endpoints. But I don't feel very strongly about it, if you want to argue the other side.

I would love to see you suggested text included.

Maybe we can clarify also that the `api_hint` MUST be completely ignored when generating response data, but MAY be used when generating response metadata (e.g. warnings or error messages).

sauliusg:
> @sauliusg, I just responded to your final comment in the thread above. Please consider my argument; and if you still think it would be better to adjust the text to allow warnings based on `api_hint` on versioned endpoints, I'll make the change.

Answered in the thread above...

optimade.rst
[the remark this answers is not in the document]
rartino: Ok. You seem unconvinced of my argument, and I won't the possibility of an undesired warning in a special circumstance hold this up. I've inserted the version above. (And, I don't think further clarification is necessary; there is already a MUST in the first sentence.)

rartino:
@sauliusg, @giovannipizzi, @merkys, @CasperWA 

I think we are at the final version of this now (?). Please take a final look.

Please don't merge this until v1.0.0-rc.2 is out. Rather, please at your earliest convenience also go to #293 and approve that one.

optimade.rst
```diff
@@ -255,7 +267,56 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+Access via the unversioned URL is primarily intended for (i) convenience when manually interacting with the API, and (ii) to provide version agnostic permanent links to resource objects.
+Clients that perform automated processing of responses SHOULD access the API via versioned base URLs.
+
+Implementations serving the API on the unversioned base URL have a few alternative options:
+
+1. Direct access MAY be provided to the full API.
+2. Requests to endpoints under the unversioned base URL MAY be redirected using an HTTP 307 temporary redirect to the corresponding endpoints under a versioned base URL.
+3. Direct access MAY be limited to only single entry endpoints (see section `Single Entry Endpoints`_), i.e., so that this form of access is only available for permanent links to resource objects.
+
+Implementations MAY combine direct access to Single Entry Endpoints with redirects for other API queries.
```
merkys, line 285: "Single Entry Endpoints" should be lowercase, as in the previous sentence.

optimade.rst
```diff
@@ -255,7 +267,56 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+1. The :endpoint:`versions` endpoint served directly under the unversioned base URL allows a client to discover all major API versions supported by a server in the order of preference (see section `Versions Endpoint`_).
+
+2. A client can access the API under versioned base URLs.
+   In this case, the server MUST respond according to the specified version or return an error if the version is not supported (see section `Versioned Base URLs`_).
+  
+3. When accessing the API under the unversioned base URL, clients are encouraged to append the OPTIONAL query parameter :query-param:`api_hint` to hint the server about a preferred API version for the request.
+   This parameter is described in more detail below.
+
+The :query-param:`api_hint` query parameter MUST be accepted by all API endpoints.
+However, for endpoints under a versioned base URL the request MUST be served as usual according to the version specified in the URL path segment regardless of the value of `api_hint`.
```
merkys, line 307: Should be ":query-param:`api_hint`"

optimade.rst
```diff
@@ -255,7 +267,56 @@ Database providers SHOULD strive to implement the latest released version of thi
...
+In this case, the server MAY issue a warning if the value of `api_hint` suggests that the query may not be properly supported.
```
merkys, line 308: Should be ":query-param:`api_hint`"

rartino:
The latest version has been available for review since yesterday and has now been sitting idle for 7 hours. I'll merge, and if someone comes with something to change, we'll take that as a separate PR.

PR #295, opened 2020-06-29
Merge develop into master for release v1.0.0-rc.2

rartino:
As per the [release procedure in our Wiki](https://github.com/Materials-Consortia/OPTIMADE/wiki/New-release-%28How-to%29), this is the merging of develop into master for release v1.0.0-rc.2


rartino:
@CasperWA I followed the wiki instructions, which I think are primarily written by you; meaning I merged on my own machine. This meant I had to merge in my own clone, and then do this as a PR the normal way.

However, is there a reason we are doing this and not using the github interface to merge develop into master just in the main repo? That would be "safer" in the sense one cannot smuggle in a commit into master that no one has seen.

I suppose the reason we documented not to do that is that it wouldn't be possible to specify -ff-only. But, is it a major problem if the merge cannot be done as -ff-only if github still thinks the merge can be resolved automatically?

CasperWA:
> @CasperWA I followed the wiki instructions, which I think are primarily written by you; meaning I merged on my own machine. This meant I had to merge in my own clone, and then do this as a PR the normal way.
> 
> However, is there a reason we are doing this and not using the github interface to merge develop into master just in the main repo? That would be "safer" in the sense one cannot smuggle in a commit into master that no one has seen.

This is safe due to the checks on the commits already. I do not believe you can pollute your commit history locally and it will still allow you to push to `master`.
It's the GitHub concept of "protected branches".

There is no way to fast-forward merge (`git merge --ff-only`) via GitHub web interface. Hence this workaround.
However, if it's necessary, we can create a GitHub Actions workflow for all of this. But, I think that's spilt labor, since we shouldn't be releasing very often (I presume)?

> I suppose the reason we documented not to do that is that it wouldn't be possible to specify -ff-only. But, is it a major problem if the merge cannot be done as -ff-only if github still thinks the merge can be resolved automatically?

Yes, then there is suddenly an extra "merge" commit on the `master` branch that doesn't exist on the `develop` branch.
It then becomes a merging hell back and forth between the two branches, instead of this "simple" fast-forward merge, and then we can update `develop` afterwards as we please. No need to "fold back".

CasperWA:
@rartino I wouldn't mind doing this release in a call together if you wish?

rartino:
@CasperWA 
> I wouldn't mind doing this release in a call together if you wish?
Skype? I wrote a line in our usual Skype channel, but I feel weird about starting a call there, given that everyone will then be called...
> Yes, then there is suddenly an extra "merge" commit on the master branch that doesn't exist on the develop branch.
Ah, now I remember. I should update the wiki with even more motivation for why we do this.

PR #298, opened 2020-06-29
Fixes to meta tag description, add link to schema

### LINKED ISSUE #37: API response should include reference to schema, to be completed first
We (Fawzi, Andrius, Markus, Matt, Leopold) propose that the API response should include a reference to  a schema that can be used to validate its responses.

This can include one or two of:
 * a JSON schema
 * a NOMAD meta-info schema

Advantages:
 * machine-readable way for clients to understand the (intended) response format, including custom extensions

Disadvantages:
 * none that we are aware of

### LINKED ISSUE #292: Clarify api_version in response specification
Currently the `api_version` field is described in two places:

- [Section 4.2](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#json-response-schema-common-fields) as:
  > **api_version**: a string containing the version of the API implementation.

- [Section 5.3.1](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#base-info-endpoint) as:
  > **api_version**: Presently used version of the OPTIMADE API.

Could we make this stricter? basically matching the `version` field of `available_api_versions`, i.e.

> **version**: a string containing the full version number of the API served at that versioned base URL. The version number string MUST NOT be prefixed by, e.g., "v".

Happy to make the PR if people agree. We've just started validating this field with a semantic version type in optimade-python-tools.

Tagging @CasperWA we had already discussed it.

### LINKED ISSUE #297: Remove requirements on index_base_url in meta, or remote it completely?
As I was working on the specification, I came across the `index_base_url` field inside the top-level `meta`:
>  If the index meta-database (see section `Index Meta-Database`_) is implemented by the provider, the :field:`index_base_url` field MUST be included.

I read this to say that if there exists an index meta-database for a provider, and the response includes a top-level meta (which, strictly speaking, is optional on the SHOULD level, but one nearly always want to do), then that index meta-database MUST be linked here.

This seems problematic now that we, for some providers, handle their index meta-databases for them in the providers repository. It seems we are saying that, e.g., `cod` and `tcod` MUST link back to the index meta-db in the providers repository in the meta field of *every response* they provide.

I don't see why we should have this requirement at all.

Even further, why is this redundant `index_base_url` in the meta at all? We implemented `root`-type links in the `/links` endpoint that seem to cover this perfectly. Shouldn't we altogether remove the `index_base_url` field from the meta tag?

rartino:
This started as a PR to allow an OPTIONAL link to a schema to be included in responses, as requested in #37.

However, in editing this, I noticed some things regarding the top-level `meta` field (and was reminded of other things from recent discussions):
1. The `meta` field itself is optional (at SHOULD level). This means it is possible for a response to completely skip including an `api_version` in the response, which can make life difficult for clients (see, e.g., #290, #277), and is a bad idea if one considers that users may want to save responses and have them (re-)processed by future clients.
2. However, according to the present specification, **if** one includes `meta`, then many subfields are MANDATORY, leading to an "all or nothing" requirement that seems unnecessarily strict (?)
3. The definition of `more_data_available` mentions "this filter query" which makes it unclear what applies to this field for non-filter-queries.
4. A strict reading of the mandatory field `representation` inside the mandatory `query` makes it difficult/impossible to include `meta` in any endpoints served statically, since, in that case, the provided query parameters cannot be echoed here. 

My attempts at addressing these issues:
- The `meta` field is made MANDATORY. (Addresses point 1)
- Only three of the previously mandatory subfields remain MANDATORY: `api_version`, `query`, and `more_data_available`. The two first are sufficient to tell a client exactly how to interpret the data, so it would be good if clients can rely on them. The last one we can discuss, but it seems highly useful for the client to get a binary true/false flag to know if it needs to engage in pagination or not. (Addresses point 2) 
- The rest of the previously MANDATORY subfields are instead made optional on SHOULD-level. This give them the same status as previously, since the server could omit the `meta` field entirely. (Addresses point 2)
- The description of `more_data_available` has been updated to hopefully be more clear (Addresses point 3)
- The definition of `query`→`representation` has been clarified/relaxed so that one can skip query parameters that are ignored by the implementation. (Addresses point 4)

Finally, a new optional (on SHOULD level) field `schema` has been added that allows linking a schema in OpenAPI format. (See discussion in #37).

Closes #37 
Closes #297

Notes:
- If anyone sees a reason to reject/wait with any sub-point, I'm happy to break it out as a separate PR to take later. But since they all interfere with each other in the same piece of text it seemed easier to handle them as a group.
- If we add this feature to link schemas, we probably also should set up permalink URLs for our present and future schemas. ~~How about, e.g., `https://schemas.optimade.org/openapi/v1.0/structures.json` ?~~
**Edit:** after having toyed around a bit with schemas, I realize that there really is no need to separate schemas based on endpoints; all responses should just link the same schema, e.g.:  `https://schemas.optimade.org/openapi/v1.0/optimade.json`?

**Edit:** This PR now also:
- Consistently clarifies the version string format throughout the specification (Fixes #292).
- Adds a `homepage` field to the `implementation` subfield of the top-level `meta`.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -383,12 +383,12 @@ In the JSON response format, property types translate as follows:
 - **dictionary** is represented by the JSON object type.
 - **unknown** properties are represented by either omitting the property or by a JSON :field-val:`null` value.
 
-Every response SHOULD contain the following fields, and MUST contain at least :field-val:`meta`:
+Every response SHOULD contain the following fields, and MUST contain at least :field:`meta`:
 
 - **meta**: a `JSON API meta member <https://jsonapi.org/format/1.0/#document-meta>`__ that contains JSON API meta objects of non-standard meta-information.
   It MUST be a dictionary with these fields:
 
-  - **api\_version**: a string containing the version of the API implementation.
+  - **api\_version**: a string containing the full version of the API implementation.
 
   - **query**: information on the query that was requested.
     It MUST be a dictionary with this field:
@@ -401,10 +401,10 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
     
   :field:`meta` SHOULD also include these fields:
 
-  - **schema**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ that point to a schema for the response.
-    If it is a string, or a dictionary containing no :field:`meta` field, the provided URL MUST point at an OpenAPI schema.
+  - **schema**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ that points to a schema for the response.
+    If it is a string, or a dictionary containing no :field:`meta` field, the provided URL MUST point at an `OpenAPI <https://swagger.io/specification/>`__ schema.
     It is possible that future versions of this specification allows for alternative schema types.
-    Hence, if the :field:`meta` field of the JSON API links object is provided and contains a field **schema\_type** that is not equal to the string "OpenAPI" the client MUST not handle failures to parse the schema or to validate the response against the schema as errors.
+    Hence, if the :field:`meta` field of the JSON API links object is provided and contains a field :field:`schema_type` that is not equal to the string :field-val:`"OpenAPI"` the client MUST not handle failures to parse the schema or to validate the response against the schema as errors.
     
   - **time\_stamp**: a timestamp containing the date and time at which the query was executed.
   - **data\_returned**: an integer containing the total number of data resource objects returned for the current :query-param:`filter` query, independent of pagination.
--- a/optimade.rst
+++ b/optimade.rst
@@ -475,6 +475,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 	     "representation": "/structures/?filter=a=1 AND b=2",
 	   },
 	   "api_version": "1.0.0",
+	   "schema": "http://schema.optimade.org/openapi/v1.0/optimade.json",
 	   "time_stamp": "2007-04-05T14:30:20Z",
 	   "data_returned": 10,
 	   "data_available": 10,
--- a/optimade.rst
+++ b/optimade.rst
@@ -389,13 +389,16 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   It MUST be a dictionary with these fields:
 
   - **api\_version**: a string containing the full version of the API implementation.
+    The version number string MUST NOT be prefixed by, e.g., "v".
+    Example: :field-val:`1.0.0`.
 
   - **query**: information on the query that was requested.
     It MUST be a dictionary with this field:
 
-    - **representation**: a string with the part of the URL following the versioned base URL.
+    - **representation**: a string with the part of the URL following the versioned or unversioned base URL that serves the API.
       Query parameters that have not been used in processing the request MAY be omitted.
       In particular, if no query parameters have been involved in processing the request, the query part of the URL MAY be excluded.
+      Example: :field-val:`/structures?filter=nelements=2`.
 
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -586,8 +587,7 @@ An example of a full response:
 	     "meta": {
 	       "_exmpl_title": "This is an example site"
 	     }
-	   },
-	   "index_base_url": "http://example.com/optimade"
+	   }
```
CasperWA, line 590: Was this the only occurence of an example use of `index_base_url`? I'm presuming yes, this comment is just here to make extra extra sure :)

rartino: I don't find any more...

rartino:
@CasperWA's changes incorporated.

Plus:
- I tried to make our description of version strings even more unambiguously clear also in other parts of the specification for consistency.
- I noted the inconsistency that data for implementations in the `meta` field only allow for a `source_url` link to link the "implementation source" and no `homepage` (a field we generally have in other cases). So, I added that as an OPTIONAL field.

optimade.rst
```diff
@@ -894,11 +899,16 @@ The single resource object's response dictionary MUST include the following fiel
 - **id**: :field-val:`"/"`
 - **attributes**: Dictionary containing the following fields:
 
-  - **api\_version**: Presently used version of the OPTIMADE API.
+  - **api\_version**: Presently used full version of the OPTIMADE API.
+    The version number string MUST NOT be prefixed by, e.g., "v".
+    Example: :field-val:`1.0.0`.
```
CasperWA, line 904: Similar comment to the previous one.
```suggestion
    Examples: :field-val:`1.0.0`, :field-val:`1.0.0-rc.2+develop`.
```

optimade.rst
```diff
@@ -894,11 +899,16 @@ The single resource object's response dictionary MUST include the following fiel
...
+
   - **available\_api\_versions**: MUST be a list of dictionaries, each containing the fields:
 
     - **url**: a string specifying a versioned base URL that MUST adhere to the rules in section `Base URL`_
-    - **version**: a string containing the full version number of the API served at that versioned base URL. The version number string MUST NOT be prefixed by, e.g., "v".
+    - **version**: a string containing the full version number of the API served at that versioned base URL.
+      The version number string MUST NOT be prefixed by, e.g., "v".
+      Example: :field-val:`1.0.0`.
```
CasperWA, line 911: Again, the same as previous
```suggestion
    Examples: :field-val:`1.0.0`, :field-val:`1.0.0-rc.2+develop`.
```

optimade.rst
```diff
@@ -385,20 +383,34 @@ In the JSON response format, property types translate as follows:
...
 - **meta**: a `JSON API meta member <https://jsonapi.org/format/1.0/#document-meta>`__ that contains JSON API meta objects of non-standard meta-information.
   It MUST be a dictionary with these fields:
 
-  - **query**: information on the query that was requested.
-    It MUST be a dictionary with these fields:
-
-    - **representation**: a string with the part of the URL following the versioned base URL.
+  - **api\_version**: a string containing the full version of the API implementation.
+    The version number string MUST NOT be prefixed by, e.g., "v".
+    Example: :field-val:`1.0.0`.
```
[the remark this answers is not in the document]
rartino: I think we definitely shouldn't have develop versions here, because it is very debatable whether you can have them here. (It should be "~develop", and that intentionally breaks semantic versioning...) We could have "-rc.2" I suppose.

rartino:
Sigh... no idea how we managed to get a merge conflict as result of this. Now we need re-approvals @sauliusg, @giovannipizzi, @CasperWA.

optimade.rst
```diff
@@ -452,7 +452,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   - **api\_version**: a string containing the full version of the API implementation.
     The version number string MUST NOT be prefixed by, e.g., "v".
     Examples: :field-val:`1.0.0`, :field-val:`1.0.0-rc.2`.
-
+
```
CasperWA, line 455:
```suggestion

```

PR #299, opened 2020-06-30
Define units per The Unified Code for Units of Measure

### LINKED ISSUE #283: Units (and their machine-readability)
At the moment, units can be specified but there is no explanation of what they should be except that they are a string. In #282 there is a slight improvement with a clarification.

In any case, they remain a string, so human-readable, rather than machine-readable.

Do we want to make this field better specified?

Some options:
* We create a long list of accepted units (I think the Pauling file has already such a list?) - possibly out of the specification so it's easier to add more?
* We explain well how to specify units (e.g. "μm" rather than "um", or what to do with exponents: "m/s" vs "m s^-1" vs "m*s^{-1}" vs ...)
* We just accept that it's just a human-readable field
* other ideas

merkys:
This PR defines the way to unify the indications of units, as well as makes them machine-readable, using [The Unified Code for Units of Measure](http://unitsofmeasure.org/ucum.html). Fixes #283.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1029,6 +1029,7 @@ The response for these endpoints MUST include the following information in the :
 
   - :field:`unit`: String.
     The physical unit symbol in which the property's value is given.
+    This MUST be a valid representation of units according to version 2.1 of `The Unified Code for Units of Measure <https://unitsofmeasure.org/ucum.html>`__.
     It is RECOMMENDED that non-standard (non-SI) units are described in the description for the property.
   - :field:`sortable`: Boolean.
     Whether the property can be used for sorting (see `Entry Listing URL Query Parameters`_ for more information on this field).
--- a/optimade.rst
+++ b/optimade.rst
@@ -1058,7 +1058,7 @@ Example:
           },
           "lattice_vectors": {
             "description": "Unit cell lattice vectors",
-            "unit": "Å",
+            "unit": "Ao",
             "sortable": false,
             "type": "list"
           }


merkys:
> There is of course the unit segment in unicode: http://xahlee.info/comp/unicode_units.html ; but it would be hilariously confusing to end-users to standardize on that only those MUST be used for all specification of units :-)

In addition, this Unicode segment is not exhaustive and unambiguous.

rartino:
> This single line presents a lot of work for the validation in optimade-python-tools.
> Is there a reason why it is such a hard requirement? Could it not be a SHOULD or RECOMMENDED instead?

Due to concern for future-proofing this, I was initially about to propose adding a line saying something like this:

---

Clients MUST treat strings that cannot be interpreted according to The Unified Code for Units of Measure as an unknown but valid unit, and MUST NOT treat this as an unrecoverable validation error.

----

This means that we could upgrade the standard to a later version (as long as it stays backward compatible) without triggering a backward-incompatible change for us. It seems this would also resolve your woes about validation, since this means any string formally validates.

(I decided against proposing it, because I was worried that I am just complicating the specification without anyone really understanding the purpose, but if you want this also to simplify the validation, perhaps that tips the scale.)

Changing MUST to SHOULD is of course also an option, but is a bit "dangerous" for the attempt to standardize units. I'm pretty sure it will mean everyone puts 'Å' instead of 'Ao'... :-)

merkys:
> This single line presents a lot of work for the validation in `optimade-python-tools`.

Indeed. However, there is a [Python package to validate UCUM units](https://github.com/stomioka/ucum), maybe it could be reused?

> Changing MUST to SHOULD is of course also an option, but is a bit "dangerous" for the attempt to standardize units. I'm pretty sure it will mean everyone puts 'Å' instead of 'Ao'... :-)

Precisely. I am afraid that _soft_ requirements will just make the `unit` field unreliable.

CasperWA:
> Changing MUST to SHOULD is of course also an option, but is a bit "dangerous" for the attempt to standardize units. I'm pretty sure it will mean everyone puts 'Å' instead of 'Ao'... :-)

It's also just sooo much prettier :star_struck: :sweat_smile: 

> > This single line presents a lot of work for the validation in `optimade-python-tools`.
> 
> Indeed. However, there is a [Python package to validate UCUM units](https://github.com/stomioka/ucum), maybe it could be reused?

Great! Let's hope it's up to date and compatible with (or adaptable to) our current setup (i.e., `pydantic` models).

PR #300, opened 2020-06-30
Minor text fixes in preparation for v1.0.0 release

rartino:
I'll use this PR to collect all scattered minor text fixes I find in preparation for the v1.0.0 release.

Others can feel free to push to this as well.

PR #302, opened 2020-06-30
Fix text referencing API being served (only) under versioned URLs

rartino:
It seems I missed a few places sprinkled throughout the specification about how endpoints are present (only) under versioned URLs when preparing #290. 

This PR fixes these.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -454,7 +454,7 @@ Every response SHOULD contain the following fields, and MUST contain at least on
   - **query**: information on the query that was requested.
     It MUST be a dictionary with these fields:
 
-    - **representation**: a string with the part of the URL following the versioned base URL.
+    - **representation**: a string with the part of the URL following the versioned or unversioned base URL that serves the API.
 
   - **api\_version**: a string containing the version of the API implementation.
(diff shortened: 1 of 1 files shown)


rartino:
> however, from Section "Base URL" I understand that term "Base URL" stands for both versioned and unversioned base URLs. Why not replace "versioned or unversioned base URL" with just "base URL"?

That is not how I've used these terms. To me the "base URL" does not contain a version number path segment and is somewhat synonymous with "unversioned base URL". To cover the case that includes the version number path segment, one has to specify "versioned base URL".

merkys:
> That is not how I've used these terms. To me the "base URL" does not contain a version number path segment and is somewhat synonymous with "unversioned base URL". To cover the case that includes the version number path segment, one has to specify "versioned base URL".

Right, this distinction becomes more evident to me after re-reading "Definition of Terms". There seems to be one more occurrence where "base URL" should be replaced by "versioned or unversioned base URL":

https://github.com/Materials-Consortia/OPTIMADE/blob/f0a2912d190c9ff32a6d0382eed5676f6b3de272/optimade.rst#L771

rartino:
@merkys Good catch, I've commited a fix to #300

PR #305, opened 2020-07-01
Adding pre-commit tests (for traling whitespaces)

### LINKED ISSUE #239: Add formatting checks, e.g., remove whitespace
In consideration of #235 and other previous similar PRs, it would be good to have formatting checks, i.e., checking for trailing whitespace and more.

This may either be implemented as a pre-commit check that one would need to install locally or as a Travis check (or both).

giovannipizzi:
This is also adds a GitHub Actions test. The rest for now are still on Travis and can be moved at a later stage; in any case they require a different environment (java vs. python) so I didn't feel worth trying to do this in travis.

Fixes #239 

## Note

If this is approved, we can add a note to [this wiki page](https://github.com/Materials-Consortia/OPTIMADE/wiki/How-to-contribute) a note on how to run the pre-commit check locally, that amounts to the following text:

> To install locally the pre-commit tests so they automatically run right before you commit, run the following two commands (you need python 3):
> 
> - `pip install pre-commit`
> - `pre-commit install`
> 
> Now at every commit, checks will be run automatically on the modified files.
> For trailing whitespace, problems are automatically fixed on your files, and you just need to commit again.
> If you really want to commit and skip tests, just run `git commit --no-verify`.
> 
> If you want to run the checks on all files, you can run `pre-commit run --all-files`.


giovannipizzi:
Replaced by #306 that starts from a branch on the main action - for the first commit, we need this to have the action run

PR #325, opened 2020-10-14
Remove a space in an example URL

merkys:
I found a space in URL, which I guess is unintended, although the URL is used to illustrate partially URL-encoded request. If showcasing usage of spaces instead of `+` characters is intended, I suggest replacing all `+` with spaces in this particular example.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1409,7 +1409,7 @@ In particular, this means the client MUST escape special characters in string va
 
 Examples of syntactically correct query strings embedded in queries:
 
--  :query-url:`http://example.org/optimade/v1/structures?filter=_exmpl_melting_point%3C300+AND+ nelements=4+AND+elements="Si,O2"&response_format=xml`
+-  :query-url:`http://example.org/optimade/v1/structures?filter=_exmpl_melting_point%3C300+AND+nelements=4+AND+elements="Si,O2"&response_format=xml`
 
 Or, fully URL encoded :
 


optimade.rst
```diff
@@ -1409,7 +1409,7 @@ In particular, this means the client MUST escape special characters in string va
 
 Examples of syntactically correct query strings embedded in queries:
 
--  :query-url:`http://example.org/optimade/v1/structures?filter=_exmpl_melting_point%3C300+AND+ nelements=4+AND+elements="Si,O2"&response_format=xml`
+-  :query-url:`http://example.org/optimade/v1/structures?filter=_exmpl_melting_point%3C300+AND+nelements=4+AND+elements="Si,O2"&response_format=xml`
```
sauliusg, line 1412: True, for the filter language spaces are optional, the style employed encodes spaces as '+', and spaces should not be used in URLs. The correction is good, IMHO! Other URLs need to scanned via regexps, probably.

sauliusg (approved):
True, for the filter language spaces are optional, the style employed encodes spaces as '+', and spaces should not be used in URLs. The correction is good, IMHO! Other URLs need to scanned via regexps, probably.
@sauliusg

rartino (approved):
Good catch. And this is indeed a reminder that we should get on the automatic validation of things like this, which is possible since the swap to rst markup that added context info to, e.g., URLs.

merkys:
Thanks for prompt reviews! Merged.

PR #330, opened 2020-10-31
Updated OpenAPI schemas to those from latest python-tools (v0.12.9)

ml-evs:
This is another schema update PR, including changes from the latest optimade-python-tools release, namely:

- [x] SHOULD fields are now OpenAPI `nullable: true` (pending acceptance of one more PR in python-tools)
- [x] Regexps (`pattern` in OpenAPI) have been added to the chemical formula fields. Currently the regexp is `^([A-Z][a-z]?\\d*)*$`, which will match on empty strings, and on formulae that contain nonsense elements.


schemas/openapi_schema.json
```diff
@@ -3020,48 +3020,58 @@
...
+            "nullable": true
           },
           "chemical_formula_reduced": {
             "title": "Chemical Formula Reduced",
+            "pattern": "^([A-Z][a-z]?\\d*)*$",
```
merkys, line 3058: Patterns for chemical formulas might be improved to explicitly forbid proportions equal to 1, but not necessary for this PR.

merkys (approved):

rartino (approved):
I looked through the changes and did not see anything odd.

schemas/openapi_schema.json
CasperWA: Great catch. Can you make an issue for it in `optimade-python-tools`?

PR #332, opened 2020-11-06
Added Zenodo citation

### LINKED ISSUE #318: Publish releases to Zenodo
We should consider publishing each release of this repository on Zenodo with their GitHub integration. Various papers are beginning to mention OPTIMADE, and until the main paper is out, Zenodo can provide a citeable record. I imagine we all have the permissions to do this as admins of Materials-Consortia, I'm happy to do it if we think it's a good idea...

ml-evs:
As discussed on the mailing list, this PR simply adds the Zenodo link as the current preferred citation, feel free to tweak the wording.

Closes #318.


rartino:
Should this actually be to http://doi.org/10.5281/zenodo.4195050
or should it be http://doi.org/10.5281/zenodo.4195050
(Are we linking the latest version, or specifically version 1.0)?

ml-evs:
> Should this actually be to http://doi.org/10.5281/zenodo.4195051
> or should it be http://doi.org/10.5281/zenodo.4195050
> (Are we linking the latest version, or specifically version 1.0)?

That's a good question, I would prefer to link to the unversioned DOI (which I thought was the Zenodo default when exporting the citation) i.e. your second suggestion. Any disagreement?

README.md
```diff
@@ -22,3 +22,9 @@ The [develop branch of the repository](https://github.com/Materials-Consortia/OP
 
 API and client implementations are encouraged to support the latest release or pre-release of the specification.
 If this is a pre-release, implementations are also encouraged to support the latest stable release.
+
+## How to cite
+
+Pending the release of an OPTIMADE paper, we kindly ask that you cite the Zenodo archive for this repository:
+
+> Andersen *et al*, The OPTIMADE Specification, http://doi.org/10.5281/zenodo.4195051
```
ml-evs, line 30:
```suggestion
> Andersen *et al*, The OPTIMADE Specification, http://doi.org/10.5281/zenodo.4195050
```

README.md
```diff
@@ -22,3 +22,9 @@ The [develop branch of the repository](https://github.com/Materials-Consortia/OP
 
 API and client implementations are encouraged to support the latest release or pre-release of the specification.
 If this is a pre-release, implementations are also encouraged to support the latest stable release.
+
+## How to cite
+
+Pending the release of an OPTIMADE paper, we kindly ask that you cite the Zenodo archive for this repository:
```
merkys, line 28: I assume "release" here means "opening to public" by either publication or upload to preprint server. Otherwise "publication" would be more precise.

merkys (changes requested):
I would suggest using HTTPS, otherwise good to go.

README.md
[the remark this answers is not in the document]
ml-evs: Well spotted, added to the suggestion above. https://github.com/Materials-Consortia/OPTIMADE/pull/332#discussion_r518791983

README.md
ml-evs: That's what I was going for, agree it could be more precise. I think "publication" works for a preprint too, so I'll change it.

```suggestion
Pending the publication of an OPTIMADE paper, we kindly ask that you cite the Zenodo archive for this repository:
```

README.md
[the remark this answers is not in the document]
ml-evs: Turns out Zenodo redirects all HTTP to HTTPS anyway, but I've changed it everywhere :)

merkys:
Regarding versioned and unversioned links. I would prefer using unversioned link in `develop` branch, but at the moment of a release a link should be made versioned in order to get to stable release as such. But I guess this will lead us to egg-and-chicken problems, or not?

ml-evs:
> Regarding versioned and unversioned links. I would prefer using unversioned link in `develop` branch, but at the moment of a release a link should be made versioned in order to get to stable release as such. But I guess this will lead us to egg-and-chicken problems, or not?

This would lead to egg-chicken problems if we used the Zenodo/Github integration to automatically add a Zenodo entry for each new GitHub release, but we could take this approach if we wanted to release manually. This *may* actually be preferable as manual releases keep the manual authorship. I think the automated releases use the GitHub contributors alone, which is suboptimal...

merkys (approved):
Good to go! As before, I prefer versioned links for releases if this is not too difficult to do. If we can claim the DOI, release OPTIMADE and then update the content under DOI, this would be perfect.

ml-evs:
> Good to go! As before, I prefer versioned links for releases if this is not too difficult to do. If we can claim the DOI, release OPTIMADE and then update the content under DOI, this would be perfect.

This is technically possible, though it must be done manually. I'm a bit worried about the bus factor of all this as the Zenodo archive is done through my Zenodo account (and I don't think there's any way to give others access...). What I can do is reserve the DOI for the next versions (both patch and minor release) so we can consider this option next release.

ml-evs:
> I'm a bit worried about the bus factor of all this as the Zenodo archive is done through my Zenodo account (and I don't think there's any way to give others access...).

Relevant issue, hopefully not a problem for long:

https://github.com/zenodo/zenodo/issues/151#issuecomment-585738162

A shared account would be the other option, but my Zenodo is already linked to GH/ORCID with personal projects, so I think that's a non-starter. If the proverbial bus were to hit me, feel free to link this comment to the Zenodo admins to somehow give access to my Zenodo account to anyone nominated by Materials-Consortia. :upside_down_face:

merkys:
> A shared account would be the other option, but my Zenodo is already linked to GH/ORCID with personal projects, so I think that's a non-starter. If the proverbial bus were to hit me, feel free to link this comment to the Zenodo admins to somehow give access to my Zenodo account to anyone nominated by Materials-Consortia.

A shared account sounds really good. It would be cool if people would be allowed to transfer ownership of projects as in GitHub, but from your message I understand that this is not yet implemented in Zenodo. Hopefully, it's not for long.

shyamd (approved):

Issue #331, opened 2020-11-06
Travis changing OSS billing


ml-evs:
As you may have seen, [Travis is ending free builds for open source projects](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing). 

As part of the transition, we will get 1000 CI minutes to use before we have to start paying; each push to this repository uses ~3 minutes (as PR and push runs separately), so it will take us quite a while to run this down, but we should think about migrating (presumably to GitHub actions, which is used for part of our CI here already).

CasperWA:
👍 for moving to GH Actions as the solution.

ml-evs:
I think it would be a useful time to add swagger validation of the OpenAPI schemas to this repository too to avoid any errors arising when transferring them over from python-tools, or if we eventually move away from python-tools/need to change something manually.

ml-evs:
This is fixed by #333, which migrated the CI to GitHub Actions, but Travis has been left enabled for a little while just to make sure things are working. Let's keep this issue open until we disable it for good.

CasperWA:
> This is fixed by #333, which migrated the CI to GitHub Actions, but Travis has been left enabled for a little while just to make sure things are working. Let's keep this issue open until we disable it for good.

I'd probably just make the move ASAP. No reason to run both CIs.
What timeline did you have in mind for this though?

PR #333, opened 2020-11-06
Migrate CI from Travis to GitHub Actions

ml-evs:
Following the Travis OSS billing changes mentioned in #331, this PR migrates the Java build and checks to GH actions. It seems to work perfectly... 

We won't run out of Travis minutes for quite a while, so I've left Travis enabled until we're convinced the new Actions agree with Travis.

GitHub shows an ugly diff given the file rename, the actual changes are all in 4b14f26.

Downsides: more eggs in GH basket, but Travis is basically forcing our hand...


merkys (commented):
I have insufficient knowledge of both Travis and GitHub Actions. Maybe someone having more experience here could review and approve?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        pip install pre-commit==2.2.0
+    - name: Run pre-commit tests
+      run: |
+        pre-commit run --all-files
```
CasperWA, line 28:
```suggestion
      run: pre-commit run --all-files
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        pip install pre-commit==2.2.0
```
CasperWA, line 25:
```suggestion
      run: pip install pre-commit==2.2.0
```

CasperWA: Maybe we should update this version to the latest version of `2.8.2`?
```suggestion
      run: pip install pre-commit==2.8.2
```

CasperWA (approved):
Looks good, thanks @ml-evs !

Some slightly unrelated suggested changes. I've approved, since these suggested changes can be argued to be out-of-scope of this PR...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    name: Test grammar and formatting
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
```
CasperWA, line 40:
```suggestion
        java-version: 8
```

CasperWA (changes requested):
Ah, missed this - should actually be java version 8, right? According to what is used currently in Travis?

.github/workflows/ci.yml
CasperWA: Never mind, just saw [here](https://github.com/actions/setup-java#basic) that it's an alias...

CasperWA (approved):

.github/workflows/ci.yml
ml-evs: Yeah, I loved that too...

shyamd (approved):

ml-evs:
> Some slightly unrelated suggested changes. I've approved, since these suggested changes can be argued to be out-of-scope of this PR...

Agree with all your suggestions, but to save another round of reviewing I'll merge this as is. I'll raise another issue tracking the swagger validation of schema in the CI and can fix those minor blips then.

Issue #314, opened 2020-08-11
Swagger OpenAPI validator yields an error


blokhin:
See https://validator.swagger.io/validator/debug?url=https://raw.githubusercontent.com/Materials-Consortia/OPTIMADE/develop/schemas/openapi_schema.json

Sorry if already known.

ml-evs:
Do you know why it might fail in your link (which contains "debug"), but seems to validate fine in the GUI version (unless I'm missing something)? 

https://validator.swagger.io/?url=https://raw.githubusercontent.com/Materials-Consortia/OPTIMADE/master/schemas/openapi_schema.json

ml-evs:
Just a heads-up that we've fixed this upstream at https://github.com/Materials-Consortia/optimade-python-tools/pull/480 

(see e.g. https://validator.swagger.io/validator/debug?url=https://raw.githubusercontent.com/Materials-Consortia/optimade-python-tools/master/openapi/openapi.json).

We should prepare a PR for this repo fairly soon that updates the schema according to our changes since the 1.0 release.

blokhin:
@ml-evs Matthew, should the [main schema](https://raw.githubusercontent.com/Materials-Consortia/OPTIMADE/master/schemas/openapi_schema.json) be also PR-ed?

blokhin:
There are still validation errors: https://validator.swagger.io/validator/debug?url=https%3A%2F%2Fraw.githubusercontent.com%2FMaterials-Consortia%2FOPTIMADE%2Fmaster%2Fschemas%2Fopenapi_schema.json

ml-evs:
These are all fixed on develop (and the schemas are now checked against this URL in the CI), and will be fixed on master when we decide to do a v1.0.1 release (or if we decide to reorganise our branches following #340).

https://validator.swagger.io/validator/debug?url=https%3A%2F%2Fraw.githubusercontent.com%2FMaterials-Consortia%2FOPTIMADE%2Fdevelop%2Fschemas%2Fopenapi_schema.json

blokhin:
@ml-evs is there any update on that?

PS my motivation is https://github.com/APIs-guru/openapi-directory/issues/722

ml-evs:
Hi @blokhin, the schemas have been fixed for a while now, we just haven't pushed to master as it is meant to be reserved for releases only. Once #351 is merged the link to master should work fine (still waiting on some reviews of blocking PRs; hopefully by the next workshop we will have 1.0.1...)

[...]

PR #335, opened 2020-11-08
Validate OpenAPI schemas in CI with validator.swagger.io

### LINKED ISSUE #331: Travis changing OSS billing

### LINKED ISSUE #334: Validation of OpenAPI schema files in CI
I think it would be a useful time to add swagger validation of the OpenAPI schemas to this repository too to avoid any errors arising when transferring them over from python-tools, or if we eventually move away from python-tools/need to change something manually.

_Originally posted by @ml-evs in https://github.com/Materials-Consortia/OPTIMADE/issues/331#issuecomment-723115759_

ml-evs:
This PR updates our CI such that:

- [x] The two schema files in `./schemas/` now get validated by validator.swagger.io (closes #334, see also #314)
- [x] The CI run has been restructured so that each sub-task of `make audit` is run in a separate job, so that top-level check hierarchy has nicer error messages.
    - This means a bit of extra work is done by generating the grammar tests at every step (which also fills the output unfortunately), but it also allows steps to run in parallel. This is now significantly faster than it was on Travis.
- [x] Bumped version of `pre-commit` used in CI
- [x] Disabled Travis as GH actions seems to work perfectly (closes #331).


shyamd (approved):

tests/scripts/validate_swagger.sh
```diff
@@ -0,0 +1,27 @@
...
+set -e
+
+if [ $# -eq 0 ]; then
+    echo "Please provide an OpenAPI schema filename!"
+    exit 126
+fi
+
+echo "Validating schema at ${1}":
+
+response=$(curl -sS -H "Content-Type: application/json" -d @${1} https://validator.swagger.io/validator/debug)
```
merkys, line 15: I see that for now we depend on https://validator.swagger.io/validator/debug for OpenAPI validation. This is fine, but I would prefer having a local solution, for example, validating against OpenAPI JSON schema. Do you know, are there any blockers for doing so?

merkys (approved):

tests/scripts/validate_swagger.sh
ml-evs: We tried a few local OpenAPI validators for optimade-python-tools but ended up settling on this no-install approach. Whilst there are no blockers from doing this with local tools, they are either more fiddly to set up or do not support all features:

- Official OpenAPI parser/validator (Java) https://github.com/swagger-api/swagger-parser
- Unofficial validator from IBM (Javascript) https://github.com/IBM/openapi-validator
- Unofficial validator (Python) https://github.com/p1c2u/openapi-spec-validator

merkys: I see, thanks a lot for explanation. No-install approach is just fine for the time being. I will give these local alternatives a look sometime.

PR #336, opened 2020-11-08
Fix spelling errors in specification and update spellcheck tooling

ml-evs:
This PR adds a spellcheck to the audit tests run by the CI.

Previously, had `make audit_spelling`, which launched an interactive `aspell` session. This clearly hadn't be run for a while, as a few minor spelling errors had crept into the specification. To remedy this, this PR does the following:

- [x] Changed `make audit_spelling`/`make spell` to call `aspell` non-interactively, returning a list of suspected spelling errors. ~This is now run in the CI.~ (see discussion below https://github.com/Materials-Consortia/OPTIMADE/pull/336#discussion_r542266198).
- [x] Created a new make target `make fix_spelling` that calls the previous interactive `aspell`.
- [x] Used the aforementioned `make fix_spelling` to update the specification and the allowed `.words.lst` (see the diff for 9c42850 for new words added to `.words.lst`, the overall PR diff is messy as it also includes automatic sorting `.words.lst` file by `aspell`).


merkys (dismissed):

ml-evs:
Does anyone else have time to review this?

CasperWA (commented):
Looks all right to me, thanks @ml-evs 👍 

However, I've reviewed this as a Comment, since I am confused about the results of the spell audit CI (see the messages [here](https://github.com/Materials-Consortia/OPTIMADE/runs/1513056635?check_suite_focus=true#step:3:4)) from the latest check.
Some files are missing/not found and is it actually changing stuff, but still exiting with return code 0?

ml-evs:
> However, I've reviewed this as a Comment, since I am confused about the results of the spell audit CI (see the messages [here](https://github.com/Materials-Consortia/OPTIMADE/runs/1513056635?check_suite_focus=true#step:3:4)) from the latest check.
> Some files are missing/not found and is it actually changing stuff, but still exiting with return code 0?

Well spotted, unfortunately this has been the case in the CI for a while. There is some Makefile magic going on before the tests actually run that (I think) extracts parts of the specification to be tested as filters, but this seems to have been broken for some time (see e.g. https://travis-ci.org/github/Materials-Consortia/OPTIMADE/builds/704036295) and I didn't want to break it further in this PR...

rartino:
There is some info about the GNU make build system we have in the discussions around PR #195 that covers why there is "unrelated printout" when running make on rules like, e.g., spellcheck. But, in short, the make system dynamically generates make rules which requires these commands that extracts info from the specification to run before anything else runs.

I don't have any idea why they seem to fail in this case, though.

rartino:
About the review of this PR: if I read the code correctly, it does not avoid spell-checking code blocks etc. I've previously argued not to add spellcheck to the automatic audit until we handle skipping over code sections (https://github.com/Materials-Consortia/OPTIMADE/pull/122#pullrequestreview-249928842). Are we okay with every PR that edits code having to deal with the aspell wordlist?

tests/makefiles/Makelocal-audit
```diff
@@ -28,7 +28,7 @@ audit_authors: test_authors
 .PHONY: audit audit_spelling
 
 audit_spelling:
-	@echo "Spellcheck is not yet part of the audit of the specification."
+	$(MAKE) spell
```
ml-evs, line 31: Thanks @rartino, I wasn't aware of the previous discussion. The commit suggestion below will disable spell check in the CI once again, if that is what we want. I understand the argument that it would be annoying to have to update the wordlist for every PR (potentially introducing false negatives too), though it seems a shame to throw out all spellchecking as a side effect.

A compromise might be to enable the spellchecker in the CI but make it always pass, just printing spellcheck suggestions for PR reviewers to check manually.

```suggestion
  @echo "Spellcheck is not yet part of the audit of the specification."
```

rartino: So, I will approve this PR if you mere the suggested change, since then it is completely uncontroversial, but I indeed like the idea of a "warning" approach for spelling errors.

Is there some way we can get a mandatory passed spellcheck in for things that gets merged into main? I don't think it is a good idea to put it as a pass criteria on PRs against main, because we don't want to discover things breaking that late. But, if nothing else, perhaps as the first manual step in our release workflow on the wiki.

ml-evs: I'll remove it for now and add a reminder to run `make fix_spelling` when doing releases to the wiki.

rartino (approved):
Great work with this! Nice to see multiple spelling errors fixed.

merkys (approved):
Looks good to me!

PR #339, opened 2020-11-17
Add field for implementation issue tracker

### LINKED ISSUE #337: New `issue_tracker` field in `meta->provider`?
In a discussion with @shyamd and @csadorf in CasperWA/voila-optimade-client#215 I suggested to create a field `issue_tracker` in the `provider` JSON object under the top-level `meta` field.
The field would be OPTIONAL.
The value should be a URL, pointing to the provider's issue tracker for their OPTIMADE implementation, e.g., for Materials Cloud the `issue_tracker` value would be https://github.com/aiidateam/aiida-optimade/issues, while for COD it would be https://projects.ibt.lt/repositories/issues (AFAIK, @sauliusg and @merkys can correct me if I'm wrong).

It may be that this field is more relevant under the `implementation` JSON object in the top-level `meta` field, if one has an issue tracker specific to each different implementation from the same provider?

merkys:
Fixes #337


optimade.rst
```diff
@@ -501,6 +501,8 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
       - **email** with the maintainer's email address.
 
+    - **issue\_tracker**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ pointing to the implementation issue tracker.
```
CasperWA, line 504:
```suggestion
    - **issue\_tracker**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ pointing to the implementation's issue tracker.
```

CasperWA (changes requested):
Extremely minor change request and then good to go on my part, thanks @merkys !

CasperWA (approved):
Thanks @merkys !

optimade.rst
merkys: Thanks @CasperWA! Shouldn't the same change be applicable to "implementation source" (as is given in `source_url` a couple of lines above)?

CasperWA: I guess, yes. But I think one should add `URL` explicitly also then? I.e., it will become `implementation's source URL` or something. Or "source code"?

CasperWA: Strictly speaking, I would probably create another PR for this - especially because I don't want it to hold up this PR ;)

merkys: I think "downloadable archive or version control system" implicitly means URL, and the fact that this is in JSON links object even makes it more clear. But of course we may add "URL" to make it even more explicit.

I agree that description fix for `source_url` should be on a separate PR :)

ml-evs (approved):
I really like this idea, thanks for working it up @merkys.

merkys:
> I really like this idea, thanks for working it up @merkys.

Thanks @CasperWA for coming up with it in #337 in first place! :)

Issue #327, opened 2020-10-28
Chemical symbols D and T


merkys:
OPTIMADE specification v1.0.0 contains quite vague guidelines about allowed values for chemical symbols. Implicitly it is understood that they are symbols from the periodic table. However, deuterium (`D`) and tritium (`T`), which are present in the COD, are two common symbols not appearing in the periodic table. Maybe the specification could have an explicit clause allowing `D` and `T` in element listings as well as formulae? Of course deuterium and tritium could be detected by different masses in `species.mass`, however, as query support for `species` is OPTIONAL, there is no other way to filter deuterium and tritium in or out.

merkys:
In the web meeting of 2020-10-28 the following points were raised:

* isotopes can already be identified via atomic masses, therefore, there is no need to duplicate that information in chemical names;
* searches for `H` atoms have to return structures with deuterium and tritium too.

AFAIR there was no clear consensus about the following:

* usage of `D` and `T` in searches (should they match structures with deuterium and tritium, if a provider differentiates between hydrogen and them?);
* usage of `D` and `T` in formulae.

I personally prefer homogeneity: if we decide that `D` and `T` are not allowed in `species`, then they should not appear in formulae and should not be supported in searches too.

merkys:
> * isotopes can already be identified via atomic masses, therefore, there is no need to duplicate that information in chemical names

Actually, the specification defines `mass` as an average mass on a site, thus it is not defined as mass-per-species-per-site, but as mass-per-site. This additionally complicates the task of identifying deuterium and tritium, but should still be doable for sites with no more than two species in them.

CasperWA:
> Actually, the specification defines `mass` as an average mass on a site, thus it is not defined as mass-per-species-per-site, but as mass-per-site. This additionally complicates the task of identifying deuterium and tritium, but should still be doable for sites with no more than two species in them.

Would it make sense to change the value of `mass` to a list instead? If the average value of the mass on the site is needed it can easily be calculated by the user or we can propose an additional field?

merkys:
> Would it make sense to change the value of `mass` to a list instead? If the average value of the mass on the site is needed it can easily be calculated by the user or we can propose an additional field?

I would go for redefining `mass` as a list of floats. From a list one can easily calculate the average mass. For the same reason I would not keep an additional field for a value so easily calculated.

merkys:
[...]

Issue #340, opened 2020-11-20
Separate git branches for bug fixes and features


merkys:
Maybe this is already too late for v1.0.0++, but later on it may be wise to collect bug fixes to one branch, and new features on another. This way we may in parallel develop two releases: next minor and next patch. So if at some point we decide to make a patch release (v1.0.1 as of the time of writing), there would not be a need to check whether `develop` already includes additions requiring next minor release (v1.1.0).

ml-evs:
I support this idea, and think we should even consider applying this retroactively. As far as I can see, the only "feature" added since 1.0 is the `issue_tracker` field.

Would the suggestion be to continue with `develop` as it is, but backport bug fixes to a new branch `stable`, or is your suggestion to go the other way and merge bug fixes from `stable` into `develop` when required?

Either way, this would have to be done per PR, though there may be some GH action we can use that looks for a PR label, e.g. `backport-v1`, that could do this automatically.

merkys:
> I support this idea, and think we should even consider applying this retroactively. As far as I can see, the only "feature" added since 1.0 is the `issue_tracker` field.

Indeed, it was #339. Then there should not be too much of cherry-picking.

> Would the suggestion be to continue with `develop` as it is, but backport bug fixes to a new branch `stable`, or is your suggestion to go the other way and merge bug fixes from `stable` into `develop` when required?

I support the first option. Next minor release will surely contain the bug fixes, so `develop` should sync from `stable`.

> Either way, this would have to be done per PR, though there may be some GH action we can use that looks for a PR label, e.g. `backport-v1`, that could do this automatically.

I think the submitter of every PR would indicate whether the PR is a bug fix or a feature by choosing the merge target branch (`stable` or `develop`, accordingly). We can think about GH action too.

ml-evs:
We just discussed this in the meeting. It sounded like there was consensus that we should at least do something to tackle this.

As far as I see it, the options are either:

- create a new branch that is specifically for new features, which has fixes cherry-picked from develop as-and-when
- treat `develop` as a real development branch, and backport fixes to `master`, relying on git tags to deal with stable versioning.

I would suggest that the latter makes more sense, but it leaves more work for PR reviewers to dictate whether changes should be backported (see upcoming issue on code owners).

merkys:
I think that backporting approach would need more effort that the two-branch approach. So I vote for two-branch approach.

rartino:
> As far as I see it, the options are either:
> 
[...]

Issue #197, opened 2019-11-08
Add a ChangeLog


merkys:
A ChangeLog would help both the spec developers and the implementers to track spec modifications between releases without the need to resort to reading git logs.

ml-evs:
Are the summaries in the [released tags](https://github.com/Materials-Consortia/OPTIMADE/releases) sufficient, or should we make a changelog file containing them too?

merkys:
> Are the summaries in the [released tags](https://github.com/Materials-Consortia/OPTIMADE/releases) sufficient, or should we make a changelog file containing them too?

Good point. Summaries in released tags are not visible outside the GitHub (I mean they don't come with the source), so I would prefer having them in a changelog file too.

merkys:
Changelog is already present in the source (:partying_face:), closing the issue.

PR #344, opened 2020-12-17
Redefine species.mass as a list of floats

merkys:
When looking into means to differentiate hydrogen, deuterium and tritium (see #327), I found out that information about isotopes in multi-species sites can easily be lost, because `species.mass` defined per-site, not per-species. Thus I suggest redefining `species.mass` as a list of floats.

Redefinition brings a problem of backwards-incompatibility. Here are ways to solve it:

1. Use different property name. We may retain `species.mass` with its current definition, and make `species.masses` a list of floats. This retains backwards compatibility, but does not sound good (all subproperties of `species` are in singular). Another suggestion is `species.element_mass`.
2. Declare that defining `species.mass` with a single float value was a bug. Then redefining is a bugfix, thus does not qualify as a breaking change.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2119,7 +2119,7 @@ species
   - :property:`concentration`: list of float (REQUIRED)
   - :property:`attached`: list of strings (OPTIONAL)
   - :property:`nattached`: list of integers (OPTIONAL)
-  - :property:`mass`: float (OPTIONAL)
+  - :property:`mass`: list of floats (OPTIONAL)
   - :property:`original_name`: string (OPTIONAL).
 
 - **Requirements/Conventions**:
@@ -2153,7 +2153,7 @@ species
       The implementation MUST include either both or none of the :field:`attached` and :field:`nattached` keys, and if they are provided, they MUST be of the same length.
       Furthermore, if they are provided, the `structure_features`_ property MUST include the string :val:`site_attachments`.
 
-    - **mass**: OPTIONAL. If present MUST be a float expressed in a.m.u.
+    - **mass**: OPTIONAL. If present MUST be a list of floats, with same length as :property:`chemical_symbols`, providing element masses expressed in a.m.u. Elements denoting vacancies MUST have masses equal to 0.
     - **original\_name**: OPTIONAL. Can be any valid Unicode string, and SHOULD contain (if specified) the name of the species that is used internally in the source database.
 
           Note: With regards to "source database", we refer to the immediate source being queried via the OPTIMADE API implementation.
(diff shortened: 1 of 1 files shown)


rartino:
I agree that this seems more a bug (and I am trying to take into account my bias from not wanting to bump major version numbers.)

Hey, we should keep an errata somewhere about "known bugs in 1.0.0" until we actually release an 1.0.1 that addresses them. But I agree with merging this.

In the errata we could point out that a workaround for this bug while retaining full 1.0.0 compatibility is to define different species for the different isotopes and then instead use `assemblies` to represent the disorder.

rartino (dismissed):
Good job finding this and writing it up as a PR. I've looked it over and don't see any issues (apart from the discussion of backwards compatibility).

CasperWA:
> **Edit:** also as far as I know, no one we know of has implemented mass yet?

That's not true. It's part of the OPTIMADE Python tools, hence part of the Materials Project, ODBX, and Materials Cloud.
It is simply not applied in the way @merkys refers to.

> Hey, we should keep an errata somewhere about "known bugs in 1.0.0" until we actually release an 1.0.1 that addresses them. But I agree with merging this.
> 
> In the errata we could point out that a workaround for this bug while retaining full 1.0.0 compatibility is to define different species for the different isotopes and then instead use `assemblies` to represent the disorder.

This sounds like a great idea!
Either create a GitHub Project named "Known bugs in 1.0.0", adding issues and PRs there, create a single issue to collect them all, or create a wiki page focusing on this. Either of these are equally acceptable in my opinion.

merkys:
> I agree that this seems more a bug (and I am trying to take into account my bias for not wanting to bump major version numbers.)

Same bias here :)

> Hey, we should keep an errata somewhere about "known bugs in 1.0.0" until we actually release an 1.0.1 that addresses them. But I agree with merging this.
> 
> In the errata we could point out that a workaround for this bug while retaining full 1.0.0 compatibility is to define different species for the different isotopes and then instead use `assemblies` to represent the disorder.

I think a changelog (as suggested in #197) could cater for this purpose.

optimade.rst
```diff
@@ -2153,7 +2153,7 @@ species
       The implementation MUST include either both or none of the :field:`attached` and :field:`nattached` keys, and if they are provided, they MUST be of the same length.
       Furthermore, if they are provided, the `structure_features`_ property MUST include the string :val:`site_attachments`.
 
-    - **mass**: OPTIONAL. If present MUST be a float expressed in a.m.u.
+    - **mass**: OPTIONAL. If present MUST be a list of floats, with same length as :property:`chemical_symbols`, providing element masses expressed in a.m.u. Elements denoting vacancies MUST have masses equal to 0.
```
CasperWA, line 2156:
```suggestion
    - **mass**: OPTIONAL. If present MUST be a list of floats, with the same length as :property:`chemical_symbols`, providing element masses expressed in a.m.u.
      Elements denoting vacancies MUST have masses equal to 0.
```

optimade.rst
```diff
@@ -2168,9 +2168,9 @@ species
...
   - :val:`[ {"name": "Ti", "chemical_symbols": ["Ti", "vacancy"], "concentration": [0.9, 0.1]} ]`: any site with this species is occupied by a Ti atom with 90 % probability, and has a vacancy with 10 % probability.
-  - :val:`[ {"name": "BaCa", "chemical_symbols": ["vacancy", "Ba", "Ca"], "concentration": [0.05, 0.45, 0.5], "mass": 88.5} ]`: any site with this species is occupied by a Ba atom with 45 % probability, a Ca atom with 50 % probability, and by a vacancy with 5 % probability. The mass of this site is (on average) 88.5 a.m.u.
-  - :val:`[ {"name": "C12", "chemical_symbols": ["C"], "concentration": [1.0], "mass": 12.0} ]`: any site with this species is occupied by a carbon isotope with mass 12.
-  - :val:`[ {"name": "C13", "chemical_symbols": ["C"], "concentration": [1.0], "mass": 13.0} ]`: any site with this species is occupied by a carbon isotope with mass 13.
+  - :val:`[ {"name": "BaCa", "chemical_symbols": ["vacancy", "Ba", "Ca"], "concentration": [0.05, 0.45, 0.5], "mass": [0, 137.327, 40.078]} ]`: any site with this species is occupied by a Ba atom with 45 % probability, a Ca atom with 50 % probability, and by a vacancy with 5 % probability.
```
CasperWA, line 2171: To underline it should be a `float`, I've added a decimal point here :) But this change is merely a suggestion.
```suggestion
  - :val:`[ {"name": "BaCa", "chemical_symbols": ["vacancy", "Ba", "Ca"], "concentration": [0.05, 0.45, 0.5], "mass": [0.0, 137.327, 40.078]} ]`: any site with this species is occupied by a Ba atom with 45 % probability, a Ca atom with 50 % probability, and by a vacancy with 5 % probability.
```

CasperWA (changes requested):
This is a very sensible _fix_ 👍 😃 Thanks @merkys !

ml-evs (dismissed):
Thanks for spotting this @merkys! Happy to accept, pending the changes suggested by @CasperWA (just ping me another review request if I need to re-accept)

> I think a changelog (as suggested in #197) could cater for this purpose.

+1 for a rolling changelog that includes known/fixed bugs, I don't think a GH wiki page is visible enough (feels more like an internal knowledge base). I'm happy to spin up a manual one for now, though we might also consider auto-generating one based on merged PRs.

I think we really need to decide on our branching in #340 too...

optimade.rst
merkys: Thanks! I like retaining consistency with other examples.

optimade.rst
merkys: Agree. Thanks!

merkys:
> Thanks for spotting this @merkys! Happy to accept, pending the changes suggested by @CasperWA (just ping me another review request if I need to re-accept)

Done!

> > I think a changelog (as suggested in #197) could cater for this purpose.
> 
> +1 for a rolling changelog that includes known/fixed bugs, I don't think a GH wiki page is visible enough (feels more like an internal knowledge base). I'm happy to spin up a manual one for now, though we might also consider auto-generating one based on merged PRs.

For starters, manual changelog would be just great!

> I think we really need to decide on our branching in #340 too...

Agree. Consensus for that was not a strong one (from my perspective), thus I'd be happy to hear other opinions.

ml-evs (approved):

CasperWA (approved):
Looks good now, thanks @merkys !

Let's move the changelog/bug list tracking discussion to an appropriate issue.

merkys:
Since there were no opposition to the status of a "bug fix", I have merged the PR. Regarding #340: this PR should appear in both the next patch and the next minor release.

PR #348, opened 2021-01-04
Allow boolean values in queries

### LINKED ISSUE #345: Boolean value representation in queries
Hi!

First of all thanks for all your hard work on this great project!

I'm just getting to know the different standards that Optimade provides, and I bumped into a problem with boolean values. The documentation (optimade.rst) at  2.1 says that there are the following basic types: *string, integer, float, boolean, timestamp*. In section 4.2. it is mentioned that booleans in responses should use the JSON standard. However, the standard is unclear on how boolean values should be represented in the queries. Section 6.1 discusses the different tokens, but also there is no mention of how boolean values should be presented. The grammar itself also does not seem to define a token for booleans.

So the question is whether I just somehow missed how booleans are to be defined, or if the official standard is missing? I guess none of the current entries are of boolean type, so this was never discussed. I would, however, like to apply the Optimade query language to a domain which also contains boolean values. My naive suggestion would be to go with the JSON style lowercase `true`/`false` values.

Thanks fo your time!

merkys:
Fixes #345 by introducing `TRUE` and `FALSE` grammar tokens, which can be used everywhere where values are used.

TO DO:

- [x] Describe tokens and their usage in the text.


optimade.rst
```diff
@@ -1566,6 +1568,19 @@ Examples:
 - :filter:`chemical_formula_anonymous CONTAINS "C2" AND chemical_formula_anonymous STARTS WITH "A2"`
 - :filter:`chemical_formula_anonymous STARTS "B2" AND chemical_formula_anonymous ENDS WITH "D2"`
 
+Comparisons of boolean values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Apart from straightforward comparisons ('=' and '!=') other usual comparison operators ('<', '>', '<=', '>=') MUST be supported.
+In such comparisons TRUE is held greater than FALSE.
```
rartino, line 1575: My opinion in #345 was to disallow relative comparison with booleans. What are the arguments for allowing this and not just say it is an error? It just seems a way to make mistyped queries evaluate.

rartino (commented):
This looks great - thanks for putting in the effort - but I wait with approval until we've discussed inequivalences of booleans.

Also, does this have to be a major version bump? (Note the discussion I just added to #366 about this - but I'm not convinced if the distinction matters here - could the addition of boolean constants in the grammar technically break clients?)

optimade.rst
```diff
@@ -1566,6 +1568,19 @@ Examples:
...
+
+Examples:
+
+- :filter:`true = TRUE`
+- :filter:`true != FALSE`
+- :filter:`true > FALSE`
+- :filter:`anything <= TRUE`: matches all entries.
```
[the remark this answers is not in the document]
rartino: I think the intent is that true is the name of a property that is equal to TRUE, but you are right that this is not so clear. I don't see why that property couldn't just be given another generic name, e.g.:
```suggestion
- :filter:`property = TRUE`
- :filter:`property != FALSE`
- :filter:`property > FALSE`
- :filter:`property <= TRUE` (note: this will match all entries.)
```

optimade.rst
ml-evs: I would prefer to disallow them too, unless someone can provide a good use case! To me, these comparison operators sort of imply that there is some expression to be evaluated rather than simply matching on a token. 

It may be worth adding a note that makes the distinction between query negation with `NOT` and `= FALSE` too, e.g., that `(nsites = 3 AND nelements = 3) = FALSE` is not a valid

fekad: I also think that only '=' and '!=' should be allowed to use (even I know it is a little bit harder to implement the grammar )

@ml-evs As far as I know we do not need to worry about the example that you presented because it is "grammatically" not even allowed. You can use parentheses only with `not`,`and` or `or`.

ml-evs: > @ml-evs As far as I know we do not need to worry about the example that you presented because it is "grammatically" not even allowed. You can use parentheses only with `not`,`and` or `or`.

It's a question of clarity/ergonomics for the human making queries, we can't expect them to read the grammar... the implementation will hopefully return an error but they are not necessarily intelligible. We can guard against it with a short note that bools cannot be used with expressions.

optimade.rst
```diff
@@ -2563,6 +2578,10 @@ The Filter Language EBNF Grammar
 
     Operator = ( '<', [ '=' ] | '>', [ '=' ] | [ '!' ], '=' ), [Spaces] ;
 
+    (* Boolean values *)
+
+    BooleanValue = ( 'TRUE' | 'FALSE' ), [Spaces] ;
+
```
fekad, line 2584: It might be worth writing the rules by following previous conventions (even if it looks unnecessary eg: tokens are defined separately...)
Something like this when you would allow only (not) equal operators:
```
PropertyFirstComparison = Property, ( ValueOpRhs
                                    | BooleanOpRhs
                                    | KnownOpRhs
                                    | FuzzyStringOpRhs
                                    | SetOpRhs
                                    | SetZipOpRhs
                                    | LengthOpRhs ) ;



BooleanOpRhs = BooleanOperator, ( TRUE | FALSE ) ;

(* Comparison operator tokens: *)

BooleanOperator = [ '!' ], '=' , [Spaces] ;

(* TOKENS *)

TRUE = 'TRUE', [Spaces] ;
FALSE = 'FALSE', [Spaces] ;
```

optimade.rst
merkys: I agree with you all that we should only support `=` and `!=` comparison operators for now.

> It may be worth adding a note that makes the distinction between query negation with `NOT` and `= FALSE` too, e.g., that `(nsites = 3 AND nelements = 3) = FALSE` is not a valid

A short note could be added, as @ml-evs suggests.

I will adjust the PR.

merkys:
> Also, does this have to be a major version bump? (Note the discussion I just added to #366 about this - but I'm not convinced if the distinction matters here - could the addition of boolean constants in the grammar technically break clients?)

I think this PR precipitates only minor version bump. Note that the changes in the filter grammar are merely an addition, thus all queries that worked before should work after merging this PR.

optimade.rst
merkys: While I am a little reluctant to further complicate the filter language grammar, I agree that forbidding other types of comparisons for boolean values in grammar makes sense. I will follow @fekad's suggestion right after I deal with #379.

optimade.rst
merkys: Note added in f14ed70ebb5697555b6e9c9e9579a7d492495ae1.

optimade.rst
merkys: I have tested and @fekad's suggestion seems to break the grammar: tests now emit the following error:
```
> Error: in tests/generated/Filter.g: line 136:
>     inherent ambiguity in production 'PropertyFirstComparison(7)'
>     starting with tokens "!", and "="
> 
> PropertyFirstComparison = Property ( ValueOpRhs | KnownOpRhs | FuzzyStringOpRhs | BooleanOpRhs | SetOpRhs | SetZipOpRhs | LengthOpRhs );
```
Thus it seems that the parser encounters an ambiguity between `ValueOpRhs` and `BooleanOpRhs` whenever it encounters `!` or `=` token. This could be remediated by something like this:
```
PropertyFirstComparison = Property, ( EqNeqOpRhs | LessGreaterOpRhs | KnownOpRhs | FuzzyStringOpRhs | SetOpRhs | SetZipOpRhs | LengthOpRhs ) ;
EqNeqOpRhs = [ '!' ], '=', [Spaces], ( Value | TRUE | FALSE ) ;
LessGreaterOpRhs = ( '<', '>' ), [ '=' ], [Spaces], Value ;
```
However, as said, I am not sure if this is worth the complication. It should be noted that not only `PropertyFirstComparison` has to be changed, but also `ValueList` and `ValueZip` leading to many additional ad-hoc rules.

fekad: It is unfortunate that the grammatica labels the rule as ambiguity (probably it uses some greedy method ). Lark parser doesn't have any issue building the syntax tree in python .

fekad (approved):
In practice, I agree that it is perfectly fine if the grammar allows some syntax even it is not a valid filter expression. I think it is mostly a question of implementation to decide where to handle the exceptions (in the grammar level or in the "filter parser")

ml-evs (approved):
Thanks for this @merkys! There's a draft implementation in https://github.com/Materials-Consortia/optimade-python-tools/pull/879/ (may not be up to date with your latest changes) so hopefully we can support this soon.

optimade.rst
rartino: I think Grammatica's limit for the grammar to be LL parsable is actually good for making sure we avoid computational overhead and require overly complex parsers.

sauliusg: > I have tested and @fekad's suggestion seems to break the grammar: tests now emit the following error:
> 
> ```
> > Error: in tests/generated/Filter.g: line 136:
> >     inherent ambiguity in production 'PropertyFirstComparison(7)'
> >     starting with tokens "!", and "="
> > 
> > PropertyFirstComparison = Property ( ValueOpRhs | KnownOpRhs | FuzzyStringOpRhs | BooleanOpRhs | SetOpRhs | SetZipOpRhs | LengthOpRhs );
> ```
> 
> Thus it seems that the parser encounters an ambiguity between `ValueOpRhs` and `BooleanOpRhs` whenever it encounters `!` or `=` token. This could be remediated by something like this:

Very strange, but I can not reproduce the problem – for me the definitions like @fekad suggested works. I'll ask @merkys for his specific version...

merkys: @sauliusg I have pushed my grammar with only @fekad's suggestion to https://github.com/merkys/OPTIMADE/tree/boolean-values-in-queries-schema. I can reproduce my initial problem with `inherent ambiguity in production 'PropertyFirstComparison(7)` by running `make` on Ubuntu-18.04 machine with #387 applied in order to get Grammatica compile successfully.

rartino: I've posted a PR against https://github.com/merkys/OPTIMADE/tree/boolean-values-in-queries-schema to implement something along the lines as proposed by @fekad above to fix the ambiguity.

optimade.rst
```diff
@@ -2479,7 +2494,7 @@ The Filter Language EBNF Grammar
 
     Constant = String | Number ;
 
-    Value = String | Number | Property ;
+    Value = String | Number | BooleanValue | Property ;
```
JPBergsma, line 2497: A minor point, but why do you use the term "BooleanValue" instead of the shorter "Boolean" ?
You dont write "StringValue" or "NumberValue" so this seems inconsistent.

merkys: Fair point. But I will wait for https://github.com/merkys/OPTIMADE/pull/1 before I rename grammar rules.

merkys: Changed in bc34afb.

optimade.rst
merkys: > I've posted a PR against https://github.com/merkys/OPTIMADE/tree/boolean-values-in-queries-schema to implement something along the lines as proposed by @fekad above to fix the ambiguity.

The consensus of today's web meeting was to merge the aforementioned branch here. I am marking the issue with the grammar as resolved now.

optimade.rst
```diff
@@ -1566,6 +1568,19 @@ Examples:
 - :filter:`chemical_formula_anonymous CONTAINS "C2" AND chemical_formula_anonymous STARTS WITH "A2"`
 - :filter:`chemical_formula_anonymous STARTS "B2" AND chemical_formula_anonymous ENDS WITH "D2"`
 
+Comparisons of boolean values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Straightforward comparisons ('=' and '!=') MUST be supported for boolean values.
+Other usual comparison operators ('<', '>', '<=', '>=') MUST NOT be supported.
```
JPBergsma, line 1575: I think you can leave out usual here, as unusual comparison operators are also not allowed. 
```suggestion
Other comparison operators ('<', '>', '<=', '>=') MUST NOT be supported.
```

merkys: Right, thanks for the suggestion.

optimade.rst
```diff
@@ -1566,6 +1568,19 @@ Examples:
 - :filter:`chemical_formula_anonymous CONTAINS "C2" AND chemical_formula_anonymous STARTS WITH "A2"`
 - :filter:`chemical_formula_anonymous STARTS "B2" AND chemical_formula_anonymous ENDS WITH "D2"`
 
+Comparisons of boolean values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Straightforward comparisons ('=' and '!=') MUST be supported for boolean values.
+Other comparison operators ('<', '>', '<=', '>=') MUST NOT be supported.
+It should be noted that boolean values are supposed to be used only in direct comparisons with properties, but not compound comparisons.
```
JPBergsma, line 1576: If we want the specification to be brief, it is perhaps better to write:
```suggestion
Boolean values should only be used in direct comparisons with properties, but not in compound comparisons.
```

merkys: Thanks for the suggestion! Accepted.

merkys:
I have merged in https://github.com/merkys/OPTIMADE/pull/2 introducing the operator-less boolean property comparison into the grammar, as suggested by @sauliusg in a web meeting a week ago. I believe the proposal now reflects the consensus achieved in that we meeting. Thus I ask for yet another review.

ml-evs (approved):

rartino (approved):
The grammar change has not been fully checked yet, but we merge this now because there are many other PRs depending on Boolean comparisons being available, and we are going to support them one way or the other. If the grammar as given here turns out to not be perfect, another PR should be filed to fix the issue in the grammar. Likely this shouldn't affect the end syntax at all.

PR #349, opened 2021-01-04
Explicitly declare relations between 'elements', 'nelements' and 'elements_ratios'

merkys:
Reading the specification once more I find the relations between `elements`, `nelements` and `elements_ratios` lacking explicit declaration:

1. Lists of `elements` and `elements_ratios` are correlated;
2. `nelements` is equal to lengths of `elements` and `elements_ratios`.

These relations can be derived from the examples, or the general knowledge, but I believe it would be clearer to have them written explicitly.


ml-evs (dismissed):
Thanks @merkys, I think this is a useful clarification.

ml-evs:
Did anyone else want to take a look at this before 1.0.1? @rartino/@CasperWA?

optimade.rst
```diff
@@ -1821,6 +1821,7 @@ elements
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
   - The strings are the chemical symbols, i.e., either a single uppercase letter or an uppercase letter followed by a number of lowercase letters.
   - The order MUST be alphabetical.
+  - MUST be correlated with the list property `elements\_ratios`_.
```
CasperWA, line 1824: What do you mean by/is your distinct definition of "correlated" here?

optimade.rst
```diff
@@ -1841,6 +1842,7 @@ nelements
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
+  - MUST be equal to the lengths of the list properties `elements`_ and `elements\_ratios`_.
```
CasperWA, line 1845:
```suggestion
  - MUST be equal to the lengths of the list properties `elements`_ and `elements\_ratios`_, if they are provided.
```

optimade.rst
```diff
@@ -1863,6 +1865,7 @@ elements\_ratios
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
   - Composed by the proportions of elements in the structure as a list of floating point numbers.
   - The sum of the numbers MUST be 1.0 (within floating point accuracy)
+  - MUST be correlated with the list property `elements`_.
```
CasperWA, line 1868: Same question as above.

I think we have other examples of this happening, e.g., `species_at_sites` and the `cartesian_site_positions`. Maybe we can write here something similar as to what has been done for those? (Or vice versa?)

CasperWA (commented):
Two things:
- I am not completely comfortable with the "correlated" sentences, as the "correlation" is not defined.
- I think we should make clear that the `MUST` requirement between properties are only valid as long as both properties are given, i.e., not `null`.
  This could also be done via an introductory sentence at the top of the section, stating something like "all inter-property requirements are only valid as long as both properties are provided, i.e., neither have an unknown value"?

optimade.rst
ml-evs: We do refer to "correlated" with this meaning in the specification already, e.g. in [Comparisons of list properties](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#id47) when defining the `list1:list2:... HAS val1:val2:...` syntax, and in the section [Filtering in relationships](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#id49)

merkys:
> * I am not completely comfortable with the "correlated" sentences, as the "correlation" is not defined.

As @ml-evs wrote, term "correlation" was already in use in the specification of v1.0.0. In addition to use in section for filtering, it is also mentioned in [Structures entries](https://github.com/Materials-Consortia/OPTIMADE/blob/39f2a8841bd43f875b9e76567716c7f02b001707/optimade.rst#structures-entries) section to emphasize such relation between various lists.

The situation could be possibly improved by describing "correlated lists" in the definition of terms section. Maybe it could also cite Wikipedia page on [parallel array](https://en.wikipedia.org/w/index.php?title=Parallel_array&oldid=1008210954) which describes the idea pretty well IMO. Maybe we could even switch to use "parallel lists" if this term is more intuitive/used more widely.

> * I think we should make clear that the `MUST` requirement between properties are only valid as long as both properties are given, i.e., not `null`.
> This could also be done via an introductory sentence at the top of the section, stating something like "all inter-property requirements are only valid as long as both properties are provided, i.e., neither have an unknown value"?

I agree, this would bring some more clarity to the specification. Of course, first we have to make sure whether this is a truth global to the specification :)

optimade.rst
merkys: I would probably go with describing a global rule:
> the MUST requirement between properties are only valid as long as both properties are given, i.e., not null.

We may add "if not described otherwise" just to be safe.

optimade.rst
rartino: I understand the formal connection to "correlated lists", but agree that it isn't easy to understand for someone who isn't as familiar with the specification as we are. Why not just put these criteria like this?:
```suggestion
  - MUST refer to the elements in the same order as `elements\_ratios`_.
```
(There is actually a precedent on using this type of formulation in the description of `species_at_sites`)

rartino: Ping @CasperWA Since you lifted this issue, when you have time I am curious about your thought about my suggestion here to hopefully get this issue to move forward.

merkys: So it seems we have two ways to define the same relation between properties, either using term "correlated lists" or avoiding it. Shall we phase out the term "correlated list" altogether? If so, does the definition of "same order" conveys the meaning that both lists have to be of the same length? If not, I would vote to promote "correlated lists" to an in-specification term with meaning "lists of the same length with the same order of elements".

CasperWA: TL;DR: I vote to not use "correlated", but instead spell out the requirements for _all_ fields in the way suggested by @rartino and echoed by @merkys in his suggested definition of how "correlated lists" might be defined.

I prefer @rartino's suggestion over "correlated", with some exceptions. So I think going to a field and reading its requirements should be self-contained, i.e., the requirement should be listed for "both ends" of the correlated fields. Even though it's repetition, it will (hopefully) be painfully clear to implementers, as well as being nice when looking up specific fields to understand how they interact with any of the other fields.

Concerning "correlated". The issue with that word is that, yes, we can define it in the way you suggest @merkys, but that is a poor representation of what the word _can_ mean, and in the end I still think it will be more confusing than explanatory of what we want to convey.
Correlated can also be read to understand that the "meaning" of the values are in some way theoretically related to each other. It can be that when one content value changes, a content value in the other field should/might also change (which is definitely not the case for all fields). In short, I think it's simply the wrong word choice for this narrow and very specific meaning.

CasperWA: Note, if you truly despise spelling this out, I am not opposed to defining a different term and using that instead (like "parallel list" suggested by @merkys below). My only issue is with the use of the word "correlated", which I find to be a poor choice for this particular use case.

optimade.rst
merkys: For now I have accepted @CasperWA's suggestion. I feel that tampering with global rules and definitions would further complicate the acceptance of this PR. It is better to leave global definitions to later PRs.

optimade.rst
merkys: Fixed this instance. I have also replaced "correlated" with a more intuitive formulation here.

CasperWA (commented):
So `nelements` only has to be equal length but not refer to the other fields with the same order and `elements` and `elements_ratios` does not have to have equal length to the others? 😅

merkys:
> So `nelements` only has to be equal length but not refer to the other fields with the same order and `elements` and `elements_ratios` does not have to have equal length to the others? sweat_smile

Oops, will change this as well!

rartino (dismissed):
I like this version; this way of expressing the "correlation" criteria is fully clear to me.

CasperWA:
> > So `nelements` only has to be equal length but not refer to the other fields with the same order and `elements` and `elements_ratios` does not have to have equal length to the others? sweat_smile
> 
> Oops, will change this as well!

Did you implement these changes yet? I don't think I've seen it yet. I can do it though? :)

merkys:
> > > So `nelements` only has to be equal length but not refer to the other fields with the same order and `elements` and `elements_ratios` does not have to have equal length to the others? sweat_smile
> > 
> > 
> > Oops, will change this as well!
> 
> Did you implement these changes yet? I don't think I've seen it yet. I can do it though? :)

I just fixed them. Sorry for the delay.

optimade.rst
```diff
@@ -1821,6 +1821,7 @@ elements
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
   - The strings are the chemical symbols, i.e., either a single uppercase letter or an uppercase letter followed by a number of lowercase letters.
   - The order MUST be alphabetical.
+  - MUST be of equal length as `elements\_ratios` and refer to the elements in the same order as `elements\_ratios`_, if the latter is provided.
```
rartino, line 1824: Sorry for doing this... Would this be shorter and even more clear (?) written as this:
```suggestion
  - MUST refer to the same elements in the same order as `elements\_ratios`, if the latter is provided.
```

optimade.rst
```diff
@@ -1863,6 +1865,7 @@ elements\_ratios
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
   - Composed by the proportions of elements in the structure as a list of floating point numbers.
   - The sum of the numbers MUST be 1.0 (within floating point accuracy)
+  - MUST be of equal length as `elements` and refer to the elements in the same order as `elements`_, if the latter is provided.
```
rartino, line 1868:
```suggestion
  - MUST refer to the same elements in the same order as `elements`_, if the latter is provided.
```
(I would suggest waiting for @CasperWA to agree/disagree on this modification, unless @merkys already has a reason to outright dismiss it.)

merkys: I myself would prefer "correlated lists" or "parallel lists" wordings due to their brevity and already existing usage (of the former) in the specification. However, I admit that these wordings may not be as clear to reader as they are to me.

rartino: But this discussion we already had? I agree with @CasperWA that "X MUST be correlated to Y" just reads less clearly than "X MUST refer to the same elements in the same order as Y".

(If we really want to go in the direction of defining "correlated lists" as a concept throughput the specification, then it needs to be defined under 'Definition of terms', the query section updated to refer to that definition, and other uses of 'correlated' mostly removed - e.g., in the definition of assemblies. The last point is going to be tricky.)

CasperWA: See my answer above [here](https://github.com/Materials-Consortia/OPTIMADE/pull/349#discussion_r596133053).

optimade.rst
rartino: No... I think you miss the formal difference of the addition of an extra "same" in the sentence: "MUST refer to the elements in the same order as `elements\_ratios`" vs. "MUST refer to the **same** elements in the same order as `elements\_ratios`".

I'm arguing that latter implies that the lengths must be equal and is possibly even stronger. (Just saying that the lengths must be equal and the order the same could allow referring to the same element twice and leaving another one out...)

optimade.rst
ml-evs: Do we have any consensus here? This PR is now the last hold-up for 1.0.1, I think, and it would be nice to get this in before the next meeting so we can agree on the release.

merkys: There doesn't seem to be a consensus here, I think. However, I doubt that this PR should be blocking 1.0.1. It merely attempts to make the specification a little clearer. So I guess it's OK to merge it in later.

rartino: If there are strong feelings that 

  - MUST be of equal length as `elements` and refer to the elements in the same order as `elements`_, if the latter is provided.

is better than

  - MUST refer to the same elements in the same order as `elements`_, if the latter is provided.

then let us go with the former. This comment thread that we have this discussion in was started by me, but it was just an attempt to clean up what seemed redundant and unnecessarily complex. If people do not agree with that cleanup, then we have consensus for the old formulation - I think.

There are strong feelings against the early "correlated lists" formulation, so I assume we won't go with that unless @merkys feel strongly enough for that formulation to champion the issue further (?)

merkys: > There are strong feelings against the early "correlated lists" formulation, so I assume we won't go with that unless @merkys feel strongly enough for that formulation to champion the issue further (?)

I will decouple the issue of "correlated lists" to a separate discussion.

rartino: New suggestion from call:
```suggestion
  - MUST refer to the same elements in the same order as `elements`_, and therefore be of the same length, if the latter is provided.
```
(I would suggest waiting for @CasperWA to agree/disagree on this modification, unless @merkys already has a reason to outright dismiss it.)

merkys: I think the new suggestion is really an improvement with regard to earlier wording. I will change all occurrences of the previous wording to this, and hope that this will be OK for now.

ml-evs (approved):
I have just fixed the conflict that I added inadvertently in #358, and I am happy with this wording.

shyamd:
I'll throw a wrench into the discussion. 
`nelements` should equal `len(elements)`, this is fine.
In an experiment with known impurities, a DB might want to list those impurities but not include them in the `element_ratios`, so then the list length would not match between `elements` and `element_ratios`

ml-evs:
> I'll throw a wrench into the discussion.
> `nelements` should equal `len(elements)`, this is fine.
> In an experiment with known impurities, a DB might want to list those impurities but not include them in the `element_ratios`, so then the list length would not match between `elements` and `element_ratios`

This is a good point that could do with some clarification (I would hope in another PR ;)). The statistical aspects of compositions are handled by `species` and `species_at_sites`, but there is currently no linkage from `species_at_sites` to `elements`, e.g. we do not require that `set(symbol for species in entry.attributes.species for symbol in species.chemical_symbols) == entry.attributes.elements`. Of course, impurity could mean 10% or it could mean 0.00001%...

I would suggest that given the existing interpretation (or lack thereof) of the specification, we should still merge this pending further discussion. This requirement has been implicit (and is coded up in the python-tools models, for example); if we decide to relax it based on this impurity use-case then that will not be a breaking change.

CasperWA (approved):
Let's get this in. It's been discussed extensively and I agree with the current edition. Cheers @merkys ! :)

rartino (approved):
Oh, didn't we merge this yet? I approve the present version.

merkys:
Thanks for approving and merging this!

ml-evs:
I have just opened #361 to carry on @shyamd's discussion point

PR #352, opened 2021-03-04
Update OpenAPI schemas according to OPTIMADE Python tools v0.13.3

CasperWA:
This is yet another update to the OpenAPI schemas.

The changes are related to the update of the Python package used to create the schemas, `pydantic`, where OPTIMADE Python tools now depends on the latest version (v1.8.1) instead of being fixed at an older version (v1.6).

The related OPTIMADE Python tools PR: Materials-Consortia/optimade-python-tools#731.
The changes are related to enumerations (Python Enum sub-classes), extending information about them and defining the default value for the `aggregate` field attribute for a `links` resource.

### DIFF
--- a/schemas/index_openapi_schema.json
+++ b/schemas/index_openapi_schema.json
@@ -2,7 +2,7 @@
   "openapi": "3.0.2",
   "info": {
     "title": "OPTIMADE API - Index meta-database",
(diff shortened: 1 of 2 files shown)


merkys (approved):

CasperWA:
> I'm holding off reviewing this as I'm quite closely involved in the changes...

So you don't agree with them? 😏

rartino (approved):
The changes look reasonable to me. It isn't super easy to properly review these, as that would require importing them and run validations against them (which I did not do). I think that eventually I will be able to do that with my implementation, though.

CasperWA:
> (...) I think that eventually I will be able to do that with my implementation, though.

Intriguing. Will this implementation be part of a repo in this GH organization or is it something different?

rartino:
> Intriguing. Will this implementation be part of a repo in this GH organization or is it something different?

There will be a validator as part of the httk OPTIMADE implementation capable of validating responses based on a schema (there actually already was such a validator in the old optimade-python-candidate-reference-implementation...)

PR #356, opened 2021-04-20
Fix examples involving comparing `elements` to string

### LINKED ISSUE #355: elements="Si,O2" in filter language examples
There are examples in the specification having filter values containing `elements="Si,O2"` (legal filter language expression). As `elements` is a list property and per "Type handling and conversions in comparisons" cannot be compared to string values (MUST be reported as HTTP 501), I suggest replacing `elements="Si,O2"` with something that would normally be supported by providers, for example, `chemical_formula_reduced="ClNa"`.

merkys:
Fixes #355.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1412,11 +1412,11 @@ In particular, this means the client MUST escape special characters in string va
 
 Examples of syntactically correct query strings embedded in queries:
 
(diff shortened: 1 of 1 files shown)


ml-evs (approved):
This looks like a good correction, thanks @merkys

CasperWA (approved):
Cheers @merkys

PR #359, opened 2021-05-28
improved layout section 7.4

JPBergsma:
The section requirements previously had the same indentation level as the listed properties. 
I changed the indentation level so it would be clear that the requirements do not belong to the properties.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2328,24 +2328,24 @@ References Entries
 
 - **Description**: The :entry:`references` entries describe bibliographic references. The following properties are used to provide the bibliographic details:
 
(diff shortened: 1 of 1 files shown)


ml-evs:
Thanks for this @JPBergsma, but I'm not entirely sure that this is correct. Section 7.4 References describes a separate entry type (following 7.2 Structure and 7.3 Calculations) and thus should be indented less than e.g. 7.2.12 structure_features. The formatting of 7.4 is a bit different to 7.2 because we list all of the references attributes inline, as opposed to giving them a subsection each.

We could use this PR to embolden "Example:" and fix the formatting of the `structure_features` example (which seems to appear as quoted text after a ">" for me), if you wish, but the indent changes should not be merged.

Issue #343, opened 2020-12-09
Extending the list of standard response formats (i.e. CIF, POSCAR)


merkys:
There have been discussions about exposing CIF data (#329) or any entry-related files (#211) through OPTIMADE. Now I remembered the provision of endpoint-specific `response_format` which could be used to ask for specific representations of entries or their collections. Up to now OPTIMADE only supports `json` response format. I suggest standardizing other optional response formats that could be used by clients to retrieve representations of entries, and my immediate suggestions would be CIF (`cif`, native structure data format for the COD) and POSCAR (`poscar`). Upon request for `response_format=cif` client would receive plain CIF file for an entry (or several entries) and could directly use it in calculations etc.

It may also be true that I misunderstood the intention of `response_format`. Maybe it is supposed to specify a carrier format for OPTIMADE data. Then foreseen additions to existing `json` would be `xml` and `yaml`, which would carry the same property information wrapped in their specific syntax. If this is the case, then #211 is worth pursuing to provide other representations of entries instead.

rartino:
I think this is a very nice idea, and very much in line with what I imagine `response_format` is useful for.

However, to actually include `response_format=cif` in the standard requires a bit of work. IMO we must specify exactly how all standard OPTIMADE entries are mapped into the cif format, including things like relationships between entries (although, for anything optional it could say `output_format=cif` does not support it and it MUST be omitted for now). This to ensure that every database returns consistent cif output. You may say "cif is already a standard" and obviously the mapping should adhere completely to that standard. But there are definitely things in the OPTIMADE -> cif mapping that isn't obvious, and it would be a shame to end up with 43 dialects of cif output all using their own representation of, e.g., `assemblies`.

May I suggest that you start with creating a `response_format=_cod_cif` to work out the details of this mapping. I'd be more than happy to discuss ideas about how to best do these various mappings, but the decisions would be all yours since this is under your namespace.

Then, at some future point, you'll come to OPTIMADE with your design and propose to adopt it as the standard for `response_format=cif`. Perhaps then we are happy to adopt it exactly as designed, or we will see a need to consider some adjustments to meet requirements of others in the consortium. But we can take that discussion at that point.

gmrigna:
I also think that it is a very nice idea... It might be worth looking at some previous work by @merkys (https://github.com/cod-developers/CIF2JSON)

merkys:
[...]

Issue #34, opened 2018-06-14
Trajectories


davidwaroquiers:
Trajectories should be a simple extension of the structure API (see Issue #23)

Suggestion :

`lattice_vectors_trajectory` : list of each lattice_vectors-like (as defined in structure, see Issue #23) at each time step or relaxation step
`cartesian_site_positions_trajectory` : list of each cartesian_site_positions-like (as defined in structure) at each time step or relaxation step

For trajectories, the following (optional) fields can be given :
`md_start_time` : Starting time of the trajectory (suggested unit : fs)
`md_time_step` : Time step of the MD trajectory (if time step is constant), in suggested unit fs
`md_times` : If variable time steps are allowed, list of times of the MD trajectory, in suggested unit fs

We suggest that the rest of the structure field stays the same (dimension_types, species, ...). Assemblies could also be used for e.g. Path Integral Molecular Dynamics.

rartino:
So, I guess the suggestion is that the above defines a `trajectory` entry, i.e., the above + what is in structure, but with `lattice_vectors` *replaced* by `lattice_vectors_trajectory` and `cartesian_site_positions` *replaced* by `cartesian_site_positions_trajectory`?

(Or is the proposal that these can be added on top of a structure?)

It was brought up in the discussion that this may lead to single records that are too big to send over 'in one go'. My suggestion was, in such a situation, to rely on the mechanisms in http for partial transfers, e.g., https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206

Issue #24, opened 2018-06-14
Calculation entry standardization


rartino:
We need to standardize mandatory and optional the fields in the calculation entry type.

rartino:
Here is an attempt at a concrete suggestion:
```
{
  ... <other response items> ...
  "data": [
    {
      "type": "calculation",
      "id": "example.db:calcs:0001",
      "attributes": {
        "local_id": "example.db:calcs:0001",
        "url": "http://example.db/calcs/0001",
        "immutable_id": "http://example.db/calcs/0001@123",
        "last_modified": "2007-04-05T14:30Z",
        "calculation_description" = "relaxation",
        "related_structures_descriptions": ["initial", "final"],
        "related_structure_ids" = ["example.db:struct:0042", "example.db:struct:4711"], 
      }
    }
  ]
}
```

rartino:
It was said during the discussion that this probably cannot be decided without another meeting.

kxyyang:
Should we think about how we make a query ? For example, if I want to get the data of BAND GAP, should we enter the keyword  "band_gap" , or " BandGap" , or "BGap", ... etc. Is it discussed somewhere else ?   Or we have already had the specification formulated ?

rartino:
Looking at other parts of the specification, e.g., #23, makes it seem likely to become `band_gap` over the other alternatives you suggest. But, we presently have no agreement of a common definition of `band_bap`. I suspect we will go down the road of, e.g., `band_gap_kohn_sham`, `band_gap_optical`, `band_gap_fermi_liquid_quasiparticle`, etc. But I expect that we'll need a discussion in a meeting to sort this out.

PR #360, opened 2021-06-08
Files endpoint

### LINKED ISSUE #211: Links to files
In the COD the primary description of a structure is a Crystallographic Information Framework (CIF) file. Thus we at the COD desire a standard method to present a link to such file in OPTiMaDe entries. In JSON API format I suggest using [links](https://jsonapi.org/format/#document-links), for example:
```json
{
  "type" : "structure",
  "id" : "1234567",
  "links" : {
    "file" : {
      "href" : "https://www.crystallography.net/cod/1234567.cif",
      "meta" : {
        "type" : "chemical/x-cif"
      }
    }
  }
}
```
Thus `links/file` would be a JSON API *link object*. A *link object* would contain `href` field with URL and `meta/type` field with [media type string](https://en.wikipedia.org/wiki/Media_type).

merkys:
In #211 the need to provide links to files from entries has been discussed. @rartino proposed introducing `files` endpoint as the most universal way to deal with files, recycling JSON API relationships to provide links between files and other types of entries.

Differences from suggestions in #211:
* <s>`url` attribute is merely a string instead of JSON API links object, as I cannot think of a use case where the links object would be of use;</s>
* I did not formalize means to describe link attributes between other entries and `files`. This could be done later on in other PRs.

Closes #211. Addresses the original intent of #343 superseding it.

**Edit**: `url` redefined as JSON API links object.


optimade.rst
```diff
@@ -2462,6 +2522,41 @@ Relationships with calculations MAY be used to indicate provenance where a struc
 
     At the moment the database providers are suggested to extend their API the way they choose, always using their database-provider-specific prefix in non-standardized fields.
 
+Files
+~~~~~
+
+Relationships with files MAY be used to relate an entry with any number of :entry:`files` entries.
+Alternatively, they MAY be used to provide file representations of entries.
```
CasperWA, line 2529: I'm not sure `MAY` is necessary in these sentences? In that I'm not sure these lines specify anything other than present an example, or rather, outline the example presented below?

Also, how are these two sentences different? To me it just sounds like you're saying a relationship with `files` can include one or many `files`. What it means if it's a single one or several will eventually be up to the client to decide, I guess?

Also - perhaps add an example with a reference and an article file? Maybe even one to our own reference of the OPTIMADE API? 😅

CasperWA (commented):
A single comment for now. I'll do another review later with pedantic syntax corrections when we all start to agree that this should go in.

Great effort on this though - and I'm not at all against an explicit `/files` endpoint. It makes it more specific and clear where to place these kind of entries - instead of in a `meta` or as a `links` value... Although, the latter there might be a solution if a single file represents any one entry (e.g., a CIF file for a structure, or a single paper for a reference).

Finally, for completeness I'd like to keep the JSON API philosophy of keeping all links or rather URL-related values as a mix of a string, a Links object, and `null`. For the specific attribute `url` here, it's fine to not allow `null`, but it doesn't make sense to me to disallow the use of a Links object - other than just because. It might be that one wants to specify a `meta` value for the links that this links only works locally on a specific computer (which is very weird) or that the link points to an external resource. E.g., I could think of a case where a structure in an AiiDA DB comes from COD, so the file is kept and the `url` can then both point to the COD link and state as much.

merkys:
> A single comment for now. I'll do another review later with pedantic syntax corrections when we all start to agree that this should go in.

Thanks!

> Great effort on this though - and I'm not at all against an explicit `/files` endpoint. It makes it more specific and clear where to place these kind of entries - instead of in a `meta` or as a `links` value... Although, the latter there might be a solution if a single file represents any one entry (e.g., a CIF file for a structure, or a single paper for a reference).

I agree that single file could be linked more easier, but I prefer more homogeneous approach.

> Finally, for completeness I'd like to keep the JSON API philosophy of keeping all links or rather URL-related values as a mix of a string, a Links object, and `null`. For the specific attribute `url` here, it's fine to not allow `null`, but it doesn't make sense to me to disallow the use of a Links object - other than just because. It might be that one wants to specify a `meta` value for the links that this links only works locally on a specific computer (which is very weird) or that the link points to an external resource. E.g., I could think of a case where a structure in an AiiDA DB comes from COD, so the file is kept and the `url` can then both point to the COD link and state as much.
> 
> To counter-argue my last point, one could say that this information could also be put directly into the `files`-entry's `meta` field (that's at the same level as `id`, `type`, `attributes`, etc.)

My motivation for describing `url` as string is both the lack of a use case for URL-less file and the wish to promote stuff from `meta` to `attributes`. But I understand the wish to be homogeneous here too (BTW, `references`.`url` is also defined as a string to remain close to BibTeX).

optimade.rst
merkys: > I'm not sure `MAY` is necessary in these sentences? In that I'm not sure these lines specify anything other than present an example, or rather, outline the example presented below?

I took these `MAY`s from the description of relations with Calculations entries. But you are right, there is no need to use them.

> Also, how are these two sentences different? To me it just sounds like you're saying a relationship with `files` can include one or many `files`. What it means if it's a single one or several will eventually be up to the client to decide, I guess?

Right, they could be condensed into one short sentence.

> Also - perhaps add an example with a reference and an article file? Maybe even one to our own reference of the OPTIMADE API? sweat_smile

Hmm. Not really sure this would be of much use to anyone. References entries have `url` attribute which can link to either publication landing page or the article file itself. What would probably be more useful is a Calculations entry with both input and output file - something closer to the AiiDA provenance graph.

ml-evs:
Looks like the CI was initially failing here due to the Fastly outage (which impacted PyPI), everything seems okay now

optimade.rst
```diff
@@ -2373,6 +2373,66 @@ Example:
...
+- **Type**: string.
+- **Requirements/Conventions**:
+
+  - **Support**: MUST be supported by all implementations, MUST NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - **Response**: REQUIRED in the response.
+
+- **Examples**:
+
+  - :val:`"https://www.crystallography.net/cod/1000000.cif"`
```
rartino, line 2395: We have so far avoided putting URLs to specific databases in the specification, which is a policy I think we should adhere to for several reasons. So I suggest:
```suggestion
  - :val:`"https://example.org/files/cifs/1000000.cif"`
```

optimade.rst
```diff
@@ -2464,6 +2524,40 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+          }
+        }
+      },
+      "included": [
+        {
+          "type": "files",
+          "id": "example.db:files:1234",
+          "attributes": {
+            "media_type": "chemical/x-cif",
+            "url": "https://example.db/files/1234.cif"
```
rartino, line 2555: Always use example.org / .com for example URLs rather than making up URLs that could exist.
```suggestion
            "url": "https://example.org/files/cifs/1234.cif"
```
(Or is there a rule that *every* top-level domain MUST have a reserved domain on the form example.top-domain?)

optimade.rst
```diff
@@ -2373,6 +2373,66 @@ Example:
...
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+- **Examples**:
+
+  - :val:`"chemical/x-cif"`
+
+version
+~~~~~~~
+
+- **Description**: version information of a file (e.g. commit, revision, timestamp)
```
rartino, line 2428: Would it not be a good idea to have a separate optional, but standardized and computer-interpret able, `modification_timestamp`field for modification time? With something along the lines that if the field is provided it MUST change when the file contents changes. (However, the guarantee does not apply the other way, the file is allowed to remain the same even if the timestamp changes.)

rartino (dismissed):
Looks good, except for some URL adjustments and a suggested addition.

Also a question: what is it in the present specification that says which entry types/endpoints are optional to implement? Based on what in the specification can I presently decide not to implement the `/files` endpoint?

optimade.rst
merkys: I will use `example.org` from now on, thanks. As for `example.<TLD>` I don't know, but I guess we can get around with `example.org`.

optimade.rst
merkys: Indeed. Section "Properties Used by Multiple Entry Types" lists `last_modified` as the modification timestamp for an entry, maybe it could be recycled for Files as well. Probably in this case there is a need to repeat the description from "Properties Used by Multiple Entry Types" also listing additional constraints?

optimade.rst
```diff
@@ -2371,6 +2371,66 @@ Example:
...
+- **Type**: string.
+- **Requirements/Conventions**:
+
+  - **Support**: MUST be supported by all implementations, MUST NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - **Response**: REQUIRED in the response.
+
+- **Examples**:
+
+  - :val:`"https://example.org/files/cifs/1000000.cif"`
```
ml-evs, line 2393: Should we also allow [JSON:API links objects](https://jsonapi.org/format/#document-links) too? i.e., the type is either a string, or a links object (which can have a `meta` field for arbitrary metadata)

I think we take this approach in all other places in the specification.

optimade.rst
```diff
@@ -2371,6 +2371,66 @@ Example:
...
+
+name
+~~~~
+
+- **Description**: base name of a file.
```
ml-evs, line 2398: Is the intention that the file MUST have this name if I were to download it? We should make that explicit if so.

optimade.rst
```diff
@@ -2371,6 +2371,66 @@ Example:
...
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+- **Examples**:
+
+  - :val:`"1000000.cif"`
+
+media\_type
+~~~~~~~~~~~
+
+- **Description**: media type identifier for a file.
```
ml-evs, line 2412: Can we add a link to e.g. [RFC 6838](https://datatracker.ietf.org/doc/html/rfc6838) (or something more appropriate) and indicate that this media type "SHOULD" follow IANA standards (i.e. follow your example with `chemical/x-cif` rather than just `cif`).

optimade.rst
ml-evs: I would like to serve files on-the-fly by adding relationships to all structures such that various file types can be spat out - what value should I use for `last_modified` here? I guess the query time stamp might work...

optimade.rst
ml-evs: And as a follow-up, using media types is fine but if someone wants to serve e.g. input/output files for various DFT codes, using `text/plain` might not have high enough fidelity to be useful, yet defining something custom like `application/vnd.vasp.poscar` application might be confusing to user's not versed in all of the various standards!

ml-evs (commented):
Thanks for writing this up @merkys, we should definitely add this functionality to OPTIMADE somehow.

I have two general comments (with more specific comments below):

1) It looks like the aim isn't really to allow querying on a `files` endpoint, so perhaps we could simplify this by just adding the fields described here into the `meta` of a "special" kind of JSON:API link entry? That way, instead of needing to define a new entry type (with corresponding endpoints), files can be served as JSON:API `links` (per structure entry), rather than the `relationships`/`included` route.
2) If we decide to go ahead with this approach (rather than my point above), does that make a `/files` endpoint mandatory? I would rather not maintain a database of files but instead serve everything on-the-fly (I think this would be a common use case for people with computational data). In that case, my `/files` endpoint might return no data, but a particular file id `/files/structure/1.cif` might have a response (where the file ID is `structure/odbx/1.cif`)...

optimade.rst
merkys: @CasperWA had the same question, to which I answered in https://github.com/Materials-Consortia/OPTIMADE/pull/360#issuecomment-856627845:
> My motivation for describing `url` as string is both the lack of a use case for URL-less file and the wish to promote stuff from `meta` to `attributes`. But I understand the wish to be homogeneous here too (BTW, `references`.`url` is also defined as a string to remain close to BibTeX).

optimade.rst
merkys: I think that servers should avoid [`Content-Disposition: attachment; filename="..."`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) HTTP header, using `Content-Disposition: inline` (default, so could be omitted) instead. With latter HTTP header it is up to the browser to set the file name, and I believe most of the browsers will give the base name of the URL.

optimade.rst
merkys: I think a "MUST" for RFC 6838 would be more appropriate here. IMO, allowing custom values undermines the benefit of such fields. I understand that some file formats would be left out by a "MUST", but I cannot think of a better solution now.

optimade.rst
merkys: > I would like to serve files on-the-fly by adding relationships to all structures such that various file types can be spat out - what value should I use for `last_modified` here? I guess the query time stamp might work...

In theory, the version of a generated file depends on versions of a bunch of things:
* input
* generator software
* environment

As the last two change rarely (I suppose), it is often safe to say that contents of the generated file depend (bit-by-bit) only on the contents of the input file. Thus you may give the version of the input, and the client will be able to cache instead of downloading every file anew each time. However, query time stamp is a safe bet when we really care about users not missing any changes.

optimade.rst
ml-evs: Perhaps then a second field for a human-readable file type or description? This info could be incorporated into a general description field, I guess.

optimade.rst
ml-evs: Apologies, I didn't dig through the other review comments.

optimade.rst
merkys: I like the idea of general `description` field. Will introduce it.

merkys:
> I have two general comments (with more specific comments below):
> 
> 1. It looks like the aim isn't really to allow querying on a `files` endpoint, so perhaps we could simplify this by just adding the fields described here into the `meta` of a "special" kind of JSON:API link entry? That way, instead of needing to define a new entry type (with corresponding endpoints), files can be served as JSON:API `links` (per structure entry), rather than the `relationships`/`included` route.

I think we have discussed this approach in #211 (it was my initial suggestion), and the consensus there was that `/files` endpoint would be more homogeneous with the rest of the specification.

> 2. If we decide to go ahead with this approach (rather than my point above), does that make a `/files` endpoint mandatory? I would rather not maintain a database of files but instead serve everything on-the-fly (I think this would be a common use case for people with computational data). In that case, my `/files` endpoint might return no data, but a particular file id `/files/structure/1.cif` might have a response (where the file ID is `structure/odbx/1.cif`)...

I see your point in not serving the `/files` endpoint (entry listing might just be pointless). However, they seem mandatory as per the specification:
> Entry listing endpoints return a list of resource objects representing entries of a specific type.

> "Entry listing" endpoint response dictionaries MUST have a :field:`data` key.
> The value of this key MUST be a list containing dictionaries that represent individual entries.

On the other hand, if served on-the-fly, entry listing could be easily populated with just minimal data:
```json
{
    "id": "1.cif",
    "type": "files",
    "attributes": {
        "url": "/files/structure/1.cif"
    }
}
```

sauliusg:
Just a thought – it is usually very useful when files in any file system have as a subset attributes mandated by POSIX; in particular atime, mtime, ctime. Should we maybe include them?

merkys:
> Just a thought – it is usually very useful when files in any file system have as a subset attributes mandated by POSIX; in particular atime, mtime, ctime. Should we maybe include them?

I can in principle introduce provisions for optional `atime`, `mtime`, `ctime` fields.

optimade.rst
merkys: > I like the idea of general `description` field. Will introduce it.

Introduced in f3ced3e9ff92513e8acbbbfb75c3f4ae85fe331a. Marking the conversation as resolved, hope it is OK with you, @ml-evs?

merkys:
@CasperWA @ml-evs @rartino @sauliusg Are there any blockers left for this PR? Please let me know if I did not address any of your comments.

I think standardizing of the way we link in files is of relatively high priority. For example, we at the COD very much would want to provide CIF files for the COD entries, since CIF files are "the ultimate source of truth" for the crystal data at the COD.

optimade.rst
```diff
@@ -2371,6 +2371,110 @@ Example:
...
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+- **Examples**:
+
+  - :val:`"POSCAR format file"`
+
+atime
+~~~~~
+
+- **Description**: time of last access of a file.
```
ml-evs, line 2451: I'm not sure I see the use-case for `atime` in this API - does it track the last time the file was downloaded (which seems like a heavy implementation burden) or the last time it was accessed on disk (which seems irrelevant to an API user)?

optimade.rst
```diff
@@ -2371,6 +2371,110 @@ Example:
...
+
+- **Description**: time of last status change of a file.
+- **Type**: timestamp.
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+mtime
+~~~~~
```
ml-evs, line 2469: Whilst I get the appeal of using the precise `stat`-derived terms, we already have the `last_modified` field for entries. Do they really need to be distinct? I guess `last_modified` refers to the OPTIMADE entry, should we consider the **entry** to be modified by default if the file itself has been modified?

optimade.rst
ml-evs: I think I also have the same issue with `ctime`... if you think this is useful metadata then we can add it, but my comment about `last_modified` below still stands.

optimade.rst
```diff
@@ -2462,6 +2566,40 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+
+Relationships with files may be used to relate an entry with any number of :entry:`files` entries.
+
+.. code:: jsonc
+
+    {
+      "data": {
+        "type": "structures",
+        "id": "example.db:structs:1234",
```
ml-evs, line 2579: Are colons and periods allowed in `id`'s? I'm not necessarily against it (though not sure how well the reference server implementation would work with `.`'s in single entry URLs), but currently the spec states:

> Identifier
    Names that MUST start with a lowercase letter ([a-z]) or an underscore ("_") followed by any number of lowercase alphanumerics ([a-z0-9]) and underscores ("_").

optimade.rst
ml-evs: I now see the discussion with @rartino above along these lines too, not sure whether these `xtime` fields were added before or after that discussion.

optimade.rst
ml-evs: I guess I am okay with this, though I am a bit worried that the `name` may contain metadata itself that is not captured by other fields, thus downloading it with an arbitrary name could cause issues. I'm not totally au fait with `Content-Disposition` headers, would you be against a "SHOULD" for using `name` as the default served when accessing when downloading from the relevant `url`?

optimade.rst
ml-evs: I would still prefer allowing JSON API links, e.g. (following the example of `homepage`):

```
    url: a JSON API links object, at which the file contents of the file can be dowloaded, either directly as a string, or as a link object which can contain the following fields:
        href: a string containing the file URL.
        meta: a meta object containing non-standard meta-information about the file download.
```

There are probably metadata to do with *downloading* the file that we would not want to promote to attributes, so `meta` could be a useful thing to use here (e.g., whether you can stream the download, gzippable, restartable etc., say if someone was hosting very large files).

optimade.rst
```diff
@@ -2371,6 +2371,110 @@ Example:
...
+
+- **Description**: version information of a file (e.g. commit, revision, timestamp)
+- **Type**: string.
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - If provided, it MUST be guaranteed that file contents pertaining to the same combination of :field:`id` and :field:`version` are the same.
+
+description
```
ml-evs, line 2434: Should we also include a field for file size (or an upper bound if the server/browser is automatically gzipping reponses)? Alternatively can we add a "SHOULD" that indicates that the file size should be indicated upon download? (avoiding the case where e.g. the client cannot tell the user how much is left to download)

ml-evs (commented):
Hi @merkys, a few more comments below, but I would definitely like this added!

Perhaps this could be discussed (alongside properties) at a specification meeting before the next monthly meeting? I'm not sure if someone volunteered to organise such a meeting...

One general formatting comment, it looks like through the spec we treat the description as a new sentence, so the first words of all your fields should be capitalised.

optimade.rst
merkys: I have added `xtime` per [@sauliusg request](https://github.com/Materials-Consortia/OPTIMADE/pull/360#issuecomment-926692722), thus I will leave it to @sauliusg to answer.

optimade.rst
merkys: Having file size makes sense indeed.

optimade.rst
ml-evs: My mistake (a Ctrl+F failure)... for clarity here is the actual ID definition in the spec:

>ID
    The ID entry property is a unique string referencing a specific entry in the database. The following constraints and conventions apply to IDs:
        Taken together, the ID and entry type MUST uniquely identify the entry.
        Reasonably short IDs are encouraged and SHOULD NOT be longer than 255 characters.
        IDs MAY change over time.

optimade.rst
merkys: OK, I see the point in maintaining the consistency. And this exception after all would look inelegant. Will fix soon.

merkys: Redefined as JSON API links object in fb187a74e4ae0e6d4395244d362520487684aba3. Now I am a bit unsure how querying on this property should work. It seems that we did not have JSON API links objects as property types before.

optimade.rst
merkys: I do not quite understand the issue. Could you please provide a use case where this would be important?

I think the bottom line is that we cannot (and do not want to) control what user/software does on the client side. We cannot tell the client side how to name the file after downloading it. Experienced user/proper software will name it either after the `id`, or another entry to which the file is attached, or according to some other scheme retaining at least some link to the original file entry. Others may just store the file as `optimade.dat` despite all the recommendations.

optimade.rst
merkys: Added `size` in 82f3d11f32b08d86c9b13fde09d36db7acf4564c. Please mark as resolved if you are happy with the description.

merkys:
> Hi @merkys, a few more comments below, but I would definitely like this added!

Thanks @ml-evs!

> Perhaps this could be discussed (alongside properties) at a specification meeting before the next monthly meeting? I'm not sure if someone volunteered to organise such a meeting...

This is a nice idea to bring together people interested in this PR. I could lump together other my PRs to the same meeting to fast-track their acceptance.

> One general formatting comment, it looks like through the spec we treat the description as a new sentence, so the first words of all your fields should be capitalised.

Noted and fixed in e535aa14e1d0caa0231e577531db54409c347c9c.

optimade.rst
rartino: Sorry for not picking this up earlier, but I'm with @merkys on this - the request for *data* fields to be JSON API links objects breaks our data type paradigm. In the definition of endpoints, the data types under `**Type**` refers to OPTIMADE data types, because they need to be representable in any output format.

If we feel that it is crucial for this URL to be a JSON API Links object in JSON-formatted responses, then we need to introduce a new OPTIMADE URL data type in section 2.1, and then in 4.2 specify that that OPTIMADE URLs are represented in JSON responses as JSON API Links objects, and then in the files section specify the type of `url` to be an OPTIMADE URL.

But, IMO, I don't see why we need a special URL type for the data we communicate.

optimade.rst
rartino: Isn't the point here that the files endpoint represent "files" in the normal sense of the word, with the usual meta-data? That would absolutely include 'name' as important meta-data.

Typical usage would be when we communicate the orginal cif file from a journal that is tied to an entry. In that case we really want to be able to communicate the original file name, so it can be preserved if desired. @ml-evs Where otherwise would we communicate that information?

optimade.rst
rartino: @ml-evs What I wrote in my suggestion was that the only requirement on `last_modified` is that if the file content is different between two accesses to the API, then `last_modified` MUST be different. So, as long as you can fulfill that, you can essentially do whatever you want.

optimade.rst
rartino: @ml-evs: Aren't all these times at least technically referring to different things? 

- `last_modified` = when the OPTIMADE data *entry* changed.
- `modification_timestamp` = a timestamp that must advance when the file contents or file meta-data changes of the file that can be downloaded via `url`.
- `mtime`/`ctime` = POSIX file metadata communicated alongside the file, which we in OPTIMADE essentially do not prescribe any specific handling of - they are part of the underying file.

To clarify the last point, I think the role of `mtime`, `ctime` should be to make it possible for someone with a POSIX file system to download a file from me and get exactly the same copy that I have - including all the POSIX metadata. However, POSIX doesn't come with absolute guarantees about these timestamps. It is quite easy in practice to modify a bit in a file, and then re-set `mtime` to be what it was before. I may still want to communicate *that* file (with different contents but the same `mtime`) via OPTIMADE.

optimade.rst
merkys: @rartino very well emphasized the problem I merely felt we may run into. Thus it seems that we either rollback fb187a7 or define the URL data type.

Rolling back fb187a7 would render the specification not so much JSON:API-sh. However, redefining string URLs as JSON:API Links objects later (post-PR) would be backwards-compatible change. As I would like to avoid bundling this PR with any more heavyweight changes, I am in favor of getting back to string URLs.

On the other hand, URL data type should not be too difficult to implement. Since JSON:API Links object can have at most one URL, comparisons for both `"https://example.org"` and `{ "href": "https://example.org", "meta": { "info": null } }` could work as string comparisons on the actual URL value (`"https://example.org"` in both cases).

optimade.rst
merkys: When I was putting together this proposal, I had in mind "files" as streams of bytes, without the additional metadata which usually is subject to change (name and all POSIX inode attributes which are lost when copying/renaming). I understand now, however, the wish to preserve these details as well. Some providers may not care (or not have these concepts at all), thus I defined `name` and `xtime` as optional.

I am now leaning towards making `name` mandatory, though. This would remove the need for the client to "invent" a name for each downloaded file (without any provisions to resolve clashes on the client side), and even if names are not used by the provider (think of on-the-fly generated files), it is easy for the provider to come up with something. And sometimes file names are metadata which cannot be expressed in other ways, yes.

optimade.rst
rartino: +1 for a rollback for this URL to just be an OPTIMADE String for now (and that we adopt this as the policy for URL data items as we go forward.)

The only reason I can see for not representing OPTIMADE data item URLs as Strings is if we need some special non-string search semantics on them. Do we have that?

merkys: > +1 for a rollback for this URL to just be an OPTIMADE String for now (and that we adopt this as the policy for URL data items as we go forward.)

I have reverted fb187a7 in d68037073f099a7be290296cf8ae15f1be92ad84. Thus the data type of `url` is again String. I am convinced that JSON:API Links object data type overcomplicates things for now.

> The only reason I can see for not representing OPTIMADE data item URLs as Strings is if we need some special non-string search semantics on them. Do we have that?

I do not thing so. Most likely searches will be limited to string search operators as supported by OPTIMADE (exact and fuzzy matches). Cannot think of anything else.

optimade.rst
merkys: In 467d117369111aa75c6d28100a3bbd61cf42a260 I made the support for file name mandatory. I hope I can mark this discussion subject as resolved. Anyway we may discuss it again in the upcoming web meeting.

optimade.rst
ml-evs: Sorry for missing this last week, I now understand where you are both coming from.  I guess the JSON:API way of doing this would be that each `files` resource has a top-level `links` to avoid having the URL in `attributes`, e.g.

```
// ...
{
  "type": "files",
  "id": "1",
  "attributes": {
    "name": "1.cif"
  },
  "relationships": {
    "structure": {
      "data": { "type": "structures", "id": "1" }
    }
  },
  "links": {
    "self": "http://raw_files.example.org/1.cif",
    "related": "http://optimade.example.org/files/1"
  }
}
// ...
```

but I think this is convoluted, masks the intent of this endpoint and requires API users to have JSON:API knowledge.

I will resolve this now.

optimade.rst
ml-evs: > @ml-evs What I wrote in my suggestion was that the only requirement on `last_modified` is that if the file content is different between two accesses to the API, then `last_modified` MUST be different. So, as long as you can fulfill that, you can essentially do whatever you want.

Fine by me.

optimade.rst
merkys: @ml-evs my early proposition in #211 was quite similar to this. However, my idea was to get around without using `/files` endpoint by adding links to files *directly* in top-level `links` of other entry types. `/files` endpoint, however, is a bit more homogeneous solution, not requiring changing the way we handle queries.

JPBergsma:
During the meeting the consensus seemed to be that the modification time would be used to determine whether a file has been updated. In my experience there are quite some computers that do not have a correct system time. So it may occur that the mtime of a file is older that the mtime of the previous version. So perhaps we should mention that the database provider should check this or some other means to handle this situation.

merkys:
> During the meeting the consensus seemed to be that the modification time would be used to determine whether a file has been updated. In my experience there are quite some computers that do not have a correct system time. So it may occur that the mtime of a file is older that the mtime of the previous version. So perhaps we should mention that the database provider should check this or some other means to handle this situation.

@JPBergsma `mtime` is not meant to be used as a substitute for `last_modified` or version. But the problem might persist with `last_modified` probably. I assume you are talking about the computers with out-of-sync clocks (i.e., lagging behind/running ahead of the Coordinated Universal Time), please correct me if my assumption is incorrect.

As far as I understand, Network Time Protocol and similar techniques do not cause computer clocks to "jump back in time" (this is the sole problem I see, as "leap forward in time" will still give the subsequent writes to a file ever-increasing modification timestamps). Instead, clock frequencies are adjusted to reduce the deltas between computer clocks and the Coordinated Universal Time. Thus I cannot see how writing/touching a file may render its modification timestamp to actually decrease, unless someone intentionally "turns back time".

Thus I do not think we should put in the specification that modification timestamps for subsequent modifications MUST each be greater (or equal?) to the previous one. But as said, I may have misunderstood the problem.

merkys:
OK, I believe I have implemented the consensus emerging in yesterday's web meeting:

* Introduced checksum properties, `md5sum` and `sha1sum`.
* Introduced `modification_timestamp` for the changes to the file contents. [This comment](https://github.com/Materials-Consortia/OPTIMADE/pull/360#discussion_r759138258) by @rartino summarizes well why `last_modified` is not enough.
* Introduced `url_stable_until` to declare the time-to-live for the URL in `url`.
* Improved descriptions of `xtime` and relations between `last_modified`, `modification_timestamp` and `mtime`.
* Explained that `name` has to include a suffix if a file comes with a suffix.

Please let me know if I missed something.

JPBergsma:
> As far as I understand, Network Time Protocol and similar techniques do not cause computer clocks to "jump back in time" (this is the sole problem I see, as "leap forward in time" will still give the subsequent writes to a file ever-increasing modification timestamps). Instead, clock frequencies are adjusted to reduce the deltas between computer clocks and the Coordinated Universal Time. Thus I cannot see how writing/touching a file may render its modification timestamp to actually decrease, unless someone intentionally "turns back time".

I do not know how cif files get generated exactly, but I can imagine that some x-ray machines are quite old.
An empty CMOS battery is quite a common problem in old computers, and this will cause the system time to be reset every time the computer is turned off.(I have had at least 2 computers with this problem).
Windows and Linux also (disagree)[https://mashtips.com/fix-linux-windows-dual-boot-clock-different-time/] about how time should be kept, so this can also lead to time differences, travelling across the date line may also cause issues. 
The mtime may thus indicate a time before the creation time of the previous version of the file. 
 
When I posted this there was no modification_timestamp field yet in the PR, so you may only be able to use the mtime to check whether a file had changed.   
I assume that the servers have a consistent time. It however would not hurt to specify that each new file version should have a greater modification time than the previous version of the file.

optimade.rst
```diff
@@ -2371,6 +2371,170 @@ Example:
...
+mtime
+~~~~~
+
+- **Description**: Time of last modification of a file as per POSIX standard.
+- **Type**: timestamp
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - It should be noted that the values of :field:`last_modified`, :field:`modification_timestamp` and :field:`mtime` do not necessary match.
```
JPBergsma, line 2534: I would interpret `modification_timestamp` as the time at which the file was changed in the database. As this is the value that determines whether I have to redownload the file.
'mtime' would be the time at which the file itself was modified, but it may still take some time to upload the file and publish it in the database, so the  `modification_timestamp` will generally indicate a later time.

JPBergsma (changes requested):
It seems we find it important that a client can check whether a file has been updated.
There now however are multiple fields that can indicate whether a file has changed. e.g. version,  md5sum, sha1sum, mtime,  modification_timestamp. With the current proposal, each database could use a different method for showing that a file has been updated. This makes it more difficult for clients to determine whether there is a new file version. I therefore think it would be good to upgrade one of these, to the SHOULD or MUST level. The modification_timestamp is probably the most appropriate field for this.

optimade.rst
merkys: > 'mtime' would be the time at which the file itself was modified, but it may still take some time to upload the file and publish it in the database, so the `modification_timestamp` will generally indicate a later time.

Not necessarily. For example, let's say an administrator decided to switch the hard disk from which the files are served. Thus they copied files from one disk to another and rewrote the links in OPTIMADE response. Thus the `modification_timestamp` stay the same, but `mtime` will be newer.

merkys:
> It seems we find it important that a client can check whether a file has been updated. There now however are multiple fields that can indicate whether a file has changed. e.g. version, md5sum, sha1sum, mtime, modification_timestamp. With the current proposal, each database could use a different method for showing that a file has been updated. This makes it more difficult for clients to determine whether there is a new file version. I therefore think it would be good to upgrade one of these, to the SHOULD or MUST level. The modification_timestamp is probably the most appropriate field for this.

How about this:
> Values of `md5sum`, `sha1sum` and `modification_timestamp` MUST NOT change in between changes to `version`.

It would be nice to say:
> Values of `md5sum`, `sha1sum`, `modification_timestamp` and `version` MUST all change together.

But there may be rare cases when `version` is bumped without affecting the other three aforementioned fields. An example of such case is when a database uses the same version name for all its contents. I am not sure that all the providers will be happy about the requirement to retain the old versions for unmodified files.

optimade.rst
JPBergsma: That is why I used the word generally.
This was however not the point that I was making.

The approach I had in mind was that I keep track of the last time I checked for updates. After a certain period, I may check if there are any updates to the file and then look if `modification_timestamp` is later than the last time I checked for updates. If so, I would have to redownload the file. For this method to work, `modification_timestamp` should indicate the time the file became available on the server. 

I now think that the approach you have in mind is: that one stores the `modification_timestamp` of each file when it is downloaded.  Once in a while, one could then check whether the `modification_timestamp` of the file is still the same as the `modification_timestamp` on the server. If not, you would download the file. With that approach, is does not matter whether the `modification_timestamp` is the time the file was modified or the time the file was made available on the server.

merkys: > I now think that the approach you have in mind is: that one stores the `modification_timestamp` of each file when it is downloaded. Once in a while, one could then check whether the `modification_timestamp` of the file is still the same as the `modification_timestamp` on the server. If not, you would download the file. With that approach, is does not matter whether the `modification_timestamp` is the time the file was modified or the time the file was made available on the server.

Right, I would probably do like that. Also, instead of `modification_timestamp` one could as well rely on checksums.

I really expect situations when subsequent versions of a file acquire decreasing timestamps would be rare, but you are right, adding a sentence would not hurt.

JPBergsma:
> How about this:
> 
> > Values of `md5sum`, `sha1sum` and `modification_timestamp` MUST NOT change in between changes to `version`.
> 
> It would be nice to say:
> 
> > Values of `md5sum`, `sha1sum`, `modification_timestamp` and `version` MUST all change together.
> 
> But there may be rare cases when `version` is bumped without affecting the other three aforementioned fields. An example of such case is when a database uses the same version name for all its contents. I am not sure that all the providers will be happy about the requirement to retain the old versions for unmodified files.

That does not quite address my point, although it does improve things. All those fields are optional. Each database can therefore choose which one it implements to indicate whether a file has changed. As a result, a client would have to implement multiple ways to check whether a file has been updated. It would be nice if we could agree on one method that can be used for every database to see whether a file has been updated.

merkys:
> That does not quite address my point, although it does improve things. All those fields are optional. Each database can therefore choose which one it implements to indicate whether a file has changed. As a result, a client would have to implement multiple ways to check whether a file has been updated. It would be nice if we could agree on one method that can be used for every database to see whether a file has been updated.

We may make `version` mandatory as probably most of the providers will be versioning their files. (However, some providers may not; think for example unversioned blobs in SQL databases). Then change of `version` could trigger download, but as said it might be too intensive for providers having a single version for all the files.

I am really reluctant to enforce new policies and conventions that may be cross-cutting concerns for some of the providers. I understand that diversity of the providers will boil down to clients having to implement multiple ways to cover those differences. I do not think a win-win solution exists, probably the best way is to have some meet-in-the-middle situation. Putting too much work on either end will stall the adoption.

ml-evs:
A comment arising from a discussion I just had with @CasperWA: how do we see this endpoint interacting with the proposed #364 endpoint? In my mind, this endpoint already fulfills many of the criteria we discussed at the workshop for `/archives`, we might just need a way to codify that a particular `files` entry points to an OPTIMADE archive type, and a way to add a `relationship` between the resource and the database. Maybe this could be done via a `relationships` field in the `info` endpoint of the database.

Do we feel that this is overloading/missing the point of the current `files` endpoint design?

merkys:
@ml-evs, @CasperWA: I gave #364 a glance, and to me `/files` and `/archives` entry types seem to be largely overlapping. However, in `/archives`, file tree hierarchy seems to be important, whereas in `/files` it is not mandated, as the organization of actual files is left for the provider to decide on. Maybe this could indeed be represented using relationships.

Additionally, `/archives` has means to specify licenses. I see communication of licenses as an issue on its own (I opened #102 quite some time ago), and would be happy to see `license` becoming property supported by all entry types.

optimade.rst
```diff
@@ -2404,6 +2404,173 @@ Example:
...
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+- **Examples**:
+
+  - :val:`"POSCAR format file"`
+
+md5sum
```
ml-evs, line 2521: @merkys and I discussed making a generic `checksum` field which keys corresponding to common hashing algorithms like `md5`, `sha1` etc. with a set of enumerated values that are officially supported.

optimade.rst
```diff
@@ -2495,6 +2662,40 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+Relationships with files may be used to relate an entry with any number of :entry:`files` entries.
+
+.. code:: jsonc
+
+    {
+      "data": {
+        "type": "structures",
+        "id": "example.db:structs:1234",
+        "attributes": {
+          "formula": "H2O"
```
ml-evs, line 2677:
```suggestion
          "chemical_formula_descriptive": "H2O"
```

optimade.rst
ml-evs: Add provider prefix if not in the supported list

JPBergsma:
During the discussions about the trajectories, a point was raised that it could be a good idea to include data about the "role" of a file. 
A file could for example be an input structure, an output structure, or a file with settings used to do the calculation.
We thought it would therefore be good to somehow indicate this "role" a file has. This does not necessarily have to be done at the files endpoint. We may also be able to do this in the relationships field. 

I was wondering whether you prefer to discuss this as part of the Files endpoint PR or you would prefer if I open a seperate Issue for this.

merkys:
@JPBergsma I see the "role" as a relationship attribute. Thus I think "role" belongs in `meta`. If there is a need to standardize on that, I would suggest opening a separate PR (or amend #24 or #34) after current PR is merged.

sauliusg (approved):

rartino (approved):
Great work!

PR #365, opened 2021-06-11
Staging of v1.1.0

ml-evs:
From discussion in #366 and at the workshop, it looks like the most reasonable option is to release v1.1.0 and skip v1.0.1. This PR stages those changes.


ml-evs:
I have just bumped the version number in this PR (i.e. `develop`) to 1.1.0, so this is now ready for a final review and release.

@rartino: in the wiki release instructions, it talks about doing a -ff-only merge locally. I think this is equivalent to a GitHub PR rebase+merge, which I think is preferable...

optimade.rst
```diff
@@ -1838,6 +1844,7 @@ nelements
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: MUST be a queryable property with support for all mandatory filter features.
+  - MUST be equal to the lengths of the list properties `elements`_ and `elements\_ratios`_, if they are provided.
```
JPBergsma, line 1847: For me, it was not clear what "list properties" are the first time I read this sentence.
Maybe something like "MUST be equal to the number of values in `elements`_ and `elements\_ratios`" is clearer.

rartino:
> @rartino: in the wiki release instructions, it talks about doing a -ff-only merge locally. I think this is equivalent to a GitHub PR rebase+merge, which I think is preferable...

I think the instructions were devised by @CasperWA originally, and were created before GitHub had these merge options. However, reading [the documentation](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#rebase-and-merge-your-pull-request-commits), I'm actually not sure that 'Rebase and merge' would be equivalent, because the documentation says the commit hashes will be updated, which we would like to avoid.

optimade.rst
rartino: We should strive to be as clear as possible, but for these sentences specifically there were a lot of discussion to arrive at the present formulations (see #349). Hence, as long as the edit is just for clarity, I'd suggest taking it in a separate PR to merge after people have had more room to agree on the change (and thus after v1.1).

rartino (approved):
The v1.1.0 release looks ready to me.

ml-evs:
> I think the instructions were devised by @CasperWA originally, and were created before GitHub had these merge options. However, reading [the documentation](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#rebase-and-merge-your-pull-request-commits), I'm actually not sure that 'Rebase and merge' would be equivalent, because the documentation says the commit hashes will be updated, which we would like to avoid.

I'm 90% sure that GitHub will do a ff-only rebase when the history is linear (as it is here) but I will triple check in another repo. The GitHub docs do allude to this too. Let's see what @CasperWA  says.

merkys (approved):
Looks good!

rartino:
> I'm 90% sure that GitHub will do a ff-only rebase when the history is linear (as it is here) but I will triple check in another repo.

[They indeed say](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#rebase-and-merge-your-pull-request-commits) that the merge is done fast-forward:

*When you select the Rebase and merge option on a pull request on GitHub, all commits from the topic branch (or head branch) are added onto the base branch individually without a merge commit. Pull requests with rebased commits are merged using the fast-forward option. [...]*

However, this other paragraph just below makes it sound as if that fast-forward is done using re-written (rebased) commits with different hashes:

*The rebase and merge behavior on GitHub deviates slightly from git rebase. Rebase and merge on GitHub will always update the committer information and create new commit SHAs, whereas git rebase outside of GitHub does not change the committer information when the rebase happens on top of an ancestor commit.*

optimade.rst
JPBergsma: Ok, then I will create a separate PR for this in the future.

ml-evs: Just a comment that "list properties" are also discussed outside of the diff of this PR in [section 6.2.4](https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#624comparisons-of-list-properties), not sure if that clarifies it

CasperWA (approved):

ml-evs:
Final steps:

- I've written a [little script](https://gist.github.com/ml-evs/da75034dea475bde994b2e2d488107a7) to check all external links in the specification, we might consider adding this (or something similar) to the CI.
- The release will need release notes: I think it makes sense to also add these to a top-level CHANGELOG.md in this repository (I believe this was discussed at one point, but not actioned upon). I will copy the notes from 1.0.0 and write new ones for 1.1 in a new PR, so we will need one more round of reviews for this.
- Once this PR is approved again, I will locally merge the branch.

JPBergsma (approved):
I did not see any major issues in this draft PR.
And great, that you made a script to test external links.

optimade.rst
JPBergsma: Thanks for pointing this out. As it now seems to become more than simply changing one line, I will postpone creating an issue for this until after I have created a draft for the trajectories endpoint.

ml-evs:
Proceeding with the local merge... wish me luck!

ml-evs:
:tada:

PR #370, opened 2021-07-06
Add v1.1.0 schemas from optimade-python-tools v0.16.0

ml-evs:
This PR is blocking #365. It updates the OpenAPI schemas to the latest version, ready for the v1.1.0 specification release.


CasperWA (approved):
Cheers @ml-evs

schemas/index_openapi_schema.json
```diff
@@ -2,8 +2,8 @@
   "openapi": "3.0.2",
   "info": {
     "title": "OPTIMADE API - Index meta-database",
-    "description": "The [Open Databases Integration for Materials Design (OPTIMADE) consortium](https://www.optimade.org/) aims to make materials databases interoperational by developing a common REST API.\nThis is the \"special\" index meta-database.\n\nThis specification is generated using [`optimade-python-tools`](https://github.com/Materials-Consortia/optimade-python-tools/tree/v0.15.3) v0.15.3.",
-    "version": "1.0.0~develop"
+    "description": "The [Open Databases Integration for Materials Design (OPTIMADE) consortium](https://www.optimade.org/) aims to make materials databases interoperational by developing a common REST API.\nThis is the \"special\" index meta-database.\n\nThis specification is generated using [`optimade-python-tools`](https://github.com/Materials-Consortia/optimade-python-tools/tree/v0.16.0) v0.16.0.",
+    "version": "1.1.0"
```
ml-evs, line 6: Note, this will be merged almost immediately as part of #365, so we will only be missing the `~develop` suffix for one commit.

rartino (dismissed):
I do not see any issue here.

I note again that I really think we should consider adjusting the specification so that the same schema applies for non-index and index endpoints. IMO the difference should just be whether you've decided to serve data through that end-point or not.

CasperWA:
> I do not see any issue here.
> 
> I note again that I really think we should consider adjusting the specification so that the same schema applies for non-index and index endpoints. IMO the difference should just be whether you've decided to serve data through that end-point or not.

This would require an assimilation of the currently separate models for the `/info` endpoint (for the index and standard implementations, respectively) in OPTIMADE Python tools.
Just noting this here, as this would be the only thing necessary to make this change in the schema - in the server implementation of OPTIMADE Python tools a more rigorous testing for whether an implementation is an index meta-db or not must be implemented though.

ml-evs:
We should hold off on this until #371 is merged, which updates some descriptions. Rather than going through the whole optimade-python-tools release process again, I'm happy to just patch the descriptions in this PR.

rartino (approved):
Looks good.

Our present handling of the version number in the schemas is not great. But there is a very-long-term plan to fix that.

Note to self: this is the version of the schemas that adds the issue_tracker, and hence should not be merged as-is into v1.0. (However, it would be nice, but not critical, to get the Chemical Symbols description update into v1.0; but perhaps it is easiest if we live without it.)

JPBergsma (approved):
Thanks for changing "name" to "symbol" in the appropriate places. I did not know that it occurred here as well.

PR #371, opened 2021-07-06
Replaced "name" with "symbol" when refering to the abreviation of the element names

### LINKED ISSUE #367: Use "Symbols" instead of "Names" in the description of the elements field
I noticed that in the description of the field elements under the structures endpoint it says:
"**Description**: Names of the different elements present in the structure."
If I, however, look at the example, the symbols of the elements are used, e.g. "Fe", "H", instead of the names, e.g. "Iron", "Hydrogen". I was therefore wondering whether it would be ok to change the description to:
 "**Description**: Symbols of the different elements present in the structure."

JPBergsma:
As mentioned in issue #367 we often write that a certain field should have the name of an element. In practice, we however use the elemental symbols, e.g. "K", "Fe", instead of the names: Potassium, Iron. 
This PR changes the text in OPTIMADE standard to reflect this.

Closes #367

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1813,7 +1813,7 @@ Structures Entries
 elements
 ~~~~~~~~
 
-- **Description**: Names of the different elements present in the structure.
+- **Description**: Symbols of the different elements present in the structure.
 - **Type**: list of strings.
 - **Requirements/Conventions**:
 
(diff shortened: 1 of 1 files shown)


rartino (approved):
Thanks for reporting and fixing this. I fully agree with the change.

ml-evs (dismissed):
Thanks @JPBergsma, I have a minor suggestion (below) but I don't think it is super important so will not hold up the review approval.

ml-evs (approved):

rartino (approved):
Indeed a good adjustment by @ml-evs

PR #375, opened 2021-07-08
Add CHANGELOG containing release notes

ml-evs:
This PR adds a CHANGELOG that contains the release notes that I would have typed into GitHub for the upcoming release. These notes can instead be copied from here. I think we discussed adding a CHANGELOG at some point, so hopefully this is not controversial (though please make suggestions on the text, if you have any).

I also took the release notes (written by @rartino?) for v1.0.0 and added them here.

**This** *should* be the final PR/commit before locally merging #365 and releasing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ This is a minor release that primarily clarifies the specification, with only tw
 
 ### Changes
 
-- The `mass` field of the `species` attribute for the `structures` type has been updated from a float to a list of floats [#344](https://github.com/Materials-Consortia/OPTIMADE/pull/344). This is a breaking change for any implementations that serve this optional field.
-- The `implementation` field of the general `meta` response has been updated to include an `issue_tracker` field [#339](https://github.com/Materials-Consortia/OPTIMADE/pull/339).
+- The `mass` field of the `species` attribute for the `structures` type has been updated from a float to a list of floats ([#344](https://github.com/Materials-Consortia/OPTIMADE/pull/344)). This is a breaking change for any implementations that serve this optional field.
+- The `implementation` field of the general `meta` response has been updated to include an `issue_tracker` field ([#339](https://github.com/Materials-Consortia/OPTIMADE/pull/339)).
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,35 @@
+# Changelog
+
+## v1.1.0 (08/07/2021)
+
+This is release v1.1.0 of the OPTIMADE API specification.
+
+This is a minor release that primarily clarifies the specification, with only two additions.
+
+### Changes
+
+- The `mass` field of the `species` attribute for the `structures` type has been updated from a float to a list of floats [#344](https://github.com/Materials-Consortia/OPTIMADE/pull/344). This is a breaking change for any implementations that serve this optional field.
+- The `implementation` field of the general `meta` response has been updated to include an `issue_tracker` field [#339](https://github.com/Materials-Consortia/OPTIMADE/pull/339).
+
+### Fixes
+
+- The specification text has been clarified in several places.
+- Multiple typos, grammatical errors, and incorrect API examples have been fixed.
+- The OpenAPI schemas are now fully compliant with the Swagger validator.
+
+
+## v1.0.0 (01/07/2020)
+
+This is release v1.0.0 of the OPTIMADE API specification.
+
+The specification has undergone a few changes since v1.0.0-rc.2, some not backward compatible.
+
+A few highlighted changes directly affecting the API:
+- Several changes related to version negotiation for future releases of the API have been added.
+- The API may now be served on both the unversioned and versioned URLs.
+- The top-level `meta` field has undergone major revision:
+   - It is now mandatory, with a few mandatory subfields (e.g., `api_version`), and most other fields made optional.
+   - A link to a schema may be provided.
+   - It no longer contains an `index_base_url`, since this functionality is covered by `root` links in the `/links` endpoint.
+
+- The `unit` field for describing properties in the `/info` endpoints is now standardized to use the Unified Code for Units of Measure.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ This is a minor release that primarily patches minor specification errors and in
 
 ### Patches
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
+# Changelog
+
+## v1.1.0 (08/07/2021)
```
rartino, line 3: Date formats is a curious thing, and one persons standard date format is another persons indecipherable scribbles. How do I know if this represents 7:th of August or 8:th of July? I usually suggest the one and only ISO format (see https://xkcd.com/1179/), but with time I've become more humble (eh...) and I am going to suggest that we adopt what I think is the least controversial format in this context:
```suggestion
## v1.1.0 (July 8, 2021)
```

CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
...
+- The `implementation` field of the general `meta` response has been updated to include an `issue_tracker` field ([#339](https://github.com/Materials-Consortia/OPTIMADE/pull/339)).
+
+### Fixes
+
+- The specification text has been clarified in several places.
+- Multiple typos, grammatical errors, and incorrect API examples have been fixed.
+- The OpenAPI schemas are now fully compliant with the Swagger validator.
+
+
+## v1.0.0 (01/07/2020)
```
rartino, line 21:
```suggestion
## v1.0.0 (July 1, 2020)
```

CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
...
+### Changes
+
+- The `mass` field of the `species` attribute for the `structures` entry type has been updated from a float to a list of floats ([#344](https://github.com/Materials-Consortia/OPTIMADE/pull/344)). This is a breaking change for any implementations that serve this optional field.
+- The `implementation` field of the general `meta` response has been updated to include an `issue_tracker` field ([#339](https://github.com/Materials-Consortia/OPTIMADE/pull/339)).
+
+### Fixes
+
+- The specification text has been clarified in several places.
+- Multiple typos, grammatical errors, and incorrect API examples have been fixed.
+- The OpenAPI schemas are now fully compliant with the Swagger validator.
```
rartino, line 18: Is the distinction between *Changes* and *Fixes* clear? Since this changelog entry is likely going to be the format we copy+paste going forward, perhaps we should divide it into the semver categories? So, how about:

```suggestion
### New features

- The `implementation` field of the general `meta` response has been updated to include an `issue_tracker` field ([#339](https://github.com/Materials-Consortia/OPTIMADE/pull/339)).

### Patches

- The `mass` field of the `species` attribute for the `structures` entry type has been updated from a float to a list of floats ([#344](https://github.com/Materials-Consortia/OPTIMADE/pull/344)). The was deemed a specification bug that now is fixed in both the specification text and the schemas. (Note: software implemented to strictly adhere to the specification bug could break by this change.)
- The specification text has been clarified in several places without change of intended meaning.
- Multiple typos, grammatical errors, and incorrect API examples have been fixed.
- The OpenAPI schemas are now fully compliant with the Swagger validator.
```

CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
...
+
+A few highlighted changes directly affecting the API:
+- Several changes related to version negotiation for future releases of the API have been added.
+- The API may now be served on both the unversioned and versioned URLs.
+- The top-level `meta` field has undergone major revision:
+   - It is now mandatory, with a few mandatory subfields (e.g., `api_version`), and most other fields made optional.
+   - A link to a schema may be provided.
+   - It no longer contains an `index_base_url`, since this functionality is covered by `root` links in the `/links` endpoint.
+
+- The `unit` field for describing properties in the `/info` endpoints is now standardized to use the Unified Code for Units of Measure.
```
rartino, line 35: I'd argue just remove this and simply start with the "This is release v1.0.0 of the OPTIMADE API specification" for the first release. Going forward, it doesn't seem particular important to start the file with what was changed since specifically the v1.0.0-rc.2 prerelease. Isn't the most sensible place to formally start logging changes more carefully the v1.0.0 version?
```suggestion
```

CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
+# Changelog
+
+## v1.1.0 (08/07/2021)
+
+This is release v1.1.0 of the OPTIMADE API specification.
+
+This is a minor release that primarily clarifies the specification, with only two additions.
```
rartino, line 7: (Due to changes suggested below)
```suggestion
This is a minor release that primarily patches minor specification errors and introduces one new feature.
```

rartino (changes requested):

CHANGELOG.md
ml-evs: I was toying with this too, I'll remove it but I think it makes sense to leave this in the actual GitHub release notes.

CasperWA: I agree with this sentiment.
Hopefully no one has implemented versions prior to v1.0.0 at this stage - at least not in production.
On the other hand, it's neat to have this information somewhere for our sake as the developers.

CHANGELOG.md
ml-evs: I prefer your suggestion, I made a couple of edits to your suggestion to fix typos and split the species.mass point into a few bulletpoints.

CasperWA (approved):
Looks good to me, thanks @ml-evs !

rartino (approved):
Looks good; thanks for the effort with this. It is good to go.

As a rather non-important point I have a slight reaction to your formulation change from my: *"software implemented to strictly adhere to the specification bug could break by this change"* into *"could constitute a breaking change for software implemented to strictly adhere to the v1.0.0 specification"*, since, from the perspective that this really is a specification bug, the *intent* of the v1.0.0 specification is that the mass field should NOT be implemented as a single float, even though it seemingly says so.

I get the impression that you are struggling to suppress a personal feeling that the change is actually a breaking change (and thus, should be a major version number bump...) rather than a specification bug :-)

Nevertheless, it isn't like people read the changelog to this degree, so go ahead with the formulation you prefer.

ml-evs:
> I get the impression that you are struggling to suppress a personal feeling that the change is actually a breaking change (and thus, should be a major version number bump...) rather than a specification bug :-)

Just trying to call a spade a spade! As you say, no-one is really going to act on the information here, but I think it is worth acknowledging... thanks for the suggestions!

Issue #347, opened 2020-12-21
Host specification on optimade.org


CasperWA:
Also another thing. Now that we are getting into a bit more of a serious phase: no one knows if GitHub will suddenly change their URL structure. Should we not stop doing these deep-links into GitHub, and instead make sure our released specifications are hosted at permanent links somewhere? 

My immediate thought is under the 'schema' repository I just created. But maybe it would be confusing with: `https://schemas.optimade.org/specification/v1.0/optimade.html`
although, in a very generalized sense, the specification is a schema (for humans) :-)

Did I do a mistake naming that domain 'schemas'? Should it has been something like 'permanent.optimade.org'?...

_Originally posted by @rartino in https://github.com/Materials-Consortia/providers/issues/43#issuecomment-653394380_

merkys:
Decoupling OPTIMADE's permanent URLs from GitHub is good indeed!

> My immediate thought is under the 'schema' repository I just created. But maybe it would be confusing with: `https://schemas.optimade.org/specification/v1.0/optimade.html`
> although, in a very generalized sense, the specification is a schema (for humans) :-)

Seeing 'schemas', I myself tend to think about OpenAPI/JSON schemas, but maybe that's just me :smile:. As a project OPTIMADE needs to establish common terms for things. I would prefer calling human-readable specification a "specification", and computer-readable specification a "schema". To me it seems that we have done so up to now, but I may be wrong.

> Did I do a mistake naming that domain 'schemas'? Should it has been something like 'permanent.optimade.org'?...

I don't think we need a specific domain for permanent URLs. I think that everything that falls under versioned URLs must be permanent.

rartino:
> I don't think we need a specific domain for permanent URLs. I think that everything that falls under versioned URLs must be permanent.

Well, I have seen websites be restructured to a point where I think it is a good idea to generally isolate things that absolutely needs permanent URLs far away from the "end-user facing" URL structure. That was why I directly pushed for a separate `https://schemas.optimade.org/` so that many years from now, my saved OPTIMADE output results can still be validated with the embedded schema URLs.

But, I suppose the permalinks to the specification texts are *a bit* less sensitive then the machine-readable schemas.

So, what URL do we place them under then? https://www.optimade.org/s/specification/v1.0/optimade.html

(Where the extra '/s/' for "static" is a somewhat common trick to avoid issues that could show up if we at a future website restructuring change our minds and think that the website really needs to start use the "/specification" URL tree.)

[...]

PR #378, opened 2021-07-27
Adding an HTML build of the specification

ml-evs:
This PR experiments with using GitHub actions, pages + pandoc to generate HTML specifications based on the latest development version (following #347, tagging @rartino, @CasperWA and @merkys from that discussion).

It will need to be tweaked to handle multiple versions and could use `git tags` for this instead (e.g. update the `gh_pages` branch on every pushed tag, and make a new folder with that tag name to host on GitHub pages). We would need to manually run the action for existing tags to populate the old versions.

You can see what the generated HTML looks on the Pages for my fork, over here https://ml-evs.github.io/OPTIMADE/optimade.html.


JPBergsma:
Could you perhaps also add a table of contents? It is quite a long document to scroll through when you are looking for a particular entry.

ml-evs:
> Could you perhaps also add a table of contents? It is quite a long document to scroll through when you are looking for a particular entry.

Added in 53c0646, along with some of the other stuff we discussed when making the PDF (section numbers etc).

There's a chance to really customise the HTML output to make the spec more readable, e.g. a contents sidebar, links to other versions and so on, but think this is a reasonable start.

JPBergsma:
Thank you for processing my suggestion so quickly.

rartino:
Why does this implement its own rst -> html conversion, when we have a 
```
make html
```
(I'm fine with edits that make that html conversion nicer, but it seems redundant to have two different code paths that does the same thing.)

ml-evs:
I hadn't spotted `make html`: https://github.com/Materials-Consortia/OPTIMADE/blob/6d35c8c66a5cb80e8c378f65f866efc849039489/tests/makefiles/Makelocal-output#L18-L21

Since we are using pandoc for the `make pdf` rule above it, I will just update the current `rst2html` approach to use pandoc and leave the docker call for the CI only.

ml-evs:
After some discussion I think @merkys and I decided that it makes more sense to perform this procedure *outside* of the main specification repo (perhaps in the homepage repo instead) so we don't run into the problem of needing to version this automation (and perform extra releases of the spec to trigger it).

I will close this PR for now and will try to find some time to move it to another repo.

ml-evs:
Just revived this somewhat after comments by @rartino in the meeting. The built HTML now tries to follow browser prefs for dark mode, and I tweaked the docutils CSS to (hopefully) improve readability. You can see the deployed version at https://ml-evs.science/OPTIMADE/optimade.html

I think it would be good to merge this (or something similar) that deploys the spec for all pushes to develop, then I can set up another workflow that deploys released versions under the appropriate tagged URLs. We could also have a simple template that lists all versions but I don't want to over-engineer things too much at this stage.

tests/makefiles/Makelocal-output
```diff
@@ -18,5 +18,4 @@ HTML_FILES = ${RST_FILES:%.rst=%.html}
 html: ${HTML_FILES}
 
 %.html: %.rst
-	rst2html --syntax-highlight=none $< $@
-
+	rst2html5.py -d --syntax-highlight=none --stylesheet=./tests/makefiles/style.css $< $@
```
vaitkus, line 21:
```suggestion
	rst2html5 -d --syntax-highlight=none --stylesheet=./tests/makefiles/style.css $< $@
```

vaitkus: Could this be changed to `rst2html5`? Then it would work out-of-the box on Debian/Ubuntu machines if the `python3-docutils` package is installed.

tests/makefiles/style.css
```diff
@@ -0,0 +1,542 @@
...
+/* =========== */
+
+/* Literal Blocks */
+pre.literal-block, pre.doctest-block,
+pre.math, pre.code {
+  font-family: monospace;
+  background-color: var(--code-bg-color);
+  border: thin solid var(--fg-color);
+  color: var(--code-fg-color);
+  border-radius: 0.2em;
```
vaitkus, line 515:
```suggestion
  border-radius: 0.2em;
  padding-left: 0.5em;
```

vaitkus: Without the padding some syntax example like `identifier <operator> constant` are a bit harder to read.

tests/makefiles/Makelocal-output
```diff
@@ -18,5 +18,4 @@ HTML_FILES = ${RST_FILES:%.rst=%.html}
 html: ${HTML_FILES}
 
 %.html: %.rst
```
vaitkus, line 20:
```suggestion
%.html: %.rst ./tests/makefiles/
```

tests/makefiles/Makelocal-output
vaitkus, line 21:
```suggestion
	rst2html5 -d --syntax-highlight=none --stylesheet=$(word 2, $^) $< $@
```

tests/makefiles/style.css
```diff
@@ -0,0 +1,542 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap');
+
+:root {
+    --bg-color: white;
+    --fg-color: #505050;
```
rartino, line 5: Great work on the stylesheet in general! However, I'm going to be a bit of a bore and suggest that for the body text of a document this long, it seems more reader-accommodating to go with higher contrast:
```suggestion
    --fg-color: #000000;
```
(Compare with e.g., [JSON:API](https://jsonapi.org/format/), [Semantic Versioning](https://semver.org/). To compensate the headlines could be set in grey or a color.)

tests/makefiles/style.css
```diff
@@ -0,0 +1,542 @@
...
+
+/* Literal Blocks */
+pre.literal-block, pre.doctest-block,
+pre.math, pre.code {
+  font-family: monospace;
+  background-color: var(--code-bg-color);
+  border: thin solid var(--fg-color);
+  color: var(--code-fg-color);
+  border-radius: 0.2em;
+
```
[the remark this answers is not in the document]
ml-evs: I think it's better than the alternative of not wrapping -- will integrate this

tests/makefiles/Makelocal-output
ml-evs: I guess that's fine; maybe a switch that checks for rst2html5.py first, as that is the name of the executable in the official Python package

vaitkus (approved):
Looks great, thanks!

rartino (approved):
Great work! Thanks!

PR #382, opened 2021-08-12
Update citation in README

ml-evs:
Following the publication of the Sci. Data article, this PR adds the reference to the README as the preferred citation and tweaks the text about Zenodo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,10 @@ If this is a pre-release, implementations are also encouraged to support the lat
 
 ## How to cite
 
-Pending the publication of an OPTIMADE paper, we kindly ask that you cite the preprint and Zenodo archive for this repository:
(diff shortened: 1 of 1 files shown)


rartino (approved):
Excellent!

Should v1.0.1 also find its way onto Zenodo? (It is important, however, to only add it in services where we still get a sane ordering of versions - I'm not sure how that works at Zenodo).

merkys (approved):

ml-evs:
> Should v1.0.1 also find its way onto Zenodo? (It is important, however, to only add it in services where we still get a sane ordering of versions - I'm not sure how that works at Zenodo).

I had a go at this earlier, unfortunately v1.0.1 would become the "latest" release, so I didn't bother.

Issue #390, opened 2021-11-23
Does dimension_types refer to the periodicity of the volume or of the structure?


JPBergsma:
I just looked at some data in 2Dmatpedia. They simulate 2D materials. I however noticed that in one of their structures, some of the atoms stick through the top of the simulation volume. Apparently their simulation was performed in a box that was periodic in three dimensions yet the structure itself is only periodic in 2 directions. I was therefore wondering what the correct value of the "dimension_types" field would be in this case.
Once we reach a conclusion about this, I can update the description in the OPTIMADE standard.

merkys:
Good question. I have always thought `dimension_types` referred to the volume. But after re-reading the examples in the specification I am now a bit confused.

Looking back, I can find it was #23 where `dimension_types` has been introduced. A comment there points to the definition of the same property in ESCDF, which now seems to be described here: https://gitlab.com/ElectronicStructureLibrary/escdf/escdf-specifications/-/blob/master/source/system.rst (link in OPTIMADE issue is broken and [Wayback Machine does not help](https://web.archive.org/web/*/https://esl.cecam.org/ESCDF_-_System)).

Since ESCDF does not say anything about molecules, I assume the periodicity of the volume is meant. But I may be wrong.

JPBergsma:
After reading  the definition in ESCDF, I also think that the periodicity of the simulation volume is the intended periodicity of the dimension_types field. I'll try to make a quick PR to clarify this in the OPTIMADE standard.

merkys:
> Edit: Under nperiodic_dimensions it is specifically mentioned that it is about the structures, So I think I will wait until I have heard a few more opinions on this.

To me the following sentence from the description explains quite the contrary:
> This property only reflects the treatment of the lattice vectors provided for the structure, and not any physical interpretation of the dimensionality of its contents.

Here "structure" means the simulated/real crystal (not molecule), I think. At least it seems that the specification is using the term "structure" to mean that.

You may ping authors of #23 or 13a766cd226925974183761fcc7f19de5f19d26d for their opinions.

JPBergsma:
I looked at the examples under nperiodic_dimensions which use the word structure. And I interpreted structure as: "a group of atoms that are attached to each other".  Hence my doubt. The sentence you highlighted makes it a lot clearer, though. So I will make a short PR about this.

merkys:
[...]

PR #394, opened 2021-12-12
Clarify description dimension_types + small corrections

JPBergsma:
I tried to clarify the definition of the `dimension_types` field as described in issue #390.
I made several small corrections and replaced pertains with belongs, as belongs is a more common word in English.


optimade.rst
```diff
@@ -1564,7 +1564,7 @@ OPTIONAL features:
 Examples:
 
 - :filter:`chemical_formula_anonymous CONTAINS "C2" AND chemical_formula_anonymous STARTS WITH "A2"`
-- :filter:`chemical_formula_anonymous STARTS "B2" AND chemical_formula_anonymous ENDS WITH "D2"`
+- :filter:`chemical_formula_anonymous STARTS WITH "A2" AND chemical_formula_anonymous ENDS WITH "D2"`
```
merkys, line 1567: Great catch!

merkys (commented):
This PR contains a bundle of not-so-much related changes, some of them are straightforward, some are a bit more controversial thus may slow down PR's acceptance. I cannot comment on "pertains" -> "belongs" change as non-natural English speaker. Defining `dimension_types` as "periodicity of boundaries" sounds good to me as I know the original intent, but it would be great to hear other opinions as well.

optimade.rst
ml-evs: I think the "WITH" is optional anyway: https://github.com/Materials-Consortia/OPTiMaDe/blob/develop/optimade.rst#623substring-comparisons

> identifier STARTS WITH x: Is true if the property starts with the substring value x. The WITH keyword MAY be omitted.

merkys: Oh, right! Then there is no need to change that. We would probably want our specification to showcase also the alternative syntax constructs.

JPBergsma: Ok, I removed "WITH".
I also changed "D2" to "D1" as a chemical_formula_anonymous can never start with "A2" and end with "D2".

optimade.rst
```diff
@@ -2004,10 +2004,10 @@ dimension\_types
 
 - **Examples**:
 
-  - For a molecule: :val:`[0, 0, 0]`
-  - For a wire along the direction specified by the third lattice vector: :val:`[0, 0, 1]`
-  - For a 2D surface/slab, periodic on the plane defined by the first and third lattice vectors: :val:`[1, 0, 1]`
-  - For a bulk 3D system: :val:`[1, 1, 1]`
+  - A non periodic unit cell, for example, for a single molecule : :val:`[0, 0, 0]`
```
ml-evs, line 2007:
```suggestion
  - A nonperiodic structure, for example, for a single molecule : :val:`[0, 0, 0]`
```

optimade.rst
```diff
@@ -2004,10 +2004,10 @@ dimension\_types
...
+  - A unit cell that is periodic in the direction of the third lattice vector, for example for a carbon nanotube: :val:`[0, 0, 1]`
+  - For a 2D surface/slab, who's unit cell is periodic in the direction of the first and third lattice vectors: :val:`[1, 0, 1]`
```
ml-evs, line 2009:
```suggestion
  - For a 2D surface/slab, with a unit cell that is periodic in the direction of the first and third lattice vectors: :val:`[1, 0, 1]`
```

ml-evs (approved):

merkys (approved):

Issue #389, opened 2021-11-05
Using OPTIMADE for Biomolecular data


JPBergsma:
I recently had a discussion with Adam Hospital about making the OPTIMADE standard suitable for biomolecular data.

The most important data to share, that is not standard in Optimade, would be:

- The name of the atom. In PDB files, each atom type has a name. These names are not unique, but indicate atoms in similar positions. For example “CA“ for the carbon atom to which the side chains in an amino acid are connected. 
This for example allows only displaying atoms in the side chains or in the backbone.      

- Proteins often contain multiple subunits. Each subunit is an individual chain and these chains are in turn build up out of amino acids which in turn are build up of atoms. 
      In the PDB file format, there is a row for each atom which contains:
        ◦ The kind of amino acid/residue the atom belongs to.
        ◦ The Residue sequence number/rank of this amino acid in the chain. (the first residue does not need to be residue 1)
        ◦ The chain it belongs to (Chain ID). 

In the PDB files, this information is stored for each atom. There is however some redundancy with this approach. If you know to which residue an atom belongs, you would only need to store the rank of this residue and the chain to which this residue belongs once, instead of storing it for every atom. 

#### Possible implementations:

My idea would be to use the species field for the atoms, just as normal in OPTIMADE.  
If we want to include the number of hydrogen atoms attached to an atom, which could be useful to distinguish between single and double bonds, we would need to have more types of species as the number of attached hydrogen atoms is defined on the species level. Instead of writing “CA” for the alpha carbon atom in glycine, we could name the species “CAH2” and for other amino acids it could be “CAH”. In that case we could use the original name subfield to hold the original name: “CA”.

To include the information about the chains, I see two options. 

One option would be to add fields with a list of values belonging to each atom. 
For example, for each atom/site a property called “residue” could specify to what kind of residue the atom belongs, e.g. “Leu” for Leucine or “DA” for the nucleic acid adenosine etc. Similarly, a property called “chainid” can be used to specify to which chain each atom belongs. And finally, a “residue_sequence_number” property to store the number/rank of the residue in the sequence.

This would be very similar to the way it is currently implemented in the PDB files and would be easy to implement because of the similarity with the data structure in the PDB file. 
The disadvantages would be that there is some redundant data with this approach and that there is no top-down information. You would need to visit the values for all the atoms before you would know the number of chains and the sequence of the residues in a chain. 
[...]

PR #395, opened 2021-12-24
Jp bergsma add bio fields

JPBergsma:
After the discussion in issue #389 I have created this suggestion to incorporate biomolecular data in OPTIMADE.
The change here describes the first option mentioned in issue #389 and is in my opinion the simplest way to include this data in OPTIMADE, but not necessarily the most elegant. The fields match directly to the corresponding fields in the PDB format.
If I have enough time I will also create a second draft PR in which I will describe a second more hierarchical way to include this data.

As these fields will not be used by most OPTIMADE providers I have placed them in an appendix.


optimade.rst
```diff
@@ -2465,6 +2465,84 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+- **Examples**:
+
+  - For example for glycine and a water molecule :val:`["GLY","GLY","GLY","GLY","GLY","HOH"]`
+
+residue_sequence_number
+~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: A list of integers describing the position of the monomer/residue to which the atom belongs in a chain.
+  This matches the residue sequence number field in the of the PDB file format. 
+  There is therefore no guarantee that these numbers are ordered or unique.
```
JPBergsma, line 2515: If a variant of an already known protein is found, the same numbering is kept as in the original protein. This can cause two residues to have the same sequence_number when an extra amino acid was inserted. A number of amino acids may also be missing, causing some numbers to be skipped.

optimade.rst
```diff
@@ -2465,6 +2465,84 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+
+- **Examples**:
+
+  - For example for glycine, with implicit hydrogens and a water molecule :val:`["A","A","A","A","A","H2O_1"]`
+
+
+monomer_type
+~~~~~~~~~~~~
+
+- **Description**: A list of strings each describing the type of monomer to which an atom belongs.
```
JPBergsma, line 2496: Do we want to allow an atom to belong to multiple monomer types? For example, you could make a benzene monomer group which is in turn part of a styrene monomer?

JPBergsma:
It seems most people prefer the other approach, so I am closing this one.

PR #396, opened 2021-12-29
Jp bergsma add hierarchical bio fields

JPBergsma:
This is the second option on how to include biomolecular data within Optimade. It is the second option that is discussed in issue #389.  The first option is described in PR #395.
It introduces a `group_type` that represents kinds of molecules or groups of groups and sites. 
A group of groups is for example a protein, here the individual amino acids are the subgroups of the much larger protein strand. 
Properties that are specific for a particular molecule or group of atoms, such as a molecular formula or a mass, can be stored here.

The `groups` property is introduced to describe instances of these group_types, i.e. a particular molecule or chemical group. 

It would be nice if you could tell me which method you prefer, the one described here or the one described in PR #395, and any other remarks about these the options are of course also welcome. 

For now, I have placed these fields in the appendix, as I am not sure whether they will  be used by many OPTIMADE providers.
They can be moved to the structures section if you think they are general enough.


optimade.rst
```diff
@@ -2465,6 +2465,117 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
 
+Domain Specific Fields
+----------------------
+
+The fields below are all optional and are only used within specific research fields.
+
+group_type
+~~~~~~~~~~
+
+- **Description**: For each type of chemical group/molecule in the system there is a dictionary that describes this group/molecule.
```
JPBergsma, line 2476: Do we want to have some way of indicating that a group is at the top level, i.e. that it is a molecule?

optimade.rst
```diff
@@ -2465,6 +2465,119 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+   - :property:`mass`: float (OPTIONAL)
+   - :property:`subgroups`: list of strings (OPTIONAL)
+- **Requirements/Conventions**:
+   - **Support**: OPTIONAL support in implementations. When the :property:`groups` property is present the :property:`group_type property` MUST, however, be present as well.
+   - **Query**:  Support for queries on this property is OPTIONAL.
+     If supported, only a subset of the filter features MAY be supported.
+   - **name**: REQUIRED; The name of the group_type;
+      - The **name** value MUST be unique in the :property:`group_types` list.
+      - Strings of 3 characters or less MUST match the strings belonging to this group as defined by wwPDB at `<ftp://ftp.wwpdb.org/pub/pdb/data/monomers>`_.
+   - **molecular_formula**: OPTIONAL; The molecular formula of the molecule/group described in this group type.
```
JPBergsma, line 2491: In the list of groups of wwPDB, the groups are described without taking any bonding to other groups into account. When amino acids are polymerized, each amino acid loses a water molecule. 
I am therefore wondering whether I should have different group types for bound and unbound amino acids. In the list of wwPDB there is however just one entry for each type of amino acid.
For now, I am still wondering about how to handle this case.

PR #400, opened 2022-02-23
New biomol fields

d-beltran:
This is the third option on how to include biomolecular data within Optimade.
This has been discussed in [issue 389](https://github.com/Materials-Consortia/OPTIMADE/issues/389).

It introduces two new main fields: biomol_chains and biomol_residues. These fields describe how atoms are grouped in "chains" and "residues", two classifiers widely used in the biomolecular field.

In addition two more fields are suggested: biomol_sequences and biomol_sequence_types. These fields describe sequences of residues and they are useful for queries.

New fields are placed in the appendix, as @JPBergsma did in previous options [PR395](https://github.com/Materials-Consortia/OPTIMADE/pull/395) and [PR396](https://github.com/Materials-Consortia/OPTIMADE/pull/396)

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2568,6 +2568,48 @@ biomol_residues
     ]
   }
 
+biomol_sequences
+~~~~~~
+
+- **Description**: A list of residue sequences in current structure. It may be any type of sequence, as this type is further specified in :field:`biomol_sequence_types`.
+  Sequences may be grouped and ordered in any form (e.g. by chains, by fragments of covalently bonded atoms, etc.) as long as they make sense when querying structures by sequence.
+- **Type**: list of strings
+- **Requirements/Conventions**:
+   - **Query**:  Support for queries on this property is OPTIONAL.
+   - There MUST be the same number of values that in :field:`biomol_sequence_types`.
+   - Values SHOULD be in capital letters.
+
+- **Examples**:
+
+.. code:: jsonc
+  {
+    "biomol_sequences":[
+      'MSHHWGYG',
+      'GATTACA'
+    ]
+  }
+
+
+biomol_sequence_types
+~~~~~~
+
+- **Description**: A list of tags specifying the type of each sequence in the :field:`biomol_sequences` field.
+  The type of a sequence is defined by its components (e.g. 'aminoacids').
+- **Type**: list of strings
+- **Requirements/Conventions**:
+   - **Query**:  Support for queries on this property is OPTIONAL.
+   - There MUST be the same number of values that in :field:`biomol_sequences`.
+
+- **Examples**:
+
+.. code:: jsonc
+  {
+    "biomol_sequence_types":[
+      'aminoacids',
+      'nucleotides'
+    ]
+  }
+
 
 The Filter Language EBNF Grammar
 --------------------------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -2465,6 +2465,110 @@ Relationships with calculations MAY be used to indicate provenance where a struc
 Appendices
 ==========
 
+Domain Specific Fields
+----------------------
+
+The fields below are all optional and are only used within specific research fields.
+
+Every field has a standard domain-specific prefix.
+
+
+biomol_chains
+~~~~~~
+
+- **Description**: For each chain in the system there is a dictionary that describes this chain. Chains are groups of related residues (e.g. a polymer).
+  Databases are allowed to add more properties as long as the properties are prefixed with the database specific prefix.
+- **Type**: list of dictionaries with the properties:
+   - :property:`name`: string (REQUIRED)
+   - :property:`residues`: list of integers (REQUIRED)
+   - :property:`types`: list of strings
+   - :property:`sequences`: list of strings
+   - :property:`sequence_types`: list of strings
+- **Requirements/Conventions**:
+   - **Query**:  Support for queries on this property is OPTIONAL.
+     If supported, only a subset of the filter features MAY be supported.
+   - **name**: The chain name/letter.
+   - **residues**: A list of integers referring to the index of :field:`biomol_residues`, that belong to this chain.
+     The list SHOULD NOT be empty. The index of the first residue is 0.
+   - **types**: A list of tags specifying the type of molecules this chain contains.
+   - **sequences**: A list of residue sequences in current chain.
+   - **sequence_types**: A list of tags specifying the type of each sequence in the :property:`sequences` field. The type of a sequence is defined by its components (e.g. 'aminoacids').
+   - There SHOULD NOT be two or more chains with the same :property:`name`.
+   - Values in :property:`name` SHOULD be in capital letters.
+   - Values in :property:`name` SHOULD NOT be longer than 1 character when the number of chains is not greater than the number of letters in English alphabet (26).
+   - Values in :property:`sequences` SHOULD be in capital letters.
+   - Number of values in :property:`sequences` and :property:`sequence_types` MUST match.
+
+- **Examples**:
+
+.. code:: jsonc
+  {
+    "biomol_chains":[
+      {
+        "name": "A",
+        "residues":[0,1,2,3, ...],
+        "types": ['protein', 'ions'],
+	      "sequences": ['MSHHWGYG'],
+	      "sequence_types": ['aminoacids']
+      },
+      {
+        "name": "B",
+        "residues":[54,55,56,57, ...],
+        "types": ['nucleic acid'],
+	      "sequences": ['GATTACA'],
+	      "sequence_types": ['nucleotides']
+      },
+    ]
+  }
+
+biomol_residues
+~~~~~~
+
+- **Description**: For each residue in the system there is a dictionary that describes this residue. Residues are groups of related atoms (e.g. an aminoacid).
+  Databases are allowed to add more properties as long as the properties are prefixed with the database specific prefix.
+- **Type**: list of dictionaries with the properties:
+   - :property:`name`: string (REQUIRED)
+   - :property:`number`: integer (REQUIRED)
+   - :property:`insertion_code`: string or null (REQUIRED)
+   - :property:`sites`: list of integers (REQUIRED)
+- **Requirements/Conventions**:
+   - **Query**:  Support for queries on this property is OPTIONAL.
+     If supported, only a subset of the filter features MAY be supported.
+   - **name**: The residue name
+   - **number**: The residue number according to source notation.
+   - **insertion_code**: The residue insertion code. It MUST NOT be longer than 1 character. It MAY be null.
+   - **sites**: A list of integers referring to the index of :field:`cartesian_site_positions`, that belong to this residue.
+     The list SHOULD NOT be empty. The index of the first site is 0.
+   - There MUST NOT be two or more residues with the same integer in :property:`sites`.
+   - All :property:`name` and :property:`insertion_code` values SHOULD be in capital letters.
+
+- **Examples**:
+
+.. code:: jsonc
+  {
+    "biomol_residues":[
+      {
+        "name": "PHE",
+	      "number": 17,
+	      "insertion_code": null,
+        "sites":[0,1,2,3, ...]
+      },
+      {
+        "name": "ASP",
+	      "number": 18,
+	      "insertion_code": null,
+        "sites":[17,18,19,20, ...]
+      },
+      {
+        "name": "LEU",
+	      "number": 18,
+        "insertion_code": "A",
+        "sites":[29,30,31, ...]
+      },
+    ]
+  }
+
+
 The Filter Language EBNF Grammar
 --------------------------------
 


optimade.rst
```diff
@@ -2465,6 +2465,152 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+Domain Specific Fields
+----------------------
+
+The fields below are all optional and are only used within specific research fields.
+
+Every field has a standard domain-specific prefix.
+
+
+biomol_chains
+~~~~~~
```
JPBergsma, line 2477:
```suggestion
biomol_chains
~~~~~~~~~~~
```
I think the line with tildes is supposed to be as long as the title

optimade.rst
```diff
@@ -2465,6 +2465,154 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+        "insertion_code": "A",
+        "sites":[29,30,31, ...]
+      },
+    ]
+  }
+
+biomol_sequences
+~~~~~~~~~~~~~~~~
+
+- **Description**: A list of residue sequences in current structure. It may be any type of sequence, as this type is further specified in :field:`biomol_sequence_types`.
```
JPBergsma, line 2576: Are you not duplicating data here? It seems that the sequences already occur in the biomol_chains field. It would perhaps be better to make a reference from the biomol_chains to these sequences.

Is the main point of this field not to enable querying on the sequences?
If that is the case, it may be better to set the query ability to SHOULD.

optimade.rst
```diff
@@ -2490,7 +2490,9 @@ biomol_chains
    - **name**: The chain name/letter.
    - **residues**: A list of integers referring to the index of :field:`biomol_residues`, that belong to this chain.
      The list SHOULD NOT be empty. The index of the first residue is 0.
-   - **types**: A list of tags specifying the type of molecules this chain contains.
+   - **types**: A list of custom tags/labels specifying the type of molecules this chain contains (e.g. 'protein').
+     This field is useful as an overview of every chain and as a query target for the structure.
+     Labels in this field are non-standard. Every implementation may use different labels according to its needs.
```
merkys, line 2495: I would advise standardizing labels at least for the most common molecule types to benefit the queryability of the field. Implementations could use their own labels, but prefixed with their own database-specific prefixes.

optimade.rst
d-beltran: You are right, data would be duplicated.

I was thinking that it could be better specifying which chains (and even which residues) are included on each sequence, by their indices, and then removing the _sequences_ and _sequence_types_ fields on _biomol_chains_.

Actually, instead of making new fields for this we could reshape sequences like a list of dictionaries, to make it coherent with the other biomol field formats. What do you think?

```rst

biomol_sequences
~~~~~~~~~~~~~~~~

- **Description**: A list of residue sequences in current structure.
 Every sequence is a dictionary which includes the sequence itself and the type of sequence it is.
 Every sequence may include a list of chain and residue indices.
 Sequences may be grouped and ordered in any form (e.g. by chains, by fragments of covalently bonded atoms, etc.) as long as they make sense when querying structures by sequence.
- **Type**: list of dictionaries with the properties:
   - :property:`sequence`: string (REQUIRED)
   - :property:`type`: string (REQUIRED)
   - :property:`chains`: list of integers
   - :property:`residues`: list of integers
   - **Query**:  Queries on this property SHOULD be supported.
   - **sequence**: A string with a letter for each residue in the sequence. Letters SHOULD be capital letters.
   - **type**: The type of a sequence is defined by its components (e.g. 'aminoacids').
   - **chains**: A list of integers referring to indices in :field:`biomol_chains` for chains which include this sequence totally or partially.
   - **residues**: A list of integers referring to indices in :field:`biomol_residues` for residues included in the sequence.
   Indices start the count at 0.
   There MUST NOT be repeated indices both in :property:`chains` and :property:`residues`.



- **Examples**:

.. code:: jsonc
  {
    "biomol_sequences":[
      {
        sequence: 'MSHHWGYG',
        type: 'aminoacids'
      },
      {
        sequence: 'GATTACA',
        type: 'nucleotides'
      }
    ]
  }

```

optimade.rst
d-beltran: Good idea. This way we could save a lot of work.

I have been searching for references for our labels in the current mmCIF format (the future PDB standard) and [this](https://mmcif.wwpdb.org/dictionaries/mmcif_em.dic/Items/_em_entity_assembly.type.html) is the best I have found. They are meant for assemblies and they do not totally suit me, but I will try to resemble them.

So I suggest the standard labels to be the following: 'PROTEIN', 'NUCLEIC ACID', 'CARBOHYDRATES', 'LIPID', 'MEMBRANE', 'LIGAND', 'ION', 'SOLVENT', 'OTHER'.

If you agree I will commit changes soon.

JPBergsma: Yes, it seems like a good idea to define these labels.

merkys: I very much like basing on mmCIF. Maybe there is already a JSON representation for mmCIF data?

> So I suggest the standard labels to be the following: 'PROTEIN', 'NUCLEIC ACID', 'CARBOHYDRATES', 'LIPID', 'MEMBRANE', 'LIGAND', 'ION', 'SOLVENT', 'OTHER'.
> 
> If you agree I will commit changes soon.

These labels sound very good. I would just render them in lowercase and describe the use of prefixes for custom labels in the form of `<prefix>:<label>`.

d-beltran: > Maybe there is already a JSON representation for mmCIF data?

In a fast search I found several mmCIF to JSON parsers and one of them seems to be the [official](https://pdbj.org/tools/cif2json) one.

> I would just render them in lowercase and describe the use of prefixes for custom labels in the form of <prefix>:<label>

Allright

optimade.rst
JPBergsma: We could indeed turn the sequences into a dictionary.

I was still wondering whether there is a clear hierarchical structure? (That a chain can have one or more sequences, but a sequence can not contain more than one chain? In that case, we would not need to duplicate the residues. 

I was also wondering how the situation is handled, where there are multiple chains with the same sequence.
Will there be multiple sequence entries with the same sequence ? My idea was to make a sequence unique, so it only occurs once in the biomol_sequence field. But if you include the residues, you would need to have a separate sequence entry for each chain.

optimade.rst
```diff
@@ -2465,6 +2465,156 @@ Relationships with calculations MAY be used to indicate provenance where a struc
...
+        "name": "B",
+        "residues":[54,55,56,57, ...],
+        "types": ['nucleic acid'],
+	      "sequences": ['GATTACA'],
+	      "sequence_types": ['nucleotides']
+      },
+    ]
+  }
+
+biomol_residues
```
JPBergsma, line 2528:
```suggestion
_biomol_residues
```
I think the conclusion of the meeting last Friday was that it would be better to prepend the domain specific fields in the same way as the database providers prefix. So the names of the fields should prepended with an underscore.

d-beltran: I tried to do the commit from your suggestion but I did not got it.
I did a new commit to change this.

optimade.rst
d-beltran: > I was still wondering whether there is a clear hierarchical structure? (That a chain can have one or more sequences, but a sequence can not contain more than one chain? In that case, we would not need to duplicate the residues.

In my previous suggestion the chains property is a list of integers so a sequence may contain more than one chain. This is important since a polymer may be splitted in several chains. In the other hand, a chain may contain the whole sequence of a polymer and more things at the same time, so not all residues in the chain would be part of the sequence. Then listing residues in the chain makes sense.

> I was also wondering how the situation is handled, where there are multiple chains with the same sequence.
Will there be multiple sequence entries with the same sequence ? My idea was to make a sequence unique, so it only occurs once in the biomol_sequence field. But if you include the residues, you would need to have a separate sequence entry for each chain.

I did not think about this and you are right. Then we can forget about residues to make everything easier.

optimade.rst
JPBergsma: That is ok, sometimes it is also easier to include it in your own branch first. I just wanted to make sure we would not forget to make the change.

JPBergsma:
I just came across another issue. I am trying to implement the standard we described here to aid our discussion. To have some example data, I downloaded a random trajectory from the internet. This trajectory has a non-standard amino acid in it. 
How do you suggest we handle this case? It seems a one letter code is not sufficient to describe all amino acids in a sequence.

JPBergsma:
Yes, I can do that. It would make it harder to search for sequences with non-standard amino acids, but those are probably quite rare anyway.

optimade.rst
```diff
@@ -3148,6 +3151,129 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+- **Requirements/Conventions**:
+   - **Query**:  Support for queries on this property is OPTIONAL.
+     If supported, only a subset of the filter features MAY be supported.
+   - **name**: The residue name
+   - **number**: The residue number according to source notation.
+   - **insertion_code**: The residue insertion code. It MUST NOT be longer than 1 character. It MAY be null.
+   - **chain**: The chain number this residue belongs to.
+   - Values in :property:`chain` SHOULD be in capital letters.
+   - Values in :property:`chain` SHOULD NOT be longer than 1 character when the number of chains is not greater than the number of letters in English alphabet (26).
+   - There MUST NOT be two or more residues with the same integer in :property:`sites`.
```
JPBergsma, line 3180: We no longer have a sites property so I think this line can go.

optimade.rst
```diff
@@ -3148,6 +3151,129 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Index of the residues at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_).
+  The properties of the residues are found in the property `_biomol_residues`_.
+- **Type**: list of integers.
+- **Requirements/Conventions**:
+  - **Support**: SHOULD be supported by all biomol implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - MUST have length equal to the number of sites in the structure (first dimension of the list property `cartesian_site_positions`_).
```
JPBergsma, line 3219:
```suggestion
  - The number of values MUST be equal to :property: `nsites`, i.e. the number of sites in the structure.
```

optimade.rst
```diff
@@ -3148,6 +3151,129 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+  - Residue indices mentioned in the `_biomol_residues_at_sites`_ list MUST be lower than the length of the list property `_biomol_residues`_ (i.e. for each value in the `_biomol_residues_at_sites`_ list there MUST exist one dictionary in the `_biomol_residues`_ list with the index equal to the corresponding `_biomol_residues_at_sites`_ value).
```
JPBergsma, line 3220:
```suggestion
  - Each value in the `_biomol_residues_at_sites`_ list MUST correspond to  the index of one the dictionaries in the `_biomol_residues`_ list.
```
I think this point should be clear from the previous text. So I tried to shorten it. I am also ok with leaving this line out entirely.

optimade.rst
d-beltran:
```suggestion
```

optimade.rst
```diff
@@ -3148,6 +3151,129 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+  }
+  
+_biomol_residues_at_sites
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Index of the residues at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_).
+  The properties of the residues are found in the property `_biomol_residues`_.
+- **Type**: list of integers.
+- **Requirements/Conventions**:
+  - **Support**: SHOULD be supported by all biomol implementations, i.e., SHOULD NOT be :val:`null`.
```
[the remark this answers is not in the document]
d-beltran:
```suggestion
  - **Support**: MUST be supported when `_biomol_residues`_ is present as well, i.e., MUST NOT be :val:`null`.
```

optimade.rst
```diff
@@ -3148,6 +3151,132 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+
+- **Examples**:
+
+  - :val:`[0,0,0,0,0,0,0,0,1,1,1,1,1,1,1,1,1, ... ]` indicates that the first 8 sites belong to the first residue in the the `residues`_ list, while the 9 following atoms belong to the second residue.
+
+_biomol_site_sequences
+~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: A list of dictionaries, each representing a linear segment of covalently-linked standard or modified amino acids or nucleotides having atoms with coordinates in sites. 
+  The order of the elements in the `_biomol_site_sequences`_ list is not relevant.
```
JPBergsma, line 3229:
```suggestion
  The elements in the `_biomol_site_sequences`_ list are unordered.
```
I think this has the same meaning and is a bit shorter.

optimade.rst
```diff
@@ -3148,6 +3151,132 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+  Each dictionary in the list holds two keys: sequence and type. The sequence is a string of one-letter codes identifying each amino acid or nucleotide as defined by the `mmCIF standard <https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_entity_poly.pdbx_seq_one_letter_code.html>`__.
+  The type is a string defining the monomers of the sequence. Accepted values are “polypeptide” for amino acids, “polydeoxyribonucleotide”  for deoxyribonucleotides and “polyribonucleotide” for ribonucleotides, according to the `mmCIF standard <https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_pdbx_reference_linked_entity.link_to_entity_type.html>`__.
```
JPBergsma, line 3231:
```suggestion
  The type is a string defining what kind of monomers are in the sequence.
    Accepted values are “polypeptide” for amino acids, “polydeoxyribonucleotide”  for deoxyribonucleotides and “polyribonucleotide” for ribonucleotides, according to the `mmCIF standard <https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_pdbx_reference_linked_entity.link_to_entity_type.html>`__.
```

optimade.rst
```diff
@@ -3148,6 +3151,132 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+
+.. code:: jsonc
+  {
+    "_biomol_site_sequences":[
+      {
+        sequence: 'MSHHWGYG',
+        type: 'polypeptide'
+      },
+      {
+        sequence: 'GATTACA',
```
JPBergsma, line 3250:
```suggestion
        sequence: '(DG)(DA)(DT)(DT)(DA)(DC)(DA)',
```
If this sequence represents DNA, should the DNA nucleotides not be encoded by two letter codes according to the mmCIF standard ?

optimade.rst
```diff
@@ -3148,6 +3151,132 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+        type: 'polydeoxyribonucleotide'
+      }
+    ]
+  }
+  
+_biomol_full_sequences
+~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: A list of dictionaries, each representing a linear segment of covalently-linked standard or modified amino acids or nucleotides including residues without coordinates in sites.
+  The order of the elements in the `_biomol_full_sequences`_ list is not relevant.
```
JPBergsma, line 3260:
```suggestion
  The elements in the `_biomol_full_sequences`_ list are unordered.
```

optimade.rst
```diff
@@ -3148,6 +3151,132 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+
+_biomol_residues
+~~~~~~~~~~~~~~~~
+
+- **Description**: For each residue in the system there is a dictionary that describes this residue. Residues are groups of related atoms (e.g. an aminoacid).
+  Databases are allowed to add more properties as long as the properties are prefixed with the database specific prefix.
+- **Type**: list of dictionaries with the properties:
+   - :property:`name`: string (REQUIRED)
+   - :property:`number`: integer (REQUIRED)
+   - :property:`insertion_code`: string or null (REQUIRED)
```
JPBergsma, line 3169: Would icode be clear enough? It is a bit shorter than insertion_code.

optimade.rst
```diff
@@ -3148,6 +3151,133 @@ Relationships with files may be used to relate an entry with any number of :entr
...
+_biomol_residues
+~~~~~~~~~~~~~~~~
+
+- **Description**: For each residue in the system there is a dictionary that describes this residue. Residues are groups of related atoms (e.g. an aminoacid).
+  Databases are allowed to add more properties as long as the properties are prefixed with the database specific prefix.
+- **Type**: list of dictionaries with the properties:
+   - :property:`name`: string (REQUIRED)
+   - :property:`number`: integer (REQUIRED)
+   - :property:`icode`: string or null (REQUIRED)
+   - :property:`chain`: string (OPTIONAL)
```
merkys, line 3170: What does it mean when `chain` is missing?

merkys: Thanks for explanation. But maybe then it would make sense to make `chain` mandatory and faithfully retain the space character (` `) as its value?

d-beltran: Sure, it also works to me.

We also talked about not getting constrained by the limits of PDB format regarding the 1 character string in the chain name, so the missing chain could also be 'Not defined', '', null or many others. As you prefer.

PR #401, opened 2022-03-02
Allow provider-specific nested properties within Optimade properties.

### LINKED ISSUE #397: Database-provider-specific keys in all dictionaries
As of v1.1.0, providers are allowed to add custom entry properties, provided they prefix them with provider-specific prefixes. I suggest extending this to allow custom keys in all dictionaries. Thus if a provider wants to add something to each of every species, for example, isotope, they will be able to do so, viz.:
```json
"species": [
  { "name": "13C", "chemical_symbols": [ "C" ], "_exmpl_isotope": 13 }
]
```
This would allow passing additional nonstandard information via the same OPTIMADE interface without dragging possibly huge parallel lists or tables. Viz. how isotope information would be carried conforming to v1.1.0:
```jsonc
"_exmpl_isotopes": { "13C": 13 } // here the key is the name of species
```

JPBergsma:
In the discussion on the new fields for Biomolecular data #389, it was not clear whether it was allowed to add database specific fields within OPTIMADE properties. With this PR I hope to clarify that this indeed allowed.
This has also been pointed out by @merkys in #397

Closes #397


merkys:
@JPBergsma you may also specify that this PR closes #397.

optimade.rst
```diff
@@ -1783,7 +1783,10 @@ last\_modified
 database-provider-specific properties
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- **Description**: Database providers are allowed to insert database-provider-specific entries in the output of both standard entry types and database-provider-specific entry types.
+- **Description**: Database providers are allowed to add database-provider-specific properties in the output of both standard entry types and database-provider-specific entry types.
+  These database-provider-specific properties can be nested within other fields.
```
merkys, line 1787: "other fields" -> "other properties"

optimade.rst
```diff
@@ -1783,7 +1783,10 @@ last\_modified
...
+  For example, the database-provider-specific property :property:`_exmpl_oxydation_state`, can be placed within the OPTIMADE field :field:`species`.
```
merkys, line 1788: "OPTIMADE field" -> "OPTIMADE property"

merkys (dismissed):
Thanks for putting this together, @JPBergsma. This is a small and, IMO, uncontroversial change which opens up new possibilities for providers to deliver nonstandard data in compact manner.

I would suggest small changes to the terms in use in this PR. "Field" per OPTIMADE terms means a key in associative-array-type (a.k.a. dictionaries). I think it is more precise to talk about nested *properties* as this already happens in the specification.

JPBergsma:
> @JPBergsma you may also specify that this PR closes #397.

Sorry, I had not thought about your issue. I will update the description.

optimade.rst
JPBergsma: I have changed this in the text.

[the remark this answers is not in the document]
JPBergsma: Thanks for spotting the spelling mistake.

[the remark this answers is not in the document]
JPBergsma: If this PR is accepted, we would indeed have to add it to the Optimade python tools.
Shall I make an issue about this now, or do you want to wait until this PR is accepted?

optimade.rst
```diff
@@ -1783,7 +1783,10 @@ last\_modified
 database-provider-specific properties
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- **Description**: Database providers are allowed to insert database-provider-specific entries in the output of both standard entry types and database-provider-specific entry types.
+- **Description**: Database providers are allowed to add database-provider-specific properties in the output of both standard entry types and database-provider-specific entry types.
+  These database-provider-specific properties can be nested within other properties.
```
rartino, line 1787: I don't think we can call keys in a dictionary "properties" with the current terminology we have. How about something like this?:
```suggestion
  Similarly, an implementation MAY also add keys prefixed by a database-provider-specific prefix to dictionary properties, as well as their sub-dictionaries to arbitrary level.
```

merkys: Term "nested property" is used throughout the specification, thus my first instinct was to reuse it here as well.

JPBergsma: I think the term nested is a bit broader than "key" and "dictionary". We also use if for lists within lists.
But I do not think that we should allow adding properties to any data type, other than dictionaries. So we could use the terms "key" and "dictionary" as well 

I have tried to shorten @rartino's  suggestion a bit. Would this be ok?

"Similarly, an implementation MAY add keys with a database-provider-specific prefix to dictionary properties and their sub-dictionaries."

merkys: I tend to think term "property" stands for something limited to values of `attributes` key in JSON:API objects. @JPBergsma's suggestion "dictionary properties and their sub-dictionaries" sounds good to me, if "nested properties" do not seem obvious.

I now see that in #397 I suggested allowing provider-specific key-value pairs in any dictionary anywhere in OPTIMADE JSON output. Thus "nested properties" is just a subset of my original suggestion. I do not intend to stall this nice PR, so maybe the extension could be made later.

rartino (approved):
This version looks good to me. Thanks for the work on this!

optimade.rst
rartino: > I tend to think term "property" stands for something limited to values of attributes key in JSON:API objects.

But we use 'property' throughout the specification to refer to data communicated via OPTIMADE entries (admittedly the term we've defined for this is 'entry property', but it is being abbrevated to 'property' in many places.) IMO we need a clear term to refer to that, so I still think it makes sense to try to stay away from using 'property' with other meanings.

> I now see that in #397 I suggested allowing provider-specific key-value pairs in any dictionary anywhere in OPTIMADE JSON output. Thus "nested properties" is just a subset of my original suggestion. I do not intend to stall this nice PR, so maybe the extension could be made later.

Hmm... I'm confused by this. Above you say "in OPTIMADE JSON output", but that context isn't clear in #397 itself. So there are two possible readings:
1. #397 refers to dictionaries in the entry data. But in that case I don't see the difference between #397 and what the present PR implements.
2. If you really mean to suggest that specifially the "json" output format should allow these extra keys in any dictionaries, that could be considered. However, in that case it almost orthogonal to the present PR that explains how to add *output-format-agnostic* extra data in entries?

ml-evs (approved):
LGTM, feel free to raise an issue for python-tools if you want to support this there too

optimade.rst
merkys: > > I tend to think term "property" stands for something limited to values of attributes key in JSON:API objects.
> 
> But we use 'property' throughout the specification to refer to data communicated via OPTIMADE entries (admittedly the term we've defined for this is 'entry property', but it is being abbrevated to 'property' in many places.) IMO we need a clear term to refer to that, so I still think it makes sense to try to stay away from using 'property' with other meanings.

I agree. Sorry about dragging JSON and JSON:API into this - I find it difficult to express myself when trying to decouple from them. What I do not quite understand now is whether things like relationships fall under the umbrella of "entry property" or not. In other words, does this PR allow provider-specific keys in relationship dictionaries, `meta`, `links` and so on.

> > I now see that in #397 I suggested allowing provider-specific key-value pairs in any dictionary anywhere in OPTIMADE JSON output. Thus "nested properties" is just a subset of my original suggestion. I do not intend to stall this nice PR, so maybe the extension could be made later.
> 
> Hmm... I'm confused by this. Above you say "in OPTIMADE JSON output", but that context isn't clear in #397 itself. So there are two possible readings:
> 
> 1. [Database-provider-specific keys in all dictionaries #397](https://github.com/Materials-Consortia/OPTIMADE/issues/397) refers to dictionaries in the entry data. But in that case I don't see the difference between [Database-provider-specific keys in all dictionaries #397](https://github.com/Materials-Consortia/OPTIMADE/issues/397) and what the present PR implements.

Yes, this is what I meant.

> 2. If you really mean to suggest that specifially the "json" output format should allow these extra keys in any dictionaries, that could be considered. However, in that case it almost orthogonal to the present PR that explains how to add _output-format-agnostic_ extra data in entries?

No. As said, I mentioned JSON just because OPTIMADE does not support any other output format beside that.

merkys:
Glad to see this PR in! I was late to approve it, but I would still like to express even post factum support.

optimade.rst
rartino: > What I do not quite understand now is whether things like relationships fall under the umbrella of "entry property" or not. In other words, does this PR allow provider-specific keys in relationship dictionaries, meta, links and so on.

Feel free to disagree - but my reading of the present specification is that relationships is an output format-agnostic feature of OPTIMADE that is separate from (entry) properties. But, as such, it is defined as simply tying together an entry with a list of ids of entries of different (or the same) entry type + a human readable description. Nowhere in that definition appears a output format-agnostic "place" to put your database-provider specific keys.  

However, for specifially the JSON API output format, this simply references relationships to be encoded via a [JSON API relationship](https://jsonapi.org/format/1.0/#document-resource-object-relationships) which defines a `meta` field that points to a [JSON API meta object](https://jsonapi.org/format/1.0/#document-meta), that can contain any members. I see no reason this freedom isn't inherited by the OPTIMADE specification.

>> #397 refers to dictionaries in the entry data. But in that case I don't see the difference between #397 and what the present PR implements.
> Yes, this is what I meant.

But, then, what is still missing after this PR? Is it "only" provider-specific-fields in relationships you are still missing?

merkys: > Feel free to disagree - but my reading of the present specification is that relationships is an output format-agnostic feature of OPTIMADE that is separate from (entry) properties. But, as such, it is defined as simply tying together an entry with a list of ids of entries of different (or the same) entry type + a human readable description. Nowhere in that definition appears an output format-agnostic "place" to put your database-provider specific keys.
> 
> However, for specifially the JSON API output format, this simply references relationships to be encoded via a [JSON API relationship](https://jsonapi.org/format/1.0/#document-resource-object-relationships) which defines a `meta` field that points to a [JSON API meta object](https://jsonapi.org/format/1.0/#document-meta), that can contain any members. I see no reason this freedom isn't inherited by the OPTIMADE specification.

I agree with your conclusions. Current specification does not seem to standardize fields in `meta` dictionary of JSON API relationships. There is, though, a suggestion to optionally accompany relationships with human-readable descriptions, but the specification does not standardize that.

> > > #397 refers to dictionaries in the entry data. But in that case I don't see the difference between #397 and what the present PR implements.
> 
> > Yes, this is what I meant.
> 
> But, then, what is still missing after this PR? Is it "only" provider-specific-fields in relationships you are still missing?

I do not think anything is missing. I am just trying to understand how to correctly interpret this PR. In #397 I was thinking about all the dictionaries supported by an output format, but I currently see no need to go beyond dictionary properties and their sub-dictionaries.

Issue #35, opened 2018-06-14
Structure symmetry in OPTIMADE


rartino:
We need to think a bit about how we represent symmetry in relation to structures if we are going to include that. There is indeed good reason to allow queries on symmetry data when available.

If symmetry information is given, it may be best to require giving the list of symmetry operations, which is what strict CIF files do to avoid ambiguity. In addition, spacegroup number, ML, and Hall symbol can also be allowed as optional go give. However, it is probably unwise to allow giving those, *without* also giving symmetry operations, since that would replicate issues found for less strict CIF files.

The question is how to represent symmetry operations. Some alternatives:
1. The cif x,y,z format
2. The cif x,y,z format, but with canonicalized individual symmetry operations (e.g., require translations to be [0-1), etc.
3. The cif x,y,z format but with the complete set of symmetry operations canonicalized, e.g., by sorting alternative 2.
4. Some other, more compact, format we come up with.

rartino:
Here is an argument for requiring a fully canonicalized output:

One of the reasons for a non-canonical output was that otherwise one would frequently have to deal with implementations that return incorrect output. However, 'unsorted' symmetry operations is far from the only error an implementation can do when returning symmetry operators. One easy mistake is to return a set of symmetry operators that just cannot be correct together. (Because there is only a finite number of total possibilities). If we require translation coefficients to be [0-1), it is easy to return operations that do not fulfill this, etc. Hence, to avoid incorrect output it would be good if this field can be properly validated by our json schemas to belong exactly to the set of possibilities.

If we require a canonical output (e.g., by a sorted list of operations), then one of us with a framework that uses complete sets of symmetry operations can just generate a list of ALL valid combinations of symmetry operations and we can put that in a schema and say: "The output MUST be exactly one of these alternatives". This is easy to validate. An unsorted list of operations (and possibly with several possibilities for coefficients that mean the same thing) is not easy to validate.

sauliusg:
> can just generate a list of ALL valid combinations of symmetry operations and we can put that in a schema and say: "The output MUST be exactly one of these alternatives".

I suspect this will be impractical since the list of combinations will be huge beyond manageability...

rartino:
> I suspect this will be impractical since the list of combinations will be huge beyond manageability...

[...]

PR #405, opened 2022-05-30
Space group additions: `space_group_hall` and `space_group_it_number`

merkys:
Partially addresses #35  This PR introduces two symmetry space group notations, Hall symbol and IT number, closely following their usage in CIF. Both Hall symbol and IT number have their drawbacks: in some cases both need additional information to uniquely define symmetry space group of a crystal. To cover this drawback, a list of symmetry operations is needed, and its introduction is envisaged in #35. Nevertheless standardizing Hall symbol and IT number would facilitate searches on these widely used notations.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2056,6 +2056,17 @@ lattice\_vectors
 
   - :val:`[[4.0,0.0,0.0],[0.0,4.0,0.0],[0.0,1.0,4.0]]` represents a cell, where the first vector is :val:`(4, 0, 0)`, i.e., a vector aligned along the :val:`x` axis of length 4 Å; the second vector is :val:`(0, 4, 0)`; and the third vector is :val:`(0, 1, 4)`.
 
+space\_group\_it\_number
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Symmetry space group number as defined in International Tables for Crystallography Vol. A.
+- **Type**: integer
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - The integer value MUST be between 1 and 230.
+
 cartesian\_site\_positions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2056,6 +2056,22 @@ lattice\_vectors
 
   - :val:`[[4.0,0.0,0.0],[0.0,4.0,0.0],[0.0,1.0,4.0]]` represents a cell, where the first vector is :val:`(4, 0, 0)`, i.e., a vector aligned along the :val:`x` axis of length 4 Å; the second vector is :val:`(0, 4, 0)`; and the third vector is :val:`(0, 1, 4)`.
 
+space\_group\_hall
+~~~~~~~~~~~~~~~~~~
+
+- **Description**: Symmetry space group name as defined in Hall, S. R. (1981), Acta Cryst. A37, 517-525 and erratum (1981), A37, 921.
+- **Type**: string
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - Each component of the space group name MUST be separated by a single space symbol.
+
+- **Examples**:
+
+  - :val:`P 2c -2ac`
+  - :val:`-I 4bd 2ab 3`
+
 space\_group\_it\_number
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2065,7 +2065,9 @@ space\_group\_hall
 
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: Support for queries on this property is OPTIONAL.
-  - Each component of the space group name MUST be separated by a single space symbol.
+  - Each component of the Hall symbol MUST be separated by a single space symbol.
+  - If there exists a standard Hall symbol which represents the symmetry it SHOULD be used.
+  - MUST be null if :property:`nperiodic_dimensions` is not equal to 3.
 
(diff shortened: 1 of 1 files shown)


rartino:
Correct me if I am wrong, but I have come to the conclusion that it is *not* 100% safe to assume there is a standard Hall symbol for all possibilities - in case a structure isn't placed with a "proper" origin. Is it absolutely safe to assume there is a non-standard Hall symbol that describes structures given with a completely arbitarily origin?

Also, I think I'm for MAY support level on these. If you don't know the symmetry information it can be very expensive (arguably sometimes even impossible) to assign them.

merkys:
> Correct me if I am wrong, but I have come to the conclusion that it is _not_ 100% safe to assume there is a standard Hall symbol for all possibilities - in case a structure isn't placed with a "proper" origin. Is it absolutely safe to assume there is a non-standard Hall symbol that describes structures given with a completely arbitary origin?

Right, but maybe we can say that only standard Hall symbols are supported, and they should be provided only for structures with "proper" origin? My point is that if we cannot use standard Hall symbol/IT number to unambiguously define symmetry, these fields should be `null`, and in future symmetry operator lists should be used to define such situations.

> Also, I think I'm for MAY support level on these. If you don't know the symmetry information it can be very expensive (arguably sometimes even impossible) to assign them.

Sure, I am fine with OPTIONAL/MAY support level.

optimade.rst
```diff
@@ -2056,6 +2056,33 @@ lattice\_vectors
...
+space\_group\_hall
+~~~~~~~~~~~~~~~~~~
+
+- **Description**: Symmetry space group name as defined in Hall, S. R. (1981), Acta Cryst. A37, 517-525 and erratum (1981), A37, 921.
+- **Type**: string
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - Each component of the space group name MUST be separated by a single space symbol.
```
rartino, line 2068:
```suggestion
  - Each component of the Hall symbol MUST be separated by a single space symbol.
  - If there exists a standard Hall symbol which represents the symmetry it SHOULD be used.
  - MUST be null if :property:`nperiodic_dimensions` is not equal to 3.
```

optimade.rst
```diff
@@ -2056,6 +2056,33 @@ lattice\_vectors
 
   - :val:`[[4.0,0.0,0.0],[0.0,4.0,0.0],[0.0,1.0,4.0]]` represents a cell, where the first vector is :val:`(4, 0, 0)`, i.e., a vector aligned along the :val:`x` axis of length 4 Å; the second vector is :val:`(0, 4, 0)`; and the third vector is :val:`(0, 1, 4)`.
 
+space\_group\_hall
+~~~~~~~~~~~~~~~~~~
+
+- **Description**: Symmetry space group name as defined in Hall, S. R. (1981), Acta Cryst. A37, 517-525 and erratum (1981), A37, 921.
```
rartino, line 2062: To be closer to the phrasing of the IUCr, and strictly expressing that it is not just any Hall symbol, but the one that represents the structure:
```suggestion
- **Description**: A Hall space group symbol representing the symmetry of the structure as defined in Hall, S. R. (1981), Acta Cryst. A37, 517-525 and erratum (1981), A37, 921.
```

optimade.rst
```diff
@@ -2056,6 +2056,33 @@ lattice\_vectors
...
+
+- **Examples**:
+
+  - :val:`P 2c -2ac`
+  - :val:`-I 4bd 2ab 3`
+
+space\_group\_it\_number
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Symmetry space group number as defined in International Tables for Crystallography Vol. A.
```
rartino, line 2078:
```suggestion
- **Description**: Space group number for the structure assigned by the International Tables for Crystallography Vol. A.
```

optimade.rst
```diff
@@ -2056,6 +2056,33 @@ lattice\_vectors
...
+- **Type**: integer
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - The integer value MUST be between 1 and 230.
```
rartino, line 2084:
```suggestion
  - The integer value MUST be between 1 and 230.
  - MUST be null if :property:`nperiodic_dimensions` is not equal to 3.
```

rartino (changes requested):
A few suggested changes

optimade.rst
```diff
@@ -2056,6 +2056,33 @@ lattice\_vectors
...
   - :val:`[[4.0,0.0,0.0],[0.0,4.0,0.0],[0.0,1.0,4.0]]` represents a cell, where the first vector is :val:`(4, 0, 0)`, i.e., a vector aligned along the :val:`x` axis of length 4 Å; the second vector is :val:`(0, 4, 0)`; and the third vector is :val:`(0, 1, 4)`.
 
+space\_group\_hall
+~~~~~~~~~~~~~~~~~~
+
+- **Description**: Symmetry space group name as defined in Hall, S. R. (1981), Acta Cryst. A37, 517-525 and erratum (1981), A37, 921.
+- **Type**: string
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
```
rartino, line 2066:
```suggestion
  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
```

optimade.rst
```diff
@@ -2056,6 +2056,33 @@ lattice\_vectors
...
+  - :val:`-I 4bd 2ab 3`
+
+space\_group\_it\_number
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: Symmetry space group number as defined in International Tables for Crystallography Vol. A.
+- **Type**: integer
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
```
rartino, line 2084:
```suggestion
  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
```

ml-evs (approved):

PR #411, opened 2022-06-01
Backoff time

### LINKED ISSUE #409: Backoff time
During the workshop it has come up that some databases are concerned about the amount of traffic. It has been suggested that we could add an optional field to `meta` to indicate a suggested backoff time that an implementation should wait until the next query. If your client is looping over something with multiple re-queries, it is then meant to include a pause of this length between the queries.

merkys:
Fixes #409 by introducing `request_delay`, an optional field in `meta` of any response to ask the client to wait specified number of seconds (float) before issuing subsequent request.


dwinston:
Given that HTTP is assumed as the API access protocol, we can also leverage the standard `Retry-After` response header: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After>.

merkys:
@dwinston fair point. I will adjust the proposal to add `Retry-After`.

rartino (dismissed):
Consider moving the error code suggestion as discussed.

optimade.rst
```diff
@@ -493,6 +493,9 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   - **data\_available**: an integer containing the total number of data resource objects available in the database for the endpoint.
   - **last\_id**: a string containing the last ID returned.
   - **response\_message**: response string from the server.
+  - **request\_delay**: a non-negative float giving time in seconds a client should wait before issuing a subsequent request.
+    An implementation MAY respond with :http-error:`429 Too Many Requests` error if a client refuses to respect the indicated delay.
+    In that case the response SHOULD contain ``Retry-After` HTTP header <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After>`__ to instruct the client to wait before retrying.
```
ml-evs, line 498:
```suggestion
    In that case the response SHOULD contain a ` `Retry-After` HTTP header <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After>`__ to instruct the client to wait before retrying.
```

The formatting here doesn't  seem to work (linking from monospaced text), perhaps the above fixes it?

optimade.rst
```diff
@@ -493,6 +493,9 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   - **data\_available**: an integer containing the total number of data resource objects available in the database for the endpoint.
   - **last\_id**: a string containing the last ID returned.
   - **response\_message**: response string from the server.
+  - **request\_delay**: a non-negative float giving time in seconds a client should wait before issuing a subsequent request.
```
rartino, line 496: I don't like the lowercase should, as it makes it a bit ambiguous exactly how standards-breaking it is for a client to ignore this (it should be a pure suggestion, i.e., not standards-breaking at all.)
```suggestion
  - **request\_delay**: a non-negative float giving time in seconds that the client is suggested to wait before issuing a subsequent request.
```

optimade.rst
rartino, line 498: I think this belongs elsewhere - the SHOULD-level requirement that server return the Retry-After http header in a 429 response should apply regardless of whether the request_delay field is used or not.

rartino (commented):
Thanks for the work with this. Some suggestions to consider.

optimade.rst
merkys: This did not seem to work. I just removed the monospace styling for now.

merkys: Right, this has already been discussed, and we agreed it should be moved elsewhere.

merkys:
@ml-evs
> Do we prefer `request_delay` instead of just using the same nomenclature as `retry_after`? One minor formatting comment below.

I would indeed prefer sticking to the same nomenclature, but "retry" part does not sound right here. I would expect "retry" to mean "issue the exact same request". Maybe `next_request_after` or just `request_after` sounds better?

JPBergsma (dismissed):
This PR seems good to me.

optimade.rst
merkys: @rartino yes, I intend to do so, but it would help me a lot to know where exactly should I move this text.

rartino:
@ml-evs, @merkys  
> > Do we prefer `request_delay` instead of just using the same nomenclature as `retry_after`? One minor formatting comment below.
> 
> I would indeed prefer sticking to the same nomenclature, but "retry" part does not sound right here. I would expect "retry" to mean "issue the exact same request". Maybe `next_request_after` or just `request_after` sounds better?

As a remark: I first though the http header `retry_after` was not appropriate at all outside delivering an http errors. However, digging a bit in the RFCs made me realize that while the latest RFC allows it to be used also the way we say. The reason for the name is thus "historical debt".

Nevertheless, I agree with @merkys to rather just use a more helpful term.

optimade.rst
rartino: I think the only place that makes sense is https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst#http-response-status-codes

merkys: Ah, right, will move soon.

ml-evs (approved):
Fine by me!

optimade.rst
rartino: I'd say no, but there were others pushing for this info. I think the primary problem I have with just leaving what you have as it is, is that it adds a requirement on HTTP headers in a section that is, in a sense, not strinctly tied to HTTP as the transmission protocol. I'll try to suggest a requirement-free version.

optimade.rst
```diff
@@ -493,6 +493,9 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   - **data\_available**: an integer containing the total number of data resource objects available in the database for the endpoint.
   - **last\_id**: a string containing the last ID returned.
   - **response\_message**: response string from the server.
+  - **request\_delay**: a non-negative float giving time in seconds that the client is suggested to wait before issuing a subsequent request.
+    An implementation MAY respond with :http-error:`429 Too Many Requests` error if a client refuses to respect the indicated delay.
+    In that case the response SHOULD contain a `Retry-After HTTP header <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After>`__ to instruct the client to wait before retrying.
```
rartino, line 498:
```suggestion
  - **request\_delay**: a non-negative float giving time in seconds that the client is suggested to wait before issuing a subsequent request.
  
        Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
```

rartino: Note: I realized that you had a very non-standard link for the Retry-After header, so I've also changed that to link the apropriate section of the HTTP rfc.

optimade.rst
merkys: Thanks for the suggestion, I have accepted it. I like that HTTP 429 and `Retry-After` are now in an implementation note. This way OPTIMADE specification does not add specific interpretation on top of HTTP standards.

rartino (approved):
Good to go!

ml-evs (approved):
I think I slightly preferred the SHOULD/MAY formulation, but this still looks good to me.

PR #414, opened 2022-06-02
Means to specify database license

### LINKED ISSUE #102: Data(base) licenses

merkys:
Fixes #102. After discussions with @rartino in #102 I became convinced that for now a single license for all the data in a database should be sufficient. Should there be parts of database belonging under a different license, they should either be described in database-wide license document, or moved to a separate "sibling" database under a different license.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1022,9 +1022,9 @@ The single resource object's response dictionary MUST include the following fiel
   - **formats**: List of available output formats.
   - **entry\_types\_by\_format**: Available entry endpoints as a function of output formats.
   - **available\_endpoints**: List of available endpoints (i.e., the string to be appended to the versioned or unversioned base URL serving the API).
-  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a pointer to a license covering all the data and metadata provided under this database.
-    If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with :field-val:`https://spdx.org/licenses/`, but MUST NOT include the terminating :field-val:`.html`.
-    Example: :field-val:`https://spdx.org/licenses/CC0-1.0`.
+  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a URL to a web page containing a human-readable text describing the license (or licensing options if there are multiple) covering all the data and metadata provided by this database.
+    Clients are advised not to try automated parsing of this link or its content, but rather rely on the field :field:`available_licenses` instead.
+    Example: :field-val:`https://example.com/licenses/example_license.html`.
 
   :field:`attributes` MAY also include the following OPTIONAL fields:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1021,15 +1021,16 @@ The single resource object's response dictionary MUST include the following fiel
   - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a pointer to a license covering all the data and metadata provided under this database.
     If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with :field-val:`https://spdx.org/licenses/`, but MUST NOT include the terminating :field-val:`.html`.
     Example: :field-val:`https://spdx.org/licenses/CC0-1.0`.
-  - **compatible\_licenses**: List of SPDX license identifiers giving licenses the data and metadata in a database is compatible with.
-    In case the data and metadata is multiply-licensed, identifiers of these multiple licenses SHOULD be provided to let clients know under which conditions the data and metadata could be used.
-    If :field:`license` and :field:`compatible_licenses` contradict, :field:`compatible_licenses` has precedence.
 
   :field:`attributes` MAY also include the following OPTIONAL fields:
 
   - **is\_index**: if :field-val:`true`, this is an index meta-database base URL (see section `Index Meta-Database`_).
 
     If this member is *not* provided, the client MUST assume this is **not** an index meta-database base URL (i.e., the default is for :field:`is_index` to be :field-val:`false`).
+
+  - **compatible\_licenses**: List of SPDX license identifiers giving licenses the data and metadata in a database is compatible with.
+    In case the data and metadata is multiply-licensed, identifiers of these multiple licenses SHOULD be provided to let clients know under which conditions the data and metadata could be used.
+    If :field:`license` and :field:`compatible_licenses` contradict, :field:`compatible_licenses` has precedence.
 
 If this is an index meta-database base URL (see section `Index Meta-Database`_), then the response dictionary MUST also include the field:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1021,8 +1021,9 @@ The single resource object's response dictionary MUST include the following fiel
   - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a pointer to a license covering all the data and metadata provided under this database.
     If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with :field-val:`https://spdx.org/licenses/`, but MUST NOT include the terminating :field-val:`.html`.
     Example: :field-val:`https://spdx.org/licenses/CC0-1.0`.
-  - **license_is_compatible_with_cc_by_4_0**: Boolean value indicating compatibility with `Creative Commons Attribution 4.0 International license <https://spdx.org/licenses/CC-BY-4.0>`__.
-    This field may be used to check whether a particular database allows aggregation of its data and metadata.
+  - **compatible\_licenses**: List of SPDX license identifiers giving licenses the data and metadata in a database is compatible with.
+    In case the data and metadata is multiply-licensed, identifiers of these multiple licenses SHOULD be provided to let clients know under which conditions the data and metadata could be used.
+    If :field:`license` and :field:`compatible_licenses` contradict, :field:`compatible_licenses` has precedence.
 
   :field:`attributes` MAY also include the following OPTIONAL fields:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1018,11 +1018,11 @@ The single resource object's response dictionary MUST include the following fiel
   - **formats**: List of available output formats.
   - **entry\_types\_by\_format**: Available entry endpoints as a function of output formats.
   - **available\_endpoints**: List of available endpoints (i.e., the string to be appended to the versioned or unversioned base URL serving the API).
-  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a pointer to a license covering all the data provided under this database.
+  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a pointer to a license covering all the data and metadata provided under this database.
     If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with :field-val:`https://spdx.org/licenses/`, but MUST NOT include the terminating :field-val:`.html`.
     Example: :field-val:`https://spdx.org/licenses/CC0-1.0`.
   - **license_is_compatible_with_cc_by_4_0**: Boolean value indicating compatibility with `Creative Commons Attribution 4.0 International license <https://spdx.org/licenses/CC-BY-4.0>`__.
-    This field may be used to check whether a particular database allows aggregation of its data.
+    This field may be used to check whether a particular database allows aggregation of its data and metadata.
 
   :field:`attributes` MAY also include the following OPTIONAL fields:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -1019,6 +1019,9 @@ The single resource object's response dictionary MUST include the following fiel
   - **entry\_types\_by\_format**: Available entry endpoints as a function of output formats.
   - **available\_endpoints**: List of available endpoints (i.e., the string to be appended to the versioned or unversioned base URL serving the API).
 
+  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ providing a pointer to a license covering all the data provided under this database.
+    If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with `https://spdx.org/licenses/`.
+
   :field:`attributes` MAY also include the following OPTIONAL fields:
 
   - **is\_index**: if :field-val:`true`, this is an index meta-database base URL (see section `Index Meta-Database`_).
--- a/optimade.rst
+++ b/optimade.rst
@@ -1020,7 +1020,8 @@ The single resource object's response dictionary MUST include the following fiel
   - **available\_endpoints**: List of available endpoints (i.e., the string to be appended to the versioned or unversioned base URL serving the API).
 
   - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ providing a pointer to a license covering all the data provided under this database.
-    If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with `https://spdx.org/licenses/`.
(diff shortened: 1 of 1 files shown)


sauliusg (commented):
The stanza says "all data", but shouldn't we include also "all metadata"?
I was slightly inclined to have a possibility to specify license per record, but I agree that this wold be unnecessary complication for now. We can always upgrade if there is a genuine need.

I would approve the PR if/when the question about metadata is resolved.

merkys:
> The stanza says "all data", but shouldn't we include also "all metadata"? I was slightly inclined to have a possibility to specify license per record, but I agree that this wold be unnecessary complication for now. We can always upgrade if there is a genuine need.
> 
> I would approve the PR if/when the question about metadata is resolved.

@sauliusg I have added this in 3a678c3.

sauliusg (approved):

merkys:
This is an important change (please note `MUST` level of inclusion), thus I would like to make sure that providers-consortium members check it out and let us know should there be a problem with conforming to it.

merkys:
I have updated the PR to reflect workshop discussions: `license_is_compatible_with_cc_by_4_0` has been replaced by `compatible_licenses` list of SPDX identifiers. The new property is OPTIONAL.

blokhin:
For the case of the MPDS, we might have multiple licenses for the different sections of our data: CC BY 4.0, commercial/proprietary, per-vendor custom license, etc. We currently use per-entry custom field `_mpds_data_license`, which I would feel strong to recommend as a standard `data_license` / `entry_license` field.

merkys:
> For the case of the MPDS, we might have multiple licenses for the different sections of our data: CC BY 4.0, commercial/proprietary, per-vendor custom license, etc. We currently use per-entry custom field `_mpds_data_license`, which I would feel strong to recommend as a standard `data_license` / `entry_license` field.

This is a scenario quite closely fitting my reasoning which I have presented in #102 in my discussions with @rartino. If MPDS uses per-entry custom field, why not promote it to the standard? But this probably could be introduced in a follow-up PR in order not to block the current one.

According to this PR, MPDS licensing situation could be solved in two ways:
1. Separate databases (one per each license);
2. Specify all licenses and their governed domains in the top-level license file.

@blokhin Is any of these solutions suitable for MPDS? If so, maybe per-entry licensing could wait for the follow-up PR?

optimade.rst
```diff
@@ -1018,13 +1018,20 @@ The single resource object's response dictionary MUST include the following fiel
...
 
   :field:`attributes` MAY also include the following OPTIONAL fields:
 
   - **is\_index**: if :field-val:`true`, this is an index meta-database base URL (see section `Index Meta-Database`_).
 
     If this member is *not* provided, the client MUST assume this is **not** an index meta-database base URL (i.e., the default is for :field:`is_index` to be :field-val:`false`).
 
+  - **compatible\_licenses**: List of SPDX license identifiers giving licenses that the data and metadata in a database is compatible with.
+    In case the data and metadata is multiply-licensed, identifiers of these multiple licenses SHOULD be provided to let clients know under which conditions the data and metadata could be used.
+    If :field:`license` and :field:`compatible_licenses` contradict, :field:`compatible_licenses` has precedence.
```
rartino, line 1033: Based on our Workshop discussion on this and Wikipedia's "definition" of what license compatibility means, I don't think "compatible licenses" is the right term. For this to work, this field simply is a list of licenses that the database owner grant you the data under. So, how about naming it `available_licenses`? 

Also, if there ever was a time to be overly verbose for clarity in this text, this is it; so I suggest adding a bit of text to try to make the commitment of including something in this field more stringently clear.

```suggestion
  - **available\_licenses**: List of SPDX license identifiers specifying a set of alternative licenses under which the client is granted access to the data and metadata in this database.
    Inclusion of a license identifier in the list is a commitment of the database that the rights are in place to grant clients access to all the data and metadata according to the terms of these licenses.
    If the licensing information provided via the field :field:`license` omits licensing options specified in :field:`available_licenses`, or if it otherwise contradicts them, a client MUST still be allowed to interpret the inclusion of a license in :field:`available_licenses` as a full commitment from the database that the data and metadata is available, without exceptions, under the respective licenses.
    If the database cannot make that commitment, e.g., if only part of the data is available under a license, the corresponding license identifier MUST NOT appear in :field:`available_licenses` (but, rather, the field :field:`license` is to be used to clarify the licensing situation.)
    In case the data and metadata is multiply-licensed, identifiers of these multiple licenses SHOULD be provided to let clients know under which conditions the data and metadata can be used.
```

optimade.rst
```diff
@@ -1018,13 +1018,20 @@ The single resource object's response dictionary MUST include the following fiel
   - **formats**: List of available output formats.
   - **entry\_types\_by\_format**: Available entry endpoints as a function of output formats.
   - **available\_endpoints**: List of available endpoints (i.e., the string to be appended to the versioned or unversioned base URL serving the API).
+  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a pointer to a license covering all the data and metadata provided under this database.
+    If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with :field-val:`https://spdx.org/licenses/`, but MUST NOT include the terminating :field-val:`.html`.
+    Example: :field-val:`https://spdx.org/licenses/CC0-1.0`.
```
rartino, line 1023: It isn't fully clear to me how to evaluate the "MUST point to [the] SPD-hosted license fulltext". Is this saying that if I make my data available under CC-BY-4.0 I MUST put "https://spdx.org/licenses/CC-BY-4.0" in this field? That is to say, I MUST NOT link to my own website where I say, e.g., that ~ "all the structures are indeed fully available under CC-BY-4.0, but it would also be really nice if you guys could please also cite our paper." Because, in that case I don't really like this.

I get a feeling there is an attempt here to make this field 'interpretable' if the license is a standard license. I would rather suggest pointing to `compatible_licenses` (suggested: `available_licenses`) for all client needs of a machine-readable license field.

rartino (commented):
Some input on the overall design.

rartino:
@blokhin Even with your licensing situation, I hope that you do not have anything against merging this PR that allow linking a database-wide license text, where you can clarify the maybe not so simple licensing situation in MPDS? At the same time, perhaps we can continue the discussion of separate licenses for individual entries in #102? I'll copy your comment there to continue that discussion.

optimade.rst
```diff
@@ -1018,13 +1018,20 @@ The single resource object's response dictionary MUST include the following fiel
   - **formats**: List of available output formats.
   - **entry\_types\_by\_format**: Available entry endpoints as a function of output formats.
   - **available\_endpoints**: List of available endpoints (i.e., the string to be appended to the versioned or unversioned base URL serving the API).
+  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a pointer to a license covering all the data and metadata provided under this database.
+    If the license is included in the `SPDX License List <https://spdx.org/licenses/>`__, then the URL MUST point to SPDX-hosted license fulltext, i.e., it MUST start with :field-val:`https://spdx.org/licenses/`, but MUST NOT include the terminating :field-val:`.html`.
```
JPBergsma, line 1022: There may be more databases besides MPDS that use a per entry licencing system. So I think it would be good to specify that in that case, the link should point to a document describing how to find the licence for each entry. Otherwise someone may fill in a null value here, because there is no server wide licence.

rartino:
@merkys This seems a bit stalled? To move things forward, if you disagree with my suggestions, but don't have time to debate this, perhaps we can simplify the PR back to just a single "free-form" link without any machine-readable version and add the machine readable version in a separate PR?

merkys:
> @merkys This seems a bit stalled? To move things forward, if you disagree with my suggestions, but don't have time to debate this, perhaps we can simplify the PR back to just a single "free-form" link without any machine-readable version and add the machine readable version in a separate PR?

Yes, I wanted to give some more time for opinions like @blokhin's. But sure, if we can agree on a common subset of this PR, I am fine with stepping back. I would like to have licensing sorted out before the next release, one way or another.

blokhin:
I'm okay with the database-wide `available_licenses` / `license` fields. But what do you think about the optional per-entry license field?

optimade.rst
rartino: @merkys But, this semantic difference is exactly the point. To be meaningful, this field needs to mean that the copyright owner of the database is willing to allow you to use the data under any one of these licenses which is closer to ~= "licenses used by this database" than "licenses compatible with the usage of data in this database" with the accepted meaning of "license compatibility".

I think this is a point we probably should try to agree on before moving forward; because this should at least be understandable by us, if it is to be understandable by anyone else...

rartino:
@blokhin

My opinion on the per-entry field is that it is best kept as a database-specific field, because it should always have a database-specific meaning defined in the global license. As soon as we standardize a `license` field, rather than `mdps_license`, it means both servers and clients can try to give it meaning outside of the main license info. And IMO they shouldn't.

optimade.rst
```diff
@@ -1022,13 +1022,22 @@ The single resource object's response dictionary MUST include the following fiel
...
+    Example: :field-val:`https://spdx.org/licenses/CC0-1.0`.
```
rartino, line 1027:
```suggestion
  - **license**: A `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__ giving a URL to a web page containing a human-readable text describing the license (or licensing options if there are multiple) covering all the data and metadata provided in this database.
    Clients are advised not to try automated parsing of this link or its content, but rather rely on the field :field:`licenses` instead.
    Example: :field-val:`https://example.com/licenses/example_license.html`.
```

optimade.rst
```diff
@@ -1022,13 +1022,22 @@ The single resource object's response dictionary MUST include the following fiel
...
+    Inclusion of a license identifier in the list is a commitment of the database that the rights are in place to grant clients access to all the data and metadata according to the terms of either of these licenses (at the choice of the client).
+    If the licensing information provided via the field :field:`license` omits licensing options specified in :field:`available_licenses`, or if it otherwise contradicts them, a client MUST still be allowed to interpret the inclusion of a license in :field:`available_licenses` as a full commitment from the database that the data and metadata is available, without exceptions, under the respective licenses.
+    If the database cannot make that commitment, e.g., if only part of the data is available under a license, the corresponding license identifier MUST NOT appear in :field:`available_licenses` (but, rather, the field :field:`license` is to be used to clarify the licensing situation.)
+    In case the data and metadata is multiply-licensed, identifiers of these multiple licenses SHOULD be provided to let clients know under which conditions the data and metadata can be used.
+
```
rartino, line 1040:
```suggestion
  - **available\_licenses**: List of `SPDX license identifiers <https://spdx.org/licenses/>` specifying a set of alternative licenses under which the client is granted access to all the data and metadata in this database.
    if the data and metadata is available under multiple alternative licenses, identifiers of these multiple licenses SHOULD be provided to let clients know under which conditions the data and metadata can be used.
    Inclusion of a license identifier in the list is a commitment of the database that the rights are in place to grant clients access to all the data and metadata according to the terms of either of these licenses (at the choice of the client).
    If the licensing information provided via the field :field:`license` omits licensing options specified in :field:`available_licenses`, or if it otherwise contradicts them, a client MUST still be allowed to interpret the inclusion of a license in :field:`available_licenses` as a full commitment from the database that the data and metadata is available, without exceptions, under the respective licenses.
    If the database cannot make that commitment, e.g., if only part of the data is available under a license, the corresponding license identifier MUST NOT appear in :field:`available_licenses` (but, rather, the field :field:`license` is to be used to clarify the licensing situation.)
        An empty list indicates that none of the SPDX licenses apply for the entirety of the database and that the licensing situation is clarified in human readable form in the field :field:`license`.
```

rartino (changes requested):
The adjustments I think are needed to address my other point -- i.e., that the link really should be a free link with no requirement to be computer parsed, since `available_licenses` takes that role.

(I'm happy to hear contradicting opinions on this, though - I really don't mean to "steamroll" this PR, just make sure it is will work in a way that can be handled semi-automatically by my client implementation...)

optimade.rst
merkys: > It isn't fully clear to me how to evaluate the "MUST point to [the] SPD-hosted license fulltext". Is this saying that if I make my data available under CC-BY-4.0 I MUST put "https://spdx.org/licenses/CC-BY-4.0" in this field? That is to say, I MUST NOT link to my own website where I say, e.g., that ~ "all the structures are indeed fully available under CC-BY-4.0, but it would also be really nice if you guys could please also cite our paper." Because, in that case I don't really like this.

Yes, this was my intention here. I understand the need to specify other usage guidelines, but I see these as a separate thing, provided a standard license is used.

> I get a feeling there is an attempt here to make this field 'interpretable' if the license is a standard license. I would rather suggest pointing to `compatible_licenses` (suggested: `available_licenses`) for all client needs of a machine-readable license field.

Yes, having a machine-readable way to specify a license was my intention. But I agree your suggestion of better relying on `compatible_licenses`/`available_licenses` is more natural here.

optimade.rst
merkys: I would suggest resolving this by explicitly stating that there is no default value/interpretation of `null` for these licensing properties. In case no pointers are provided in `license` (or `*_licenses`), this means no rights are granted for the client (as is with the [source code](https://softwareengineering.stackexchange.com/questions/148146/open-source-code-with-no-license-can-i-fork-it)). I do not want to entangle too much logic around this PR - there is a field for provider to put a link to their license/policies, and it is up to a provider to describe their license/policies in it.

optimade.rst
merkys: On second thought, I guess the precise name of the property is not that much of an issue, provided the description is clear. Thus I accepted your subsequent change suggestion.

optimade.rst
merkys: As I accepted your change suggestions, I mark this as resolved now.

optimade.rst
rartino: Also, with the latest edits, the text for `license` says that the text it points to may express "(or licensing options if there are multiple)", which I think is the appropriate level of clarification about this.

rartino (approved):
I think it would be great if @sauliusg has a chance to take a look at the present version before we merge it, since I think he has a good handle on the formalities around licensing. I also recall @giovannipizzi having some thoughts/input on this on the workshop.

optimade.rst
merkys: But maybe we need to explicitly state what missing/`null` value for `license` means too?

rartino: Sure, it does not hurt to clarify for both the fields that null/missing means "no license". But, I suppose, in particular `available_licenses` present, and `license` missing should still be interpret able.

blokhin (approved):

rartino:
This has sitting approved for a week waiting for further input. In the interest of cleaning up our PRs, I'm going to merge it.

optimade.rst
```diff
@@ -1022,13 +1022,22 @@ The single resource object's response dictionary MUST include the following fiel
...
+    Inclusion of a license identifier in the list is a commitment of the database that the rights are in place to grant clients access to all the data and metadata according to the terms of either of these licenses (at the choice of the client).
+    If the licensing information provided via the field :field:`license` omits licensing options specified in :field:`available_licenses`, or if it otherwise contradicts them, a client MUST still be allowed to interpret the inclusion of a license in :field:`available_licenses` as a full commitment from the database that the data and metadata is available, without exceptions, under the respective licenses.
+    If the database cannot make that commitment, e.g., if only part of the data is available under a license, the corresponding license identifier MUST NOT appear in :field:`available_licenses` (but, rather, the field :field:`license` is to be used to clarify the licensing situation.)
+    An empty list indicates that none of the SPDX licenses apply for the entirety of the database and that the licensing situation is clarified in human readable form in the field :field:`license`.
```
ml-evs, line 1040: Without opening a new issue, I just wanted to clarify something here @merkys (and others). The field `license` is simply a Links object, so when you say "situtation is clarified in human readable form in the field `license`", are you referring to the content served at the URL pointed to be license, or the `meta` section of the Links object?

JPBergsma: If I remember the discussion correctly, we were indeed referring to the content of the URL to which the licence link points.

merkys: It is `license->href` indeed, worth clarifying.

PR #415, opened 2022-06-02
Support fuzzy comparison lists

### LINKED ISSUE #393: Support substring comparisons on list elements
While discussing #368 with @JPBergsma I noticed that comparisons on list properties having string elements can only use numeric comparison operators. This means, for example, that the following query is not allowed even on filter language grammar level:
```
species_at_sites HAS STARTS WITH "Si"
```
Since as of v1.1.0 there are not much of list properties with string elements, the issue is minor. However, such properties may arise in the future, thus I propose extending the grammar to allow such constructs.

Fix seems rather simple: replace
```
ValueList = [ Operator ], Value, { Comma, [ Operator ], Value } ;
```
with
```
ValueList = ( [ Operator ], Value | FuzzyStringOpRhs ), { Comma, ( [ Operator ], Value | FuzzyStringOpRhs ) } ;
```
(Similarly adjust `ValueZip` and `SetOpRhs` as well). I have not tried this for real, thus have not checked for grammar conflicts.

Pinging people usually involved in developing the filter language: @sauliusg, @rartino and @fekad.

merkys:
Fixes #393 by allowing `FuzzyStringOpRhs` in list comparisons and zip comparisons. All additions in this PR are OPTIONAL.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2580,7 +2580,7 @@ The Filter Language EBNF Grammar
                      | STARTS, [ WITH ], Value
                      | ENDS, [ WITH ], Value ;
 
-    SetOpRhs = HAS, ( ( Value | EqualityOperator, Value | RelativeComparisonOperator, OrderedValue ) | ALL, ValueList | ANY, ValueList | ONLY, ValueList ) ;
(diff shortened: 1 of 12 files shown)


rartino (approved):
Looks good to me and seems to work in grammatica, no reason not to merge this.

ml-evs (approved):
Also makes sense to me from a grammar standpoint, though glad we don't have an example in the spec to encourage this kind of query!

PR #417, opened 2022-06-09
Explain our "media_type" is the same as MIME type

JPBergsma:
I think MIME type is the more common term for the media type definition we give here.
Using "MIME type" instead of "media type" in a search, should make it easier for users/servers to find the correct media type for a particular file type.


merkys:
I personally see no need for a synonym (RFC reference is provided), but if it makes the specification clearer for a reader, then sure we may add it.

PR #424, opened 2022-09-29
Add `database` field to `meta` for describing the current database

### LINKED ISSUE #422: Add metadata field for describing the current database
Currently the only place where a *database* can be given a human-readable description is at the links endpoint of the relevant index meta-database. It might be an idea to add a field either to `/info` (my preference) or in the `meta` of every response where provider's can serve this description "in band".

ml-evs:
Closes #422.

This PR adds the optional `database` field to all `meta` responses. It can be used to provide the same data on an individual database as that found at the index meta-database, without requiring several requests and potential filtering of the index meta-database response.

----
Only other thought when writing this up is that this info could also be provided as a `self` entry in the `links` field for every response (though the fields would be limited to just using a `meta` field as part of a JSON API Links object).

If we prefer this route, then an alternative note could be added to the "JSON Response Schema: Common Fields" section of the spec that reserves the `self` keyword or something similar for this use, the only complication being that JSON:API `self` links should just return the exact URL of the current response.

----
Or one final option is just to add this kind of info at `/info/` directly, and exclude it from `meta`.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -643,6 +643,17 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
 
+  - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
+    If provided, the dictionary fields SHOULD match those provided in the corresponding the links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
+    The dictionary should contain the OPTIONAL fields:
+
+    - **name**: a human-readable name for the database, e.g., for use in clients.
+    - **description**: a human-readable description of the database, e.g., for use in clients.
+    - **homepage**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage for the particular database.
+    - **maintainer**: a dictionary providing details about the maintainer of the database, which MUST contain the single field:
+
+      - **email** with the maintainer's email address.
+
   - **implementation**: a dictionary describing the server implementation, containing the OPTIONAL fields:
 
     - **name**: name of the implementation.
@@ -709,6 +720,14 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
                "email": "admin@example.com"
              },
              "issue_tracker": "http://tracker.example.com/exmpl-optimade"
+           },
+           "database": {
+             "name": "Example database 1 (of many)",
+             "description": "The first example database in a series hosted by the Example Provider.",
+             "homepage": "http://database_one.example.com",
+             "maintainer": {
+               "email": "science_lead@example.com"
+             }
            }
          }
          // ...
--- a/optimade.rst
+++ b/optimade.rst
@@ -647,6 +647,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
     If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
     The dictionary can contain the OPTIONAL fields:
 
+    - **id**: the identifier of this database within those served by this provider, i.e., the ID under which this database is served in this provider's index meta-database.
     - **name**: a human-readable name for the database, e.g., for use in clients.
     - **version**: a string describing the version of the database.
     - **description**: a human-readable description of the database, e.g., for use in clients.
--- a/optimade.rst
+++ b/optimade.rst
@@ -644,8 +644,8 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
 
   - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
-    If provided, the dictionary fields SHOULD match those provided in the corresponding the links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
-    The dictionary should contain the OPTIONAL fields:
+    If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
+    The dictionary can contain the OPTIONAL fields:
 
     - **name**: a human-readable name for the database, e.g., for use in clients.
     - **description**: a human-readable description of the database, e.g., for use in clients.
--- a/optimade.rst
+++ b/optimade.rst
@@ -723,6 +723,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
              "issue_tracker": "http://tracker.example.com/exmpl-optimade"
            },
            "database": {
+               "id": "example_db",
              "name": "Example database 1 (of many)",
              "description": "The first example database in a series hosted by the Example Provider.",
              "homepage": "http://database_one.example.com",
--- a/optimade.rst
+++ b/optimade.rst
@@ -651,7 +651,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
     - **name**: a human-readable name for the database, e.g., for use in clients.
     - **version**: a string describing the version of the database.
     - **description**: a human-readable description of the database, e.g., for use in clients.
-    - **homepage**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage for the particular database.
+    - **homepage**: a `JSON API link <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage for the particular database.
     - **maintainer**: a dictionary providing details about the maintainer of the database, which MUST contain the single field:
 
       - **email** with the maintainer's email address.
--- a/optimade.rst
+++ b/optimade.rst
@@ -648,6 +648,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
     The dictionary can contain the OPTIONAL fields:
 
     - **name**: a human-readable name for the database, e.g., for use in clients.
+    - **version**: a string describing the version of the database.
     - **description**: a human-readable description of the database, e.g., for use in clients.
     - **homepage**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage for the particular database.
     - **maintainer**: a dictionary providing details about the maintainer of the database, which MUST contain the single field:
--- a/optimade.rst
+++ b/optimade.rst
@@ -724,7 +724,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
              "issue_tracker": "http://tracker.example.com/exmpl-optimade"
            },
            "database": {
-               "id": "example_db",
+             "id": "example_db",
              "name": "Example database 1 (of many)",
              "description": "The first example database in a series hosted by the Example Provider.",
              "homepage": "http://database_one.example.com",


optimade.rst
```diff
@@ -537,7 +548,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   - Example for a request made to :query-url:`http://example.com/optimade/v1/structures/?filter=a=1 AND b=2`:
 
     .. code:: jsonc
-
+
```
ml-evs, line 551:
```suggestion
```

optimade.rst
```diff
@@ -497,6 +497,17 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
 
+  - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
+    If provided, the dictionary fields SHOULD match those provided in the corresponding the links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
```
JPBergsma, line 501:
```suggestion
    If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
```
I think this the is not necessary.

optimade.rst
```diff
@@ -497,6 +497,17 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
...
+    The dictionary should contain the OPTIONAL fields:
```
JPBergsma, line 502: To me, it sounded a bit conflicting that you use both should and OPTIONAL here. 
If all the properties of the database dictionary are optional "can" would perhaps be more fitting.

JPBergsma (commented):
Hi ml-evs,

I think it is good that we finally have a field to describe the database.
I still have a few small remarks but overall it looks good.

optimade.rst
```diff
@@ -497,6 +497,17 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
...
   Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
 
+  - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
+    If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
+    The dictionary can contain the OPTIONAL fields:
+
+    - **name**: a human-readable name for the database, e.g., for use in clients.
+    - **description**: a human-readable description of the database, e.g., for use in clients.
+    - **homepage**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage for the particular database.
+    - **maintainer**: a dictionary providing details about the maintainer of the database, which MUST contain the single field:
```
JPBergsma, line 507: Sorry I did not spot this the first time, but what is the meaning of the word "single" in this line.
"- **maintainer**: a dictionary providing details about the maintainer of the database, which MUST contain the single field: "
Does this imply that the database can not add extra database specific fields here? If so, it does not seem necessary to have a separate dictionary for the maintainer. If it does not, I think it would be better to remove the word single here.

JPBergsma (commented):
On line 655 there is still an example of a response. I think your database field should be added to this example.

ml-evs:
> On line 655 there is still an example of a response. I think your database field should be added to this example.

Not sure it's super important, as `database` is an optional field (and other optional fields are also omitted in that example, like `implementation`).

optimade.rst
ml-evs: I used same wording as `implementation->maintainer` but agree it is a bit clunky.

JPBergsma:
Yes you are right. As it is an optional field it is not so bad that it is not in the example on line 655.

optimade.rst
JPBergsma: I think I would prefer if it is removed there as well.
@rartino do you know why you originally chose this wording?

rartino: Maybe bad form from my part to dig through the git blame log to reflect blame :-), but this phrasing seems to originate from @merkys ( https://github.com/Materials-Consortia/OPTIMADE/blame/ddd33aca159c22c7c8760bbd3aa2381164056d61/optimade.md#L250 ). I think it was only meant as a helper word to clarify that the reader should only expect a single field below. I do not see the harm in it, but also do not object to it being removed.

JPBergsma: I did a git blame from my terminal and there it listed you as the author. But perhaps you merely moved some lines around. 
You definitely looked at an older version of the text than me.

Perhaps I am also adding too much importance to it.  "single" just seemed an unnecessary word in this sentence to me.

rartino: About the formulation: even with my most strict reading, I just see it saying "this one specific single field is required" with no implications for other, e.g., data-base specific, fields 

About git blame: you have to dig backwards in the git blame log (on GitHub, use the "stack of papers" icon next to the line number) to find who originally made the contribution. Otherwise I'm going to be blamed for almost everything due to the PR where I reformatted the specification from .md to .rst, and a few others where I moved whole sections.

merkys: This indeed might be me to blame :) I added this construction with the aim to forbid empty `maintainer` dictionaries. At the time of writing there could be only one field in `maintainer`, thus I used this wording. Also at that time I was not aware that prefixed database-specific fields were allowed in every dictionary (or did we change that a bit later?).

merkys:
Suggestion in #422 describes two places where this info can be added: `/info` endpoint (preferred by @ml-evs) or `meta` of every response. I would also prefer `/info` to retain smaller responses, but maybe there are more arguments backing `meta` proposal?

optimade.rst
```diff
@@ -497,6 +497,17 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
 
+  - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
+    If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
+    The dictionary can contain the OPTIONAL fields:
+
+    - **name**: a human-readable name for the database, e.g., for use in clients.
+    - **description**: a human-readable description of the database, e.g., for use in clients.
```
rartino, line 505:
```suggestion
    - **name**: a human-readable name for the database, e.g., for use in clients.
    - **version**: a string describing the version of the database.
    - **description**: a human-readable description of the database, e.g., for use in clients.
```

rartino (changes requested):
For parity with implementation, I think we should also allow to specify a version for the database from which the response comes.

I agree that it make sense to (also?) allow database information in /info, but I can see reasons why people may want it in every response from their database. It isn't many bytes.

ml-evs:
> For parity with implementation, I think we should also allow to specify a version for the database from which the response comes.

Fine by me.

> I agree that it make sense to (also?) allow database information in /info, but I can see reasons why people may want it in every response from their database. It isn't many bytes.

I guess as it stands these fields could also be provided in the `meta` response from `/info`. If we are happy having them in `meta` then we should not duplicate it in `/info` too. I wouldn't be against moving it to just `/info` but perhaps that is a v2.0 change.

We could also consider adding a query parameter that elides the `meta` response, which could, for example, be included in pagination links from an implementation by default. `?meta=false` or `?hide_meta` etc. We have added a few unbounded fields to `meta` (e.g., `description` could be the full text of a paper describing the provider/database) so perhaps this would be helpful as an optional parameter.

I'll re-request review from everyone now to see if anyone has remaining objections to adding this to `meta`.

rartino (approved):
I'm approving this - but I think the discussion shows that we have no policy with regards to what should go in the meta field of any response, and what belongs in `/info`. The lack of such policy is a problem, because it is going to make us - and our users - unsure where to find stuff.

optimade.rst
```diff
@@ -497,6 +497,18 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
 
+  - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
```
JPBergsma, line 500: Could you perhaps use in the word OPTIONAL in the description of the database field. That would be more consistent with how we describe optional fields elsewhere in the database.
```suggestion
  - **database**: an OPTIONAL dictionary describing the specific database accessible at this OPTIMADE API.
```

rartino: Isn't this placed under "meta MAY also include these fields"?

JPBergsma: Yes you are right, I had not scrolled up far enough.

JPBergsma (dismissed):

optimade.rst
```diff
@@ -519,6 +519,18 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
...
+    If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
+    The dictionary can contain the OPTIONAL fields:
+
+    - **name**: a human-readable name for the database, e.g., for use in clients.
```
ml-evs, line 526:
```suggestion
    - **id**: the identifier of this database within those served by this provider, i.e., the ID under which this database is served in this provider's index meta-database.
    - **name**: a human-readable name for the database, e.g., for use in clients.
```

optimade.rst
```diff
@@ -585,6 +597,14 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
                "email": "admin@example.com"
              },
              "issue_tracker": "http://tracker.example.com/exmpl-optimade"
+           },
+           "database": {
```
ml-evs, line 601:
```suggestion
           "database": {
               "id": "exmpl",
```

ml-evs (commented):
Reviewing my own PR here... I think we should also provide the database `id` in the meta of every response. This is another piece of data that is no provided anywhere but in the index meta-database. This can be used to construct globally unique IDs for entries within this database via `{provider_prefix}/{database_id}/{entry_id (or immutable_id}` (the provider prefix still needs a call to `/info` but at least its accessible from the same base URL).

One missing field other includes link aggregation, but that can probably be omitted here.

optimade.rst
JPBergsma: Would it perhaps be better to use a string other than "exmpl", as we also use that for the provider prefix?

ml-evs: Fair, just changed it to "example"

JPBergsma: Perhaps "example_db" would be even clearer, although I guess I should have suggested that in my first comment.

merkys:
Regarding the placement, I noticed that `"provider"` is allowed to appear in `"meta"` of any response, thus it makes sense to me for `"database"` to follow suit (as is suggested here). Another point in favor of current proposal is provenance preservation - in case a database does not provide immutable IDs or URLs, this could be extracted from the response as a last-resort means. Therefore I am leaning to accepting (after merging `develop` in).

merkys (approved):

JPBergsma (approved):
I do still have a preference for moving the description field to the info endpoint, as there is no reason to share it in every response. 
As we have however already placed the description for the provider under meta in every response, it seems logical to do the same for the description field of the database.
So for now I will approve this change. But I do think that we should move it for the 2.0 release.

ml-evs:
If we are happy with the field names etc here then I think @rartino this is ready for a final review+merge?

optimade.rst
```diff
@@ -585,6 +598,15 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
...
+               "id": "example_db",
```
vaitkus, line 603:
```suggestion
             "id": "example_db",
```

merkys (approved):

optimade.rst
```diff
@@ -519,6 +519,19 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
...
 
+  - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
+    If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
+    The dictionary can contain the OPTIONAL fields:
+
+    - **id**: the identifier of this database within those served by this provider, i.e., the ID under which this database is served in this provider's index meta-database.
+    - **name**: a human-readable name for the database, e.g., for use in clients.
+    - **version**: a string describing the version of the database.
+    - **description**: a human-readable description of the database, e.g., for use in clients.
+    - **homepage**: a `JSON API links object <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage for the particular database.
```
vaitkus, line 530: The `homepage` should probably also be allowed to be as string, similarly to the `homepage` field under the `provider`. Note, that the currently provided example already uses it as a string field.

rartino: Oh, wait - does it make sense for these things (here and in multiple other places in the specification) to be "links objects", which allow multiple links with different keys? Or, did we actually mean to say "a link as defined in the JSON API specification under links objects http://jsonapi.org/format/1.0/#document-links ?

ml-evs: Yes @rartino, I think you are right... Maybe I can suggest this in a separate PR

ml-evs: https://github.com/Materials-Consortia/OPTIMADE/pull/474#issuecomment-1591806840

rartino: You're right, that first "(a “links object”)" parenthesis makes no sense to me either. Everything else makes sense as:

- "links object": an object where each member is a "link".
-  "link": a string or "link object".
- "link object": an object with a href member; in v1.1 it has its own subsection: https://jsonapi.org/format/1.1/#auto-id--link-objects

So, when we are trying to represent single links in our API, I'd say our options in increasing "JSON API-iness" (and perhaps decreasing OPTIMADE-iness") are:

1. "A string" equal to the URL.
2. "A link as described in http://jsonapi.org/format/1.1/#document-links" (i.e., string or "link object").
3. "A link object" as described in https://jsonapi.org/format/1.1/#auto-id--link-objects (must be an object with a href memeber, not allowed to be a string).
4. Move the link over to the closest JSON API-mandated links object, and define it there with a key that makes sense. For this PR it would be the topmost link under a key named "database").

Not sure what to do about this. Given how non-JSON-API-formatted this response already is, "1" seems attractive...

rartino: Seeing you https://github.com/Materials-Consortia/OPTIMADE/pull/474, I suspect you mean to do this before merging then?
```suggestion
    - **homepage**: a `JSON API link <http://jsonapi.org/format/1.0/#document-links>`__, pointing to a homepage for the particular database.
```
(Or do you want to merge this with the bug first, and then correct it in #474?)

ml-evs: Sorry, this one got lost in the notifications... I've committed your suggestion, I think we're at least self-consistent now...

ml-evs: Perhaps @rartino can re-review and I'll just merge and stop bothering everyone else? (unless they also review before I get the chance to merge...)

rartino (approved):

optimade.rst
```diff
@@ -643,6 +643,19 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
 
+  - **database**: a dictionary describing the specific database accessible at this OPTIMADE API.
+    If provided, the dictionary fields SHOULD match those provided in the corresponding links entry for the database in the provider's index meta-database, outlined in `Links Endpoint JSON Response Schema`_.
+    The dictionary can contain the OPTIONAL fields:
```
vaitkus, line 648:
```suggestion
    The dictionary can contain the following OPTIONAL fields:
```

vaitkus (approved):

optimade.rst
ml-evs: I think to avoid another review round-trip I'll merge this as is; this can be added to a pre-release "tweaks" PR though.

vaitkus: Sure thing.

PR #436, opened 2022-12-07
SMILES data type

### LINKED ISSUE #368: Add property for SMILES that allows '@','%','/','\','#' and '$'
Do we want to allow the use of smiles string in the field chemical_formula_descriptive ?
The SMILES notation for molecular formulas uses '#' and '$' to indicate triple and quadruple bonds,
the characters '/' and '\' to indicate whether the bonds are in the cis or trans orientation and '@' and '@@' to differentiate enantiomers. Finally, ring numbers with more than one digit have to be preceded by a '%' sign.  
It, therefore, seems reasonable to me to add these to the allowed characters for the  chemical_formula_descriptive field.

Or do you think we should add a separate SMILES field instead?

merkys:
In #368 SMILES property for structures was proposed. PR #392 was proposed to introduce string-valued SMILES property, however, there were opinions that internal semantics of SMILES strings have to be respected, what does not fit nicely with "plain string" data type. Thus the current PR introduces both `smiles` data type and an associated property.

Fixes #368.


optimade.rst
```diff
@@ -1647,6 +1648,21 @@ Examples:
 - :filter:`property != FALSE`
 - :filter:`_exmpl_has_inversion_symmetry AND NOT _exmpl_is_primitive`
 
+Comparisons of SMILES values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Equality comparisons ('=' and '!=') MUST be supported for SMILES values.
+When handling equality comparisons of SMILES values, an implementation SHOULD NOT regard them as simple strings.
+Instead, an implementation SHOULD either compare the described chemical structures or canonicalize SMILES representations and then perform direct string matching.
+In addition to equality comparison operators, :val:`CONTAINS` MAY be supported optionally as an operator to check whether one structure is a substructure of another.
```
JPBergsma, line 1657: To me, this line is not entirely clear. Does this mean that we support querying for chemical groups ? As would be defined with SMARTS query language ? In that case we should probably mention the SMARTS query language. 
Or do we only support searching for whole molecules in the SMILES string which could be separated by a "."?

optimade.rst
```diff
@@ -2448,6 +2466,24 @@ chemical\_formula\_anonymous
...
+
+- **Description**: The SMILES (Simplified Molecular Input Line Entry System) representation of the structure.
+- **Type**: smiles
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+  - Value MUST adhere to the `OpenSMILES specification v1.0 <http://opensmiles.org/opensmiles.html>`__.
+  - When structures or their parts cannot be unambiguously represented in SMILES according to OpenSMILES recommendations, using the guidelines from `Quirós et al. 2018 <https://doi.org/10.1186/s13321-018-0279-6>`__ is RECOMMENDED.
+  - Providers MAY canonicalize (i.e., use rules to establish stable order of atoms) produced SMILES representations, but this is not mandatory.
```
JPBergsma, line 2480: I am not sure whether this line really adds something to the specification, for me, it could be removed.

optimade.rst
```diff
@@ -1773,11 +1789,13 @@ Type handling and conversions in comparisons
 
 The definitions of specific properties in this standard define their types.
 Similarly, for `database-provider-specific properties`_, the database provider decides their types.
-In the syntactic constructs that can accommodate values of more than one type, types of all participating values are REQUIRED to match, with a single exception of timestamps (see below).
+In the syntactic constructs that can accommodate values of more than one type, types of all participating values are REQUIRED to match, with the exception of timestamps and SMILES representations (see below).
 Different types of values MUST be reported as :http-error:`501 Not Implemented` errors, meaning that type conversion is not implemented in the specification.
 
-As the filter language syntax does not define a lexical token for timestamps, values of this type are expressed using string tokens in `RFC 3339 Internet Date/Time Format <https://tools.ietf.org/html/rfc3339#section-5.6>`__.
+As the filter language syntax does not define lexical tokens for timestamps and SMILES, values of these types are expressed using string tokens.
```
vaitkus, line 1795:
```suggestion
As the filter language syntax does not define lexical tokens for timestamps or SMILES, values of these types are expressed using string tokens.
```

optimade.rst
vaitkus, line 2480: Canonicalization of SMILES is quite an ambiguous term so I would not equate it to `stable order of atoms` (think of aromaticity representation, resonance structures, etc.). It may be more clear to explicitly specify if changes in bond order, bond type and charges are allowed/disallowed.

merkys: I believe that this line answers the immediate question for those who are aware of canonical vs. any SMILES representations.

merkys: I added the parentheses here to explain what "canonicalization" means for people who are not accustomed with SMILES and/or their canonicalization. Maybe there is a better way to explain that?

optimade.rst
```diff
@@ -1555,6 +1556,7 @@ More examples of the number tokens and machine-readable definitions and tests ca
 
 - **Operator tokens** are represented by usual mathematical relation symbols or by case-sensitive keywords.
   Currently the following operators are supported: :filter-op:`=`, :filter-op:`!=`, :filter-op:`<=`, :filter-op:`>=`, :filter-op:`<`, :filter-op:`>` for tests of number, string (lexicographical) or timestamp (temporal) equality, inequality, less-than, more-than, less, and more relations; :filter-op:`AND`, :filter-op:`OR`, :filter-op:`NOT` for logical conjunctions, and a number of keyword operators discussed in the next section.
+  Of these, SMILES data type supports only equality and inequality comparison operators (:filter-op:`=` and :filter-op:`!=`).
```
JPBergsma, line 1559:
```suggestion
  Currently the following operators are supported: :filter-op:`=`, :filter-op:`!=`, :filter-op:`<=`, :filter-op:`>=`, :filter-op:`<`, :filter-op:`>` for tests of number, string (lexicographical) or timestamp (temporal) equality, inequality, less-than, more-than, less, and more relations; :filter-op:`AND`, :filter-op:`OR`, :filter-op:`NOT` for logical conjunctions, and a number of keyword operators discussed in the next section.
  Of the equality and inequality comparison operators, SMILES data type supports only (:filter-op:`=` and :filter-op:`!=`).
```
The way this sentence was described implied, for me, that the logical conjunction operators were also not supported, so I think this is clearer.

merkys: I believe "equality" stands for `=` and "inequality" for `!=`, thus I think your suggestion implies slightly different terminology. What about just removing "Of these"?

JPBergsma: I think that would still give the suggestion that `AND`, `OR` and  `NOT` operators would not be supported for the SMILES data type. I therefore think that the first suggestion below would be a bit clearer. As an alternative, we could also move the text in line 1559 up, so it is clear it only refers to the comparison operators.
```suggestion
  Currently the following operators are supported: :filter-op:`=`, :filter-op:`!=`, :filter-op:`<=`, :filter-op:`>=`, :filter-op:`<`, :filter-op:`>` for tests of number, string (lexicographical) or timestamp (temporal) equality, inequality, less-than, more-than, less, and more relations; :filter-op:`AND`, :filter-op:`OR`, :filter-op:`NOT` for logical conjunctions, and a number of keyword operators discussed in the next section.
  Of the comparison operators, SMILES data type supports only equality and inequality operators (:filter-op:`=` and :filter-op:`!=`).
```
```suggestion
  Currently the following operators are supported:

  - :filter-op:`=`, :filter-op:`!=`, :filter-op:`<=`, :filter-op:`>=`, :filter-op:`<`, :filter-op:`>` for tests of number, string (lexicographical) or timestamp (temporal) equality, inequality, less-than, more-than, less, and more relations. 
  Of these, the SMILES data type supports only equality and inequality comparison operators (:filter-op:`=` and :filter-op:`!=`).
  - :filter-op:`AND`, :filter-op:`OR`, :filter-op:`NOT` for logical conjunctions
  - a number of keyword operators discussed in the next section. 
```

optimade.rst
vaitkus: I think I may be with @JPBergsma on this one. I am unsure if this statement adds any actionable information.

Generally, SMILES canonicalization should not change the overall interpretation of the structure. As is mentioned in the sentence, atoms may get reordered, the notation of aromatic bond may change (e.g. `c1:c:c:c:c:c1` to `c1ccccc1`), etc. It is equivalent to having two timestamps or numbers expressed in a different syntactically correct way (e.g. `0.01` and `1E-2`). Since such alternatives representations are implicitly allowed for these data types, the same could also be done with `smiles`?

Another problem with SMILES canonicalization is that there are numerous canonicalization algorithms and without knowing the specifics, the canonical representation is not very much more useful to the user than a non-canonical one. One nice property of canonical SMILES is that they always represent the same compound in the same way, so the user could apply some additional filtering  without the need to canonicalize the SMILES themselves. However, since SMILES canonicalization is not required (`MAY canonicalize`) and there are no flags to automatically indicate if the SMILES are canonical or not, the users must still recanonicalize the SMILES (especially in federated searches).

@merkys , what are your intended use cases allowed by this statement?

vaitkus: It might be better to reference a publication or something similar that describes SMILES canonicalization in more detail instead of the parentheses, but at this point I would be more for the removal of the sentence.

vaitkus:
The introduction of the SMILES data type will most likely also require some changes similar to those introduced in PR #444 for the timestamp data type.

merkys:
> The introduction of the SMILES data type will most likely also require some changes similar to those introduced in PR #444 for the timestamp data type.

Fair point. However, I am not sure JSON Schema supports custom formats, I need to have a better look at it.

ml-evs:
> > The introduction of the SMILES data type will most likely also require some changes similar to those introduced in PR #444 for the timestamp data type.
> 
> Fair point. However, I am not sure JSON Schema supports custom formats, I need to have a better look at it.

It does, see: https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-01#section-7.2

Lots of impenetrable waffle in the spec about it though -- not clear to me how we would announce what `"format": "smiles"` actually means within JSON Schema yet. We could at least include a regex for structural validation.

rartino:
> > > The introduction of the SMILES data type will most likely also require some changes similar to those introduced in PR #444 for the timestamp data type.
> > 
> > Fair point. However, I am not sure JSON Schema supports custom formats, I need to have a better look at it.
> 
> It does, see: https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-01#section-7.2
> 
> Lots of impenetrable waffle in the spec about it though -- not clear to me how we would announce what `"format": "smiles"` actually means within JSON Schema yet. We could at least include a regex for structural validation.

Reading your link:

---------------------

> Implementations MAY support custom format attributes. Save for agreement between parties, schema authors SHALL NOT expect a peer implementation to support such custom format attributes. An implementation MUST NOT fail to collect unknown formats as annotations. When the Format-Assertion vocabulary is specified, implementations MUST fail upon encountering unknown formats. Vocabularies do not support specifically declaring different value sets for keywords. Due to this limitation, and the historically uneven implementation of this keyword, it is RECOMMENDED to define additional keywords in a custom vocabulary rather than additional format attributes if interoperability is desired.

--------------------

So, if I read this correctly, they are saying: "But, don't use this, use your own field instead", e.g., `x-optimade-type`?

The third (IMO very inelegant) option is to specify the format as "regex" + the best JSON Schema compatible regex we can come up with for SMILES strings. Implementations would then have to recognize specifically that regex and reverse-map it into the knowledge that the field is of OPTIMADE SMILES-type. However, in that case I think `x-optimade-type` is about 1e10 times better as a solution.

A happy side note: it appears the JSON Schema regex situation is more clear now than when we looked into it last time. They now define a "least common denominator subset" regex standard. 

  https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-regular-expressions

To me, this seems to finally provide a solution to the regular expression dilemma discussions (#42 #160). We should now simply follow their lead and standardize this both for our filter language and for format: regex in property definitions (which presently are not allowed at all). In contrast to JSON Schema, our property definitions should say that the regex MUST only use the subset.

optimade.rst
merkys: OK, I am convinced this does not add much of a benefit. Removed.

merkys:
> So, if I read this correctly, they are saying: "But, don't use this, use your own field instead", e.g., `x-optimade-type`?
> 
> The third (IMO very inelegant) option is to specify the format as "regex" + the best JSON Schema compatible regex we can come up with for SMILES strings. Implementations would then have to recognize specifically that regex and reverse-map it into the knowledge that the field is of OPTIMADE SMILES-type. However, in that case I think `x-optimade-type` is about 1e10 times better as a solution.

`x-optimade-type` sounds reasonable to me. Should I open a separate PR to add `x-optimade-type` to the specification or should I lump it together with this one?

rartino:
> x-optimade-type sounds reasonable to me. Should I open a separate PR to add x-optimade-type to the specification or should I lump it together with this one?

Lets take it in the open PR on property definitions; I have another thing that should be adjusted in the specification there as well, moving the identifier to "$ID". If you want to hurry things along, feel free to do a PR against my branch for that PR, or don't and I'll add `x-optimade-type` as I add '$ID'.

merkys:
> Lets take it in the open PR on property definitions; I have another thing that should be adjusted in the specification there as well, moving the identifier to "$ID". If you want to hurry things along, feel free to do a PR against my branch for that PR, or don't and I'll add `x-optimade-type` as I add '$ID'.

I am OK to wait. Thanks.

optimade.rst
```diff
@@ -1554,7 +1555,7 @@ More examples of the number tokens and machine-readable definitions and tests ca
 - **Boolean values** are represented with the tokens :filter-op:`TRUE` and :filter-op:`FALSE`.
 
 - **Operator tokens** are represented by usual mathematical relation symbols or by case-sensitive keywords.
-  Currently the following operators are supported: :filter-op:`=`, :filter-op:`!=`, :filter-op:`<=`, :filter-op:`>=`, :filter-op:`<`, :filter-op:`>` for tests of number, string (lexicographical) or timestamp (temporal) equality, inequality, less-than, more-than, less, and more relations; :filter-op:`AND`, :filter-op:`OR`, :filter-op:`NOT` for logical conjunctions, and a number of keyword operators discussed in the next section.
+  Currently the following operators are supported: :filter-op:`=`, :filter-op:`!=`, :filter-op:`<=`, :filter-op:`>=`, :filter-op:`<`, :filter-op:`>` for tests of number, string (lexicographical), timestamp (temporal) or SMILES representation (structural) equality, inequality, less-than, more-than, less, and more relations; :filter-op:`AND`, :filter-op:`OR`, :filter-op:`NOT` for logical conjunctions, and a number of keyword operators discussed in the next section.
```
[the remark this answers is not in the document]
merkys: @vaitkus can we mark this as resolved?

optimade.rst
merkys: I have removed the sentence on canonicalization. Can we mark as resolved?

merkys:
With the acceptance of provider-specific data types (#529) and opening of cheminformatics namespace this PR is no longer relevant in its current form. Work on introduction of SMILES is now being done in https://github.com/Materials-Consortia/namespace-cheminformatics/pull/4. Therefore I will close the current PR.

Issue #429, opened 2022-12-05
OPTIMADE v1.2 release planning


ml-evs:
Following from @merkys's comment on the mailing list, we should begin preparations for a v1.2 release.

I think I also foolishly "volunteered" to manage this release a while ago, but please help out if you want to. As part of this process, I've been doing a bit of issue triaging -- please re-open any issues that I closed that you think are still valid. In particular, I've also been clearing up the v1.2 milestone https://github.com/Materials-Consortia/OPTIMADE/milestone/3 which should now reflect what I write below.

We can use the comments here to have some async discussion that can perhaps be finalized at the final meeting this year.

## Summary

As far as I see it, we need to decide:

1. Whether to release v1.2 before the paper
2. Whether to release and wait for providers to implement v1.2 before the paper
3. Whether to write or merge any more PRs before releasing v1.2

My suggestion would be to finalize any PRs/issues we know we want in before the next meeting, then make a release candidate before Christmas. We can write the paper based on that release candidate and hopefully servers will start picking up the changes too (we can certainly get most of it into optimade-python-tools in January). We can submit the paper whenever with references to the release candidate, ready to be updated with the real release version in time for its (hopeful) publication.
 
## Proposed changelog (updated 22/06)

I've just made a draft v1.2.0 release on GitHub (that everyone in the organization should be able to see) --- we could also convert this into a proper release candidate sooner rather than later v1.2.0-rc1, if desired. This draft has the following abridged (removed typo/testing PRs) changelog:

* Allow provider-specific nested properties within OPTIMADE properties. by @JPBergsma in https://github.com/Materials-Consortia/OPTIMADE/pull/401
* Space group additions: `space_group_hall` and `space_group_it_number` by @merkys in https://github.com/Materials-Consortia/OPTIMADE/pull/405
* Allow boolean values in queries by @merkys in https://github.com/Materials-Consortia/OPTIMADE/pull/348
* Files endpoint by @merkys in https://github.com/Materials-Consortia/OPTIMADE/pull/360
* Property definitions by @rartino in https://github.com/Materials-Consortia/OPTIMADE/pull/376 (plus several associated tweaks and fixes)
* Support fuzzy comparison lists by @merkys in https://github.com/Materials-Consortia/OPTIMADE/pull/415
* Backoff time by @merkys in https://github.com/Materials-Consortia/OPTIMADE/pull/411
[...]

PR #438, opened 2022-12-19
Clarify grouping of relationships and what key to use in JSON:API output format

### LINKED ISSUE #419: Specification unclear on the key to use in JSON:API relationships
As we've found in trying to resolve #386, the specification is presently unclear on what *key* one should use for the various JSON:API relationships one include in the `relationship` dictionary as part of a JSON:API-formatted response.

Given that we do not say anything about it, one may be led to believe that any key is permissible. However, the output-format-agnostic description of relationships in [3.6 Relationships](https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst#36relationships) makes no reference to a free-form "key" or "name" tied to an OPTIMADE relationship, so this is then some data that would only appear in a JSON:API response.

From the discussion in #386 it seems we have (at least) two options:

1. Mandate that the key must be the entry type (e.g., "structures"), and thus that there is no contextual "grouping" of related entries, but rather, all related entries of the same type are grouped, and cannot be grouped with entries of a different type. This can be done by: Edit [5.3.2 Entry Listing JSON Response Schema](https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst#532entry-listing-json-response-schema) and for the definition of the "relationships" field say that: "Relationships MUST be grouped into a single JSON API relationships object for each entry type and placed in the relationships dictionary with that entry type name as key (e.g., "structures"). 

2. Preserve the possibility of a free-form JSON:API relationship key that can "group" related entries (possibly of different types) under a name. This can be done by describing the feature of a 'relationship name' to [3.6 Relationships](https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst#36relationships). Preferably with some hint in the implementers note about how one can encode these names in more straightforward key-value based output formats (e.g. "type name + '_' + relationship name"?). It is probably also a good idea with a clarification in 5.3.2 that the key in this dictionary field represents this "relationship name".

We probably need to sort this out before merging #386.

rartino:
Resolves #419 which we found to be blocking for #386 and probably should be addressed for the v1.2 release planning (#429)

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -897,6 +897,7 @@ OPTIONALLY it can also contain the following fields:
 
 - **relationships**: a dictionary containing references to other entries according to the description in section `Relationships`_ encoded as `JSON API Relationships <https://jsonapi.org/format/1.0/#document-resource-object-relationships>`__.
   The OPTIONAL human-readable description of the relationship MAY be provided in the :field:`description` field inside the :field:`meta` dictionary of the JSON API resource identifier object.
+  All relationships to entries of the same entry type MUST be grouped into the same JSON API relationship object and placed in the relationships dictionary with the entry type name as key (e.g., :entry:`structures`).
 
 Example:
 


ml-evs (approved):
Thanks @rartino, this seems clear enough to me.

merkys (approved):
This is exactly how I had understood it before, but later on found that JSON:API did not contain such constraint. Anyway, I see benefit in adding such constraint, thanks for creating a PR for that.

PR #451, opened 2022-12-30
Fix reference to definitions.units

rartino:
Apparently I missed an 's' on definitions.units when I wrote the section for readme.md.
This probably should be fixed before release.


ml-evs (dismissed):

JPBergsma:
In optimade.rst there are also two references to definition.units (lines 2122 and 2136). Should these be fixed as well?

merkys (changes requested):
@JPBergsma is right, there are more occurrences needing replacement.

rartino:
@JPBergsma @merkys Ok, done. Thanks for checking.

ml-evs (approved):

JPBergsma (approved):
LGTM

rartino:
@merkys you have an outstanding change request that blocks this merge (and I think it is addressed).

merkys (approved):
Right, all good now!

Issue #410, opened 2022-05-31
Derived properties vs. primary properties


rartino:
@sauliusg made a remark that it would be good to know whether a property comes directly out of the underlying data source, or if it has been processed (probably to fulfill the requirements of the standard, e.g., providing Cartesian coordinates from the reduced ones).

A suggestion is to add this info to the PR for property definitions #376. There is some discussion whether it is sufficient to say this is a derived property = FALSE/TRUE, or if it is strongly desirable for a database to indicate from what primary properties a derived property has been derived from.

merkys:
This issue has not been addressed by #376. However, I wonder whether the granularity is right. Suppose a database is serving both fractional and Cartesian coordinates. Some entries have only fractional, and some have only Cartesian coordinates in the underlying data, thus the database has to perform on-the-fly conversions to present both types of coordinates. Thus the database cannot say that *all* Cartesian/fractional coordinates are primary/derived. However, such distinction would be useful, but maybe in per-entry level?

rartino:
In connection to ranged properties #452, I start to think we have a general issue related to not having any systematic way to provide *per entry* property metadata. In that PR it is about what the ranges are, etc., here it is metadata about whether the property is derived. We've been touching upon the issue of providing accuracy information, etc.

Shouldn't we just agree on some sensible suffix? - e.g.,
`cartesian_coordinates:meta`
that is a dictionary with all metadata related to the property?

The suffix could also just be `cartesian_coordinates_meta`, and we'd not have to change anything on the software side (it'll behave just as any other data property).

We could also go in the opposite direction and place each separate category of metadata as a separate top-level field:
`cartesian_coordinates:derived`
`cartesian_coordinates:ranged`
meaning that ":" is the "meta data operator".

merkys:
I think `:` in property names will break the filter syntax, unless they are not supposed to be used in filters, but I guess having them queryable would be a benefit. Suffix `_meta` looks better to me.

rartino:
Yes, ":" would require us to extend the filter syntax - but it is a quite minor extension. Using "_" means we cannot split metadata up into categories (or, at least not in more than a few categories), as every suffix restricts the namespace of property names for data.

[...]

PR #452, opened 2023-01-09
Add ranged properties

JPBergsma:
In This PR, I make a suggestion for how large “ranged” properties could be shared with OPTIMADE.
See the discussion in Issue#428

I have now placed the main body of the text in section 3.9, I am however not sure if this is the best location. Feel free to suggest another location in the text.

I have now used a prefix to indicate that a property is a ranged property. This however creates a precedent, as it is the first time we use a prefix in this manner. If you prefer, I can also add a field that lists all the properties that are ranged properties.

Do we still want to add a field in the property definition that the field is a ranged field ?

Now I have put a SHOULD level query ability on all the dictionary fields except the values field. Is this OK or do we want to be more specific? 

Now the values can only depend linear;y on the index. Do we want to define a more general field which could contain information on how to calculate the value from the index? This would be considerably more complex and probably warrants a PR of its own.

Do we want to have a negative requirement on the range_id, i.e. If the range_id is not defined, a property must not have the same range as another property?

Do we want to make the values for slicing optional? 
The default values would then be [1,last,1]
So [] would return all the values, [100] would return all the values from the hundredth one onward and [30,50] would return every, value belonging to index 30 till 50.

For now, I removed properties average, min, max, and set as this becomes more complicated with multiple dimensions and I did not want to make this PR too big / complicated.
This can be addressed in a separate PR once this one gets merged.


optimade.rst
```diff
@@ -442,6 +445,162 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+  - :property:`values`: List of any data type (OPTIONAL)
+  - :property:`nvalues`: integer (OPTIONAL)
+  - :property:`range_ids`: list of strings (OPTIONAL)
+  - :property:`n_dim`: integer(REQUIRED)
+  - :property:`dim_size`: list of integers (REQUIRED)
+  - :property:`offset_linear`: float (OPTIONAL)
+  - :property:`step_size_linear`: list of floats (OPTIONAL)
+  - :property:`offset_regular`: list of integers (OPTIONAL)
+  - :property:`step_size_regular`: list of integers (OPTIONAL)
+  - :property:`indexes`: list of list of integers(OPTIONAL)
```
rartino: So, in the rest of the specification, these parts are formatted more like this:
```suggestion

The meta data dictionary MUST include these fields:

- :field:`serialization_format`: string. 
  To improve the compactness of the data there are several ways to show to which index a value belongs.
  The string MUST take one of the following values:
  
  - `linear`: The value is a linear function of the indexes.
  - `regular`: The value is set for one out of every :property:`step_size_sparse` indexes, with :property:`offset_sparse` indicating the index of the first value.
  - `custom`: A separate list with indexes is defined in the field :property:`indexes` to indicate to which index each value belongs.
  
- :field:`n_dim`: integer.
  The number of dimensions this property has.

- :field:`dim_size`: list of integers.
  The dimensions of the range in each dimension.

The meta data dictionary MAY include these fields:

- :field:`values`: List of any data type.
  The values belonging to this property.
  The format of this field depends on the property for which data is stored.

- :field:`nvalues`: integer.
   The number of values in the field :property:`values`.   
```
etc... (More needed to be added here)

optimade.rst
```diff
@@ -442,6 +445,162 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+
+    - **indexes**: If :property:`serialization_format` is set to :val:`"custom"`, this field holds the indexes to which the values in the value field belong.
+
+      - **Type**: List of lists of integers
+      - **Requirements/Conventions**: The value MUST be present when :property:`serialization_format` is set to "custom".
+        Otherwise it SHOULD NOT be present. The order of the values must be the same as those in :property:`values`.
+
+- **Example**:
+
+         .. code:: jsonc
```
rartino, line 557: This example is in json. While I guess it would be ok to say, here is an example for how it would look in the JSON output format, I think to follow the rest of the organization of the specification, this should go in section 4 (4.2, I think) as an explanation how to represent ranged properties in JSON-formatted responses.

optimade.rst
```diff
@@ -880,6 +1038,20 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
...
+  It consists of a property name directly followed by the range that should be returned.
+  Ranges can be specified for multiple properties by separating them with a comma.
+  The ranges are 1 based. i.e. The first value is value 1.
+  The first value of the range specifies the first index in that dimension for which values should be returned.
+  The second value specifies the last index for which values should be be returned.
+  The third value specifies the step size.
+  Databases that use ranged properties SHOULD support this query parameter.
+  Example:
+
+  If there would be a structure with id: id_ property :ranged-property:`_ranged_test_field` with the values :val:`[[9.64, 7.52, 0.69, 5.69],[4.82, 8.35, 3.26, 3.25],[4.82, 2.78, 7.87, 7.42], [5.49, 3.48, 1.65, 0.75]` the query: :query-url:`http://example.com/optimade/v1/structures/id_12345?property_ranges=_ranged_test_field[[1, 3, 2],[2, 3, 1]]`
```
rartino: As far as I can see, the text doesn't say anything about bracket parentheses, which it probably should. After having read the text a couple of time, I'm still not sure how the above text precisely defines the double-brackeded expression used as an example: `[[1,3,2],[2,3,1]]`. Perhaps the text can be clarified a bit?

rartino (changes requested):
Great to see a PR of this, thanks for the work with writing this up.

I find the formatting a bit strange. Most of the text is in section 3 for general API conventions, where all other sections just write out explanations of API features in regular text. This PR adds something formatted as a property definition, even though it is not. And the list of fields is also not formatted as similar lists elsewhere in the specification. I try to explain in my review comments, but perhaps also look at how the other sections explaining similar features elsewhere in the specification is formatted?

Another question is whether we should hold this back until v1.2 is out. I lean towards "yes". Any thoughts @ml-evs?

optimade.rst
```diff
@@ -442,6 +445,162 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
 :filter:`filter=_exmpl1_band_gap<2.0 OR _exmpl2_band_gap<2.5`
 
+
+Ranged Properties
+-----------------
+
+- **Description**: Ranged properties support slicing, so the client can request that only some of the values need to be returned.
+  Likewise, the server can use paging to return the property in multiple parts.
+  This can be useful for properties that are so large that it can be inconvenient to return them in a single response.
+  In that case a link is provided, as described in `JSON Response Schema: Common Fields`_ under the `links.next` field, from which the remainder of the requested data can be retrieved.
```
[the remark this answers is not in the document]
JPBergsma: A single entry with the data of a trajectory can easily be gigabytes in size. 
Returning all that data in one response would give a long response time.   
So we need a way for the server to limit the size of a response. (Perhaps we could create some kind of streaming response, but I am not sure how the different JSON tools will handle this. We may need to create our own JSON parser.)  

The server can put an extra field in the next link that specifies to which extent the different properties have already been sent. They can slice the property in the same way the client does. 
There is, however, no need to define how servers should do this as this only applies to their own server, so they can do what they want as long as the client gets the correct data.

I could add a field to the response to make it clearer what the server has returned. e.g. which range of values has been returned, so it is easier for the client to reconstruct the entry.

We could add a field with a name different from next to the links section, but I am not sure if that is any clearer. 

Maybe paging is not the right word to use here. I will adjust the description.

optimade.rst
```diff
@@ -442,6 +445,162 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+  They also provide a method to correlate the values of two ranged properties.
+  The meta data is returned by default, the values are only returned when specifically requested via the :query-param:`ranged property` query parameter as described under `Entry Listing URL Query Parameters`_.
```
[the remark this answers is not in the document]
JPBergsma: I will remove the spaces as in your suggestion.
Do you still want to change something else about the formatting?

optimade.rst
```diff
@@ -442,6 +445,163 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+               "serialization_format": "linear",
+               "step_size_linear": 0.2,
+             },
+             "_exmpl_ranged_thermostat": {
+               "n_dim": 1,
+               "dim_size": [100],
+               "range_ids": ["mdsteps"],
+               "serialization_format": "custom",
+               "nvalues": 3,
+               "values": [20, 40, 60]
```
vaitkus, line 598:
```suggestion
               "values": [20, 40, 60],
```

optimade.rst
```diff
@@ -442,6 +445,163 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+               "step_size_regular": [1],
+               "nvalues": 3,
+               "values": ["He", "Ne", "Ar"],
+             },
+             "_exmpl_ranged_time":{
+               "n_dim": 1,
+               "dim_size": [100],
+               "range_ids": ["mdsteps"],
+               "serialization_format": "linear",
+               "step_size_linear": 0.2,
```
vaitkus, line 590:
```suggestion
               "step_size_linear": 0.2
```

optimade.rst
```diff
@@ -442,6 +445,163 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+             },
+             "_ranged_species_at_sites": {
+               "n_dim": 1,
+               "dim_size": [3],
+               "range_ids": ["particles"],
+               "serialization_format": "regular",
+               "offset_regular": [0],
+               "step_size_regular": [1],
+               "nvalues": 3,
+               "values": ["He", "Ne", "Ar"],
```
vaitkus, line 583:
```suggestion
               "values": ["He", "Ne", "Ar"]
```

optimade.rst
```diff
@@ -442,6 +445,163 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+               "range_ids": ["mdsteps","particles","xyz"],
+               "serialization_format": "regular",
+               "offset_regular": [1, 1, 1],
+               "step_size_regular": [1, 1, 1],
+               "nvalues": 900,
+               "values": [[[2.36, 5.36, 9.56],[7.24, 3.58, 0.56],[8.12, 6.95, 4.56]],
+                          [[2.38, 5.37, 9.56],[7.24, 3.57, 0.58],[8.11, 6.93, 4.58]],
+                          [[2.39, 5.38, 9.55],[7.23, 3.57, 0.59],[8.10, 6.93, 4.57]],
+                            // ...
+                         ],
```
vaitkus, line 573:
```suggestion
                         ]
```

optimade.rst
```diff
@@ -880,6 +1040,25 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
...
+  A range is written as a list with a list for each dimension.
+  Each dimensions list has three values.
+  The first value of the range specifies the first index in that dimension for which values should be returned.
+  The second value specifies the last index for which values should be returned.
+  The third value specifies the step size.
+  Ranges can be specified for multiple properties by separating them with a comma.
+  Databases MUST return the values belonging to properties listed and SHOULD use the ranges in this query parameter.
+  For properties with :property:`serialization_format` :val:`custom` indexes that fall in the requested range but for which there is no value defined should not be returned.
+  For properties with :property:`serialization_format` :val:`regular` indexes that fall in the requested range but for which there is no value defined should have the value :val:`null`.
+  The ranges are 1 based, i.e. the first value has index 1, and inclusive i.e. for a the range :val:`[10,20,1]` the last value returned belongs to index 20.
```
vaitkus, line 1055:
```suggestion
  The ranges are 1 based, i.e. the first value has index 1, and inclusive i.e. for the range :val:`[10,20,1]` the last value returned belongs to index 20.
```

optimade.rst
```diff
@@ -442,6 +445,163 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+-----------------
+
+- **Description**: Ranged properties support slicing, so the client can request that only some of the values need to be returned.
+  Likewise, the server can use paging to return the property in multiple parts.
+  This can be useful for properties that are so large that it can be inconvenient to return them in a single response.
+  If an entry is too large to be returned in a single response a link is provided, as described in `JSON Response Schema: Common Fields`_ under the `links.next` field, from which the remainder of the requested data can be retrieved.
+  Ranged properties also provide a method to correlate the values of two ranged properties via a :property:`range_ids`.
+  The metadata is returned by default, the data is only returned when specifically requested via the :query-param:`ranged property` query parameter as described under `Entry Listing URL Query Parameters`_.
+
+When a client does not use query parameters to select a range for the ranged property, the server returns a dictionary with meta data about the range property with the following format:
```
vaitkus, line 459:
```suggestion
When a client does not use query parameters to select a range for the ranged property, the server returns a dictionary with metadata about the range property with the following format:
```

optimade.rst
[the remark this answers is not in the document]
rartino: Sorry if I was unclear: I did not mean to question that range properties need to be fetched in parts.

I suppose you can refer to any mechanism for fetching partial data as "paging". However, in JSON:API paging (or actually, the specification uses "pagination") via the top-level "links object" is described as for servers that choose to "limit the number of resources returned" when returning a collection of resources as a result. I do not see a JSON:API-conforming way to "extend" that top-level links object the way I think you suggest (?)

Hence, I thought we were going to only allow fetching ranged data in parts via query parameters specifying ranges. However, I think you are right that some form of JSON:API `links`-object type paging would be better as the MANDATORY baseline mechanism. That way all the new query parameters for ranged data can be made OPTIONAL, which is perfect for us who want to support MD data but have no use case for arbitrary indexing into frames (and thus do not want to put any effort into supporting that).

So, we just need to specify a place for that range-specific a JSON:API links object. But, where? Again, I don't think it can go in the top-level `links` object. So, what about inside the ranged property metadata dictionary?

optimade.rst
```diff
@@ -442,6 +445,163 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+- **Type**: dictionary with keys:
+
+  - :property:`serialization_format`: string (REQUIRED)
+  - :property:`values`: List of any data type (OPTIONAL)
```
rartino, line 463: So, do I understand correctly that the present design is for the member `values` in the "metadata dictionary" to also contain the actual data values? (So, in a sense, "metadata" is mislabeling it?)

So, correlated with the thoughts above about pagination: how about making responses always contain *two fields* for ranged properties, one metadata field with only metadata, and one for the data. That way, even the initial response that contain a ranged property can pass the first page of data the way it would if this was just a regular (non-ranged) property. This means that if the ranged property happens to be small, the client gets all the data it needs in the first response without having to do anything else. Only in the case all data does not fit in one response does the client need to engage with the specifics of accessing ranged properties. This will actually greatly help collections (#386): I felt we were going to end up with an ugly design to force clients to fiddle with range access for all collections, when in many cases of smaller collections this would be unnecessary.

So, I propose:

* Split out `values` into its own field; i.e., edit the initial text to say that for ranged properties two fields are returned:
  `<property_name>_ranged_meta`: a dictionary with all these metadata fields. 
  `<property_name>_ranged_data`: a list of values of any type.
  The server MAY support querying on either or both.

For the field `<fieldname>_ranged_meta` all the fields you defined here are kept with the same meaning, but two new ones are needed (and perhaps these should be the only ones MANDATORY):
- 'links': a JSON:API object on the format described in the JSON:API specification section "Pagination", which is used for pagination of the ranged property.
- 'more_data_available': boolean. True if the data does not fit in a single page and this is not the final page. 

The field `<fieldname>_ranged_data` would then be defined as:

* If no range-selecting query parameters are given, it contains the first page of data (which is all of the data if and only if `more_data_available` is `false` in `<fieldname>_ranged_meta`. If `<fieldname>_ranged_meta` is `true`, then further pages of data can be fetched by the client via the URL in `next` in `<fieldname>_ranged_meta`.
* If any of the range-selecting query parameters defined below are given, then data matching that range selection is returned. The server MAY support paging also of the requested range, or MAY return an error if the requested range is too big to fit in a single response.

optimade.rst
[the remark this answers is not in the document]
rartino: To make it match other Section 3 subsections, I also marked to remove "**Description**". Other than that, I think this reads as the general text in other Section 3 subsections.

optimade.rst
JPBergsma: > So, do I understand correctly that the present design is for the member values in the "metadata dictionary" to also contain the actual data values? (So, in a sense, "metadata" is mislabeling it?)

Yes, the dictionary would also contain data when the property is listed within the `property_range` query parameter so the term "metadata dictionary" is misleading.

> how about making responses always contain two fields for ranged properties, one metadata field with only metadata, and one for the data.

I do like that all the data is grouped together in a single field. We could also make a separate sub dictionary to group the metadata separate from the data. But overall, it is not that important how the fields are organized.

> That way, even the initial response that contain a ranged property can pass the first page of data the way it would if this was just a regular (non-ranged) property. 

I am not sure, I like your proposal to return the first “page” of data of a ranged property by default.
The first request will typically be for a client to check which entries are present. It can then select the most interesting entries. And retrieve the additional data for these entries. 
With your proposal, the number of entries that a server can return in a single response would be much smaller, as most of the response would consist of the first “pages” of properties. 

If a user wants the behaviour you described, they could simply use the `property_range` query parameter to specify that they want the first x values of a property returned.

If a client would retrieve only a single property, it would make more sense to already provide some of the data, but having a different behaviour for the single entry endpoint could also be confusing/complicate things.

> I felt we were going to end up with an ugly design to force clients to fiddle with range access for all collections, when in many cases of smaller collections this would be unnecessary.

Another solution could perhaps be to define the collection field flexibly so that you could have a regular and a ranged version of this property. I already wrote that if the part after the ranged prefix is the name of a regular property, it should meet all the requirements of that property.

> 'links': a JSON:API object on the format described in the JSON:API specification section "Pagination", which is used for pagination of the ranged property. 

In the JSON API v1.0 links is a reserved key word, so we would have to give this field a different name. (or drop support for json API 1.0) see section 5.2.3 of the JSON API specification.

It is quite a big change you are suggesting. Instead of retrieving the data per entry, you want to retrieve data per property. A bit as if each property is a resource in itself. 

> If any of the range-selecting query parameters defined below are given, then data matching that range selection is returned. The server MAY support paging also of the requested range, or MAY return an error if the requested range is too big to fit in a single response. 

I do not think returning errors to the user is very nice when the user can’t know in advance that something is wrong. 
So we should somehow specify how much data the server can return, and then the client should construct a request that fits between these bounds. 
But this is difficult as the size of the properties differs from entry to entry and is unknown to the client. 
So I do think that the server should implement pagination as it has all the data needed to do so while the client has not. 

We should probably also think about creating streaming responses, so sending back large data items is less odd to the client. Now we first prepare the response and then send it. But for large responses this would be odd as the user does not get any data back for a while. If we do this, we can send larger responses, and we would not need ranged properties as often. But this is probably more of an issue for the optimade python tools rather than a change in the optimade specification.

Sara and Gian-Marco have asked me to prioritize writing a tutorial for the old trajectory proposal, so it may take a while before I get back to this topic.

optimade.rst
JPBergsma: Ok I have tried to change the formatting to fit this pattern.

optimade.rst
JPBergsma: Ok, I have moved the example to section 4.2.

optimade.rst
JPBergsma: I have tried to explain this a bit better.

optimade.rst
```diff
@@ -442,6 +445,162 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
 
+
+Ranged Properties
+-----------------
+
+- **Description**: Ranged properties support slicing, so the client can request that only some of the values need to be returned.
+  Likewise, the server can use paging to return the property in multiple parts.
+  This can be useful for properties that are so large that it can be inconvenient to return them in a single response.
+  In that case a link is provided, as described in `JSON Response Schema: Common Fields`_ under the `links.next` field, from which the remainder of the requested data can be retrieved.
+  They also provide a method to correlate the values of two ranged properties.
```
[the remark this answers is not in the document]
rartino: I was just trying to spell out more explicitly what I thought *you* meant on line 456. If this isn't it, then, what did you mean by "a method to correlate the values of two ranged properties"?

optimade.rst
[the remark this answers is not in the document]
JPBergsma: I think it would be a bit strange to place the links in the ranged property metadata dictionary. So far I considered the trajectory as one resource, so I would expect that the next link applies to the whole resource and not just one property. So I would suggest placing the link a level higher at the level of the entire trajectory resource.   
 
I am not sure why you suggested placing the link with the individual property.
I guess it is possible to turn the ranged properties into resources on their own. In that case the trajectory entry would include links to these resources like we do for the references. 
As long as the server does not support the Include parameter, this could work as well. In that case, the client would need to retrieve the data for each individual property with a separate request. And if a property is too large, we could place the links to retrieve the rest of the property within the property attributes.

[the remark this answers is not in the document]
JPBergsma: I do not think the new query parameter can be OPTIONAL. You still need a way to indicate which ranged_properties need to be returned.

optimade.rst
JPBergsma: > Where does the text of this PR presently say this, i.e., that the behavior of value depends on the query parameter. 

On line 526/527 it is mentioned that the values are only returned when the property is listed in the property_ranges query parameter.  
I have added an extra line to the description of the values field to make it a bit clearer.

> I'm trying to move this in the direction where implementations that do not support random access of the data does not have to implement property_range at all, which is fairly complex to parse the way it is currently defined.

In that case, the client would still need a way to specify which fields it wants to have. So I think you would still need the property_range query parameter.
How would you make it easier to parse ?

> I hadn't even considered that the way you've written this is meant to support a range selection over every entry that matches a filter. I do not think implementations should be forced to support that even if they do support range selection over a single entry.

Most of the code for the single entry endpoint and the multiple entry endpoint is the same. So I do not think this makes the implementation more difficult. It may actually be more work to create different behaviour for the single and multiple entry points.  

> Can you cite the part of the JSON:API specification that you read to say one cannot have links as a key in a dictionary inside attributes?

“Complex data structures involving JSON objects and arrays are allowed as attribute values. However, any object that constitutes or is contained in an attribute MUST NOT contain a relationships or links member, as those members are reserved by this specification for future use.”

> No, that is not what I meant, sorry if I was unclear. This was meant to be a small adjustment in behaviour - the URL in links is meant to give you a single entry response with the next page of data for every property using the same range id as the one your are paging in. You can implement it by a URL that adds "property_ranges" selecting the next range (but you don't have to).

So, if I understand it correctly, you want to give each resource/entry its own next link to retrieve the remainder of this entry, and use the standard JSON API next link to point to the next entry that matches the filter.  
I do not use the range_id for retrieving the data.
You use the name of the field in the query parameter property_ranges. So you can define a separate range for each property.

> I see now that I had missed how you want to allow multidimensional indexing with multiple range_ids; one would in that case have to work out how paging works in the multidimensional case. (But, do we need to support multidimensional ranges?)

The current proposal already supports paging in multiple dimensions. (The range_ids have nothing to do with indexing. It only shows that two properties use the same range, so that values with the same index belong to each other.)
I mostly started out thinking how I could make sure that the proposal was compatible with a future extension to multiple dimensions. And when I had figured that out, I thought I could just as well write it out in the specification.
I could leave out the references to multiple dimensions, although in that case some people may wonder why there are some extra pairs of square brackets needed.  
Some people have also asked me whether it would be possible to retrieve the cartesian_site_positions only for specific atoms. The current specification does not yet allow that, but I think the property_ranges query parameter could be extended to allow specific indexes. We could, for example, make a convention that the list that now contains the first last and step parameters could be replaced with a list of the indexes, placed between “(“ and “)”, for which data should be retrieved.

rartino: >> I'm trying to move this in the direction where implementations that do not support random access of the data does not have to implement property_range at all, which is fairly complex to parse the way it is currently defined.

> In that case, the client would still need a way to specify which fields it wants to have. So I think you would still need the property_range query parameter.

I don't get what you mean here, sorry. Which fields to return in a response is requested via `response_fields`?

> How would you make it easier to parse?

I'm trying to propose to make it so that the single *mandatory* feature to get all data in a ranged property is via a next-like paging system, which wouldn't require interpreting `property_range` at all.

>> Can you cite the part of the JSON:API specification that you read to say one cannot have links as a key in a dictionary inside attributes?

> “Complex data structures involving JSON objects and arrays are allowed as attribute values. However, any object that constitutes or is contained in an attribute MUST NOT contain a relationships or links member, as those members are reserved by this specification for future use.”

Oh, wow, I had missed this completely. It is a really strange rule to have for attribute data at arbitrary depth, and it seems they removed it in v1.1. Well, ok then, you are right that we would have to name that inner field something different than `links`. Perhaps just place a URL in a member named `next`.

> So, if I understand it correctly, you want to give each resource/entry its own next link to retrieve the remainder of this entry, and use the standard JSON API next link to point to the next entry that matches the filter. I do not use the range_id for retrieving the data. You use the name of the field in the query parameter property_ranges. So you can define a separate range for each property.

No, the way you describe it does not sound right. Lets me try again.

The standard JSON API next link would work as it always has: it pages over a range of returned entries. This is an absolute necessity to adhere to JSON:API.

A ranged property would, among its metadata, communicate a "special" next link. Following that next link would give you that entry (that *whole* entry with all attributes/properties as requested by `response_fields`) but where the "data" part of the ranged property contains the next range of values. This is the most straightforward way (and thus easiest to implement) I can imagine to communicate paged ranged data values within a fully valid OPTIMADE/JSON:API response.

Maybe the most clear way to express it is that this "feature" is meant to be completely compatible with your design of `property_ranges`. Someone who has implemented `property_ranges` gets precisely the right thing by simply constructing that special `next` link by adding a `property_ranges` parameter that specifies the next page of data.

> Some people have also asked me whether it would be possible to retrieve the cartesian_site_positions only for specific atoms. The current specification does not yet allow that, but I think the property_ranges query parameter could be extended to allow specific indexes.

Why would this extension belong in `property_ranges`? If so, one could only do use it for non ranged properties? It seems to me it belongs in `response_fields` or as its own query parameter, where one could imagine various ways to extend it to ask for specific subsets of a property.

optimade.rst
```diff
@@ -880,6 +1034,26 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
   If provided, these fields MUST be returned along with the REQUIRED fields.
   Other OPTIONAL fields MUST NOT be returned when this parameter is present.
   Example: :query-url:`http://example.com/optimade/v1/structures?response_fields=last_modified,nsites`
+- **property\_ranges**: specifies which data ranges should be returned for ranged properties.
+  It MUST be supported by databases having ranged properties.
+  It consists of a property name directly followed by the range that should be returned.
+  A range is a list containing a list for each dimension.
+  Each dimensions list has three integer values.
```
rartino, line 1041:
```suggestion
  Each dimension's list has three integer values.
```

optimade.rst
```diff
@@ -880,6 +1034,26 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
...
+  The first value of the range specifies the first index in that dimension for which values should be returned.
+  The second value specifies the last index for which values should be returned.
+  The third value specifies the step size.
+  A list consists of a pair of square brackets ("[", ASCII 91(0x5B)) and ("]", ASCII 93(0x5D)) enclosing a number of values separated by a comma (",", ASCII 91(0x5B))
+  Ranges can be specified for multiple properties by separating them with a comma.
```
rartino, line 1046:
```suggestion
  Ranges can be specified for multiple properties by separating them with a comma.
  The field may include optional space characters, which do no alter the meaning of the expression.
```

optimade.rst
```diff
@@ -880,6 +1034,26 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
...
+  Databases MUST return the :property:`values` and :property:`indexes` field belonging to properties listed and SHOULD use the ranges in this query parameter.
```
rartino, line 1047: Some reformulation for clarity:
```suggestion
  When a client includes `property_ranges` in a request, the response MUST include the :property:`values` and :property:`indexes` field belonging to properties listed and SHOULD use the ranges in this query parameter.
```
but the SHOULD here makes me nervous. I think supporting this parameter should be optional, but if you support it, don't you have to follow the indexing request?

optimade.rst
```diff
@@ -880,6 +1034,26 @@ Standard OPTIONAL URL query parameters not in the JSON API specification:
...
+  For properties with :property:`serialization_format` :val:`custom` indexes that fall in the requested range but for which there is no value defined should not be returned.
+  For properties with :property:`serialization_format` :val:`regular` indexes that fall in the requested range but for which there is no value defined should have the value :val:`null`.
+  The ranges are 1 based, i.e. the first value has index 1, and inclusive i.e. for the range :val:`[10,20,1]` the last value returned belongs to index 20.
+  Example:
+
+  If there would be a structure with id: id_12345 and a property :ranged-property:`_ranged_test_field` with the values :val:`[[9.64, 7.52, 0.69, 5.69], [4.82, 8.35, 3.26, 3.25], [4.82, 2.78, 7.87, 7.42], [5.49, 3.48, 1.65, 0.75]]` the query: :query-url:`http://example.com/optimade/v1/structures/id_12345?property_ranges=_ranged_test_field[[1, 3, 2], [2, 3, 1]]`
+  will return the value: :val:`[[7.52, 0.69], [2.78, 7.87]]`
+  Multiple ranges can be requested in one query. e.g. :query-param:`property_ranges=_ranged_test_field[[1, 3, 2], [2, 3, 1]], _ranged_other_field[[1,100,1]]`
```
rartino, line 1055: A bit more clarity here about how the response is supposed to be formatted (please check that it is still in line with your intent):
```suggestion
  A database has a structure with id: id_12345 and a property :ranged-property:`_ranged_test_field` with the two-dimensional data values :val:`[[9.64, 7.52, 0.69, 5.69], [4.82, 8.35, 3.26, 3.25], [4.82, 2.78, 7.87, 7.42], [5.49, 3.48, 1.65, 0.75]]`. 
  A client makes a request :query-url:`http://example.com/optimade/v1/structures/id_12345?property_ranges=_ranged_test_field[[1, 3, 2], [2, 3, 1]]`. 
  The response is then a single entry response for structure `12345` which includes the `_ranged_test_field` property where the `value` key is set to :val:`[[7.52, 0.69], [2.78, 7.87]]`.
  
  Multiple ranges can be requested in one query. e.g. :query-param:`property_ranges=_ranged_test_field[[1, 3, 2], [2, 3, 1]], _ranged_other_field[[1,100,1]]`
```

optimade.rst
JPBergsma: > I don't get what you mean here, sorry. Which fields to return in a response is requested via response_fields?

In the discussion on the trajectory endpoint, it was concluded that it would be confusing to determine whether the values should be returned based on whether the property was present in the response_fields. If the whole property (metadata + data) is present under a single field, there would be no way to specify with the response_fields parameter that you only want the metadata.
So instead we decided to trigger returning the data by the presence of the query parameters first_frame, last_frame and frame_step.
It therefore made sense to me to do the same here with the property\_ranges query parameter.

I now understand why you wanted to split the data and metadata into two separate fields. That way, you could add the data field to the response_fields to get the data. And not return it otherwise. 
If we however add by default a next link to the metadata, it would not be necessary to have a separate field for the data. 

> No, the way you describe it does not sound right. Let me try again.
The standard JSON API next link would work as it always has: it pages over a range of returned entries. This is an absolute necessity to adhere to JSON:API.
A ranged property would, among its metadata, communicate a "special" next link. Following that next link would give you that entry (that whole entry with all attributes/properties as requested by response_fields) but where the "data" part of the ranged property contains the next range of values. This is the most straightforward way (and thus easiest to implement) I can imagine to communicate paged ranged data values within a fully valid OPTIMADE/JSON:API response.
Maybe the most clear way to express it is that this "feature" is meant to be completely compatible with your design of property_ranges. Someone who has implemented property_ranges gets precisely the right thing by simply constructing that special next link by adding a property_ranges parameter that specifies the next page of data.

Ok, I think I understand you now. I’ll adjust in the proposal. 

> Why would this extension belong in property_ranges? If so, one could only use it for ranged properties? It seems to me it belongs in response_fields or as its own query parameter, where one could imagine various ways to extend it to ask for specific subsets of a property.

The cartesian_site_positions field is an example of multidimensional data.
In a trajectory the first dimension would cover the frames, the second dimension the particles and the third the dimensions(x, y, z).
In a PDB file, the atoms belonging to the protein usually come first, while the solvent atoms come later. So one could make a query, selecting only a part of the range in the second dimension, so that only the positions of the atoms of the protein and not those of the solvent would be returned.

rartino:
I apologize for feature-creep on already long on-going work; but, after some more thinking I realize we have a number of properties already defined that would benefit from the ability to be "ranged", not the least, `cartesian_site_positions`. I also suspect we will end up arguing a lot about in the future about which properties to "range" and which not to.

I think with some, actually quite minor, adjustment of this PR we can allow *any* property to be ranged on an entry-by-entry basis. If people agree this would be useful, I think this is how it could be done:

- When property data easily fits inside the response (i.e., what we always do now), the response should look exactly the same as it does now; so any entry that can be represented in OPTIMADE today remains unchanged.

- When some property does not fit, we do not want clients written according to the present specification to break, so the property field should in that case be `null` in the response. Clients that are not aware of "ranged properties" would hence conclude that the data is missing (which it, in a way, is: it is too big to represent using the standard the client uses).
 
- In the latter case, the implementation also includes the `<property definition>_ranged` field just as already is described in this PR that enable clients aware of ranged properties to page in the response.

The above seems to me as a minor change to the PR with *major* utility going forward.

JPBergsma:
This sounds like a good idea.
I was already thinking about something like this when I wrote  _If the part of the property name after the "_ranged_" prefix matches the name of a OPTIMADE field for the entry point, the values in the list of the :property:`values` MUST follow the rules of this property._

Do we still want to keep a ranged postfix as you describe above ? Or simply use the meta field as discussed in Issue #410 ?

rartino:
> Do we still want to keep a ranged postfix as you describe above ? Or simply use the meta field as discussed in Issue #410?

The idea of a general per-entry metadata field in #410 came after the above post - to some degree provoked by the above.

But, yes, my vote right now is that we'd follow the outline there and do something along the line of this as the response from the `trajectory` endpoint (trying to adhere as closely as possible to the existing PR as it now is):
```
{
  cartesian_site_positions: null,
  cartesian_site_positions_: {
    "range_ids": ["mdsteps","particles","xyz"],
    "frame_serialization_format": "linear",
    "n_dim": 3 
    "dim_size" : [10023934, 10, 5, 3],
    "returned_range": [null, null, null],
    "next": "https://optimade.example.com/trajectories/4711?property_ranges=cartesian_site_positions[[0, 999, 1],[0, 4, 1],[0, 2, 1]]"
  }
}
```
And upon following the next link:
```
{
  cartesian_site_positions: [
                          [[2.36, 5.36, 9.56],[7.24, 3.58, 0.56],[8.12, 6.95, 4.56]],
                          [[2.38, 5.37, 9.56],[7.24, 3.57, 0.58],[8.11, 6.93, 4.58]],
                          [[2.39, 5.38, 9.55],[7.23, 3.57, 0.59],[8.10, 6.93, 4.57]]
                            // ...
                         ],
  cartesian_site_positions_: {
    "frame_serialization_format": "linear",
    "range_ids": ["mdsteps","particles","xyz"],
    "n_dim": 3 
    "dim_size" : [10023934, 10, 5, 3],
    "returned_range": [[0, 999, 1],[0, 4, 1],[0, 2, 1]],
    "next": "https://optimade.example.com/trajectories/4711?property_ranges=cartesian_site_positions[[1000, 1999, 1], [0, 4, 1],[0, 2, 1]]"
  }
}
```
However, for cases like this with `range_ids` that are going to be shared over many properties describing "frames", I really would want to use those identifiers in `property_ranges` instead with the purpose of paging in all of them at the same time.

JPBergsma:
Ok, then I’ll update the trajectory proposal, so it uses the metadata field as in your example.

> However, for cases like this with range_ids that are going to be shared over many properties describing "frames", I really would want to use those identifiers in property_ranges instead with the purpose of paging in all of them at the same time.

Just to make things clear. Do you want to be able to select all properties that have a certain range_id, and/or do you want to have the option to have that the data is returned per frame? i.e. for each frame all the requested properties are present, rather than that all the data for one property is first returned.

optimade.rst
```diff
@@ -474,26 +473,29 @@ All fields in this section SHOULD be queryable with support for all mandatory fi
...
   The size of the range in each indexable dimension.
 
-Depending on the value of the :field:`serialization_format`, the following fields MUST be present or SHOULD NOT be present in the dictionary.
+Depending on the value of the :field:`serialization_format`, the following fields SHOULD/MUST be present or SHOULD NOT be present in the dictionary.
 All fields in this section SHOULD be queryable with support for all mandatory filter features.
 
-- :field:`nvalues`: integer.
-  The number of values in the field :property:`values`.
-  This field MUST be present when :property:`serialization_format` is not set to :val:`"linear"` else it SHOULD NOT be present.
+- :field:`n_values`: integer.
```
rartino, line 488: Prior OPTIMADE fields for counts have been written as, e.g., `nelements`; so regardless of what we think of that, perhaps we should stick to that?

optimade.rst
```diff
@@ -455,15 +455,14 @@ Ranged properties are used for properties that are too large to be returned by d
 - **Requirements/Conventions**:
 
   - **Support**: OPTIONAL support in implementations.
-  - A ranged property can be recognized by the presence of the :field:`range` in the metadata of the property, i.e. in the field: <property_name>_meta.
+  - A ranged property can be recognized by the presence of the :field:`range` in the metadata of the property, i.e. in the field: :field:`<property_name>_meta`.
```
rartino: If all fields are technically "ranged", I think the differentiation here has to be whether there is something along a field `ranged/more_data_available` present and `true`, meaning not all data is communicated in the current response. It must be allowed to have a `ranged` dictionary even if that is not true.

optimade.rst
```diff
@@ -503,45 +505,55 @@ Depending on the value of the :property:`serialization_format`, the following fi
...
 
 The dictionary MAY include these fields.
 
 - :field:`range_ids`: list of strings.
-  A list with an identifier for each dimension of the range. It shows that that dimension correlates to the same dimension of another range.
-  For example, when data of an MD trajectory is shared, it could be used to indicate that the energies and the cartesian_site_positions of the index in a certain dimension are correlated.
-  i.e. which energy belongs to which set of cartesian_site_positions.
+
+  A list with an identifier for each dimension of the property. If two properties have the same range_id for a dimension it means that the values at an index along this dimension belong to each other.
+  For example, if both the :property:`energy` and :property:`cartesian_site_positions` of a trajectory have the same :field:`range_ids` it indicates that the energy at an index x(in the dimension labeled by this range_ids) belongs to the cartesian_site_positions at the same index x.
```
rartino: Right, I understand now what you meant with "correlated" a few edits ago. I think we need to find a term for this, but one that hopefully is more clear. Maybe "synchronized"?

But, is your idea that one can get different ranges from different properties even for dimensions that are "synchronized"? I'd propose that one cannot. If asking ranges by range_id, all "syncronized" quantities would indeed be synchronized. (But otherwise "syncronized" is maybe not the best term)

rartino: I'm honestly not sure if my version is better - consider it a suggestion that you can accept or rework into your version for clarity.
```suggestion
  A list with an identifier for each dimension of the property.
  If any dimension in two or more properties share the same :field:`range_id` those dimensions should be thought of as the same dimension.
  For example, if both the :property:`energy` and :property:`cartesian_site_positions` of a molecular dynamics trajectory share a :field:`range_id` of :val:`frame`, it should be interpreted as them providing, for each frame, synchronized values of the coordinates and the energy.
```

rartino (commented):
A few comment (submitting because I've started to get errors about that other changes have been made)

optimade.rst
```diff
@@ -442,6 +445,120 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+The dictionary MAY include these fields.
+
+- :field:`range_ids`: list of strings.
+
+  A list with an identifier for each dimension of the property.
+  If two properties have the same range_id for a dimension, it means that the values at an index along this dimension belong to each other.
+  For example, if both the :property:`energy` and :property:`cartesian_site_positions` of a trajectory have the same :field:`range_ids` it indicates that the energy at an index x(in the dimension labelled by this range_ids) belongs to the cartesian_site_positions at the same index x.
+  SHOULD be a queryable property with support for all mandatory filter features.
+
+If the :field:`<property_name>` contains data, the following properties MUST be present or SHOULD NOT be present, depending on the value of the :property:`serialization_format`. Querying is not relevant for these properties and SHOULD NOT be supported.
```
rartino, line 530: Should `range_id` really be a per-entry metadata? Is that perhaps something we define as part of the property definition?

JPBergsma: I have not yet been able to come up with an example where defining it at the property definition would cause more problems than at the single entry level. So I think it should be possible to define it in the property definition.

optimade.rst
JPBergsma: Yes, I'll switch to the notation without underscore.

optimade.rst
JPBergsma: Well, I guess you can make all fields ranged by setting the number of indexable dimensions to zero and return the whole property in one go, so the field behaves the same as the current fields. So you could still add a ranged subfield to the metadata even if a property does not support any of the new features.  

There is already a more_data_available field that can be used to indicate whether the server has retuned all the data.
By setting the number of indexable_dimensions to 0, you as a server could also indicate to the client that it is not possible to perform slicing on the property. But perhaps it is cleaner to add a boolean to the metadata to define whether the supports that the client can request a subrange to retrieve.   

Does this answer your question/remark ?

optimade.rst
```diff
@@ -21,6 +21,7 @@ OPTIMADE API specification v1.2.0~develop
 
      entry : names of type of resources, served via OPTIMADE, pertaining to data in a database.
      property : data item that belongs to an entry.
+     ranged_property : A property that can be returned in pieces and that supports slicing.
```
rartino, line 24: This is the list of restructuredText markup text roles. Do you need a separate text role for range properties vs. regular properties? (Do you expect them to ever be marked up different in a typeset specification?) I also don't see it used anywhere?
```suggestion
```

optimade.rst
```diff
@@ -67,6 +68,8 @@ OPTIMADE API specification v1.2.0~develop
 
 .. role:: property(literal)
 
+.. role:: ranged-property(literal)
+
```
rartino, line 72: If you need this (which I still don't get why), it is inconsistently defined with the above comment that describes it as `ranged_property`.
```suggestion
.. role:: ranged-property(literal)

```

optimade.rst
```diff
@@ -442,6 +445,110 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+They can also be used by the server to support slicing, so the client can request that only a subsection of the values needs to be returned.
+
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations.
+  - A ranged property can be recognized by the presence of the field :field:`range` in the metadata of the property, i.e. in the field: :field:`<property_name>_meta`.
+  - For a ranged property, the server MAY return :val:`null` or only a part of the values of the property under the field :field:`<property_name>`, so the size of the entries remains limited, and many entries can be returned in a single response.
+    In that case, a links object MUST be provided in the field :field:`<property_name>_meta.range.next` from which the next part of the property is returned.
+
+The dictionary under :field:`<property_name>_meta.range` MUST include these fields.
```
rartino, line 463:
```suggestion
The metadata field of the ranged property, :field:`<property_name>_meta.range`, MUST include these fields:
```

optimade.rst
```diff
@@ -442,6 +445,110 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+
+  If there is still more data available for the property, this field will contain a URL pointing to the entry, to which this property belongs, which contains the next set of values for this property.
+  If all the data for this property has been returned, the value SHOULD be :val:`null`
+
+- :field:`more_data_available`: boolean.
+
+  :field-val:`false` if all the values in the requested range have been returned, and :field-val:`true` if the returned values are incomplete.
+
+
+If the :field:`<property_name>` contains data (i.e. it is not (null or an empty list)), the following additional properties can be present.
```
rartino, line 517: I don't think you should use parentheses this way in text.
```suggestion
If the :field:`<property_name>` contains data, i.e., it is neither :val:`null`, nor the empty list, the following additional properties can be present:
```

rartino (commented):
Thanks for all the work put into this.

I am generally very supportive of this version. I especially like how the property range parameter is much more readable now.

But, I would:

1. Appreciate someone else also takes a careful look to see if the present version makes sense also to them (since this is a result of you and I discussing).

2. It would be great if someone implement at least a minimal subset of this before it is merged into a version of OPTIMADE we release, since I expect little details on how things are defined may be much more clear then. I'm certainty noticing that effect when working with the property definitions now.

optimade.rst
```diff
@@ -697,6 +804,118 @@ An example of a full response:
...
+           "stop": 3,
+         },{
+           "name": "xyz",
+           "start": 1,
+           "step": 1,
+           "stop": 3,
+         }],
+         "contains_null": false,
+         "more_data_available": true,
+         "next": "https://example.com/optimade/v1/structures/id123456?response_fields=cartesian_site_positions&property_ranges=mdstep[101,200,2],particles[1,3,1],xyz[1,3,1]]"
```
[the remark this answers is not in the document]
JPBergsma: I think it is even worse than that, as I wrote on line 1105 that we would now use parenthesis ("(", ")") so there should not be any square brackets.

optimade.rst
JPBergsma: I'll remove it.

JPBergsma:
I guess I should implement the meta property first, although that should not be that much work.
I'll ask Sara and Gian-Marco whether I can prioritize creating an implementation of the ranged properties.

optimade.rst
```diff
@@ -442,6 +442,121 @@ For example, the following query can be sent to API implementations `exmpl1` and
...
+    The indexing is 1 based, so the lowest value an index can have is 1.
+
+  - :field:`step`: integer.
+    If the values are regularly spaced in this dimension, this value indicates the difference in index between subsequent values.
+    If there is no regular spacing between the values, the value of this property MUST be :val:`null`.
+
+  - :field:`stop`: integer.
+    The index of the last value of this property in this dimension.
+
+- :field:`contains_null`: boolean
```
JPBergsma, line 489: Should this not be a more generic metadata property that could also apply to other non ranged fields ?

JPBergsma:
The adding of the partial_data mechanism PR#467 has made this PR largely obsolete, so I will therefore close it for now.

PR #455, opened 2023-02-20
Add `optimization_type` property for structures

rartino:
This is a proposed solution for #406 

Before you fire up your keyboards to comment on all the many ways this can/should be extended to allow other possible aspects representing the full provenance of experimental, computational, and AI-model-created structural information, please realize that this property isn't exactly designed to do that.

This is meant to be "a flag" to address a very real and quite pressing need - which we have experienced ourselves when creating OPTIMADE use examples - to have some relatively simple way to filter out "random noise" when querying the OPTIMADE network for "all structures". Of course, what is random noise for one person is signal for another. I've tried my very best to walk the delicate balance to define *a few* reasonable categories that, based on the discussions in #406, covers actual use cases.

I'd be more than happy to see (and review) a future PR for representing a much more rich account of all the experimental and computational history that led to arriving at some structural information. But, I also believe that representation can co-exist with the "flag" proposed in this PR.

PR #456, opened 2023-02-20
Adjust handling of unknown properties

rartino:
In discussions related to issue #406 we've found that the current handling of unknown properties in OPTIMADE is not optimal for interoperability, as queries using a new property will generate errors in databases that implement an older version of the standard.

This PR is a suggestion for fixing this issue for providers implementing 1.2 and forward. (However, queries to databases that remain on 1.1 will still generate errors)."

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -426,17 +426,16 @@ In particular, filters with :filter-fragment:`IS UNKNOWN` and :filter-fragment:`
 Handling unknown property names
 -------------------------------
 
-When an implementation receives a request with a query filter that refers to an unknown property name it is handled differently depending on the database-specific prefix:
+When an implementation receives a request with a query filter that refers to an unknown property name, it MUST NOT treat this as an error.
+Instead, it should evaluate the query with the property treated as unknown, meaning comparisons are evaluated as if the property has the value :val:`null`.
+Furthermore:
 
-* If the property name has no database-specific prefix, or if it has the database-specific prefix that belongs to the implementation itself, the error :http-error:`400 Bad Request` MUST be returned with a message indicating the offending property name.
+* If the property does not have a database-specific prefix, the implementation MUST issue a warning about the unrecognized property name.
+* If the property has a database-specific prefix that the implementation does not recognize, it SHOULD return a warning indicating that the property has been handled as unknown.
+* If the property has a recognized prefix, i.e., it belongs to a known database provider, the implementation SHOULD NOT issue a warning but may issue diagnostic output with a note explaining how the request was handled.
 
-* If the property name has a database-specific prefix that does *not* belong to the implementation itself, it MUST NOT treat this as an error, but rather MUST evaluate the query with the property treated as unknown, i.e., comparisons are evaluated as if the property has the value :val:`null`.
-
-  * Furthermore, if the implementation does not recognize the prefix at all, it SHOULD return a warning that indicates that the property has been handled as unknown.
-
-  * On the other hand, if the prefix is recognized, i.e., as belonging to a known database provider, the implementation SHOULD NOT issue a warning but MAY issue diagnostic output with a note explaining how the request was handled.
-
-The rationale for treating properties from other databases as unknown rather than triggering an error is for OPTIMADE to support queries using database-specific properties that can be sent to multiple databases.
+The rationale for not triggering errors in the above situations is to enable clients to perform queries that reference properties defined in future versions of the OPTIMADE standard or database-specific properties, and have them handled only by the databases that recognize them.
+Database-specific properties (as well as properties defined in future versions of this standard) SHOULD be defined in such a way that treating them as :val:`null` in comparisons is a reasonable behavior for a provider that do not support the property.
 
 For example, the following query can be sent to API implementations `exmpl1` and `exmpl2` without generating any errors:
 


optimade.rst
```diff
@@ -426,17 +426,16 @@ In particular, filters with :filter-fragment:`IS UNKNOWN` and :filter-fragment:`
...
+* If the property has a database-specific prefix that the implementation does not recognize, it SHOULD return a warning indicating that the property has been handled as unknown.
+* If the property has a recognized prefix, i.e., it belongs to a known database provider, the implementation SHOULD NOT issue a warning but may issue diagnostic output with a note explaining how the request was handled.
 
-The rationale for treating properties from other databases as unknown rather than triggering an error is for OPTIMADE to support queries using database-specific properties that can be sent to multiple databases.
+The rationale for not triggering errors in the above situations is to enable clients to perform queries that reference properties defined in future versions of the OPTIMADE standard or database-specific properties, and have them handled only by the databases that recognize them.
+Database-specific properties (as well as properties defined in future versions of this standard) SHOULD be defined in such a way that treating them as :val:`null` in comparisons is a reasonable behavior for a provider that do not support the property.
```
vaitkus, line 438:
```suggestion
Database-specific properties (as well as properties defined in future versions of this standard) SHOULD be defined in such a way that treating them as :val:`null` in comparisons is a reasonable behavior for providers that do not support the property.
```

ml-evs (commented):
LGTM @rartino, thanks for writing this up.

I have one outstanding question before we merge: we focus here on filtering, but I think `response_fields` has the same problem.

If I ask for `structure_origin` from my own implementation (100% standards compliant, I'm sure...) at the moment (https://optimade.odbx.science/v1/structures?response_fields=structure_origin) then I get an error. Should I instead return `null` for the field (with a warning) after this PR?

ml-evs:
> 



> @ml-evs Excellent point. Actually, I don't see anything in the specification right now about returning or not returning errors for unknown `response_fields`. After looking around, I think this segment of text is the right place for that.
> 
> But, yes, should we force returning `null` + a warning? Or should we allow to _not_ include it in the response even when requested? Both?

I think returning `null` + warning would be fine, and maps to what we have already for known-not-implemented fields. The current wording for another-database-provider-field means that if I request `_random_field` (e.g., https://optimade.odbx.science/v1/structures?response_fields=_random_field&page_limit=1) then I have to return `null` but without a warning right? Do we want a warning in all cases for consistency?

rartino:
> The current wording for another-database-provider-field means that if I request _random_field (e.g., https://optimade.odbx.science/v1/structures?response_fields=_random_field&page_limit=1) then I have to return null but without a warning right? Do we want a warning in all cases for consistency?

I think the thought with suppressing that warning was that it is a somewhat expected situation even with databases all at the latest version, and thus shouldn't trigger warnings. This exception wasn't written by me and I don't feel strongly either way. But, can we remove it now? Strictly speaking this is actually a backwards-breaking change in that a clients written to the older specification do not expect to get a warning in these cases, and may now start breaking.

rartino:
@ml-evs Ok, I've updated the text to handle `response_fields`. Does this seem ok?

ml-evs (dismissed):
Thanks @rartino, this looks clear to me now!

rartino:
@ml-evs Sorry for re-editing this after approval, but I have given this some more thought.

First, I think you are right that we should (and can) remove the differences in handling between "recognized" and "unrecognized" prefixes as they complicate the specification, implementation, and the difference may be confusing for users. I also really didn't like the sentence referring to putting a note in "diagnostic output" since we do not have a place for such output, and it is not good if there is no information whatsoever to the client that an unrecognized property has been given a `null` value. So, first I started looking at introducing a **notifications** list alongside **warnings**, but then, after reading the text carefully, I realized that the formulation that no warning is to be given for unrecognized properties in ~ "others' recognized prefixes" is subjective enough to be meaningless, i.e., existing standards-conformant clients have to already accept warnings here since they cannot know what others' prefixes the implementation has chosen to recognize.

I also realized that to be able to handle the future-proofing properly, we ought to strengthen the interaction between future properties and `structure_features`. So, I have now explicitly clarified how these mechanisms are meant to work together so new properties can be introduced without affecting old data.

Finally, since I'm already editing this quite heavily and "Unknown properties" is a term we use for other things, I've systematically changed the term for this to "unrecognized property names", including renaming the whole section.

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
...
 
-Handling unknown property names
--------------------------------
+Handling unrecognized property names
+------------------------------------
 
-When an implementation receives a request with a query filter that refers to an unknown property name it is handled differently depending on the database-specific prefix:
+When an implementation receives a request with a query filter or other mechanism (e.g., :query-param:`response_fields`) that refers to a property name that is not recognized, it MUST NOT treat this as an error but rather MUST issue a warning with `code` set to `unrecognized_property`.
+Queries MUST be evaluated with the property treated as unknown, meaning comparisons are evaluated as if the property has the value :val:`null`.
+If the unrecognized property is requested to be included in the response, it MUST be included with `null` value.
```
rartino, line 431:
```suggestion
If the unrecognized property is requested to be included in the response, it MUST be included with :val:`null` value.
```

optimade.rst
```diff
@@ -719,7 +716,14 @@ Warnings
 Non-critical exceptional situations occurring in the implementation SHOULD be reported to the referrer as warnings.
 Warnings MUST be expressed as a human-readable message, OPTIONALLY coupled with a warning code.
 
-Warning codes starting with an alphanumeric character are reserved for general OPTIMADE error codes (currently, none are specified).
+Warning codes starting with an alphanumeric character are reserved for general OPTIMADE error codes.
+The following are defined:
+
+- `unrecognized_property`: the request refers to an unrecognized property in the query parameters `filter`, `response_fields`, or in some other way.
+  Implementations SHOULD indicate in the warning field `details` which properties are unrecognized.
+  For future compatability, unrecognized properties are handled as having value :val:`null`.
```
rartino, line 724:
```suggestion
  For future compatibility, unrecognized properties are handled as having value :val:`null`.
```

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
...
-
-The rationale for treating properties from other databases as unknown rather than triggering an error is for OPTIMADE to support queries using database-specific properties that can be sent to multiple databases.
+The rationale for not triggering errors for unrecognized properties is to enable clients to perform queries that reference properties defined in future versions of the OPTIMADE standard or database-specific properties, and have them handled only by the databases that recognize them.
+Database-specific properties (as well as properties defined in future versions of this standard) SHOULD be defined in such a way that treating them as :val:`null` in comparisons is a reasonable behavior for providers that do not support the property.
+If this is not possible, and if the endpoint defines a :property:`<endpoint>_features` property (e.g., :property:`structure_features` for the `structures` endpoint), the implementation MUST define a database-specific identifier and place it in this list for those entries where the distinction between implementations that do and do not support the property is important.
```
ml-evs, line 435: If only we had chosen `structures_features` :smile: `trajectory_features` vs `/trajectories` will be a fun update to explain :sweat_smile:

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
...
+Filters can then be formulated as, e.g., :filter-fragment:`_exmpl_strange_property IS UNKNOWN and structure_features HAS "strangeness"` to suppress results from implementations that do not support :property:`_exmpl_strange_property`.
```
ml-evs, line 436: So are we allowing the property `strangeness` to be completely unconstrained? Should it not follow precisely `strange_property`? e.g.,
```suggestion
Filters can then be formulated as, e.g., :filter-fragment:`_exmpl_strangeness` IS UNKNOWN and structure_features HAS "strangeness"` to suppress results from implementations that do not support :property:`_exmpl_strange_property`.
```

optimade.rst
```diff
@@ -2848,6 +2852,8 @@ structure\_features
   - MUST be sorted alphabetically.
   - If a special feature listed below is used, the list MUST contain the corresponding string.
   - If a special feature listed below is not used, the list MUST NOT contain the corresponding string.
+  - Clients SHOULD NOT assume they can represent the structure correctly if the :field:`structure_features` field indicates a feature not supported by the client or contains any unrecognized items.
+  - Implementations that use database-specific properties that must be supported by the client to correctly interpret the structural information MUST define a string with a database-provider-specific prefix and place it in this list for those entries that would be incorrectly interpreted by ignoring the property.
```
ml-evs, line 2856: The `strangeness` example above does not have such a prefix. To me, it only really makes sense to constrain this tag to the exact field name that has the issue. Is there any reason not to do that?

ml-evs (commented):
Thanks for the important clarifications @rartino! A few minor comments:

optimade.rst
rartino: I chose the names different on purpose as I do not think it is always possible to enforce a machine-understandable mapping between the feature name and the properties and still make them meaningful for humans. For example, our present "disorder" flag in `structure_features` means one or more entries inside `species` has a `chemical_symbols` list that is longer than 1. Do we need to clarify this?

rartino:
@ml-evs Is there anything still blocking the approval? I looked over your final set of comments again, but I don't see anything to act on. You argue that the features flag should be named the same as the field; but to reiterate the argument from above, the point is that a "feature" doesn't need to map cleanly to a single field; it could possibly even involve using keys or values in otherwise defined fields with unusual meaning.

A client that understands what `_exmpl_strangeness` is will know to look for and interpret all the, say, 20 different fields related to describing all aspects of the strangeness. All other clients will just se the black box `_exmpl_strangeness` in `stucture_features` and know to dismiss those structures.

optimade.rst
```diff
@@ -719,7 +716,14 @@ Warnings
 Non-critical exceptional situations occurring in the implementation SHOULD be reported to the referrer as warnings.
 Warnings MUST be expressed as a human-readable message, OPTIONALLY coupled with a warning code.
 
-Warning codes starting with an alphanumeric character are reserved for general OPTIMADE error codes (currently, none are specified).
+Warning codes starting with an alphanumeric character are reserved for general OPTIMADE error codes.
+The following are defined:
+
+- `unrecognized_property`: the request refers to an unrecognized property in the query parameters `filter`, `response_fields`, or in some other way.
+  Implementations SHOULD indicate in the warning field `details` which properties are unrecognized.
```
vaitkus, line 723: The 
```suggestion
  Implementations SHOULD indicate in the warning field `detail` which properties are unrecognized.
```

vaitkus: The description of the warning object (line 534) seems to use the singular `detail` instead of `details`.

optimade.rst
ml-evs: Fair point, following your explanation below. Happy to resolve these

optimade.rst
```diff
@@ -719,7 +716,14 @@ Warnings
 Non-critical exceptional situations occurring in the implementation SHOULD be reported to the referrer as warnings.
 Warnings MUST be expressed as a human-readable message, OPTIONALLY coupled with a warning code.
 
-Warning codes starting with an alphanumeric character are reserved for general OPTIMADE error codes (currently, none are specified).
+Warning codes starting with an alphanumeric character are reserved for general OPTIMADE error codes.
+The following are defined:
+
+- `unrecognized_property`: the request refers to an unrecognized property in the query parameters `filter`, `response_fields`, or in some other way.
```
ml-evs, line 722:
```suggestion
- :warning-code:`unrecognized_property`: the request refers to an unrecognized property in the query parameters :query-parameter:`filter`, :query_parameter:`response_fields`, or in some other way.
```

I think this is needed for the formatting, perhaps elsewhere too (will check)

optimade.rst
```diff
@@ -719,7 +716,14 @@ Warnings
...
+  Implementations SHOULD indicate in the warning field `detail` which properties are unrecognized.
```
ml-evs, line 723:
```suggestion
  Implementations SHOULD indicate in the warning field :property:`detail` which properties are unrecognized.
```

optimade.rst
ml-evs, line 435:
```suggestion
If this is not possible, and if the endpoint defines a :property:`<endpoint>_features` property (e.g., :property:`structure_features` for the :endpoint:`structures` endpoint), the implementation MUST define a database-specific identifier and place it in this list for those entries where the distinction between implementations that do and do not support the property is important.
```

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
 The interaction of properties with an unknown value with query filters is described in the section `Filtering on Properties with an unknown value`_.
 In particular, filters with :filter-fragment:`IS UNKNOWN` and :filter-fragment:`IS KNOWN` can be used to match entries with values that are, or are not, unknown for some property, respectively.
 
-Handling unknown property names
--------------------------------
+Handling unrecognized property names
+------------------------------------
 
-When an implementation receives a request with a query filter that refers to an unknown property name it is handled differently depending on the database-specific prefix:
+When an implementation receives a request with a query filter or other mechanism (e.g., :query-param:`response_fields`) that refers to a property name that is not recognized, it MUST NOT treat this as an error but rather MUST issue a warning with `code` set to `unrecognized_property`.
```
ml-evs, line 429:
```suggestion
When an implementation receives a request with a query filter or other mechanism (e.g., :query-param:`response_fields`) that refers to a property name that is not recognized, it MUST NOT treat this as an error but rather MUST issue a warning with :field:`code` set to :warning-code:`unrecognized_property`.
```

ml-evs (changes requested):
Purely formatting changes -- I think these are uncontroversial so will commit them and then add an additional commit that adds `:warning-code:` to the registered tags.

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
...
-
-  * Furthermore, if the implementation does not recognize the prefix at all, it SHOULD return a warning that indicates that the property has been handled as unknown.
-
-  * On the other hand, if the prefix is recognized, i.e., as belonging to a known database provider, the implementation SHOULD NOT issue a warning but MAY issue diagnostic output with a note explaining how the request was handled.
-
-The rationale for treating properties from other databases as unknown rather than triggering an error is for OPTIMADE to support queries using database-specific properties that can be sent to multiple databases.
+The rationale for not triggering errors for unrecognized properties is to enable clients to perform queries that reference properties defined in future versions of the OPTIMADE standard or database-specific properties, and have them handled only by the databases that recognize them.
+Database-specific properties (as well as properties defined in future versions of this standard) SHOULD be defined in such a way that treating them as :val:`null` in comparisons is a reasonable behavior for providers that do not support the property.
```
JPBergsma, line 434: Should we also place this line in section [database-provider-specific properties](https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst#id66), as that is the most logical place to look when you want to create a provider specific field?

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
...
+If this is not possible, and if the endpoint defines a :property:`<endpoint>_features` property (e.g., :property:`structure_features` for the :endpoint:`structures` endpoint), the implementation MUST define a database-specific identifier and place it in this list for those entries where the distinction between implementations that do and do not support the property is important.
+Filters can then be formulated as, e.g., :filter-fragment:`_exmpl_strange_property IS UNKNOWN and structure_features HAS "_exmpl_strangeness"` to suppress results from implementations that do not support :property:`_exmpl_strange_property`.
```
JPBergsma, line 436: To me, this section is still unclear.
Wouldn't the query `structure_features HAS "_exmpl_strangeness"` be enough to suppress results from implementations that do not support `_exmpl_strange_property`? I think that with the example you give here, you specifically want to select entries for which the field is null but whose database does support this field.

How should a client know that which value under structure features corresponds to a property ?

I am also wondering whether we really need this mechanism. Are there examples where the value null has a meaning different from unknown?
At the end of section 2 of the specification we now write: "An entry property value that is not present in the database is unknown. This is equivalently expressed by the statement that the value of that entry property is null."
So I do not see how null can mean something other than unknown.

rartino: > Wouldn't the query `structure_features HAS "_exmpl_strangeness"` be enough to suppress results from implementations that do not support `_exmpl_strange_property`? I think that with the example you give here, you specifically want to select entries for which the field is null but whose database does support this field. 

In the usual case you do not have to suppress results from implementations that do not support `_exmpl_strange_property`. A filter like `filter=nelements>3 AND _exmpl_strangeness>12` will not match any structures from databases that do not support "strangeness". *In the usual case* this is going to be the right thing.

The text right above the one you quite says that new fields should as far as possible be defined this way. Hence, the case we are discussing here is about a special case: a database has defined a field like `_exmpl_strangeness`  in such a way that a database that do not support it would return incorrect results when handling it as `null` in filters. 

So, for example, a database with no idea about "strangeness" would interpret `filter=nelements>3 AND _exmpl_strangeness IS UNKNOWN` to return all structures with more than 3 elements. If that is not what you want, you have to write the filter as: `_exmpl_strange_property IS UNKNOWN and structure_features HAS "_exmpl_strangeness"`

> How should a client know that which value under structure features corresponds to a property ?

Compare with  `assemblies`. How does a client know to look for the absence of `"assemblies"` in `structure_features` to find structures that do not have assemblies? They know it from the specifications / field definitions. 

>  So I do not see how null can mean something other than unknown.

Well, it isn't a perfect example in this context, but before I protested and reverted that edit, we were very near to define `structure_origin` that way in #455; with `null`/UNKNOWN being ambiguous between `structure_origin` not being supported, and that the origin of the structure is known to be indeterminate.

Remember that this isn't just for database-specific fields, but also for future fields we come up with as part of the standard. I'll see if I can come up with a more clear cut example. But if you really feel that strongly against this special case being handled in the specification, I guess we can simply remove the quoted text. Should we do that?

optimade.rst
rartino: This is now done. Please re-review.

optimade.rst
JPBergsma: > So, for example, a database with no idea about "strangeness" would interpret filter=nelements>3 AND _exmpl_strangeness IS UNKNOWN to return all structures with more than 3 elements. If that is not what you want, you have to write the filter as: _exmpl_strange_property IS UNKNOWN and structure_features HAS "_exmpl_strangeness"

Perhaps it is better to formulate this positively, like:
Filters can then be formulated as, e.g., :filter-fragment:`_exmpl_strange_property IS UNKNOWN and structure_features HAS "_exmpl_strangeness"` to request entries where `_exmpl_strange_property` has explicitly been defined as `null`.

> Well, it isn't a perfect example in this context, but before I protested and reverted that edit, we were very near to define structure_origin that way in https://github.com/Materials-Consortia/OPTIMADE/pull/455; with null/UNKNOWN being ambiguous between structure_origin not being supported, and that the origin of the structure is known to be indeterminate.

> Remember that this isn't just for database-specific fields, but also for future fields we come up with as part of the standard. I'll see if I can come up with a more clear cut example. But if you really feel that strongly against this special case being handled in the specification, I guess we can simply remove the quoted text. Should we do that?

I kind of like the quoted text. That null is equivalent to unknown. So I do not want to remove it.
The suggested query could perhaps be used to distinguish between entries for which a property is implicitly or explicitly defined as null. 
But I think it would be more elegant in that case to add a query parameter to determine whether a field is null or missing,
rather than have the addhock method of adding a field to structure_features.  

I also, just stumbled on a thread on handling NAN and Inf values in numerical data. It seems the JSON specification cannot handle these, and some hacky solutions people have proposed the use of `null`.   
So people may go for such a solution for optimade fields as well.

I do not think that that is a good idea, so I think we should define in the optimade specification how we should handle these cases.

merkys: I think I am with @JPBergsma on this. First of all, I do not quite understand how this is related to the filtering on unrecognized properties.

Second, I would like to see a real-life use case here. Could something like unmodeled solvent be it? In short, certain experimental structures lack some atom sites just because their coordinates cannot be measured. Instead they are modeled by a cloud of continuous electron distributions. Any DFT code not capable of defining and using such clouds would tear such structures apart. Thus adding `_cod_unmodeled solvent` to `structure_features` would signal clients to ignore such structures unless they know how to handle them. Does this sound right?

Third, imposing requirements on future/custom properties (`<entry type>_features` here) has been frowned upon before, and I think not without a reason. Could we just explicitly list `structure_features` for now and extend it for every `<entry type>_features` we define later?

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
 The interaction of properties with an unknown value with query filters is described in the section `Filtering on Properties with an unknown value`_.
 In particular, filters with :filter-fragment:`IS UNKNOWN` and :filter-fragment:`IS KNOWN` can be used to match entries with values that are, or are not, unknown for some property, respectively.
 
-Handling unknown property names
--------------------------------
+Handling unrecognized property names
+------------------------------------
 
-When an implementation receives a request with a query filter that refers to an unknown property name it is handled differently depending on the database-specific prefix:
+When an implementation receives a request with a query filter or other mechanism (e.g., :query-param:`response_fields`) that refers to a property name that is not recognized, it MUST NOT treat this as an error but rather MUST issue a warning with :field:`code` set to :field-val:`unrecognized_property`.
```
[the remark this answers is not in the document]
rartino: Yes, this change is the most vital part of this PR to improve forwards compatibility in OPTIMADE. 

The point is, when we add, e.g., `spacegroup` to the standard, and you send a query to two databases asking to include `spacegroup` among the response fields (or referencing it in the filter), you do not want an error; you want it treated as `null`.

optimade.rst
```diff
@@ -423,20 +423,17 @@ OPTIONAL properties with an unknown value, if requested explicitly via the :quer
...
+Database-specific properties (as well as properties defined in future versions of this standard) SHOULD be defined in such a way that returning them with the value :val:`null` and treating them as :val:`null` in comparisons is a reasonable behavior for providers that do not support the property.
+If this is not possible, and if the endpoint defines a :property:`<endpoint>_features` property (e.g., :property:`structure_features` for the :endpoint:`structures` endpoint), the implementation MUST define a database-specific identifier and place it in this list for those entries where the distinction between implementations that do and do not support the property is important.
+Filters can then be formulated as, e.g., :filter-fragment:`_exmpl_strange_property IS UNKNOWN and structure_features HAS "_exmpl_strangeness"` to suppress results from implementations that do not support :property:`_exmpl_strange_property`.
```
rartino, line 436:
```suggestion
```
```suggestion
If this is not possible, and if the endpoint defines a :property:`<endpoint>_features` property (e.g., :property:`structure_features` for the :endpoint:`structures` endpoint), the implementation MUST define a database-specific identifier and place it in this list for those entries where the distinction between implementations that do and do not support the property is important.
Filters can then be formulated as, e.g., :filter-fragment:`_exmpl_strange_property IS UNKNOWN and structure_features HAS "_exmpl_strangeness"` to suppress results from implementations that do not support :property:`_exmpl_strange_property`.
```

rartino: IMO it may be a colossal mistake for forwards compatibility of OPTIMADE to take this out, but since it seems to be the blocker here, lets skip it for now and get back to this.

optimade.rst
rartino: @merkys @JPBergsma Are you OK with the updated version where I have taken (most of) this out? There is still a small clarification about the features list left in, but which I think is less controversial and more clear what it tries to achieve.

rartino:
Workshop: no agreement if this is the right protocol for futureproofing requestes.

rartino:
@ml-evs: I suggest we reiterate this discussion at the release web meet? We need to do *something* to improve forward compatibility. Perhaps some general "tags" field in all endpoints to do something like what you suggested for structure_features.

rartino:
After some discussions: lets break out the way for old servers to handle references to unknown fields from newer clients and see if we can get consensus for that.

The handling with structure_features for older clients to handle newer servers (with fields that alter the data interpretation) can instead use the major version number.

optimade.rst
```diff
@@ -2856,6 +2858,9 @@ Custom properties
   Similarly, an implementation MAY add keys with a namespace prefix to dictionary properties and their sub-dictionaries.
   For example, the database-provider-specific property :property:`_exmpl_oxidation_state`, can be placed within the OPTIMADE property :property:`species`.
 
+  Database-specific properties SHOULD be defined in such a way that returning them with the value :val:`null` and treating them as :val:`null` in comparisons is a reasonable behavior for providers that do not support the property.
+  See `Handling unrecognized property names`_ for more information.
+
```
sauliusg: This is a bit ambiguous to me. How do we expect to treat 'null' values in comparisons? There are multiple possible definitions. SQL reportedly uses a "fragment of the Kleene K3" (https://en.wikipedia.org/wiki/Three-valued_logic#SQL) for AND, OR and NOT operations; "1 = NULL" is neither true (1) nor false (0) and so on. Does OPTIMADE specify operations to this degree of detail? Are we happy to specify that we "use the SQL logic"? How do other, especially non-SQL databases, handle this?

PR #445, opened 2022-12-22
Add machine-readable property definitions to repository

rartino:
This PR adds machine-readable property definitions for all the OPTIMADE standardized properties using the format we developed over the last couple of workshops.

Old PR text:

~~This PR has started the work of adding property definitions to the repository on the format that is now in the specification.~~

~~I've opened it as a pull request even though it is not yet finished to:~~
~~- Make it more visible for discussions about the format.~~
~~- To enable people to look things over while definitions are being added.~~
~~- Possibly get help from others ("maintainers" should be able to push to the PR branch of my repo)~~ 

~~Note: the referenced command: "make properties" is not yet in place.~~

Issue #442, opened 2022-12-22
Replace property definitions x-optimade-property/property-uri with just an outer "$id" key?


rartino:
JSON schema apparently has a standard feature to indicate a URI for the schema, which is to have a key "$id". It seems to me this could replace our own identifier x-optimade-property/property-uri. Should it?

ml-evs:
I'd be fine with this. What are we expecting this field to be set to for OPTIMADE properties? e.g., `https://github.com/Materials-Consortia/OPTIMADE/blob/v1.1/optimade.rst#chemical-formula-descriptive`?

With a bit of encouragement I could continue with #378 and get it hosted at `optimade.org` for a nice human-readable/linkable spec.

JPBergsma:
I am closing this issue as PR #457 has been merged.
Or do you feel your question has not been answered sufficiently yet ? @ml-evs

PR #457, opened 2023-02-20
Changes to property definitions

rartino:
Some minor adjustments to our property definitions according to what has been discussed in issues:

- Change `x-optimade-property/property-uri` into simply `$id`, which is the JSON Schema standard way of communicating the same thing, see #442.
- Add a field `x-optimade-type` to specify the OPTIMADE data type. It was concluded in #392, #436 that this is necessary if we do not want to make things very complicated when/if we want/need to support more complex data types.

The definition of the `type` field has thus been completely rewritten to take into account the quite direct relationship to `x-optimade-property`.

(Note: I mentioned earlier that I would take this in #445, but upon realizing that I had to do rather significant edits here, I decided not to mix the two up.)

**Edit:** Oh, and this PR also fixes a bug in that the mandatory `x-optimade-property/property-format` was not included in the response example.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1872,7 +1872,7 @@ A Property Definition MUST be composed according to the combination of the requi
 **REQUIRED keys for all levels of the Property Definition:**
 
 - :field:`x-optimade-type`: String
-  Specifies the OPTIMADE data type, i.e., MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, or :val:`"dictionary"`.
+  Specifies the OPTIMADE data type for this level of the defined property, i.e., MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, or :val:`"dictionary"`.
 
 - :field:`x-optimade-unit`: String.
   A (compound) symbol for the physical unit in which the value of the defined property is given or one of the strings :val:`dimensionless` or :val:`inapplicable`.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1941,33 +1945,41 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
 ..
 
   Implementation notes:
 
-    - The strings used in the :field:`type` field are JSON type names encoded as strings, but they refer to the corresponding OPTIMADE data types.
-      Nevertheless, for consistency across formats, the JSON type names MUST be used regardless of the standard type names of the output format.
-      The motivation for this design decision is that it makes the JSON representation of a Property Definition a fully valid standard JSON Schema.
+    - The field :field:`type` can be derived from the field :field:`x-optimade-type` and its role is only to provide the JSON type names corresponding to :field:`x-optimade-type`.
+      The motivation to include these type names is that it makes the JSON representation of a Property Definition a fully valid standard JSON Schema.
+      Nevertheless, for consistency across formats, these JSON type names MUST still be included when a property definition is represented in other output format (i.e., the JSON names MUST NOT be translated into the type names of that output format).
```
vaitkus, line 1970:
```suggestion
      Nevertheless, for consistency across formats, these JSON type names MUST still be included when a property definition is represented in other output formats (i.e., the JSON names MUST NOT be translated into the type names of that output format).
```

optimade.rst
```diff
@@ -1941,33 +1945,41 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
-  - timestamps are represented by setting the :field:`type` field to :val:`"string"` and the :field:`format` field to :val:`"date-time"`.
-    In this case it is MANDATORY to include the field :field:`format`.
+    * :val:`"boolean"`, `"string"`, or `"integer"` then :field:`type` is the same string.
+    * :val:`"dictionary"` then :field:`type` is `"object"`.
+    * :val:`"list"` then :field:`type` is `"array"`.
+    * :val:`"float"` then :field:`type` is `"number"`.
+    * :val:`"timestamp"` then :field:`type` is `"string"`.
 
-  Output formats that represent these OPTIMADE data types in other ways have to recognize them and reinterpret the definition accordingly.
+  - A list where the first item MUST be the string above (correlated to the field :field:`x-optimade-type` in the same way) and the second item MUST be the string :val:`"null"`.
```
vaitkus, line 1962:
```suggestion
  - A list where the first item MUST be the string described above (correlated to the field :field:`x-optimade-type` in the same way) and the second item MUST be the string :val:`"null"`.
```

vaitkus, line 1962: What are the use cases for this list?

optimade.rst
```diff
@@ -1870,6 +1871,9 @@ A Property Definition MUST be composed according to the combination of the requi
 
 **REQUIRED keys for all levels of the Property Definition:**
 
+- :field:`x-optimade-type`: String
+  Specifies the OPTIMADE data type for this level of the defined property, i.e., MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, or :val:`"dictionary"`.
```
vaitkus, line 1875:
```suggestion
  Specifies the OPTIMADE data type for this level of the defined property.
  MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, or :val:`"dictionary"`.
```

optimade.rst
```diff
@@ -1823,7 +1829,7 @@ A Property Definition MUST be composed according to the combination of the requi
 
 **REQUIRED keys for the outermost level of the Property Definition:**
 
-- :field:`title`: String and :field:`description`: String.
+- :field:`$id`, :field:`title`: String, and :field:`description`: String.
```
vaitkus, line 1832:
```suggestion
- :field:`$id`, :field:`title`: and :field:`description`: String.
```

rartino:
```suggestion
- :field:`$id`: String, :field:`title`: String, and :field:`description`: String.
```

optimade.rst
```diff
@@ -1244,22 +1248,24 @@ Example (note: the description strings have been wrapped for readability only):
...
-                 "x-optimade-requirements": {
-                   "sortable": false,
-                   "query-support": "none"
-                 }
-               }
+              "type": "array",
+              "x-optimade-type": "list",
+              "x-optimade-unit": "inapplicable",
+              "maxItems": 3
+              "minItems": 3
```
vaitkus, line 1255:
```suggestion
              "maxItems": 3,
              "minItems": 3,
```

optimade.rst
```diff
@@ -1244,22 +1248,24 @@ Example (note: the description strings have been wrapped for readability only):
             "maxItems": 3
             "minItems": 3
```
vaitkus, line 1249:
```suggestion
            "maxItems": 3,
            "minItems": 3,
```

vaitkus: Also, a comma should be added on line 1247, but I cannot seem to suggest a change there.

optimade.rst
```diff
@@ -1244,22 +1248,24 @@ Example (note: the description strings have been wrapped for readability only):
...
+              "type": "array",
+              "x-optimade-type": "list",
+              "x-optimade-unit": "inapplicable",
+              "maxItems": 3
+              "minItems": 3
+              "items": {
+                "type": "number",
+                "x-optimade-type": "float",
+                "x-optimade-unit": "angstrom",
+              "x-optimade-implementation": {
```
merkys, line 1260: Whitespace seems off

merkys (commented):
Looks good to me, I will approve once @vaitkus's suggestions are accepted.

optimade.rst
merkys: I think this is the way to specify that value may be null.

vaitkus: Thanks for the clarification, Would it make sense to add a sentence that explains this usage or is this obvious enough?

merkys: I believe this is similar to `format` in JSON Schema, but I cannot find a reference, thus I may be wrong. Since this is not *directly* taken from JSON Schema, I would support adding a sentence about the use case.

rartino: Good point - I've added a sentence clarifying this.

merkys (approved):
Looks good, thanks.

vaitkus (approved):
Thank you for the changes.

rartino:
Thanks @vaitkus @merkys for the proofreading!

Issue #459, opened 2023-02-23
Add support for $comment in property definitions


rartino:
JSON schema has in draft 7 added support to put comments in schemas via $comment: http://json-schema.org/understanding-json-schema/reference/generic.html#comments

When working on our property definitions in #445 I independently realized it would be useful to insert comments in the property definitions. So we should add this to the subset of JSON Schema that we allow.

merkys:
What would be the target audience of comments? Maybe we can just put them to `description`?

rartino:
It would be for comments that are relevant to property definition "developers" rather than "end users". Another difference is that you can put it on any level in a nested definition to comment e.g. the motivation behind setting a limit like "maximum" to something. I stumbled on the desire to insert a comment when I realized I wanted somewhere to explain why `common/element.yaml` doesn't have an `x-optimade-property` (it is because it isn't itself part of the OPTIMADE standard properties, it is just used as a target for '$ref' when we compose other standard properties.)

merkys:
OK, this makes sense.

ml-evs:
Closed by #457 (I think) or a related PR

PR #460, opened 2023-02-27
[WIP] Further updates to property definitions

rartino:
In the work on #445 a few more things have come up in need of adjustments in the specification.

This PR adopts `$comment` from JSON Schema to address #459. Even if I, in the work on definitions, ended up not needing it, I foresee possible future uses and it is a trivial extension.

This PR bundles a potentially controversial change of something that has irked me more and more when working on the definitions. The `type` field we inherit from JSON Schema is AFAIK the *only* place in the OPTIMADE standard where a dictionary field may have multiple data types, either String or List. Hence, I'm going to propose that we force Property Definitions to always use the list format. *Most* of our properties anyway allow `"null"` and are thus not affected. For the few that don't, it may seem silly to force ["string"] over "string"; but having started the work on converting these definitions  between various formats in #445, I really think the simplification of the representation format is worth it.

Further changes in this PR are:

- The addition of `x-optimade-requirements/response-default-level` to specify to what degree a property should or should not be included in a response per default. We have this information embedded in some of our current property definitions on this format: "Response: REQUIRED in the response."

- A corresponding addition in `x-optimade-implementation/response-default` for implementations to indicate what a specific implementation does for each field. (I had to name it differently, since the latter should be a boolean and the former a string.)

When I was adding these, I realized we need to clarify what is, and what is not, "the same" in terms of Property Definitions, so this is also done. Feel free to complain (and/or suggest fixes) if these formulations are not clear. They are, however, IMO absolutely important.

I should say I was initially a bit hesitant if the "default response" info really belong in the Property Definitions. However, after some thinking, I really believe they do. When groupings of people uses this format to "develop standards" e.g. for biomol, experimental, magnetism, etc., databases they will want to be able to say what a database is expected to return by "default", just as we wanted for the basic properties we have standardized so far.


rartino:
For some reason I cannot request a review from @fekad via the Reviewers system, so I ping him here instead. Since you gave much good feedback on the initial property definitions, and I suspect you may have thoughts on the `type` field change, I'd be happy if you have time to take a look at this.

optimade.rst
```diff
@@ -1884,6 +1884,10 @@ A Property Definition MUST be composed according to the combination of the requi
 - :field:`x-optimade-implementation`: Dictionary.
   A dictionary describing the level of OPTIMADE API functionality provided by the present implementation.
   If an implementation omits this field in its response, a client interacting with that implementation SHOULD NOT make any assumptions about the availability of these features.
+
+  This field, along with a few others, should be seen as annotations rather then integral parts of the Property Definition.
```
vaitkus, line 1888:
```suggestion
  This field, along with a few others, should be seen as annotations rather than integral parts of the Property Definition.
```

vaitkus: Also, I would probably rephrase the sentence to omit the `along with a few others` part and only focus on this definition. Since we do not explicitly list or reference those other definitions, it is probably of little use to the reader. The same comment applies to a few other newly added sentences with a similar structure.

optimade.rst
```diff
@@ -1918,14 +1926,23 @@ A Property Definition MUST be composed according to the combination of the requi
 
     - :val:`IS KNOWN`, :val:`IS UNKNOWN`: indicating support for filtering this property on unknown values using the respective operator.
 
+  - :field:`response-default`: boolean
```
vaitkus, line 1929:
```suggestion
  - :field:`response-default`: Boolean
```

optimade.rst
```diff
@@ -1918,14 +1926,23 @@ A Property Definition MUST be composed according to the combination of the requi
...
+    The value :val:`TRUE` means the implementation includes the property in responses by default, i.e., when not specifically requested.
```
vaitkus, line 1930: It seems that other Boolean definitions also explicitly define the semantics of the "FALSE" value.

optimade.rst
```diff
@@ -1918,14 +1926,23 @@ A Property Definition MUST be composed according to the combination of the requi
...
+    Omitting the field or :val:`null` means the implementation does not declare if the property will be included in responses by default or not.
```
vaitkus, line 1931: Could/should it default to FALSE instead under these conditions? Or does it clash with backwards compatibility?

optimade.rst
```diff
@@ -1937,6 +1954,21 @@ A Property Definition MUST be composed according to the combination of the requi
     Note: the specification by this field of whether the defined property can be :val:`null` or not MUST match the value of the :field:`type` field.
     If :val:`null` values are allowed, that field must be a list where the string :val:`"null"` is the second element.
 
+  - :field:`response-default-level`: String
+    Expresses if an implementation of this property is required or not to include or exclude it in responses when not specifically requested.
```
vaitkus, line 1958:
```suggestion
    Expresses if an implementation of this property is required to include or exclude it in responses when not specifically requested.
```

optimade.rst
```diff
@@ -1945,25 +1977,20 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
-- :field:`type`: String or List.
+- :field:`type`: List.
   Specifies the corresponding JSON type for this level of the defined property and whether the property can be :val:`null` or not.
   The value is directly correlated with :field:`x-optimade-type` as explained below.
 
-  It MUST be one of:
-
-  - A string correlated with :field:`x-optimade-type` as follows.
-    If :field:`x-optimade-type` is:
+  It MUST be a one or two element list where the first element is a string correlated with :field:`x-optimade-type` as follows; if :field:`x-optimade-type` is:
```
vaitkus, line 1984:
```suggestion
  It MUST be a list of one or two elements where the first element is a string correlated with :field:`x-optimade-type` as follows; if :field:`x-optimade-type` is:
```

optimade.rst
```diff
@@ -1990,11 +2018,23 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
   The format SHOULD be a one-line description, followed by a new paragraph (two newlines), followed by a more detailed description of all the requirements and conventions of the defined property.
   Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__.
 
+- :filed:`$comment`: String.
+  A human-readable comment relevant in the context of the raw definition data.
+  These comments should be omitted in formatted descriptions of the property shown to end users.
```
vaitkus, line 2023:
```suggestion
  These comments should be omitted in formatted descriptions of the property shown to the end users.
```

optimade.rst
```diff
@@ -1990,11 +2018,23 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
+  (Comments relevant in descriptions for end users should go into :field:`description`.)
```
vaitkus, line 2024:
```suggestion
  (Comments in descriptions relevant to the end users should go into :field:`description`.)
```

optimade.rst
```diff
@@ -1918,14 +1926,23 @@ A Property Definition MUST be composed according to the combination of the requi
...
+
 - :field:`x-optimade-requirements`: Dictionary.
   A dictionary describing the level of OPTIMADE API functionality required by all implementations of this property.
   Omitting this field means the corresponding functionality is OPTIONAL.
-  The dictionary has the same format as :field:`x-optimade-implementation`, except that it also allows the following OPTIONAL field:
+  The dictionary has the same format as :field:`x-optimade-implementation`, except that:
+
+  - :field:`response-default` SHOULD NOT appear.
+
+  and the following additional OPTIONAL fields are allowed:
```
vaitkus, line 1940: A bit of unconventional syntax here since we follow `appear.` with a lowercase `and`. Not sure how to rephrase it, though.

optimade.rst
```diff
@@ -1884,6 +1884,10 @@ A Property Definition MUST be composed according to the combination of the requi
 - :field:`x-optimade-implementation`: Dictionary.
   A dictionary describing the level of OPTIMADE API functionality provided by the present implementation.
   If an implementation omits this field in its response, a client interacting with that implementation SHOULD NOT make any assumptions about the availability of these features.
+
+  This field, along with a few others, should be seen as annotations rather then integral parts of the Property Definition.
+  Two Property Definitions that only differ by the value of :field:`x-optimade-implementation` are considered the same, and as explained in relation to the :field:`$id` field below, they SHOULD share the same :field:`$id`.
```
vaitkus, line 1889: Is there a way to reference/link a specific part of the text that defines the "`as explained in relation to the :field:$id field below`"? I tried searching for the "below" part, but could not easily locate it with a cursory glance.

ml-evs (commented):
To limit slosh/churn I will wait until Antanas' review has been addressed before taking a look, please re-request my review after that

optimade.rst
rartino:
```suggestion
    The value :val:`TRUE` means the implementation includes the property in responses by default, i.e., when not specifically requested.
    The value :val:`FALSE` means that the property is only included when requested.
```

optimade.rst
rartino: My thinking was that it could not mean either TRUE or FALSE, because that would mean that leaving this *optional* field out would make you standards-violating if you do the opposite. Meaning, it wouldn't really be optional to declare this, you'd have to do it in cases where the default is wrong.

optimade.rst
rartino, line 1940:
```suggestion
  The dictionary has the same format as :field:`x-optimade-implementation`, *except that* the :field:`response-default` SHOULD NOT appear, and the following additional OPTIONAL fields are allowed:
```

optimade.rst
```diff
@@ -1990,11 +2018,23 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
   The format SHOULD be a one-line description, followed by a new paragraph (two newlines), followed by a more detailed description of all the requirements and conventions of the defined property.
   Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__.
 
+- :filed:`$comment`: String.
+  A human-readable comment relevant in the context of the raw definition data.
+  These comments should be omitted in formatted descriptions of the property shown to end users.
+  (Comments relevant in descriptions for end users should go into :field:`description`.)
+  Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__.
+
+  This field, along with a few others, should be seen as annotations rather then integral parts of the Property Definition.
```
rartino, line 2027:
```suggestion
  This field should be seen as an annotation rather then an integral part of the Property Definition.
```

optimade.rst
vaitkus: I see. But the false-if-not-defined approach is already used with other *optional* boolean fields such as `sortable`. The default could also be wrong in those cases, albeit with slightly lesser side-effects.  Or is this case semantically different?

rartino: I think it is different, in that it just means that this declaration should only be read as the provider not *promising* you can sort on the property. It isn't meant to be standards violating to still allow a sort in that case. Here, however, FALSE means it isn't included by default, so if you do, you break that contract.

I suppose we could reformulate the definition of this boolean instead, to say that FALSE means either. But it is actually useful to know if a field definitively is not going to be included in the response.

vaitkus: Ok, so the real problem is that a reasonable default does not exist. However, the inclusion of the third "unknown" state quite limits the usability of the fields: even if one can say that certain properties will definitely be or not be there, the ones with unspecified `response-default` would still need to be manually excluded/include just to be safe one is getting the desired result.

Of course, this depends on the use case in mind. My initial impression was that this field should allow to formulate a complete expectation on what fields will be returned in the 
response by default. However, I understand that the intended purpose might be different. The use case that I described would probably require to declare the option as mandatory rather than option (something that we probably do not want to do?).

As for the FALSE value, I think that the current strict definition should be retained as it conveys more information.

optimade.rst
```diff
@@ -1990,11 +2015,23 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
   The format SHOULD be a one-line description, followed by a new paragraph (two newlines), followed by a more detailed description of all the requirements and conventions of the defined property.
   Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__.
 
+- :field:`$comment`: String.
+  A human-readable comment relevant in the context of the raw definition data.
+  These comments should be omitted in formatted descriptions of the property shown to the end users.
+  (Comments in descriptions relevant to the end users should go into :field:`description`.)
+  Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__.
+
+  This field should be seen as an annotation rather then an integral part of the Property Definition.
```
JPBergsma, line 2024:
```suggestion
  This field should be seen as an annotation rather than an integral part of the Property Definition.
```

optimade.rst
rartino: > However, the inclusion of the third "unknown" state quite limits the usability of the field: even if one can say that certain properties will definitely be or not be there, the ones with unspecified response-default would still need to be manually excluded/include just to be safe one is getting the desired result.

I think the assumption here is that a reasonable provider provides this for all the usual fields that you care about, and it is missing only on fields you are likely not to care about. In that case, it is useful. 

The typical use case for "I don't want to say" is when testing things. Lets say you are in the process of adding new information to the database, you've made a "stub" property definition to cover it, but not yet decided it should go in the default responses or not. This seems a reasonable use case for "null".

optimade.rst
rartino: Ok, I've added internal hyperlinks for this, and also to reference the definition of `x-optimade-type` from `type`.

rartino:
@ml-evs Ok, I think I've addressed everything from @vaitkus now, except the ongoing discussion we have about the default for `response-default`.

optimade.rst
vaitkus: Understood. I would probably use a different approach of not serving "stub" definitions by default and then moving on to serving them once ready, but I do understand that different people use different practices. So unless you see any benefits of declaring this field as mandatory, I consider the discussion resolved.

vaitkus (approved):
Thank you for the changes.

rartino:
So, another thing came up as I was working on the property definitions yesterday. It turns out that, since I don't want to duplicate my unit definitions in the source files, I more or less automatically also get definition files also for *units*. 

Hence, it is going to be trivial to do the same thing for physical units as we are about to do for properties, i.e., give them URIs that are URLs under `https://schemas.optimade.org/units/...`, that point at the unit definitions using the format described in the OPTIMADE specification section 7.2.

I've set things up for this in #445. For example, YAML source files for the angstrom and meter:

- https://github.com/rartino/OPTIMADE/blob/add-property-definitons-3/properties/src/units/current/atomistic/angstrom.yaml
- https://github.com/rartino/OPTIMADE/blob/add-property-definitons-3/properties/src/units/current/si/m.yaml

This actually seems very useful in itself, in that it provides a way in any context to reference our precise unit definitions via this URI.

However, this motivates a few additions I'd like to do the Unit definitions format:
- Allowing an `$id` field also inside the unit definition.
- Allowing `version` and `property-format`, and make `property-format` MANDATORY if the unit definition is not embedded inside a property definition.
- Add another tag `defining-relation` that allows us to describe how the units we define relate to others (primarily to relate them to SI units).

Any thoughts? Given that this PR isn't yet merged, can I add these changes also to this PR, or should it go in its own?

ml-evs (changes requested):
A few minor comments @rartino. Might make sense to also accept further contributions to this PR if that is easier for you. Will try to respond your most recent comment outside of this particular review.

ml-evs:
> So, another thing came up as I was working on the property definitions yesterday. It turns out that, since I don't want to duplicate my unit definitions in the source files, I more or less automatically get definition files also for _units_.
> 
> Hence, it is going to be trivial to do the same thing for physical units as we are about to do for properties, i.e., give them URIs that are URLs under `https://schemas.optimade.org/units/...`, that point at the unit definitions using the format described in the OPTIMADE specification section 7.2.
> 
> I've set things up for this in #445. For example, YAML source files for the angstrom and meter:
> 
>     * https://github.com/rartino/OPTIMADE/blob/add-property-definitons-3/properties/src/units/current/atomistic/angstrom.yaml
> 
>     * https://github.com/rartino/OPTIMADE/blob/add-property-definitons-3/properties/src/units/current/si/m.yaml
> 
> 
> This actually seems very useful in itself, in that it provides a way in any context to reference our precise unit definitions via these URIs.
> 
> However, this motivates a few additions I'd like to do the Unit definitions format:
> 
>     * Allowing an `$id` field also inside the unit definition.
> 
>     * Allowing `version` and `property-format`, and make `property-format` MANDATORY if the unit definition is not embedded inside a property definition.
> 
>     * Add another tag `defining-relation` that allows us to describe how the units we define relate to others (primarily to relate them to SI units).

I like the sound of this! 

> **Edit:** Also _one more thing_: we have inherited from JSON Schema that the text in our descriptions uses commonmark. However, it is super inconvenient not to be able to write math expressions in our context. I suppose one could embed HTML MathML. But, for our use, should we not refer to an extension (or define ourselves) that allow embedding of LaTeX expressions inside $ signs?

I think MathML makes som sense as the only real native web standard I know of (especially now that Chrome has support for it again since January 2023), however I think we are all most familiar with LaTeX. GitHub right now will render e.g., 

$$ x = y^2 $$

with MathJax, but seemingly not MathML: 

<math mode="display" xmlns="http://www.w3.org/1998/Math/MathML">
 <semantics>
  <mrow>
    <mi>x</mi>
    <mo>=</mo>
    <mfrac>
      <mrow>
        <mo form="prefix">&minus;</mo>
        <mi>b</mi>
        <mo>&pm;</mo>
        <msqrt>
          <msup><mi>b</mi><mn>2</mn></msup>
          <mo>&minus;</mo>
          <mn>4</mn><mo>&it;</mo><mi>a</mi><mo>&it;</mo><mi>c</mi>
        </msqrt>
      </mrow>
      <mrow>
        <mn>2</mn>
        <mo>&it;</mo>
        <mi>a</mi>
      </mrow>
    </mfrac>
  </mrow>
  <annotation encoding="application/x-tex">
     x = \frac{-b\pm\sqrt{b^2-4ac}}{2a}
  </annotation>
  <annotation encoding="StarMath 5.0">
     x = {-b plusminus sqrt {b^2 - 4 ac}} over {2 a}
  </annotation>
  
  
 </semantics>
</math>

Given that Mathjax supports MathML input, perhaps this is the standard we could assume (which would also enable asciimath etc)?

optimade.rst
```diff
@@ -1945,25 +1976,20 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
     * :val:`"list"` then :field:`type` is `"array"`.
     * :val:`"float"` then :field:`type` is `"number"`.
     * :val:`"timestamp"` then :field:`type` is `"string"`.
 
-  - A list where the first item MUST be the string described above (correlated to the field :field:`x-optimade-type` in the same way) and the second item MUST be the string :val:`"null"`.
-    This form specifies that the defined property can be :val:`null`.
-
-..
+  If the second element is included, it MUST be the string :val:`"null"`.
+  This two element form specifies that the defined property can be :val:`null`.
```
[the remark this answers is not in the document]
rartino: As much as I dislike the null here (and it is indeed *very much*) it seems OpenAPI 3.1 released in 2021 also has moved over to the "null" type specification: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

Seems odd for us to adopt a way of indicating a nullable field that now is deprecated in OpenAPI?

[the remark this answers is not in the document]
rartino: Oh, it is actually even worse:

*"In order to achieve the goal of letting JSON Schema tools understand OpenAPI, it was decided to remove nullable entirely instead of deprecate it."* 

We wouldn't even be compatible with OpenAPI 3.1 then, if we went with a nullable field. They also make a good point that they don't like `nullable` because it removes a constraint rather than adding one. I guess the best design would have been a `non-nullable` field.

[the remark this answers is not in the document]
ml-evs: Haha wow, thanks for checking that. I guess the current OpenAPI schemas need to be updated once this change gets down to pydantic too :sweat_smile:

ml-evs (approved):
I'll approve this for now @rartino, just let me know if you want another review if you do add the unit stuff here.

rartino:
> I think we are all most familiar with LaTeX. GitHub right now will render e.g., $x=y^2$, but seemingly not MathML [...] Given that Mathjax supports MathML input, perhaps this is the standard we could assume (which would also enable asciimath etc)?

I think this is a great way to go. How is it best expressed? Something like:

- Formatting in the text SHOULD use Markdown in the [CommonMark v0.3 format](https://spec.commonmark.org/0.30/), with mathematical expressions rendered  using the LaTeX mode of [Mathjax 3.2](https://docs.mathjax.org/en/v3.2-latest/). When possible, it is preferable if mathematical expressions use as straightforward notation as possible to make them readable also when not rendered through Mathjax.

I rather not allow people to use MathML as it really isn't very human readable in raw form.

fekad:
> Sounds good to me! I think that makes it a lot easier for us to host a really nice rendered version of the definitions and schemas somewhere outside of GitHub

In that case [KaTex](https://katex.org/) could also be a nice Latex renderer.

rartino:
Ok @ml-evs, @vaitkus, @fekad I've made the edits I intend to the Physical Unit Definitions and added the formulation about MathJax. 

@fekad: the reference to Mathjax is only there to define the format, one can of course use any library that can render that format; which is also why I think forcing Mathjax *LaTeX* is the way to go, since LaTeX is fairly well supported also in many other non-web contexts.

I may have bit off a bit much with taking on the "defining-relation" here, because it became more extensive and tricky to define than I had thought. If people do not like the design I landed in, I'll suggest we take it out entirely now (we'd simply not provide this information anywhere else than in the description text in this version of the format), and we can work out a better way to do it in a machine-readable format for a v1.3 Property Definition format, e.g., during the upcoming workshop.

optimade.rst
```diff
@@ -2016,16 +2016,19 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
   A short single-line human-readable explanation of the defined property appropriate to show as part of a user interface.
 
+.. _definition of the description field:
+
 - :field:`description`: String.
   A human-readable multi-line description that explains the purpose, requirements, and conventions of the defined property.
   The format SHOULD be a one-line description, followed by a new paragraph (two newlines), followed by a more detailed description of all the requirements and conventions of the defined property.
-  Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__.
+  Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__ format, with mathematical expressions written to render correctly with the LaTeX mode of `Mathjax 3.2 <https://docs.mathjax.org/en/v3.2-latest/>`.
+  When possible, it is preferable for mathematical expressions to use an as straightforward notation as possible to make them readable also when not rendered.
```
ml-evs, line 2026:
```suggestion
  When possible, it is preferable for mathematical expressions to use as straightforward notation as possible to make them readable also when not rendered.
```

This reads a bit strangely to me with "an" though it is it not necessarily incorrect, so feel free to ignore this suggestion if you disagree

ml-evs (commented):
Only had a second to read so will submit this review now and come back to it later!

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+
+  - :field:`offset`: Dictionary. Optional.
+    A dictionary specifying the offset value, adhering to the same format as :field:`scale`.
+    If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 0, 1, and 0.
+
+  If the fields in :field:`scale` are designated as `sn`, `sd`, and `se`; and the fields in :field:`offset` are designated as ``on``, ``od``, and ``oe``; and :field:`base-units-expression` is designated as ``b``, these fields state the following defining relation: a value ``v`` multiplied by the unit being defined is equal to the following expression ``(v * (se/sd) * 10**se + (oe/od) * 10**oe)*u``, where ``*`` designates multiplication and ``**`` designates exponentiation.
+  For example, the defining relation of the temperature unit Fahrenheit ``F`` in Celsius ``C``, that says that ``1 F = (1 - 32) * (5/9) C`` could be expressed as follows:
```
rartino, line 2328:
```suggestion
  For example, the defining relation of the temperature unit Fahrenheit ``F`` in Celsius ``C``, that says that ``x F = (x - 32) * (5/9) C`` could be expressed as follows:
```

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+  - :field:`scale`: Dictionary. Optional.
+    A dictionary specifying the scale in the defining relation, adhering to the following format:
+
+    **OPTIONAL keys:**
+
+    - :field:`numerator`: Integer.
+    - :field:`denominator`: Integer.
+    - :field:`exponent`: Integer.
+
+    These three fields specify the value as the rational number :field:`numerator` / :field:`denominator`, multiplied by 10 to the power of :field:`exponent`. If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 1, 1, and 0.
```
vaitkus, line 2322:
```suggestion
    These three fields specify the value as the rational number :field:`numerator` / :field:`denominator`, multiplied by 10 to the power of :field:`exponent`.
    If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 1, 1, and 0.
```

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+
+  - :field:`offset`: Dictionary. Optional.
+    A dictionary specifying the offset value, adhering to the same format as :field:`scale`.
+    If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 0, 1, and 0.
+
+  If the fields in :field:`scale` are designated as `sn`, `sd`, and `se`; and the fields in :field:`offset` are designated as ``on``, ``od``, and ``oe``; and :field:`base-units-expression` is designated as ``b``, these fields state the following defining relation: a value ``v`` multiplied by the unit being defined is equal to the following expression ``(v * (se/sd) * 10**se + (oe/od) * 10**oe)*u``, where ``*`` designates multiplication and ``**`` designates exponentiation.
```
vaitkus, line 2328: What does `u` stand for in the expression?

Also, `b` is defined, but not actually used in the expression?

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+          "uri": "https://units.example.com/celsius"
+        }
+      ],
+      "base-units-expression": "C",
+      "scale": {
+        "numerator": 5,
+        "denominator": 9
+      },
+      "offset": {
+        "numerator": 160,
```
vaitkus, line 2346:
```suggestion
        "numerator": -160,
```

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+        "denominator": 9
+      }
+    }
+
```
vaitkus, line 2350: Would it be possible to get another, a slightly more complex example that involves several units?

optimade.rst
vaitkus: I guess a minus sign is needed either here or in the denominator.

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+  - :field:`offset`: Dictionary. Optional.
+    A dictionary specifying the offset value, adhering to the same format as :field:`scale`.
+    If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 0, 1, and 0.
+
+  If the fields in :field:`scale` are designated as `sn`, `sd`, and `se`; and the fields in :field:`offset` are designated as ``on``, ``od``, and ``oe``; and :field:`base-units-expression` is designated as ``b``, these fields state the following defining relation: a value ``v`` multiplied by the unit being defined is equal to the following expression ``(v * (se/sd) * 10**se + (oe/od) * 10**oe)*u``, where ``*`` designates multiplication and ``**`` designates exponentiation.
+  For example, the defining relation of the temperature unit Fahrenheit ``F`` in Celsius ``C``, that says that ``x F = (x - 32) * (5/9) C`` could be expressed as follows:
```
vaitkus, line 2329:
```suggestion
  For example, the defining relation of the temperature unit Fahrenheit ``F`` in Celsius ``C``, that says that ``x F = (x - 32) * (5/9) C = 5/9 x + (-160/9) C`` could be expressed as follows:
```

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+- :field:`$comment`: String.
+  A human-readable comment relevant in the context of the raw definition data.
+  These comments should be omitted in formatted descriptions of the Physical Unit definition shown to the end users.
+  (Comments in descriptions relevant to the end users should go into :field:`description`.)
+  Formatting in the text SHOULD use Markdown using the format described in the `definition of the description field`_ of Property Definitions.
+
+  This field should be seen as an annotation rather than an integral part of the Physical Unit Definition.
+  Two Physical Unit Definitions that only differ by the value of any :field:`$comment` fields are considered the same.
+
+- :field:`defining-relation`: Dictionary.
```
rartino, line 2291: I've realized a possible need for further caution here. Many (most?) units are defined by relations to SI units today. However, there are some that are not.

The only good example I found after a quick look around is the atomic mass unit (u; dalton). It only has an approximate relation to kg (1 u ~= 1.66053906660(50)×10^-27 kg). We actually have to define this unit in OPTIMADE, since we use it for masses in species.

I'm thinking the mass unit should not be expressed with a defining relation (and maybe we should clarify in the text that `defining-relation` is NOT for that.) But then perhaps it would be nice to have some other way of expressing the best known non-defining relation to SI units?

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+    - :field:`symbol`: String.
+      The symbol used to reference this unit in the dimensional formula.
+    - :field:`uri`: String.
+      The URI of one of the units referenced in the dimensional formula for the defining relation.
+
+  - :field:`base-units-expression`: String.
+    A string expressing the base units part of the defining relation for the unit being defined.
+    It MUST adhere to the format for compound unit expression described in `Physical Units in Property Definitions`_.
+
+  - :field:`scale`: Dictionary. Optional.
```
vaitkus, line 2313: A bit of a contradiction -- this seems to fall under the **Required keys**, but is also marked as Optional.

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+
+    **OPTIONAL keys:**
+
+    - :field:`numerator`: Integer.
+    - :field:`denominator`: Integer.
+    - :field:`exponent`: Integer.
+
+    These three fields specify the value as the rational number :field:`numerator` / :field:`denominator`, multiplied by 10 to the power of :field:`exponent`. If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 1, 1, and 0.
+
+  - :field:`offset`: Dictionary. Optional.
```
vaitkus, line 2324: A bit of a contradiction -- this seems to fall under the **Required keys**, but is also marked as Optional.

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+  These comments should be omitted in formatted descriptions of the Physical Unit definition shown to the end users.
+  (Comments in descriptions relevant to the end users should go into :field:`description`.)
+  Formatting in the text SHOULD use Markdown using the format described in the `definition of the description field`_ of Property Definitions.
+
+  This field should be seen as an annotation rather than an integral part of the Physical Unit Definition.
+  Two Physical Unit Definitions that only differ by the value of any :field:`$comment` fields are considered the same.
+
+- :field:`defining-relation`: Dictionary.
+  A dictionary that encodes a defining relation to another unit or set of units, with the primary intended use of relating a unit being defined to SI units, if possible.
+  The dictionary MUST adhere to the following format:
```
rartino, line 2293:
```suggestion
  A dictionary that encodes a defining relation to another unit or set of units, with the primary intended use of relating a unit to its definition in SI units, if such a relationship exists.
  Some units, e.g., the atomic mass unit (also known as dalton, commonly denoted ``u``), only has an approximate relationship to SI units, in which case the :field:`defining-relation` MUST be omitted or :val:`null`.
  The dictionary MUST adhere to the following format:
```

optimade.rst
```diff
@@ -2203,6 +2271,82 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+  If :val:`FALSE`, the unit defined in this Physical Unit Definition is not deprecated.
+  The field not being present means :val:`FALSE`.
+
+  This field should be seen as an annotation rather than an integral part of the Physical Unit Definition.
+  Two Physical Unit Definitions that only differ by :field:`deprecated` fields are considered the same.
+
+- :field:`$comment`: String.
+  A human-readable comment relevant in the context of the raw definition data.
+  These comments should be omitted in formatted descriptions of the Physical Unit definition shown to the end users.
+  (Comments in descriptions relevant to the end users should go into :field:`description`.)
```
vaitkus, line 2285: The unit definition dictionary does not seem to define a `description` field?

vaitkus (commented):
Some initial comments.

optimade.rst
rartino:
```suggestion
  If the fields in :field:`scale` are designated as `sn`, `sd`, and `se`; and the fields in :field:`offset` are designated as ``on``, ``od``, and ``oe``; and :field:`base-units-expression` is designated as ``b``, these fields state the following defining relation: a value ``v`` multiplied by the unit being defined is equal to the following expression ``(v * (se/sd) * 10**se + (oe/od) * 10**oe)*b``, where ``*`` designates multiplication and ``**`` designates exponentiation.
```

optimade.rst
rartino: I mean... I though this really was an obvious exercise for the reader... but ok.

optimade.rst
rartino: Ok, perhaps we can add an example of a whole Physical Unit Definition and do multiple units there.

optimade.rst
rartino: It is there? Maybe you just don't see it because you are just looking at the diff? If you fold out downwards from line 2210 you should see it?

vaitkus: Sorry, my bad, totally missed that.

optimade.rst
```diff
@@ -2203,6 +2271,84 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+    - :field:`exponent`: Integer.
+
+    These three fields specify the value as the rational number :field:`numerator` / :field:`denominator`, multiplied by 10 to the power of :field:`exponent`.
+    If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 1, 1, and 0.
+
+  - :field:`offset`: Dictionary. Optional.
+    A dictionary specifying the offset value, adhering to the same format as :field:`scale`.
+    If omitted or :val:`null`, the defaults for the :field:`numerator`, :field:`denominator`, and :field:`exponent` are respectively 0, 1, and 0.
+
+  If the fields in :field:`scale` are designated as `sn`, `sd`, and `se`; and the fields in :field:`offset` are designated as ``on``, ``od``, and ``oe``; and :field:`base-units-expression` is designated as ``b``, these fields state the following defining relation: a value ``v`` multiplied by the unit being defined is equal to the following expression ``(v * (se/sd) * 10**se + (oe/od) * 10**oe)*b``, where ``*`` designates multiplication and ``**`` designates exponentiation.
```
vaitkus, line 2329:
```suggestion
  If the fields in :field:`scale` are designated as `sn`, `sd`, and `se`; and the fields in :field:`offset` are designated as ``on``, ``od``, and ``oe``; and :field:`base-units-expression` is designated as ``b``, these fields state the following defining relation: a value ``v`` multiplied by the unit being defined is equal to the following expression ``(v * (sn/sd) * 10**se + (on/od) * 10**oe)*b``, where ``*`` designates multiplication and ``**`` designates exponentiation.
```

rartino: Oh, wow, how did I get this so screwed up. Thanks for actually digging through the details here!

optimade.rst
rartino: Done.

optimade.rst
rartino: Right, this is not the syntax we use. Should be fixed now.

optimade.rst
```diff
@@ -3421,3 +3573,104 @@ The strings below contain Extended Regular Expressions (EREs) to recognize ident
...
+    "x-optimade-type": "list",
+    "x-optimade-property": {
+      "version": "1.2.0",
+      "property-format": "1.2",
+      "unit-definitions": [
+        {
+          "title": "Newton",
+          "symbol": "N",
+          "$id": "https://units.example.com/v1.2.0/N",
+          "description": "The Newton SI unit of force, defined as 1 kg m/s^2
```
vaitkus, line 3597:
```suggestion
          "description": "The newton SI unit of force, defined as 1 kg m/s^2
```

optimade.rst
```diff
@@ -3421,3 +3573,104 @@ The strings below contain Extended Regular Expressions (EREs) to recognize ident
...
+              }
+            ],
+            "base-units-expression": "kg*m*s^-2"
+          }
+        },
+        {
+          "title": "Dalton mass unit",
+          "symbol": "u",
+          "$id": "https://units.example.com/v1.2.0/u",
+          "description": "The Dalton mass unit defined as 1/12 of the mass of an
```
vaitkus, line 3627:
```suggestion
          "description": "The dalton mass unit defined as 1/12 of the mass of an
```

optimade.rst
```diff
@@ -3421,3 +3573,104 @@ The strings below contain Extended Regular Expressions (EREs) to recognize ident
...
+      "version": "1.2.0",
+      "property-format": "1.2",
+      "unit-definitions": [
+        {
+          "title": "Newton",
+          "symbol": "N",
+          "$id": "https://units.example.com/v1.2.0/N",
+          "description": "The Newton SI unit of force, defined as 1 kg m/s^2
+                          equal to $10^{-10}$ meter using the 2019
+                          redefinition of the SI base units.",
```
vaitkus, line 3599: Is this correct here? Seems like a leftover definition of angstrom.

optimade.rst
```diff
@@ -3421,3 +3573,104 @@ The strings below contain Extended Regular Expressions (EREs) to recognize ident
     #BEGIN ERE strings
     "([^\"]|\\.)*"
     #END ERE strings
+
+Property Definition Example
+---------------------------
+
+This appendix provides a more complete example of a Property Definition on the format defined in `Property Definitions`_.
+(Note: the description strings have been wrapped for readability only)
```
vaitkus, line 3581:
```suggestion
(Note: the description strings have been wrapped for readability only.)
```

optimade.rst
rartino:
```suggestion
                          using the 2019 redefinition of the SI base units.",
```

rartino: Good catch - copy-paste error when I grabbed the string about the SI redefinition.

optimade.rst
```diff
@@ -1866,24 +1874,30 @@ A Property Definition MUST be composed according to the combination of the requi
     - :field:`relation`: String.
       A human-readable description of the relationship between the property and the remote resource, e.g., a "natural language description".
 
-    - :field:`uri`: String.
-      A URI of the external resource (which MAY be a resolvable URL).
+    - :field:`resource-id`: String.
+      A IRI of the external resource, which MAY be a resolvable URL.
```
ml-evs, line 1878:
```suggestion
      An IRI of the external resource, which MAY be a resolvable URL.
```

optimade.rst
```diff
@@ -1866,24 +1874,30 @@ A Property Definition MUST be composed according to the combination of the requi
...
+- :field:`x-optimade-type`: String.
   Specifies the OPTIMADE data type for this level of the defined property.
   MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, or :val:`"dictionary"`.
 
 - :field:`x-optimade-unit`: String.
   A (compound) symbol for the physical unit in which the value of the defined property is given or one of the strings :val:`dimensionless` or :val:`inapplicable`.
   See subsection `Physical Units in Property Definitions`_ for the details on how compound units are represented in OPTIMADE Property Definitions and the precise format of this string.
 
-**OPTIONAL keys at all nested levels of the Property Definition:**
+**OPTIONAL keys at all nested levels of the Property Definition which are regarded as annotations to the Property Definition**
```
ml-evs, line 1892:
```suggestion
**OPTIONAL keys at all nested levels of the Property Definition which are regarded as annotations to the Property Definition:**
```

optimade.rst
```diff
@@ -1902,8 +1916,12 @@ A Property Definition MUST be composed according to the combination of the requi
     - :val:`partial`: the defined property MUST be queryable with support for a subset of the filter language operators as specified by the field :field:`query-support-operators`.
     - :val:`none`: the defined property does not need to be queryable with any features of the filter language.
 
+    Omitting the field or :val:`null` is equivalent to :val:`none`.
+
   - :field:`query-support-operators`: List of Strings.
     Defines the filter language features supported on this property.
+    MUST be present and not :val:`null` if and only if :field:`query-support` is :val:`partial`.
+
     Each string in the list MUST be one of :val:`<`, :val:`<=`, :val:`>`, :val:`>=`, :val:`=`, :val:`!=`, :val:`CONTAINS`, :val:`STARTS WITH`, :val:`ENDS WITH`:, :val:`HAS`, :val:`HAS ALL`, :val:`HAS ANY`, :val:`HAS ONLY`, :val:`IS KNOWN`, :val:`IS UNKNOWN` with the following meanings:
```
ml-evs, line 1925:
```suggestion
    Each string in the list MUST be one of :val:`<`, :val:`<=`, :val:`>`, :val:`>=`, :val:`=`, :val:`!=`, :val:`CONTAINS`, :val:`STARTS WITH`, :val:`ENDS WITH`, :val:`HAS`, :val:`HAS ALL`, :val:`HAS ANY`, :val:`HAS ONLY`, :val:`IS KNOWN`, :val:`IS UNKNOWN` with the following meanings:
```

optimade.rst
```diff
@@ -1918,14 +1936,20 @@ A Property Definition MUST be composed according to the combination of the requi
 
     - :val:`IS KNOWN`, :val:`IS UNKNOWN`: indicating support for filtering this property on unknown values using the respective operator.
 
+  - :field:`response-default`: Boolean
```
ml-evs, line 1939:
```suggestion
  - :field:`response-default`: Boolean.
```

optimade.rst
```diff
@@ -1945,25 +1986,25 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
 
-  - A string correlated with :field:`x-optimade-type` as follows.
-    If :field:`x-optimade-type` is:
+  It MUST be a list of one or two elements where the first element is a string correlated with :field:`x-optimade-type` as follows; if :field:`x-optimade-type` is:
 
     * :val:`"boolean"`, `"string"`, or `"integer"` then :field:`type` is the same string.
     * :val:`"dictionary"` then :field:`type` is `"object"`.
     * :val:`"list"` then :field:`type` is `"array"`.
     * :val:`"float"` then :field:`type` is `"number"`.
     * :val:`"timestamp"` then :field:`type` is `"string"`.
```
ml-evs, line 2001:
```suggestion
* :val:`"boolean"`, `"string"`, or `"integer"` then :field:`type` is the same string.
* :val:`"dictionary"` then :field:`type` is `"object"`.
* :val:`"list"` then :field:`type` is `"array"`.
* :val:`"float"` then :field:`type` is `"number"`.
* :val:`"timestamp"` then :field:`type` is `"string"`.
```

Think this is the right indent level, on the rich diff these are currently appearing as inline quotes which I guess is not desirable

ml-evs: (Same with the implementation notes bulletpoints below, but I can't make suggestions for those in this PR)

optimade.rst
```diff
@@ -1977,33 +2018,39 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
 - :field:`title`: String.
   A short single-line human-readable explanation of the defined property appropriate to show as part of a user interface.
 
+.. _definition of the description field:
+
 - :field:`description`: String.
   A human-readable multi-line description that explains the purpose, requirements, and conventions of the defined property.
   The format SHOULD be a one-line description, followed by a new paragraph (two newlines), followed by a more detailed description of all the requirements and conventions of the defined property.
-  Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__.
+  Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__ format, with mathematical expressions written to render correctly with the LaTeX mode of `Mathjax 3.2 <https://docs.mathjax.org/en/v3.2-latest/>`.
```
ml-evs, line 2041:
```suggestion
  Formatting in the text SHOULD use Markdown in the `CommonMark v0.3 format <https://spec.commonmark.org/0.30/>`__ format, with mathematical expressions written to render correctly with the LaTeX mode of `Mathjax 3.2 <https://docs.mathjax.org/en/v3.2-latest/>`__.
```

optimade.rst
```diff
@@ -2200,8 +2314,130 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
   - :field:`relation`: String.
     A human-readable description of the relationship between the unit and the remote resource, e.g., a "natural language description".
 
-  - :field:`uri`: String.
-    A URI of the external resource (which MAY be a resolvable URL).
+  - :field:`resource-id`: String.
+    A IRI of the external resource (which MAY be a resolvable URL).
```
ml-evs, line 2318:
```suggestion
    An IRI of the external resource (which MAY be a resolvable URL).
```

ml-evs (commented):
Got about halfway through, more than enough for a Sunday morning ;)  Will try to take a look at the rest next week.

I'm still a little worried about all the complexity being added, though each change in isolation seems justifiable. The most complete custom `info/structures` endpoint atm is probably https://oqmd.org/optimade/v1/info/structures, so perhaps we could offer to help them convert to this new format as a trial?

optimade.rst
rartino: The implementation notes are actually meant to be set as [block quotes](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#block-quotes) throughout the specification. I mean, I'm happy for any other suggestions - but we've been doing it semi-consistently to separate those segments from the specification text.

optimade.rst
```diff
@@ -1945,27 +1991,27 @@ The format described in this subsection forms a subset of the `JSON Schema Valid
...
-  - A string correlated with :field:`x-optimade-type` as follows.
-    If :field:`x-optimade-type` is:
+  It MUST be a list of one or two elements where the first element is a string correlated with :field:`x-optimade-type` as follows; if :field:`x-optimade-type` is:
 
-    * :val:`"boolean"`, `"string"`, or `"integer"` then :field:`type` is the same string.
-    * :val:`"dictionary"` then :field:`type` is `"object"`.
-    * :val:`"list"` then :field:`type` is `"array"`.
-    * :val:`"float"` then :field:`type` is `"number"`.
-    * :val:`"timestamp"` then :field:`type` is `"string"`.
+  * :val:`"boolean"`, `"string"`, or `"integer"` then :field:`type` is the same string.
```
JPBergsma, line 2002:
```suggestion
  * :val:`"boolean"`, :val:`"string"`, or :val:`"integer"` then :field:`type` is the same string.
```
This should give a more consistent formatting of the types.
perhaps the "object" and "array" should also be preceded by ":val:"

optimade.rst
```diff
@@ -2130,32 +2211,61 @@ Clients and servers that use other units internally thus have to do unit convers
 The physical unit of a property, the embedded items of a list, or values of a dictionary, are defined with the field :field:`x-optimade-unit` with the following requirements:
 
 - The field MUST be given with a non-:val:`null` value both at the highest level in the OPTIMADE Property Definition and all inner Property Definitions.
-- If the property refers to a physical quantity that is dimensionless (often also referred to as having the dimension 1) or refers to a dimensionless count of something (e.g., the number of protons in a nucleus) the field MUST have the value :val:`dimensionless`.
+- If the property refers to a physical quantity that is dimension and unitless (often also referred to as having the dimension 1) or refers to a dimension and unitless count of something (e.g., the number of protons in a nucleus) the field MUST have the value :val:`dimensionless`.
```
vaitkus, line 2214:
```suggestion
- If the property refers to a physical quantity that is dimensionless and unitless (often also referred to as having the dimension 1) or refers to a dimensionless and unitless count of something (e.g., the number of protons in a nucleus) the field MUST have the value :val:`dimensionless`.
```

optimade.rst
```diff
@@ -2130,32 +2211,61 @@ Clients and servers that use other units internally thus have to do unit convers
...
+  However, quantities that uses counting units, e.g., the mole, or quantities that use dimensionless units, e.g., the radian MUST NOT set the field to :val:`dimensionless`.
```
vaitkus, line 2215:
```suggestion
  However, quantities that use counting units, e.g., the mole, or quantities that use dimensionless units, e.g., the radian MUST NOT set the field to :val:`dimensionless`.
```

optimade.rst
```diff
@@ -2200,8 +2327,127 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+  **OPTIONAL keys:**
+
+  - :field:`base-units`: List of Dictionaries.
+    A list specifying the base IRIs and unit symbols for the units in which the dimensional formula for the defining relation is expressed.
+    Each item MUST be a dictionary that adheres to the following format:
+
+    **REQUIRED keys:**
+
+    - :field:`symbol`: String.
+      The symbol used to reference this unit in the dimensional formula.
```
vaitkus, line 2349:
```suggestion
      The symbol used to reference this unit in the dimensional formula.

```

optimade.rst
```diff
@@ -2200,8 +2327,127 @@ If provided, the :field:`unit-definitions` in :field:`x-optimade-property` MUST
...
+    - :field:`id`: String.
+      The IRI of one of the units referenced in the dimensional formula for the defining relation.
+
+  - :field:`base-units-expression`: String.
+    A string expressing the base units part of the defining relation for the unit being defined.
+    It MUST adhere to the format for compound unit expression described in `Physical Units in Property Definitions`_.
+    if the field is missing or :val:`null` the base-units-expression is taken to be equal to 1, i.e., the defining relation is dimensionless.
```
vaitkus, line 2356:
```suggestion
    If the field is missing or :val:`null` the base-units-expression is taken to be equal to 1, i.e., the defining relation is dimensionless.
```

optimade.rst
```diff
@@ -3421,3 +3667,103 @@ The strings below contain Extended Regular Expressions (EREs) to recognize ident
     #BEGIN ERE strings
     "([^\"]|\\.)*"
     #END ERE strings
+
+Property Definition Example
+---------------------------
+
+This appendix provides a more complete example of a Property Definition on the format defined in `Property Definitions`_.
```
vaitkus, line 3674:
```suggestion
This appendix provides a more complete example of a Property Definition in the format defined in `Property Definitions`_.
```

optimade.rst
```diff
@@ -2089,9 +2124,38 @@ A Property Definition MUST be composed according to the combination of the requi
 
 **OPTIONAL keys at all nested levels of the Property Definition:**
 
+- :field:`x-optimade-unit-definitions`: List.
+  A list of definitions of the symbols used in the Property Definition (including its nested levels) for physical units given as values of the :field:`x-optimade-unit` field.
+  This field **MUST be included at the outermost level of a property definition** if the defined property, at any level, includes an :field:`x-optimade-unit` with a value that is not :val:`dimensionless` or :val:`inapplicable`, and it MUST include definitions of all units used on all levels in the property definition.
+  The field MAY also occur at deeper nesting levels (but this is not required).
+  If it does, the unit definitions provided MUST be redundant with those provided at higher nesting levels.
+  See subsection `Physical Units in Property Definitions`_ for the details on how units are represented in OPTIMADE Property Definitions and the precise format of this dictionary.
```
rartino, line 2132: Perhaps we should make x-optimade-unit-definitions always REQUIRED in the outermost level (so, possibly an empty list). The specification just gets bulky from the OPTIONAL - oh but often REQUIRED - categorization. However, setting this to REQUIRED requires updating the property definitions in #445.

ml-evs:
Now I look at this wrt my comments on #488 and #445, I see we do not use the JSON Schema `pattern` field here. Is that intentional @rartino? It looks like we have `maxLength` and `minLength` for strings but not `pattern`.

optimade.rst
```diff
@@ -2326,7 +2432,10 @@ Depending on what string the :field:`type` is equal to, or contains as first ele
     - :val:`"time"`: the full-time production in :RFC:`3339` section 5.6.
     - :val:`"duration"`: the duration production in :RFC:`3339` Appendix A.
     - :val:`"email"`: the "Mailbox" ABNF rule in :RFC:`5321` section 4.1.2.
-    - :val:`"uri"`: a string instance is valid against this attribute if it is a valid URI, according to :RFC:`3986`.
+    - :val:`"uri"`: a string instance is valid against this attribute if it is a valid URI according to :RFC:`3986`.
+    - :val:`"iri"`: a string instance is valid against this attribute if it is a valid IRI according to :RFC:`3987`.
+
```
ml-evs, line 2437: I think the pattern field would squeeze in here (this is the only place GitHub will let me make the suggestion, so could also be reordered above `format`:

```suggestion
  - :field:`pattern`: String.
    A non-negative integer.
    This string SHOULD be a valid regular expression, according to the ECMA-262 regular expression dialect.
    A string instance is considered valid if the regular expression matches the instance successfully. Recall: regular expressions are not implicitly anchored.
```

ml-evs (changes requested):

rartino:
> I see we do not use the JSON Schema pattern field here. Is that intentional @rartino? It looks like we have maxLength and minLength for strings but not pattern.

`pattern` was indeed excluded intentionally, since we would have to agree on a regex format for OPTIMADE before we can support `pattern` (which to be used here would have to be equal to, or a subset of, ECMA since that is what JSON Schema specifies). Since we likely want to sync the regex format across OPTIMADE, this is waiting for a decision in #160.

For now, I'd suggest we merge the property definitions without 'pattern' and thus the symops would have to be validated outside of what can be done with the schemas. If anyone feels strongly enough about this omission, I think we have all pieces in place to finish #160, and then bring over the regex subset into this PR.

optimade.rst
rartino: I added this now, because I am about to merge this whole PR into #445 since they depend on each other. However, I think we need to update the description of pattern to point at whatever OPTIMADE regex subset we allow.

rartino:
This PR is now merged into #445

PR #465, opened 2023-06-06
Add `bonds` for structure type entries

merkys:
In issue Materials-Consortia/namespace-cheminformatics#14 I proposed adding more chemical properties to OPTIMADE structures. This PR implements [my suggestion](https://github.com/Materials-Consortia/namespace-cheminformatics/issues/14#issuecomment-1434603619) on representation of chemical connectivity between pairs of sites in OPTIMADE:
```json
{
    "bonds": [ {"sites": [1, 2]} ]
}
```
I intentionally omit the bond types as this might be difficult to agree upon, whereas having just the connectivity is already beneficial.

Pinging people who have expressed their interest for comments: @eimrek @BobHanson @Austin243

**Edit**: I have introduced means to express connections with translation equivalents of the sites in `sites`.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2846,7 +2846,7 @@ bonds
   - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
   - **Query**: Support for queries on this property is OPTIONAL.
     If supported, filters MAY support only a subset of comparison operators.
-  - The property SHOULD be :val:`null` for structures for which the bond perception has not been carried out by the implementation.
+  - The property SHOULD be :val:`null` for structures for which the chemical connectivity is unknown to the implementation.
   - :property:`sites` inside each dictionary contains two integers refering to different sites which are deemed chemically connected.
 
 - **Examples**:
--- a/optimade.rst
+++ b/optimade.rst
@@ -2833,6 +2833,26 @@ assemblies
     These two sites are correlated (either site 2 or 3 is present).
     However, the presence or absence of sites 0 and 1 is not correlated with the presence or absence of sites 2 and 3 (in the specific example, the pair of sites (0, 2) can occur with 0.2\*0.3 = 6 % probability; the pair (0, 3) with 0.2\*0.7 = 14 % probability; the pair (1, 2) with 0.8\*0.3 = 24 % probability; and the pair (1, 3) with 0.8\*0.7 = 56 % probability).
 
+bonds
+~~~~~
+
+- **Description**: A list describing the chemical connectivity in the structure.
+- **Type**: list of dictionary with keys:
+
+  - :property:`sites`: a list of integers (REQUIRED)
+
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - The property SHOULD be :val:`null` for structures for which the bond perception has not been carried out by the implementation.
+  - :property:`sites` inside each dictionary contains two integers refering to different sites which are deemed chemically connected.
+
+- **Examples**:
+
+  - :val:`[ {"sites": [1, 2]} ]`: a structure with a bond between sites 1 and 2.
+
 structure\_features
 ~~~~~~~~~~~~~~~~~~~
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2859,7 +2859,7 @@ bonds
 
 - **Examples**:
 
-  - :val:`[ {"sites": [0, 1]} ]`: a structure with a bond between sites 1 and 2.
+  - :val:`[ {"sites": [0, 1]} ]`: a structure with a bond between sites 0 and 1.
   - :val:`[ {"sites": [0, 0], "translations": [ [0, 0, 0], [0, 0, 1] ]} ]`: a 1D polymer.
 
 structure\_features
--- a/optimade.rst
+++ b/optimade.rst
@@ -2856,7 +2856,8 @@ bonds
 
     - *translations*: a list of two lists of three integers each, defining translations of the sites.
       Omitting this key means that both translation vectors are :val:`[0, 0, 0]`.
-
+    - *distance*: the distance in Angstrom between the two sites after applying the appropriate translation vectors.
+    
 - **Examples**:
 
   - :val:`[ {"sites": [0, 1]} ]`: a structure with a bond between sites 0 and 1.
--- a/optimade.rst
+++ b/optimade.rst
@@ -2847,7 +2847,8 @@ bonds
   - **Query**: Support for queries on this property is OPTIONAL.
     If supported, filters MAY support only a subset of comparison operators.
   - The property SHOULD be :val:`null` for structures for which the chemical connectivity is unknown to the implementation.
-  - :property:`sites` inside each dictionary contains two integers refering to different sites which are deemed chemically connected.
+  - :property:`sites` inside each dictionary contains two integers refering to sites which are deemed chemically connected.
+    The site indexes MUST be ordered in non-decreasing manner.
 
 - **Examples**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2840,6 +2840,7 @@ bonds
 - **Type**: list of dictionary with keys:
 
   - :property:`sites`: a list of integers (REQUIRED)
+  - :property:`translations`: a list of list of integers (OPTIONAL)
 
 - **Requirements/Conventions**:
 
@@ -2851,9 +2852,15 @@ bonds
 
     - *sites*: an ordered list of 0-based indexes of the two sites that form a chemical bond.
 
+  - If translations are needed by at least one of the sites of a bond, the following key SHOULD be used:
+
+    - *translations*: a list of two lists of three integers each, defining translations of the sites.
+      Omitting this key means that both translation vectors are :val:`[0, 0, 0]`.
+
 - **Examples**:
 
   - :val:`[ {"sites": [1, 2]} ]`: a structure with a bond between sites 1 and 2.
+  - :val:`[ {"sites": [1, 1], "translations": [ [0, 0, 0], [0, 0, 1] ]} ]`: a 1D polymer.
 
 structure\_features
 ~~~~~~~~~~~~~~~~~~~
--- a/optimade.rst
+++ b/optimade.rst
@@ -2859,8 +2859,8 @@ bonds
 
 - **Examples**:
 
-  - :val:`[ {"sites": [1, 2]} ]`: a structure with a bond between sites 1 and 2.
-  - :val:`[ {"sites": [1, 1], "translations": [ [0, 0, 0], [0, 0, 1] ]} ]`: a 1D polymer.
+  - :val:`[ {"sites": [0, 1]} ]`: a structure with a bond between sites 1 and 2.
+  - :val:`[ {"sites": [0, 0], "translations": [ [0, 0, 0], [0, 0, 1] ]} ]`: a 1D polymer.
 
 structure\_features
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -2833,6 +2833,26 @@ assemblies
...
+
+  - :property:`sites`: a list of integers (REQUIRED)
+
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - The property SHOULD be :val:`null` for structures for which the bond perception has not been carried out by the implementation.
+  - :property:`sites` inside each dictionary contains two integers refering to different sites which are deemed chemically connected.
```
vaitkus, line 2850:
```suggestion
  - If present, it MUST be a list of dictionaries, each of which represents a chemical bond and MUST have the following key:
    - *sites*: 0-based indexes of the two sites that form a chemical bond.
```

vaitkus: Copied the phrasing from the `assemblies` definition to have a more uniform style.

optimade.rst
```diff
@@ -2833,6 +2833,26 @@ assemblies
...
+
+- **Examples**:
+
+  - :val:`[ {"sites": [1, 2]} ]`: a structure with a bond between sites 1 and 2.
```
vaitkus, line 2854: Maybe it should also be mandated that the indexes in the array are always sorted, e.g. `[0, 8]` is allowed, but not `[8, 0]`?

optimade.rst
```diff
@@ -2833,6 +2833,26 @@ assemblies
...
+- **Type**: list of dictionary with keys:
+
+  - :property:`sites`: a list of integers (REQUIRED)
+
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - The property SHOULD be :val:`null` for structures for which the bond perception has not been carried out by the implementation.
```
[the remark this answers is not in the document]
merkys: I have attempted to address this in 1478145 by reformulating to a more generic expression.

optimade.rst
merkys: I have attempted to address this in a49da9e51677bcc08fe224e5f58b55ab79c30ec2.

vaitkus: Thank you. I guess `non-decreasing` is used instead of `ascending` since there are plans to allow sites to be bonded to symmetrical equivalents of themselves or something similar?

merkys: Right, I have just described the means for specifying translations. I have replaced "non-decreasing" with "ordered", but if you feel that the "non-decreasing" is better, I can bring it back.

optimade.rst
```diff
@@ -2833,6 +2833,35 @@ assemblies
...
+
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - The property SHOULD be :val:`null` for structures for which the chemical connectivity is unknown to the implementation.
+  - If present, it MUST be a list of dictionaries, each of which represents a chemical bond and MUST have the following key:
+
+    - *sites*: an ordered list of 0-based indexes of the two sites that form a chemical bond.
```
vaitkus, line 2853:
```suggestion
    - *sites*: a non-decreasing list of 0-based indexes of the two sites that form a chemical bond.
```

optimade.rst
vaitkus: Non-decreasing would be clearer (added relevant suggestion).

vaitkus (dismissed):

merkys:
Pinging @d-beltran for comments on how this proposal suits macromolecules.

merkys:
Pinging @utf who participated in the discussions.

optimade.rst
```diff
@@ -2833,6 +2833,35 @@ assemblies
...
+  - The property SHOULD be :val:`null` for structures for which the chemical connectivity is unknown to the implementation.
+  - If present, it MUST be a list of dictionaries, each of which represents a chemical bond and MUST have the following key:
+
+    - *sites*: a non-decreasing list of 0-based indexes of the two sites that form a chemical bond.
+
+  - If translations are needed by at least one of the sites of a bond, the following key SHOULD be used:
+
+    - *translations*: a list of two lists of three integers each, defining translations of the sites.
+      Omitting this key means that both translation vectors are :val:`[0, 0, 0]`.
+
```
ml-evs, line 2859:
```suggestion
    - *distance*: the distance in Angstrom between the two sites after applying the appropriate translation vectors.
    
```

Might be nice to also allow this as an optional field.

ml-evs (commented):
Thanks for this @merkys, here's my initial comments. I wonder if we also want this field to capture "non-chemical" connectivity, but I can't think of a good use case

optimade.rst
```diff
@@ -2833,6 +2833,38 @@ assemblies
...
+  - OPTIONAL key :property:`distance` MAY be used to provide distance in ångström (Å) between the two sites after applying the appropriate translation vectors.
```
vaitkus, line 2861:
```suggestion
  - OPTIONAL key :property:`distance` MAY be used to provide the distance in ångströms (Å) between the two sites after applying the appropriate translation vectors.
```

optimade.rst
```diff
@@ -2833,6 +2833,35 @@ assemblies
...
+
+  - If translations are needed by at least one of the sites of a bond, the following key SHOULD be used:
+
+    - *translations*: a list of two lists of three integers each, defining translations of the sites.
+      Omitting this key means that both translation vectors are :val:`[0, 0, 0]`.
+
+- **Examples**:
+
+  - :val:`[ {"sites": [1, 2]} ]`: a structure with a bond between sites 1 and 2.
+  - :val:`[ {"sites": [1, 1], "translations": [ [0, 0, 0], [0, 0, 1] ]} ]`: a 1D polymer.
```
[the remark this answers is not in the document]
vaitkus: This property could potentially be used to describe crystal topology which may involve various types of contact distances, but that seems to be a bit a beast of its own (at least according to work done in the `TOPO_CIF` dictionary [1]). I would probably limit it to chemical bonding for now, especially since currently we purposely are trying to avoid specifying the bond type/order.

[1] https://github.com/COMCIFS/TopoCif/blob/main/dictionary/Topology_0.9.5.dic

optimade.rst
```diff
@@ -2833,6 +2833,38 @@ assemblies
...
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - The property SHOULD be :val:`null` for structures for which the chemical connectivity is unknown to the implementation.
+  - If present, it MUST be a list of dictionaries, each of which represents a chemical bond and MUST have the following key:
+
+    - *sites*: a non-decreasing list of 0-based indexes of the two sites that form a chemical bond.
+
+  - If translations are needed by at least one of the sites of a bond, the following key SHOULD be used:
```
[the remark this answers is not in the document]
merkys: I assume you are talking about `translations`, as `sites` is mandatory?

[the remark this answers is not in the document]
d-beltran: Works for me :)

> I intentionally omit the bond types as this might be difficult to agree upon

The type of bond is not specified in most of the topology formats but I think this is very inconvenient. So happy to know this is provisional and the type will be specified in a future.

merkys:
@eimrek suggested leaving translation vector only for one of the sites as at least one site will stay in the unit cell, or can be translated there. How about this:
```jsonc
{
    "sites": [63, 64],
    "translation_site": 1, // the second of the two sites is translated
    "translation_vector": [0, 0, 1]
}
```
The `translation_site` is required as any of the two sites could be translated, and the order of indexes has to be respected in `sites`. Or is it better to have a more structured way: `"translation": { "site": 1, "vector": [0, 0, 1] }`?

rartino:
Workshop: We are happy to merge an explicit bond strucutre datastructure, but we must consider the exact format so the types of queries that one wants to do can be performed (with the present filter language, preferably). Inheriting the current CIF framework for this should be seriously considered.

merkys:
> @merkys I would leave at least the option to provide both of the translations. In the COD there are multiple non-polymeric molecules that spans several unit cells, so both translation vectors will be needed to correctly represent the complete molecule. I attach an example of such molecule to this comment ([1540421.cif.txt](https://github.com/Materials-Consortia/OPTIMADE/files/11696037/1540421.cif.txt), remove the txt extension before viewing), but @sauliusg could probably provide an even more extreme example (I seem to recall a molecule that spans 5 unit cells).

It is always possible to back-translate one of the sites into the primary unit cell without losing the connectivity information. Having both non-zero translation vectors is a matter of convenience, I think, or does this retain some more information?

merkys:
> Workshop: We are happy to merge an explicit bond strucutre datastructure, but we must consider the exact format so the types of queries that one wants to do can be performed (with the present filter language, preferably). Inheriting the current CIF framework for this should be seriously considered.

Unless we introduce some data redundancy here, the most powerful queries would be the ones based on correlated arrays (a.k.a. zips), as OPTIMADE does not support anything more intricate than that. To make `bonds` more friendly for zips, it should be defined as an array correlated with `species_at_sites`, possibly listing all neighbours of each site (full symmetric matrix). But again, `species_at_sites` is just an array of labels, thus one could not reliably query for, for example, structures where O atom has three bonds or more.

vaitkus:
> 
> It is always possible to back-translate one of the sites into the primary unit cell without losing the connectivity information. Having both non-zero translation vectors is a matter of convenience, I think, or does this retain some more information?

Well, if you back-translate sites into the primary unit cell you lose some information and end up with a set of disjointed bonded fragments. You could, of course, translate these fragments from the primary unit cell back into their proper place, however, it is not ye obvious to me that this is a straightforward task. What is the drawback of allowing to specify both sites?

optimade.rst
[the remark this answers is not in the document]
vaitkus: @merkys Yes, I actually meant `translations` (fixed).

sauliusg (commented):
As we are thinking about implementation of fractional coordinates in the future, in such structures only sites within an asymmetric unit will be given. Maybe we also need a key `symmetry_operation` in addition to the key `translation`?

optimade.rst
```diff
@@ -2833,6 +2833,38 @@ assemblies
...
     However, the presence or absence of sites 0 and 1 is not correlated with the presence or absence of sites 2 and 3 (in the specific example, the pair of sites (0, 2) can occur with 0.2\*0.3 = 6 % probability; the pair (0, 3) with 0.2\*0.7 = 14 % probability; the pair (1, 2) with 0.8\*0.3 = 24 % probability; and the pair (1, 3) with 0.8\*0.7 = 56 % probability).
 
+bonds
+~~~~~
+
+- **Description**: A list describing the chemical connectivity of the structure.
+- **Type**: list of dictionary with keys:
+
+  - :property:`sites`: a list of integers (REQUIRED)
+  - :property:`translations`: a list of list of integers (OPTIONAL)
```
sauliusg, line 2843: We should probably specify very clearly that translations must be applied to the coordinates of atoms as they are currently given, and not to, e.g., coordinates reduced to the [0;1) unit cell.

janosh:
just chiming in to say thanks for working on this. 👍  would be very useful! hope it still gets merged.

merkys:
Closing as no longer relevant for the main OPTIMADE specification. It should be introduced in cheminformatics namespace, if at all.

Issue #462, opened 2023-03-29
Adding per property meta_data field.


JPBergsma:
In issue #410 the idea was raised to create a per property field for metadata.
There was still some discussion on how to name such a field. 
But for the moment, appending "_meta" to the end of the field name seems the most popular way to name this field.
I would like to use this metadata field for the ranged properties (See PR #452)
Before I can write a proposal for the metadata field: I however, still have a few questions/ ideas to discuss.

- Nested metadata fields ? Can a field in metadata have a metadata field of its own ?

- Do we want to return the metadata field when the main field is requested via response_fields ?

- Do we want to be able to “override” fields set in the property definition ? 
i.e. the value in the property definition would be the default for all entries, but individual entries can deviate from this.

merkys:
@JPBergsma internal GitHub links in your post do not seem to work, could you please fix them? This way it would be easier to see the connected issues/PR, and possibly more people from related issues/PRs would find their way here.

To answer your questions:
> * Nested metadata fields ? Can a field in metadata have a metadata field of its own ?

Intuitively I would say no, but maybe there is a use case for this?

> * Do we want to return the metadata field when the main field is requested via response_fields ?

I would say yes.

> * Do we want to be able to “override” fields set in the property definition ?

I would say no. This might overcomplicate OPTIMADE.

merkys:
> Would you consider an uncertainty margin/error bar for a value metadata ?

Good point. Currently OPTIMADE has no mechanism to communicate uncertainties and metadata could be a place to do so. Alternatively we may define properties with arbitrary suffices (for example, `_SU`), but I like metadata better.

PR #466, opened 2023-06-08
InChIKey property

merkys:
This PR adds [InChIKey property](https://en.wikipedia.org/wiki/International_Chemical_Identifier#InChIKey) for structure entries. InChIKey is a chemical structure descriptor alternative to SMILES (proposed in #368). InChIKey is said to avoid the the ambiguity the SMILES possesses, moreover, it does not have any internal structure (essentially being a "chemical checksum"), thus there should be no issues with its comparisons.

Pinging people who have expressed their interest for comments: @eimrek @utf @Austin243

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2500,6 +2500,20 @@ chemical\_formula\_anonymous
 
   - A filter that matches an exactly given formula is :filter:`chemical_formula_anonymous="A2B"`.
 
+inchikey
+~~~~~~~~
+
+- **Description**: The standard InChIKey representation of the structure.
+- **Type**: string
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+- **Examples**:
+
+  - Morphine: :val:`BQJCRHHNABKAKU-KBQPJGBKSA-N`
+
 dimension\_types
 ~~~~~~~~~~~~~~~~
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2503,7 +2503,7 @@ chemical\_formula\_anonymous
 inchikey
 ~~~~~~~~
 
-- **Description**: The standard InChIKey representation of the structure, as laid out by the [InChI Trust](https://www.inchi-trust.org)
+- **Description**: The standard InChIKey representation of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
 - **Type**: string
 - **Requirements/Conventions**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2503,7 +2503,7 @@ chemical\_formula\_anonymous
 inchikey
 ~~~~~~~~
 
-- **Description**: The standard InChIKey representation of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
+- **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
 - **Type**: string
 - **Requirements/Conventions**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2503,7 +2503,7 @@ chemical\_formula\_anonymous
 inchikey
 ~~~~~~~~
 
-- **Description**: The standard InChIKey representation of the structure.
+- **Description**: The standard InChIKey representation of the structure, as laid out by the [InChI Trust](https://www.inchi-trust.org)
 - **Type**: string
 - **Requirements/Conventions**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2503,7 +2503,7 @@ chemical\_formula\_anonymous
 inchikey
 ~~~~~~~~
 
-- **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
+- **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_.
   Standard InChIKey is not guaranteed to be unique and in extremely rare cases the same InChIKey may be assigned to several different structures.
 - **Type**: string
 - **Requirements/Conventions**:
--- a/optimade.rst
+++ b/optimade.rst
@@ -2504,6 +2504,7 @@ inchikey
 ~~~~~~~~
 
 - **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
+  Standard InChIKey is non-unique, thus the same InChIKey can be assigned to several different structures.
 - **Type**: string
 - **Requirements/Conventions**:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -2504,7 +2504,7 @@ inchikey
 ~~~~~~~~
 
 - **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
-  Standard InChIKey is non-unique, thus the same InChIKey can be assigned to several different structures.
+  Standard InChIKey is not guaranteed to be unique and in extremely rare cases the same InChIKey may be assigned to several different structures.
 - **Type**: string
 - **Requirements/Conventions**:
 


ml-evs (commented):
Thanks again @merkys, this would be a useful field to have in many cases.

Just to keep a record of the discussion, could you briefly explain the domain of applicability of this representation in this PR? If it is somehow narrow (e.g., restricted to organic molecules [I know InChI does more]), perhaps this could be a target for a new molecular/cheminformatics namespace (alongside smiles, smarts etc)?

Also just picking up on the discussion yesterday, if a subcomponent of my periodic structure can be represented by an InChIKey, should that be allowed here? e.g., a database of hybrid perovskites would probably like to expose a SMILES/InChI search on the "methylammonium" part of methylammonium lead iodide (CH3NH3PbI3) rather than by reduced formula (CH6NI3Pb -- I guess in this case the other option is to use `chemical_formula_descriptive`)

optimade.rst
```diff
@@ -2500,6 +2500,20 @@ chemical\_formula\_anonymous
 
   - A filter that matches an exactly given formula is :filter:`chemical_formula_anonymous="A2B"`.
 
+inchikey
+~~~~~~~~
+
+- **Description**: The standard InChIKey representation of the structure, as laid out by the [InChI Trust](https://www.inchi-trust.org)
```
ml-evs, line 2506:
```suggestion
- **Description**: The standard InChIKey representation of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
```

Sorry, used the wrong markup...

optimade.rst
```diff
@@ -2500,6 +2500,20 @@ chemical\_formula\_anonymous
 
   - A filter that matches an exactly given formula is :filter:`chemical_formula_anonymous="A2B"`.
 
+inchikey
+~~~~~~~~
+
+- **Description**: The standard InChIKey representation of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
```
vaitkus, line 2506: InChIKey is more of a non-unique identifier than a representation of a structure since it is constructed using a one way function (there is no way to programmatically restore the structure from an InChIKey). Wikipedia also states that InChIKey is a digital representation of the InChi, not the structure (whatever "digital representation" means).

vaitkus, line 2506: Maybe we should also specify which InChI/InChiKey version should be used?

merkys: I am for replacing "representation" with an "identifier". I recall there were situations when non-unique InChIKeys were produced, but I guess that depends on the producing software which performs intermediate transformations before feeding the structure to the InChI library. Should we put a caution in the description?

merkys: Fair point. I would also like to discuss how will we handle deprecation and switching to new InChiKey versions in the future, as I do not believe it will be possible to keep using the old version when new ones are released. I can think of the following options:

1. When InChiKey version changes, put the newest version in the upcoming OPTIMADE specification vX and require all implementers of vX to recompute their InChiKey.
2. Define the InChiKey version as metadata (after #462 gets solved).

First solution is better for interoperability, but puts more strain on the implementers.

merkys: Right, adding a disclaimer for collisions makes sense.

optimade.rst
```diff
@@ -2500,6 +2500,21 @@ chemical\_formula\_anonymous
 
   - A filter that matches an exactly given formula is :filter:`chemical_formula_anonymous="A2B"`.
 
+inchikey
+~~~~~~~~
+
+- **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
+  Standard InChIKey is non-unique, thus the same InChIKey can be assigned to several different structures.
```
vaitkus, line 2507:
```suggestion
  Standard InChIKey is not guaranteed to be unique and on extremely rare the same InChIKey may be assigned to several different structures.
```

rartino, line 2507: From the workshop: must be clarified if the key MUST cover all parts of the structure or not. Consider if this should be a list so that components can be covered, or if that goes in a separate field. (May need careful definition of what a component/fragment is).

rartino:
Workshop: after the above workshop comment has been considered/handled this should be merged.

We have discussed if "complicated derived non-unique descriptors" really belongs inside the structures entry. However, before we have a decision/design on where they should go, we should accept fields that are useful and desired into the structures endpoint for now.

A similar regard holds for the desire to allow application area/subconsortia prefixes. Until we have a model for that, we merge fields without them.

rartino:
@merkys @ml-evs What do you think - should we still add this to "core OPTIMADE" as decided in the web meet. Or, now that we are hopefully somewhat close to merging #473, should all of these 'chemically oriented fields" currently waiting in PRs by @merkys instead go in its own namespace?

These PRs are affected: #466 #465 #436 #398 - and the question is if they should be labeled 'status/blocked' (blocking on #473) or 'status/waiting-for-update' (since they all currently are sitting with comments to address before merging)

ml-evs:
> @merkys @ml-evs What do you think - should we still add this to "core OPTIMADE" as decided in the web meet. Or, now that we are hopefully somewhat close to merging #473, should all of these 'chemically oriented fields" currently waiting in PRs by @merkys instead go in its own namespace?
> 
> These PRs are affected: #466 #465 #436 #398 - and the question is if they should be labeled 'status/blocked' (blocking on #473) or 'status/waiting-for-update' (since they all currently are sitting with comments to address before merging)
> 
> **Edit:** possibly also #400, #396, or would those go in a "bio" prefix?

On purely technical/scientific terms I think this field would be perfect to seed a cheminformatics namespace (along with SMARTS/SMILES etc -- would have to figure out how to allow `filter_smarts` as a custom URL param too...). However on a purely development practice, I worry that we don't have the level of engagement or scale to spread ourselves so thinly across these various namespaces, in which case just loading up the core OPTIMADE namespace is maybe preferable. Happy to discuss!

rartino:
@ml-evs maybe we can use this example to try out the infrastructure and see where we hit snags? I'm not sure we absolutely need engagement at this point, we already have 4-6 PR:s for properties to place in such a namespace-provider standard, which we can do under a "v0.1" to mark that things are highly experimental.

I've created a couple of new repos for this:

- https://github.com/Materials-Consortia/definition-provider-template : GitHub template repo to create new definition-provider repos
- https://github.com/Materials-Consortia/namespace-cheminformatics : live repo for the cheminformatics prefix where we can try to integrate @merkys cheminformatics definitions and see how far we get.

ml-evs:
Great! Thanks for this @rartino -- I definitely stalled in my attempts to do the same thing. I will try to migrate https://github.com/Materials-Consortia/optimade-stability-namespace in the same direction.

rartino:
> Now about the remaining cheminformatics PRs. #436 introduces a new SMILES OPTIMADE data type and #398 introduces a new URI query parameter. I wonder whether property definition format and namespaces are ready to accept such extensions? If not, is this something that should be allowed to be extended in namespaces?

Once the current property defs etc. are merged, lets work on a similar design for user-defined datatypes. I'm thinking a similar declarative format as for units, properties for datatypes, where one with human language declare how every operator should work. However, I don't want to draft this until the property framework is merged.

The need for user-defined filer languages should perhaps inform the design of #398. Can we instead allow some syntax for the usual filter= to provide a list of filters with some kind of specifier what kind of filter it is? Then we can allow user-defined filter languages without new query parameters.

merkys:
> Once the current property defs etc. are merged, lets work on a similar design for user-defined datatypes. I'm thinking a similar declarative format as for units, properties for datatypes, where one with human language declare how every operator should work. However, I don't want to draft this until the property framework is merged.

Makes sense.

> The need for user-defined filer languages should perhaps inform the design of #398. Can we instead allow some syntax for the usual filter= to provide a list of filters with some kind of specifier what kind of filter it is? Then we can allow user-defined filter languages without new query parameters.

For now queries in filters act on property values only. Query by SMARTS will not be bound to a specific property. A possible solution would be to allow queries on entries themselves by introducing property-less operators, viz. `/structures?filter=SMARTS "[CX4]"`.

merkys:
I suppose that by *command line parameters* you mean *URL query parameters*. Indeed, custom query parameters are already allowed. Thus #398 already can go to cheminformatics namespace.

optimade.rst
```diff
@@ -2500,6 +2500,21 @@ chemical\_formula\_anonymous
 
   - A filter that matches an exactly given formula is :filter:`chemical_formula_anonymous="A2B"`.
 
+inchikey
+~~~~~~~~
+
+- **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_
```
vaitkus, line 2506:
```suggestion
- **Description**: The standard InChIKey identifier of the structure, as laid out by the `InChI Trust <https://www.inchi-trust.org>`_.
```

merkys:
It has been decided to move cheminformatics properties to a repository of its own, which has been done in https://github.com/Materials-Consortia/namespace-cheminformatics/pull/1 and https://github.com/Materials-Consortia/namespace-cheminformatics/pull/2. Closing this PR.

Issue #470, opened 2023-06-12
Entry info endpoints have no `id`, `type`, or `attributes`


ml-evs:
I thought I'd also opened this issue a while ago, but maybe not...

As part of the OPTIMADE file format discussions, we discovered that it is currently quite hard to use the `info/structures` (for e.g.,) endpoint response in a place where is it not attached to its URL. From the response alone it is impossible to tell which entry type the info corresponds to. We kind of break the JSON:API format by allowing entry info endpoints to present just a flat dictionary under `data`, with no `attributes`, `id` or `type`, so I think I would consider this a specification bug.

By simply adding the `id` and `type` fields to be `structures` and `info` respectively, we can then formulate a line in a file that is fully descriptive, i.e., that these properties are describing a structures entry type. As this also affects all the properties definition work (cc @rartino), I don't suggest that we force all other fields to be nested under `attributes`, though perhaps for OPTIMADE 2.0 this would be nice. @giovannipizzi also suggested perhaps allowing that at this stage to nest these fields under `attributes` and deprecating the current way in 1.2.

rartino:
Indeed, these endpoints are broken and not JSON API compliant at all. But, to allow implementations to provide everything under attributes instead of data would clearly break clients. Perhaps we shouldn't be overly afraid to move to v2.0 now that so much other things are added as well (partial data and property definitions).

ml-evs:
> Perhaps we shouldn't be overly afraid to move to v2.0 now that so much other things are added as well (partial data and property definitions).

Yes, I was thinking this in the workshop, or at least that 2.0 could follow fairly soon after 1.2 (then we write another paper? :sweat_smile:). I think at the very least I would like to write a PR that *allows* `id` and `type` to be provided alongside the non-standard top-level fields in 1.2, that way we can write the file serialisation with 1.2 for servers that support it, and "impute" it for servers that don't (i.e., my client can use the fact it is from an `info/structures` to populate the field. I will try to make a PR now...

ml-evs:
Closed by #472, though we will need to remember to enact our proposed deprecation at some point (probably 2.0...)

PR #472, opened 2023-06-13
Add `type` and `id` to the entry info endpoints

ml-evs:
Partially addresses #470, in a manner suitable for 1.2 (i.e., without backwards-incompatible changes). It seemed there was consensus in #470 that this is a bug (well, there was a thumbs up and a positive comment), and I think simply adding `type` and `id` fixes it for most use cases (i.e., downloading and storing an `/info/entry_type` endpoint whilst retaining the entry's identity).

I have added a future deprecation warning, as it seems clear to me at least that OPTIMADE v2 will just follow the JSON:API format for this, though I don't see any benefit in supporting both in v1.2.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2867,8 +2867,8 @@ species
       Elements denoting vacancies MUST have masses equal to 0.
     - **original\_name**: OPTIONAL. Can be any valid Unicode string, and SHOULD contain (if specified) the name of the species that is used internally in the source database.
 
-          **Note**: With regard to "source database", we refer to the immediate source being queried via the OPTIMADE API implementation.
-          The main use of this field is for source databases that use species names, containing characters that are not allowed (see description of the list property `species_at_sites`_).
+    **Note**: With regard to "source database", we refer to the immediate source being queried via the OPTIMADE API implementation.
+    The main use of this field is for source databases that use species names, containing characters that are not allowed (see description of the list property `species_at_sites`_).
 
   - For systems that have only species formed by a single chemical symbol, and that have at most one species per chemical symbol, SHOULD use the chemical symbol as species name (e.g., :val:`"Ti"` for titanium, :val:`"O"` for oxygen, etc.)
     However, note that this is OPTIONAL, and client implementations MUST NOT assume that the key corresponds to a chemical symbol, nor assume that if the species name is a valid chemical symbol, that it represents a species with that chemical symbol.
--- a/optimade.rst
+++ b/optimade.rst
@@ -1324,12 +1324,8 @@ The response for these endpoints SHOULD also include the following information i
 - **type**: :field-val:`"info"`.
 - **id**: This MUST precisely match the entry type name, e.g., :field-val:`"structures"` for the :endpoint:`/info/structures`.
 
-    **Note**: Future versions of the OPTIMADE API will deprecate this behavior and
-    require all keys that are not :field:`type` or :field:`id` to be under the
-    :field:`attributes` key.
-    At this point, :field:`type` and :field:`id` will be made mandatory to allow
-    the :endpoint:`/info/<entry-type>` endpoints to be understood outside the
-    initial context of the request.
+    **Note**: Future versions of the OPTIMADE API will deprecate this behavior and require all keys that are not :field:`type` or :field:`id` to be under the :field:`attributes` key.
+    At this point, :field:`type` and :field:`id` will be made mandatory to allow the :endpoint:`/info/<entry-type>` endpoints to be understood outside the initial context of the request.
     The present format is introduced for backwards-compatibility only.
 
 Example (note: the description strings have been wrapped for readability only):
--- a/optimade.rst
+++ b/optimade.rst
@@ -1314,19 +1314,14 @@ Entry listing info endpoints are accessed under the versioned or unversioned bas
 They return information related to the specific entry types served by the API.
 The response for these endpoints MUST include the following information in the :field:`data` field:
 
-- **description**: Description of the entry.
-- **properties**: A dictionary describing properties for this entry type, where each key is a property name and the value is an OPTIMADE Property Definition described in detail in the section `Property Definitions`_.
-- **formats**: List of output formats available for this type of entry.
-- **output\_fields\_by\_format**: Dictionary of available output fields for this entry type, where the keys are the values of the :field:`formats` list and the values are the keys of the :field:`properties` dictionary.
-
-The response for these endpoints SHOULD also include the following information in the :field:`data` field:
-
 - **type**: :field-val:`"info"`.
 - **id**: This MUST precisely match the entry type name, e.g., :field-val:`"structures"` for the :endpoint:`/info/structures`.
+- **description**: Description of the entry.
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -1165,19 +1165,35 @@ Entry Listing Info Endpoints
...
 - **properties**: A dictionary describing properties for this entry type, where each key is a property name and the value is an OPTIMADE Property Definition described in detail in the section `Property Definitions`_.
 - **formats**: List of output formats available for this type of entry.
 - **output\_fields\_by\_format**: Dictionary of available output fields for this entry type, where the keys are the values of the :field:`formats` list and the values are the keys of the :field:`properties` dictionary.
 
+The response for these endpoints SHOULD also include the following information in the :field:`data` field:
+
+- **type**: :field-val:`"info"`.
+- **id**: This MUST precisely match the entry type name, e.g., :field-val:`"structures"` for the :endpoint:`/info/structures`.
+
+    **Note**: Future versions of the OPTIMADE API will deprecate this behavior and
```
merkys, line 1181: The lines do not seem to be folded anywhere in the specification.

optimade.rst
```diff
@@ -1165,19 +1165,35 @@ Entry Listing Info Endpoints
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Entry listing info endpoints are accessed under the versioned or unversioned base URL serving the API as :endpoint:`/info/<entry_type>` (e.g., http://example.com/optimade/v1/info/structures or http://example.com/optimade/info/structures).
+They return information related to the specific entry types served by the API.
 The response for these endpoints MUST include the following information in the :field:`data` field:
 
 - **description**: Description of the entry.
 - **properties**: A dictionary describing properties for this entry type, where each key is a property name and the value is an OPTIMADE Property Definition described in detail in the section `Property Definitions`_.
```
merkys, line 1172: It is probably not too late to move at least `properties` inside `attributes`, but not sure which is worse, having inconsistency or forwards-incompatible structure.

merkys (changes requested):
Thanks for the PR @ml-evs, I am happy with the way it fixes the incompatibility with JSON:API and says how it should be addressed in the future. Line folding should be undone, as it does not seem to happen anywhere else in the document.

optimade.rst
ml-evs: The example I found was under `original\_name`, it gets rendered as a note, see bottom of this definition https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst#species

optimade.rst
ml-evs: I think I am strongly in favour of leaving it where it is for now, otherwise e.g., my client will break and queries to `v1` URLs will require some additional logic

ml-evs: So I'd rather just resolve the inconsistency with JSON:API (which we're not completely beholden to anyway) in a later version

optimade.rst
merkys: Right, but lines are not folded there.

merkys (dismissed):

optimade.rst
```diff
@@ -1165,19 +1165,35 @@ Entry Listing Info Endpoints
...
 - **output\_fields\_by\_format**: Dictionary of available output fields for this entry type, where the keys are the values of the :field:`formats` list and the values are the keys of the :field:`properties` dictionary.
 
+The response for these endpoints SHOULD also include the following information in the :field:`data` field:
+
+- **type**: :field-val:`"info"`.
+- **id**: This MUST precisely match the entry type name, e.g., :field-val:`"structures"` for the :endpoint:`/info/structures`.
+
+    **Note**: Future versions of the OPTIMADE API will deprecate this behavior and
+    require all keys that are not :field:`type` or :field:`id` to be under the
+    :field:`attributes` key.
```
vaitkus, line 1183:
```suggestion
    **Note**: Future versions of the OPTIMADE API will deprecate this behavior and require all keys that are not :field:`type` or :field:`id` to be under the :field:`attributes` key.
```

optimade.rst
```diff
@@ -1165,19 +1165,35 @@ Entry Listing Info Endpoints
...
+    At this point, :field:`type` and :field:`id` will be made mandatory to allow
+    the :endpoint:`/info/<entry-type>` endpoints to be understood outside the
+    initial context of the request.
```
vaitkus, line 1186:
```suggestion
    At this point, :field:`type` and :field:`id` will be made mandatory to allow the :endpoint:`/info/<entry-type>` endpoints to be understood outside the initial context of the request.
```

optimade.rst
ml-evs: Ah I see what you mean now (thanks to @vaitkus's suggestions), I think I wouldcall that wrapping rather than folding. I'll fix it and request re-review

vaitkus (approved):

optimade.rst
```diff
@@ -1165,19 +1165,31 @@ Entry Listing Info Endpoints
...
 Entry listing info endpoints are accessed under the versioned or unversioned base URL serving the API as :endpoint:`/info/<entry_type>` (e.g., http://example.com/optimade/v1/info/structures or http://example.com/optimade/info/structures).
+They return information related to the specific entry types served by the API.
 The response for these endpoints MUST include the following information in the :field:`data` field:
 
 - **description**: Description of the entry.
 - **properties**: A dictionary describing properties for this entry type, where each key is a property name and the value is an OPTIMADE Property Definition described in detail in the section `Property Definitions`_.
 - **formats**: List of output formats available for this type of entry.
 - **output\_fields\_by\_format**: Dictionary of available output fields for this entry type, where the keys are the values of the :field:`formats` list and the values are the keys of the :field:`properties` dictionary.
 
+The response for these endpoints SHOULD also include the following information in the :field:`data` field:
```
rartino, line 1176: Should we consider "MUST" here? This is a rather minor thing to add to implementations, and the response is *very* JSON API non-compliant without them; and we say higher up that we are JSON API compliant, and that any discrepancies means that the implementation still should (try?) to adhere to the JSON API specification...

optimade.rst
merkys: Right, this makes sense. Marking as resolved.

optimade.rst
merkys: Oh right. It must be the name of command line tool `fold` which affected my calling of this concept :smile:

optimade.rst
merkys: Might be a good idea to do so. This does not break the backwards compatibility for clients, right?

merkys (approved):

optimade.rst
rartino: @merkys No, not as described at the top of the specification - i.e., the client shouldn't break on unrecognized fields. And implementations should be prepared to implement new forced features (that don't break old clients) if they want to advance versions.

ml-evs: Fair point, just added this and tweaked the deprecation note.

merkys (approved):
All good as before, just needs a merge from `develop`.

ml-evs:
Yes, we're stuck in a bit of a re review loop here, as we get closer to the release I think I will start using admin powers to merge in trivial cases like this.

PR #474, opened 2023-06-14
Clarify usage of links objects vs links

ml-evs:
As raised in https://github.com/Materials-Consortia/OPTIMADE/pull/424#discussion_r1229481799, we are a bit imprecise about the difference between using a links object and a member of a links object when providing URLs.

The only difference is that "Links object" describes the total `{"links": {"self": <url>"}, {"next": {"href": "<url>", "meta": "some description"}}` style dictionary, whereas we typically just want to allo either a string URL or an annotated dictionary with keys `href` and `meta`.

This PR hopefully clarifies the wording where necessary.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -489,10 +489,10 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   :field:`meta` SHOULD also include these fields:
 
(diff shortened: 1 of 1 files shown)


rartino (approved):
Looks good to me. This is clearly a specification bug being corrected, given how we repeatedly "clarify" that a links object can be a string or an object with href and meta.

merkys (approved):

PR #467, opened 2023-06-08
Partial data

rartino:
This pull request is work by me and @sauliusg to represents a rather long discussion at the meeting among @sauliusg, @JPBergsma , @giovannipizzi , @gmrigna , myself, and others.

It attempts to generate a flexible enough format with which it is possible to transmit list property data that is too big to fit in a single response.

We have decided to separate this from the related issues of providing functionality for a client to "slice" provided data. That will be done in a separate pull request, defining a `property_ranges` parameter to the "normal" OPTIMADE endpoints. The response to such a request may (but need not) lead to data being transmitted the way described here.

PR #463, opened 2023-04-11
Add per entry and per property meta data field.

JPBergsma:
During the discussion on ranged properties, we thought it would be a good idea to have a method for storing per entry metadata for a property. This has been discussed in Issue #410 and #462.
In this PR, I have added a brief description for this metadata field, so I can use it for the ranged properties in PR #452.

I explicitly allow metadata fields to also have metadata of its own.
When a ranged property would have uncertainty values. That property would be just as large as the original ranged property, so I think it makes sense to allow metadata fields to have metadata themselves.

We can discuss this further during the optimade meeting of next Friday.

PR #480, opened 2023-06-26
Small corrections for metadata and partial data

JPBergsma:
I still have some small corrections for the metadata PR #463  and partial data PR #467

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -455,7 +455,7 @@ How this list is provided is response format-dependent.
 For the JSON response format, see the description of the :field:`partial_data_links` field, nested under :field:`data` and then :field:`meta`, in the section `JSON Response Schema: Common Fields`_.
 
 The default partial data format is named "jsonlines" and is described in the Appendix `OPTIMADE JSON lines partial data format`_.
-An implementation SHOULD always include this format as one of alternative partial data formats provided for a property that has been omitted from the response to the initial query.
+An implementation SHOULD always include this format as one of the partial data formats provided for a property that has been omitted from the response to the initial query.
 Implementations MAY provide links to their own non-standard formats, but non-standard format names MUST be prefixed by a database-provider-specific prefix.
 
 Below follows an example of the :field:`data` and :field:`meta` parts of a response using the JSON response format that communicates that the property value has been omitted from the response, with three different links for different partial data formats provided.
@@ -505,7 +505,7 @@ For example, the metadata property definition of the field :field:`_exmpl_exampl
 The reason for this limitation is to avoid name collisions with metadata fields defined by the OPTIMADE standard in the future that apply also to database-specific data fields.
 
 Implementation of the :field:`meta` field is OPTIONAL.
-However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all fields present in the data part of the response which has metadata.
+However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all fields, present in the data part of the response, which have metadata.
 
 Example of a response in the JSON response format with two structure entries that each include a metadata property for the attribute field :field:`element_ratios` and the database-specific per entry metadata field :field:`_exmpl_originates_from_project` :
 
@@ -726,11 +726,11 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   - **partial_data_links**: an object used to list links which can be used to fetch data that has been omitted from the :field:`data` part of the response.
     The keys are the names of the fields in :field:`attributes` for which partial data links are available.
-    Each value is a list of items that MUST have the following keys:
+    Each value is a list of objects that MUST have the following keys:
 
     - **format**: String.
-      A name of the format provided via this link.
-      One of the items SHOULD be "jsonlines", which refers to the format in `OPTIMADE JSON lines partial data format`_.
+      The name of the format provided via this link.
+      For one of the objects this :field:`format` field SHOULD have the value "jsonlines", which refers to the format in `OPTIMADE JSON lines partial data format`_.
 
     - **link**: String.
       A `JSON API link <http://jsonapi.org/format/1.0/#document-links>`__ that points to a location from which the omitted data can be fetched.
@@ -3605,7 +3605,7 @@ The markers have been deliberately designed to be valid JSON objects but *not* v
 Since the OPTIMADE list data type is defined as a list of values of the same data type or :val:`null`, the above markers cannot be encountered inside the actual data of an OPTIMADE property.
 
   **Implementation note:** the recognizable string values for the markers should make it possible to prescreen the raw text of the JSON data lines for the reference-marker string to determine which are the lines that one can exclude from further processing to resolve references (alternatively, this screening can be done by the string parser used by the JSON parser).
-  The undelying design idea is that for lines that have reference-markers, the time it takes to process the data structure to locate the markers should be negligible compared to the time it takes to resolve and handle the large data they reference.
+  The underlying design idea is that for lines that have reference-markers, the time it takes to process the data structure to locate the markers should be negligible compared to the time it takes to resolve and handle the large data they reference.
   Hence, the most relevant optimization is to avoid spending time processing data structures to find markers for lines where there are none.
 
 The full response MUST be valid `JSON Lines <https://jsonlines.org/>`__ that adheres to the following format:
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -506,7 +506,7 @@ For example, the metadata property definition of the field :field:`_exmpl_exampl
 The reason for this limitation is to avoid name collisions with metadata fields defined by the OPTIMADE standard in the future that apply also to database-specific data fields.
 
 Implementation of the :field:`meta` field is OPTIONAL.
-However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all fields present in the data part of the response which has metadata.
+However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all fields, present in the data part of the response, which have metadata.
```
merkys, line 509:
```suggestion
However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all properties, present in the data part of the response, which have metadata.
```
I however prefer the variant without introduced commas.

optimade.rst
```diff
@@ -751,11 +751,11 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
 
   - **partial_data_links**: an object used to list links which can be used to fetch data that has been omitted from the :field:`data` part of the response.
```
merkys, line 752:
```suggestion
  - **partial_data_links**: a dictionary used to list links which can be used to fetch data that has been omitted from the :field:`data` part of the response.
```

optimade.rst
```diff
@@ -751,11 +751,11 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
...
     The keys are the names of the fields in :field:`attributes` for which partial data links are available.
-    Each value is a list of items that MUST have the following keys:
+    Each value is a list of objects that MUST have the following keys:
```
merkys, line 754:
```suggestion
    Each value is a list of dictionaries that MUST have the following keys:
```
Elsewhere in the specification we use "list of dictionaries"

optimade.rst
```diff
@@ -3693,7 +3693,7 @@ The header object MUST contain the keys:
 
 The following key is RECOMMENDED in the header object:
 
-- :field:`"returned_ranges"`: Array of Object.
+- :field:`"returned_ranges"`: Array of Objects.
```
merkys, line 3696:
```suggestion
- :field:`"returned_ranges"`: list of dictionaries.
```

optimade.rst
JPBergsma: Do you mean you would prefer if I only change has to have ?
```suggestion
However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all fields present in the data part of the response which have metadata.
```
Without the commas, I get the impression that which refers to response rather than fields.

How about:
```suggestion
However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all properties which have metadata and are present in the data part of the response.
```

optimade.rst
JPBergsma: I think we should still add capitalization as on line 2079.
```suggestion
- :field:`"returned_ranges"`: List of Dictionaries.
```

merkys: I must have confused OPTIMADE types with JSON types, apologies. I now see that
> Note: since the below definition references both JSON fields and OPTIMADE properties, the data type names depend on context: for JSON they are, e.g., "array" and "object" and for OPTIMADE properties they are, e.
g., "list" and "dictionary".

Thus here and above the JSON data types should have been given instead of OPTIMADE. I think this is confusing, but distinction should remain. It makes sense to talk in JSON terms when defining OPTIMADE JSON lines format.

I see that capitalization in this section of specification prefers initial capital letter, thus my suggestions for capitalization are inconsistent with the rest of this section. Sorry again.

JPBergsma: Ok than I'll reverse your suggestions.

merkys: Right, thank you.

optimade.rst
```diff
@@ -3684,16 +3684,16 @@ The header object MUST contain the keys:
 
   It MUST contain the following key:
 
-  - :field:`"format"`: String.
+  - :field:`"version"`: String.
```
JPBergsma, line 3687: We already used the field name "format" in section 3.10 Transmission of large property values. So I do not think we should use it here again with a different meaning.

optimade.rst
```diff
@@ -3684,16 +3684,16 @@ The header object MUST contain the keys:
...
     Specifies the minor version of the partial data format used. The string MUST be of the format "MAJOR.MINOR", referring to the version of the OPTIMADE standard that describes the format. The version number string MUST NOT be prefixed by, e.g., "v". In implementations of the present version of the standard, the value MUST be exactly :val:`1.2`.
```
vaitkus, line 3688:
```suggestion
    Specifies the minor version of the partial data format used.
    The string MUST be of the format "MAJOR.MINOR", referring to the version of the OPTIMADE standard that describes the format.
    The version number string MUST NOT be prefixed by, e.g., "v". In implementations of the present version of the standard, the value MUST be exactly :val:`1.2`.
```

optimade.rst
rartino: Good catch - it is indeed potentially confusing that these two fields are named the same. However, the use of "format" here reflects the same fieldname in the property definitions, and the design idea is the same: future versions of OPTIMADE may want to make this field more expressive than simply the OPTIMADE version (which is why the next line of the text says how to handle if the string cannot be parsed as a version number). For this, I still think "format" is the best name.

So, can we change the other "format" instead? I find it a bit tricky to come up with a good name. How about "protocol"? "data_protocol"?

optimade.rst
rartino: Commenting on the original proposed change: I see the specification currently is not consistent whether type declarations like these of lists/arrays should use plural or singular. E.g. in section "8.6.1. species" the type is "list of dictionary". Whereas in "8.2.15 species_at_sites" it is "list of strings".

My personal preference (which perhaps comes a bit from a programmer-first perspective) is to not mess around with pluralization in a declarative statement like this, just use the type name as we have defined it. So, for the above IMO it should be "Array of Object". However, in the bulk text where one uses full sentences I would write, e.g., "The value should be an array of objects". But I don't feel very strongly either way, so if no one backs me up on this view, I'm OK approving this change.

rartino (changes requested):
Thanks for making these fixes. The only thing I think needs to be handled before merging this is to agree what to do with the double "format" fields where - as you can see below - I don't agree with changing the name of the field embedded in the data. (But if others chime in and agree with plurals in the type declaration, I suggest that be reverted too).

optimade.rst
JPBergsma: I have changed "version" back into "format".
I will change the format field in section 3.10 into "file_format" as that seems to be what it describes in the example.
Protocol is more general than file format, so it would probably be better to use that for a method to share larger data items in a manner that is very different from the current partial data format.

vaitkus: What about "data_format" as an alternative name for the property in Section 3.10? This term is already used in the human-readable portion of the specification, e.g. `<...> The default partial data format is named "jsonlines" and is described <...>`?

However, I also find the name "format" in the current context a little bit unintuitive (makes me think about file formats and data types instead of version-like strings). Maybe something like "format_version" would also be viable? (I understand that it might be too late to change it due to "format" already being used in property definitions).

rartino: You are right that the naming may not be perfect here, and since we haven't released any of this yet, it isn't too late to change if we can find something that is consistently better. It may require a 'find + sed' through the not-yet-merged property definition source files, but that is not so much work.

`data_format` seems more accurate than `file_format`, but I think the other "format" field name in that case also needs a qualifier (which is why I tried to get rid of "format" completely, e.g., "protocol"). 

But, at some level I wonder if these things describing "what is the format of this thing" shouldn't just be the same? The structured way to go would be that `format` is a dictionary with a `name` and a `version`, potentially with the `version` part optional. However, `version` is in that case also not the ideal name for that subfield, since we want to chop off the patch part.

optimade.rst
JPBergsma: I have undone this change. It would probably be best to open a new issue to discuss this for the optimade specification as a whole.

rartino: Yes, please open an issue. It also concerns the auto-generated documentation pages from the property definitions published under schema.optimade.org. (Right now the human-readable pages only presents the outermost type, but that should be fixed and use whatever we think is right.)

optimade.rst
JPBergsma: I think that the format consists of two parts.

1. The part we define in Optimade, i.e. the names and definitions of the fields and how these are arranged in data structures

2. The container format, i.e. how data types and data structures are stored in a file / response.

The current format field under “partial_data_links” is most similar to the container format, so we should perhaps rename that field to “container_format”.
Elsewhere, we have however also used the term “response_format” so that is also an option. 

The “format” field that we supply in the partial data response mostly corresponds with point 1.
It is meant to describe which version of the optimade specification was used to construct the partial data response. At the moment this field gives the optimade version but not which protocol was used.

Now, we have only one way of returning large data items, so it is not necessary to specify that the partial data protocol is used. If in the future we, however, come up with more methods to return data, it could be useful to have a field specifying the used method. So I think it would be better to keep the version number and the used method separate, like @rartino already suggested. 
> The structured way to go would be that format is a dictionary with a name and a version

I am also not happy with “property-format” as a field name in the property definitions. As it describes the version of the definition of the property and not how the property is formatted.

`data_format` sound a bit vague to me, so I do not like that name.

optimade.rst
JPBergsma: I have created Issue #483 to discuss this further.

ml-evs (approved):

rartino (approved):
Looks good to me, but we need to figure out in a separate PR what to do with the double use of 'format'.

PR #488, opened 2023-12-23
Add ECMA definition of symmetry operation regexp and associated tests

ml-evs:
Closes #487.

This PR adds an explicit and simplified symmetry operation regexp that can be used with JSON Schema.
It is validated against the same test cases as the PCRE version with an exact match in the results.

PR #490, opened 2024-01-06
Define OPTIMADE regex format

rartino:
We keep coming back to the lack of regex:es in the OPTIMADE specification, in particular lately:

- In the pattern validation for OPTIMADE property definitions (see #488)
- In the more advanced string search operator #160.

The challenges with picking one specific regex format has been debated quite intensely all the way back to, e.g.,  #42, for more recent discussions see far down in #160.

This PR creates a free-standing appendix that describes what I believe works as a regex format for OPTIMADE (see in particular the discussions far down in #160 to see the motivations for this particular format).
No parts of the specification itself yet refers to this appendix. This is intended to be added in #488 and #160.

It would be desirable to merge this fairly soon since it appears we have painted ourselves into a corner to a degree with regards to the need for the `pattern` JSON Schema validation field to properly validate the new `space_group_symmetry_operations_xyz` property.


optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+
+OPTIMADE Regular Expression Format
+----------------------------------
+This section defines a string representation for regular expressions (regexes) to be referred to from other parts of the specification.
+This format will be referred to as an "OPTIMADE regex."
+Depending on the context, a delimiter may be required to enclose the regex (e.g., double quotes or a slash character), and some outer-level escape rule may apply (e.g., to distinguish an enclosing double quote from one that is part of the regex).
+Such delimiters and escape rules are not defined as part of the OPTIMADE regex format itself and has to be clarified when this format is referenced.
+
+The format is a subset of the format described in `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__.
+The format is closely inspired by the subset recommended in the JSON schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
```
rartino, line 3972:
```suggestion
The format is closely inspired by the subset recommended in the JSON Schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
```

optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
     [3,7,19, ["PARTIAL-DATA-REF", ["https://example.db.org/value2"]]]
     [4,5,19, [ [11, 110], ["PARTIAL-DATA-REF", ["https://example.db.org/value3"]], [550, 333]]]
     ["PARTIAL-DATA-END", [""]]
+
+OPTIMADE Regular Expression Format
+----------------------------------
+This section defines a string representation for regular expressions (regexes) to be referred to from other parts of the specification.
+This format will be referred to as an "OPTIMADE regex."
+Depending on the context, a delimiter may be required to enclose the regex (e.g., double quotes or a slash character), and some outer-level escape rule may apply (e.g., to distinguish an enclosing double quote from one that is part of the regex).
+Such delimiters and escape rules are not defined as part of the OPTIMADE regex format itself and has to be clarified when this format is referenced.
```
vaitkus, line 3969:
```suggestion
Such delimiters and escape rules are not defined as part of the OPTIMADE regex format itself and have to be clarified when this format is referenced.
```

optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+
+The format is a subset of the format described in `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__.
+The format is closely inspired by the subset recommended in the JSON Schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
+However, OPTIMADE has decided to restrict the subset further to better align with the features available in common database backends and clarified that the escape character token can be used with the meaning defined by ECMA-262, section 21.2.1.
```
vaitkus, line 3973:
```suggestion
However, OPTIMADE has decided to restrict the subset further to better align it with the features available in common database backends and clarified that the escape character token can be used with the meaning defined by ECMA-262, section 21.2.1.
```

optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
     [3,7,19, ["PARTIAL-DATA-REF", ["https://example.db.org/value2"]]]
     [4,5,19, [ [11, 110], ["PARTIAL-DATA-REF", ["https://example.db.org/value3"]], [550, 333]]]
     ["PARTIAL-DATA-END", [""]]
+
+OPTIMADE Regular Expression Format
+----------------------------------
+This section defines a string representation for regular expressions (regexes) to be referred to from other parts of the specification.
+This format will be referred to as an "OPTIMADE regex."
+Depending on the context, a delimiter may be required to enclose the regex (e.g., double quotes or a slash character), and some outer-level escape rule may apply (e.g., to distinguish an enclosing double quote from one that is part of the regex).
```
vaitkus, line 3968:
```suggestion
Depending on the context, a delimiter may be required to enclose the regex (e.g., double quotes or a slash character), and some outer-level escape rules may apply (e.g., to distinguish an enclosing double quote from one that is part of the regex).
```

ml-evs:
I'm not sure I see the need to couple the regexp format for search (e.g., a future `LIKE`) and the JSON Schema field `pattern` for constraining the shape of allowed values of fields (and in our only example so far, a field where we explicitly state that search is not recommended), so for me the only thing missing is to simply allow `pattern` in our property definitions. Perhaps this comes down to me not seeing a use case for the suggested `LIKE` operator given the current use of string fields though.

rartino:
@ml-evs 

> I'm not sure I see the need to couple the regexp format for search (e.g., a future LIKE) and the JSON Schema field pattern for constraining the shape of allowed values of fields

We now have (at least) two different places in the specification for regex:es where we intend to standardize on subsets of PCRE2 so that not all implementations are forced to handle the full complexity of all of PCRE2 (or possibly worse, the specific ECMA-262 flavor of it, which may not even be 100% compatibly supported by the PCRE2 library in a few years). Does it not make sense to make these regex formats exactly the same?

I suppose a property validator usually can be written with access to the full PCRE2 library. However, one may want to give a bit of consideration to the possibility of validating other data formats against these schemas - in which case it would be nice, e.g., if the `pattern` property definition regex is restrained enough to be translatable into, e.g., an [XML Schema regex](https://www.regular-expressions.info/xml.html), etc.

sauliusg (changes requested):

optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+
+The format is a subset of the format described in `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__.
+The format is closely inspired by the subset recommended in the JSON Schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
+However, OPTIMADE has decided to restrict the subset further to better align it with the features available in common database backends and clarified that the escape character token can be used with the meaning defined by ECMA-262, section 21.2.1.
+
+Hence, an OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__ (when processed with Unicode support) that only uses the following tokens and features (this list is partially quoted from the JSON Schema standard):
+
+- Individual Unicode characters, as defined by the `JSON specification <https://json-schema.org/draft/2020-12/json-schema-core#RFC8259>`__.
+- The escape character (``\``) with the functionality described in `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__.
```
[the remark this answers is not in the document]
rartino: I agree that the ECMA standards document isn't great. 

Yet, if we want this regex appendix to also describe the 'pattern' field inherited into OPTIMADE from JSON Schema, then, as I see it, we must use ECMA-262 as the defining source to make it clear that our definition adheres to the JSON Schema standard by being a subset of what is allowed by that standard.

That said, we can copy, e.g., table 34 from the ECMA standard to help clarify what that means.

If we can help it, I really would want to avoid the OPTIMADE standard encompassing two different subsets of PCRE2.

optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+- Simple character classes (e.g., ``[abc]``) and range character classes (e.g., ``[a-z]``).
+- Complemented character classes (e.g., ``[^abc]``, ``[^a-z]``)
+- Simple quantifiers: ``+`` (one or more), ``*`` (zero or more), ``?`` (zero or one).
+- The beginning-of-input (``^``) and end-of-input (``$``) anchors.
+- Simple grouping (``(...)``) and alternation (``|``).
+
+Note that compared to the JSON Schema standard, lazy quantifiers (``+?``, ``*?``, ``??``) are NOT included, nor are range quantifiers (``{x}``, ``{x,y}``, ``{x,}``).
+
+Unless explicitly anchored, the expression is evaluated unanchored.
+For example, the OPTIMADE regex "es" matches "expression".
```
rartino, line 3988:
```suggestion
The expression matches the string at any position unless it contains a leading beginning-of-input (``^``) or trailing end-of-input (``$``) anchor listed above, i.e., the anchors are not implicitly assumed.
For example, the OPTIMADE regex "es" matches "expression".
```

optimade.rst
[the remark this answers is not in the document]
rartino: Now that we dive into the ECMA standard - my appendix text has tried to stay as close as possible to [the JSON Schema standard sec 6.4](https://json-schema.org/draft/2020-12/json-schema-core#section-6.4), which specifies that the expressions should be evaluated assuming the ['u' flag of the ECMA standard](https://262.ecma-international.org/11.0/#sec-notation). However, I see that this does not sufficiently define the other possible flags in that section:
 
* s = dot matches newline
* i = case insensitive matching
* m = anchors matches newlines

We probably should clarify that the regex format we define are for all of these off (?), or more succinctly that all variables that can be set by the RegExp internal slot according to 21.2.2.1 should be false, except for Unicode.

optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+Such delimiters and escape rules are not defined as part of the OPTIMADE regex format itself and have to be clarified when this format is referenced.
+
+The format is a subset of the format described in `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__.
+The format is closely inspired by the subset recommended in the JSON Schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
+However, OPTIMADE has decided to restrict the subset further to better align it with the features available in common database backends and clarified that the escape character token can be used with the meaning defined by ECMA-262, section 21.2.1.
+
+Hence, an OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__ (when processed with Unicode support) that only uses the following tokens and features (this list is partially quoted from the JSON Schema standard):
```
rartino, line 3975:
```suggestion
Hence, an OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__.
The regex is interpreted according to the processing rules that apply for an expression where only the Unicode variable is set to true of all variables set by the RegExp internal slot described by [Sec. 21.2.2.1](https://262.ecma-international.org/11.0/#sec-notation).
Furthermore, it can only use the following tokens and features (this list is partially quoted from the JSON Schema standard):
```

optimade.rst
```diff
@@ -3960,3 +3960,29 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+- The escape character (``\``) with the functionality described in `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/>`__.
+- Simple character classes (e.g., ``[abc]``) and range character classes (e.g., ``[a-z]``).
+- Complemented character classes (e.g., ``[^abc]``, ``[^a-z]``)
+- Simple quantifiers: ``+`` (one or more), ``*`` (zero or more), ``?`` (zero or one).
+- The beginning-of-input (``^``) and end-of-input (``$``) anchors.
+- Simple grouping (``(...)``) and alternation (``|``).
+
+Note that compared to the JSON Schema standard, lazy quantifiers (``+?``, ``*?``, ``??``) are NOT included, nor are range quantifiers (``{x}``, ``{x,y}``, ``{x,}``).
+
+Unless explicitly anchored, the expression is evaluated unanchored.
```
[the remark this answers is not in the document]
rartino: Please check the updated PR.

optimade.rst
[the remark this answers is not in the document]
rartino: Pinging @sauliusg for another look here: if you agree with my argument that we need to use ECMA as the defining source (but we can further clarify what that means if we want), and if you are otherwise happy with this PR, can you please approve it as it is, and we can jointly work out and amend this appendix with clarifications at a future point? As far as I know, the property definitions PR currently is stalled *only*  on the need to refer to a regex format for the 'pattern' string validation, which this PR is meant to fix.

optimade.rst
```diff
@@ -3963,3 +3963,31 @@ An example of the sparse layout for multidimensional lists with three aggregated
     [3,7,19, ["PARTIAL-DATA-REF", ["https://example.db.org/value2"]]]
     [4,5,19, [ [11, 110], ["PARTIAL-DATA-REF", ["https://example.db.org/value3"]], [550, 333]]]
     ["PARTIAL-DATA-END", [""]]
+
+OPTIMADE Regular Expression Format
+----------------------------------
+This section defines a string representation for regular expressions (regexes) to be referred to from other parts of the specification.
+This format will be referred to as an "OPTIMADE regex."
```
ml-evs, line 3970:
```suggestion
This format will be referred to as an "OPTIMADE regex".
```

ml-evs (dismissed):
Only one formatting issue, otherwise good for me

optimade.rst
```diff
@@ -3963,3 +3963,31 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+Depending on the context, a delimiter may be required to enclose the regex (e.g., double quotes or a slash character), and some outer-level escape rules may apply (e.g., to distinguish an enclosing double quote from one that is part of the regex).
+Such delimiters and escape rules are not defined as part of the OPTIMADE regex format itself and have to be clarified when this format is referenced.
+
+The format is a subset of the format described in `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/#sec-patterns>`__.
+The format is closely inspired by the subset recommended in the JSON Schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
+However, OPTIMADE has decided to restrict the subset further to better align it with the features available in common database backends and clarified that the escape character token can be used with the meaning defined by ECMA-262, section 21.2.1.
+
+Hence, an OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/#sec-patterns>`__.
```
merkys, line 3978: To me this sounds as "OPTIMADE regex is **any** regular expression that adheres to ECMA-262, section 21.2.1" which is not true. Could this be paraphrased to dispel the possible ambiguity?

rartino: I'm not sure I follow. What would be a regex that adheres to ECMA-262 sec 21.2.1 and follows the additional clauses in the consecutive sentences and which you do not consider an OPTIMADE regex?

(It also sounds as if you want to read "A is B" to mean "All B are A"? Is that the issue?)

merkys: > (It also sounds as if you want to read "A is B" to mean "All B are A"? Is that the issue?)

Yes. In other words, I read this sentence just as "A square is a polygon that has four vertices" which is true, but not invertible, although I find the construction of the sentence hinting equivalence from both sides.

vaitkus: Would it be more clear to rephrase it as:
```
Hence, an OPTIMADE regular expression MUST adhere to `ECMA-262, section 21.2.1`.
```
or
```
Hence, all OPTIMADE regular expressions adhere to `ECMA-262, section 21.2.1`.
```
?

rartino: @merkys ... or is this what you are looking for? 
```suggestion
Hence, an OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/#sec-patterns>`__ with the additional restrictions described in the following.
```

Personally I don't see how this is needed. This type of phrasing "A is a B. A adheres to additional requirement C and D. Furthermore, restriction E also applies" seems quite common in this kind of formal text and is as far as I can see already present throughout the specification, see e.g.:

>   - **api\_version**: a string containing the full version of the API implementation.
>    The version number string MUST NOT be prefixed by, e.g., "v".
 >   Examples: :field-val:`1.0.0`, :field-val:`1.0.0-rc.2`.

I am however ok with merging the change above, if you feel strongly it clarifies the text.

optimade.rst
rartino: It turns out that this is a British vs American style difference. 
  
- https://en.wikipedia.org/wiki/Quotation_marks_in_English#Order_of_punctuation

What do we use for the OPTIMADE specification?

rartino: Ok, I'll merge this now since those that are present seem to be positive to the change.

However, we probably should more generally decide whether the specification uses British or American conventions and spelling.

Note that the American (illogical?) style usually still means placing the punctuation outside the quotations when it matters; from Wikipedia *"many American style guides explicitly permit periods and commas outside the quotation marks when the presence of the punctuation mark inside the quotation marks leads to ambiguity, such as when describing keyboard input"*

optimade.rst
merkys: I am OK with either your or @vaitkus suggestion. Also feel free to dismiss my comment. Just as an outsider wrt this PR (I missed the last meeting), the first time I read it the cited sentence threw me off the trail.

ml-evs (dismissed):

optimade.rst
sauliusg: IMHO adding the clarification "with the additional restrictions described in the following" is helpful to maintain unambiguous text.

optimade.rst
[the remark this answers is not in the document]
sauliusg: When I looked it up I actually meant table 34. Now I read that "Depending on the context, a delimiter may be required to enclose the regex (e.g., double quotes ...", so escaping a double quote will be necessary and it is not in the table 54 (at least not explicitly), but it is in the table 34.

But honestly, I can not follow the ECMA definition of REs to see what the consequences of those definitions in two different places are, not at all if I just look at their web site... And this is after 20+ years working with REs... :/ I can not see how ECMA can be useful or relevant for anything.

There is a POSIX RE standard. There are of course also PCREs, but since we take a subset of those this will be a common subset. So why not settling on:

ISO/IEC 9945-2:1993 Information technology – Portable Operating System Interface (POSIX) – Part 2: Shell and Utilities, successively revised as ISO/IEC 9945-2:2002 Information technology – Portable Operating System Interface (POSIX) – Part 2: System Interfaces, ISO/IEC 9945-2:2003, and currently ISO/IEC/IEEE 9945:2009 Information technology – Portable Operating System Interface (POSIX) Base Specifications, Issue 7

or something like that...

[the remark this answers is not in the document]
sauliusg: Are we going to support escapes like '\cJ' (meaning '\n')?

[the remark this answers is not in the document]
rartino: *For the subset we are trying to define here*, do you know if there are any differences between the POSIX RE standard you suggest and PCRE? If there is, that would be a problem, since many of us want to use PCRE engines to handle OPTIMADE RE:s without having to translate them.

Anyway, from what you suggest, I guess these are our main options:

1. Accept using a subset of the ECMA-262 standard - which - while poorly written - is widely adopted by the world as a standard underpinning (dialects of) javascript.

2. Define a RE subset that *in practice* is compatible with the ECMA-262 standard and PCRE, but not explicitly so. I suppose this does work in practice, just that it will generate doubt among our users whether the definition of `pattern` in our property definitions is compatible with the JSON schema definition of `pattern` or not.

3. Define a RE subset that isn't compatible with the ECMA-262 standard. We then, on this singular point, break the connection between our property definitions and JSON Schema.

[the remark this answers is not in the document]
sauliusg: > _For the subset we are trying to define here_, do you know if there are any differences between the POSIX RE standard you suggest and PCRE? If there is, that would be a problem, since many of us want to use PCRE engines to handle OPTIMADE RE:s without having to translate them.
> 
> Anyway, from what you suggest, I guess these are our main options:
> 
>     1. Accept using a subset of the ECMA-262 standard - which - while poorly written - is widely adopted by the world as a standard underpinning (dialects of) javascript.
> 
>     2. Define a RE subset that _in practice_ is compatible with the ECMA-262 standard and PCRE, but not explicitly so. I suppose this does work in practice, just that it will generate doubt among our users whether the definition of `pattern` in our property definitions is compatible with the JSON schema definition of `pattern` or not.
> 
>     3. Define a RE subset that isn't compatible with the ECMA-262 standard. We then, on this singular point, break the connection between our property definitions and JSON Schema.
> 
> 
> > Are we going to support escapes like '\cJ' (meaning '\n')?
> 
> I don't see that listed explicitly in the subset, so my take would be no. We can clarify that if you want.

I think it is important that we clarify it, otherwise we risk different understanding of the OPTIMADE standard. My understanding after plowing through the ECMA was that we *do* mandate these escapes (actually, all escapes...) at the moment.

rartino:
Conclusions from web meet discussions + a post-meeting discussion with @sauliusg 

1. Lets use ECMA-262 as the basis of our definitions.
2. Lets make very sure our subset is compatible with both ECMA-262 (in unicode mode) and PCRE. 
3. Lets be very clear about what we do and do not support in terms of escapes.
4. Lets support as many escapes of \ + a printable ASCII character outside a-z,A-Z as possible, for which the meaning is literally that character (in both ECMA-262 and PCRE).
5. Lets state that the behavior is undefined for REs outside our defined subset.

So, things that has to be done: 

* Research out what applies for point (4)
* Fix this PR to include (3) and (5)

Some notes from my digging:

Summarizing info [from Wikipedia](https://en.wikipedia.org/wiki/ECMAScript_version_history):
* ECMA-262 describes ECMAScript, the standard for browser "javascript" engines which has been released in 14 versions so far.
* The 9th 2018 release is the last time I find references to possible syntax changes to RegExps, documented as best I can see here: https://github.com/tc39/ecma262/releases/tag/es2016-draft-201607-25.
* There is currently no reference implementation (maybe there was back in ~ 2007), but there is an official test suite: https://github.com/tc39/test262
* Here is the Regex part of the official tests: https://github.com/tc39/test262/tree/main/test/built-ins/RegExp
* I see people being pointed to https://github.com/engine262/engine262 for a standaline standards-conforming implementation. 
* There is also a strict mode in Firefox' implementation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode
* (As an example of how tricky this whole thing is, here is thread discussing disagreement between browser implementations and how backslash + literal newline is handled: https://bugzilla.mozilla.org/show_bug.cgi?id=92944 )

From the above I have some trust in that what Mozilla documents in their [regular expression syntax cheat sheet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Cheatsheet) is a reasonable interpretation of the standard. The top table on that page has a very nice entry for `\` saying:

> Indicates that the following character should be treated specially, or "escaped". It behaves one of two ways.
>
> * For characters that are usually treated literally, indicates that the next character is special and not to be interpreted literally. For example, /b/ matches the character "b". By placing a backslash in front of "b", that is by using /\b/, the character becomes special to mean match a word boundary.
> * For characters that are usually treated specially, indicates that the next character is not special and should be interpreted literally. For example, "*" is a special character that means 0 or more occurrences of the preceding character should be matched; for example, /a*/ means match 0 or more "a"s. To match * literally, precede it with a backslash; for example, /a\*/ matches "a*".
>
> Note that some characters like :, -, @, etc. neither have a special meaning when escaped nor when unescaped. Escape sequences like \:, \-, \@ will be > equivalent to their literal, unescaped character equivalents in regular expressions. However, in regular expressions with the unicode flag, these will cause an invalid identity escape error. This is done to ensure backward compatibility with existing code that uses new escape sequences like \p or \k. 

So, according to the last paragraph - unicode mode (which JSON Schema specifies should apply) - an escaped printable character that doesn't have a defined meaning when escaped causes an "invalid identity escape error". That means that the only printable ASCII characters that are safe to escape to mean their literal versions are the ones that the ECMA-262 standard lists as ["SyntaxCharacter"](https://262.ecma-international.org/14.0/?_gl=1*1yn5qxx*_ga*MjEzNTE2ODEyNi4xNzA0NTQ1NTk5*_ga_TDCK4DWEPP*MTcwNzUzMTk2NC41LjAuMTcwNzUzMTk2NC4wLjAuMA..#prod-SyntaxCharacter): `^ $ \ . * + ? ( ) [ ] { } |`.

This also matches my tests using Firefox' strict mode by, e.g.,
```
"use strict";

var regex = /\,/u;
var str1 = 'hello, world';
var str2 = 'hello world';
var result1 = regex.test(str1);
var result2 = regex.test(str2);
console.log(result1, result2);
```
when run in the console of the Firefox web developer tool triggers `SyntaxError: invalid identity escape in regular expression error`. Removing the unicode mode (`u`) makes the RE match the literal `,`. Replacing the commas with any of the ECMA-262 "SyntaxCharacters" gives the expected literal matches.

**Hence, my conclusion is that we safely support ONLY these literal escapes: `^ $ \ . * + ? ( ) [ ] { } |`.**

(Plus any character classes that we want to support that are the same between ECMA-262 and PCRE.)

optimade.rst
```diff
@@ -3964,3 +3964,64 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+
+Hence, an OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/#sec-patterns>`__ with the additional restrictions described in the following.
+The regex is interpreted according to the processing rules that apply for an expression where only the Unicode variable is set to true of all variables set by the RegExp internal slot described by `ECMA-262, section 21.2.2.1 <https://262.ecma-international.org/11.0/#sec-notation>`__.
+Furthermore, it can only use the following tokens and features (this list is partially quoted from the JSON Schema standard):
+
+- Individual Unicode characters, as defined by the `JSON specification <https://json-schema.org/draft/2020-12/json-schema-core#RFC8259>`__.
+- A literal escape of one of the syntax characters, i.e., the escape character (``\``) followed by one of the following characters ``^ $ \ . * + ? ( ) [ ] { } |`` to represent that literal character.
+- Simple character classes (e.g., ``[abc]``) and range character classes (e.g., ``[a-z]``) with the following constraints:
+
+  * The class has to be ordered so that it does not start with the character ``[``.
```
rartino, line 3992:
```suggestion
```

optimade.rst
```diff
@@ -3964,3 +3964,63 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+  * XML Schema regexes are compatible except that they are implicitly anchored: i.e., the beginning-of-input ``^`` and end-of-input ``$`` anchors must be removed, and missing anchors replaced by ``.*``.
+  * SQLite supports regexes via libraries and thus can use a compatible format (e.g., PCRE2).
+  * Basic POSIX regular expressions differ from the defined format in that it requires groupings to be escaped, i.e. ``\(``, ``\)``.
```
rartino, line 4026:
```suggestion
  * Basic POSIX regular expressions differ from the defined format in that they require several of the metacharacters to be escaped, e.g. ``\(``, ``\)``.
```

rartino: Fix minor grammar error

optimade.rst
```diff
@@ -3964,3 +3964,63 @@ An example of the sparse layout for multidimensional lists with three aggregated
     [3,7,19, ["PARTIAL-DATA-REF", ["https://example.db.org/value2"]]]
     [4,5,19, [ [11, 110], ["PARTIAL-DATA-REF", ["https://example.db.org/value3"]], [550, 333]]]
     ["PARTIAL-DATA-END", [""]]
+
+OPTIMADE Regular Expression Format
+----------------------------------
+This section defines a unicode string representation of regular expressions (regexes) to be referenced from other parts of the specification.
```
vaitkus, line 3970:
```suggestion
This section defines a Unicode string representation of regular expressions (regexes) to be referenced from other parts of the specification.
```

optimade.rst
```diff
@@ -3964,3 +3964,63 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+The format will be referred to as an "OPTIMADE regex".
+
+Regexes are commonly embedded in a contexts where they need to be enclosed by delimiters (e.g., double quotes or slash characters).
+If this is the case, it is likely that some outer-level escape rules apply to allow the end delimiter to appear within the regex.
+Such delimiters and escape rules are *not* included in the definition of the OPTIMADE regex format itself and needs to be clarified when this format is referenced.
+The format defined in this section applies after such outer escape rules have been applied (e.g., when all occurences of ``\/`` have been translated into ``/`` for a format where an unescaped slash character is the end delimiter).
+Likewise, if an OPTIMADE regex is embedded in a serialized data format (e.g., JSON) this section documents the format of the unicode string resulting from unserialization of that format.
```
vaitkus, line 3977:
```suggestion
Likewise, if an OPTIMADE regex is embedded in a serialized data format (e.g., JSON) this section documents the format of the Unicode string resulting from unserialization of that format.
```

optimade.rst
```diff
@@ -3964,3 +3964,63 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+  * Each literal character can appear in the class at most once.
+
+- Complemented character classes (e.g., ``[^abc]``, ``[^a-z]``).
+- Simple quantifiers: ``+`` (one or more), ``*`` (zero or more), ``?`` (zero or one).
+  (These can only appear directly after a character, group, or character class.)
+- The beginning-of-input (``^``) and end-of-input (``$``) anchors.
+- Simple grouping (``(...)``) and alternation (``|``).
+
+Note that compared to the JSON Schema standard, lazy quantifiers (``+?``, ``*?``, ``??``) are NOT included, nor are range quantifiers (``{x}``, ``{x,y}``, ``{x,}``).
+Furthermore, there is no support for character class shorthands via the backslash character ``\`` and a letter, nor is there a way to represent a unicode character by its code point (i.e., one has to include it as the literal unicode character).
```
vaitkus, line 4006:
```suggestion
Furthermore, there is no support for character class shorthands via the backslash character ``\`` and a letter, nor is there a way to represent a Unicode character by its code point (i.e., one has to include it as the literal Unicode character).
```

merkys:
To answer @sauliusg concern about devising future-proof regex syntax for possible filter extensions (#160) from COD PoV. MariaDB (COD backend engine) uses PCRE, thus filter support should be straightforward as long as OPTIMADE regex format stays a subset of PCRE.

rartino:
I'm still working on this to iron out some details; but in doing so I can across that Unicode itself defines a regex format with a very readable definition document: https://unicode.org/reports/tr18/

rartino:
Ok, I'm done with this for now. The PR is now really a best effort - based on our discussions last web meet - to make the JSON Schema subset of regex:es as stringent as possible, which, in hindsight, very much lacks precision of which constructs are allowed. I'd be happy if @sauliusg can take a look at this version.

However, the consequence is  that the PR has become quite long, so I no longer have high hopes in that this will be merged swiftly. I think I should still revert all references to regex:es in #445 so that this PR no longer blocks that PR.

optimade.rst
```diff
@@ -3964,3 +3964,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
     [3,7,19, ["PARTIAL-DATA-REF", ["https://example.db.org/value2"]]]
     [4,5,19, [ [11, 110], ["PARTIAL-DATA-REF", ["https://example.db.org/value3"]], [550, 333]]]
     ["PARTIAL-DATA-END", [""]]
+
+OPTIMADE Regular Expression Format
+----------------------------------
+This section defines a Unicode string representation of regular expressions (regexes) to be referenced from other parts of the specification.
+The format will be referred to as an "OPTIMADE regex."
```
rartino, line 3971:
```suggestion
The format will be referred to as an "OPTIMADE regex".
```

optimade.rst
```diff
@@ -3964,3 +3964,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+  * SQLite supports regexes via libraries and thus can use a compatible format (e.g., PCRE2).
+  * XML Schema appears to use a compatible regex format, except it is implicitly anchored: i.e., the beginning-of-input ``^`` and end-of-input ``$`` anchors must be removed, and missing anchors replaced by ``.*``.
+  * POSIX Extended regexes (and their extended GNU implementations) are incompatible because ``\`` is not a special character in character classes. POSIX Basic regexes also have further differences, e.g., the meaning of some escaped syntax characters is reversed.
```
rartino, line 4035:
```suggestion
  * POSIX Extended regexes (and their extended GNU implementations) are incompatible because ``\`` is not a special character in character classes. 
     POSIX Basic regexes also have further differences, e.g., the meaning of some escaped syntax characters is reversed.
```

optimade.rst
```diff
@@ -3964,3 +3964,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+Furthermore, there is no support for escapes designating shorthand character classes as ``\`` and a letter or number, nor is there any way to represent a Unicode character by specifying a code point as a number, only via the Unicode character itself.
+(However, the regex can be embedded in a context that defines such escapes, e.g., in serialized JSON a string containing the character ``\u`` followed by four hexadecimal digits is deserialized into the corresponding Unicode character.)
+
+An OPTIMADE regex matches the string at any position unless it contains a leading beginning-of-input (``^``) or trailing end-of-input (``$``) anchor listed above, i.e., the anchors are not implicitly assumed.
+For example, the OPTIMADE regex "es" matches "expression".
+
+Regexes that utilize tokes and features documented by ECMA-262 beyond the designated subset are allowed to have an undefined behavior, i.e., they MAY match or not match *any* string or MAY produce an error.
+Implementations that do not produce errors in this situation are RECOMMENDED to generate warnings if possible.
```
rartino, line 4017:
```suggestion
Regexes that utilize tokes and features beyond the designated subset are allowed to have an undefined behavior, i.e., they MAY match or not match *any* string or MAY produce an error.
Implementations that do not produce errors in this situation are RECOMMENDED to generate warnings if possible.
```

sauliusg:
> I'm still working on this to iron out some details; but in doing so I can across that Unicode itself defines a regex format with a very readable definition document: https://unicode.org/reports/tr18/

Looked at the Unicode Report and I like it much better then ECMA-262, indeed, Unicode definition is much clearer. Also, I regard Unicode and ISO/IEC 10646 are more fundamental (I expect Unicode to be around even when JSON goes away :)

Should we cite https://unicode.org/reports/tr18/ instead of or in addition to ECMA-262?

sauliusg:
> > I'm still working on this to iron out some details; but in doing so I can across that Unicode itself defines a regex format with a very readable definition document: https://unicode.org/reports/tr18/
> 
> Looked at the Unicode Report and I like it much better then ECMA-262, indeed, Unicode definition is much clearer. Also, I regard Unicode and ISO/IEC 10646 are more fundamental (I expect Unicode to be around even when JSON goes away :)
> 
> Should we cite https://unicode.org/reports/tr18/ instead of or in addition to ECMA-262?

After a more careful reading of the UTS#18 (https://unicode.org/reports/tr18/) I realize that it is more concerned with descriptions of character classes, with the definitions of the REs themselves, so we probably can't replace UTS#18 for ECMA-262. That's a pity.

rartino:
@sauliusg 

>>  I can across that Unicode itself defines a regex format with a very readable definition document: https://unicode.org/reports/tr18/
> Looked at the Unicode Report and I like it much better then ECMA-262, indeed, Unicode definition is much clearer. Also, I regard Unicode and ISO/IEC 10646 are more fundamental (I expect Unicode to be around even when JSON goes away :)
> Should we cite https://unicode.org/reports/tr18/ instead of or in addition to ECMA-262?

I had not realized it when pointing at UTS#18, but that document does not view itself as a regex standard on its own, but a standard way to accommodate unicode in existing regex standards; so I don't think is can be used instead of ECMA-262. We could still cite it to be helpful, but in the end the degree to which the OPTIMADE regex standard adhere to UTS#18 will depend on the degree to which ECMA-262 has embraced it for its "unicode on" flag mode.

optimade.rst
```diff
@@ -3964,3 +3964,73 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+- Simple grouping (``(...)``) and alternation (``|``).
+
+Note that compared to the JSON Schema standard, lazy quantifiers (``+?``, ``*?``, ``??``) are *not* included, nor are range quantifiers (``{x}``, ``{x,y}``, ``{x,}``).
+Furthermore, there is no support for escapes designating shorthand character classes as ``\`` and a letter or number, nor is there any way to represent a Unicode character by specifying a code point as a number, only via the Unicode character itself.
+(However, the regex can be embedded in a context that defines such escapes, e.g., in serialized JSON a string containing the character ``\u`` followed by four hexadecimal digits is deserialized into the corresponding Unicode character.)
+
+An OPTIMADE regex matches the string at any position unless it contains a leading beginning-of-input (``^``) or trailing end-of-input (``$``) anchor listed above, i.e., the anchors are not implicitly assumed.
+For example, the OPTIMADE regex "es" matches "expression".
+
+Regexes that utilize tokes and features beyond the designated subset are allowed to have an undefined behavior, i.e., they MAY match or not match *any* string or MAY produce an error.
```
sauliusg, line 4016: "tokens"?

optimade.rst
```diff
@@ -4002,3 +4002,73 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+Implementations that do not produce errors in this situation are RECOMMENDED to generate warnings if possible.
+
+  Compatibility notes:
+
+  The definition tolerates (with undefined behavior) regexes that use tokens and features beyond the defined subset.
+  Hence, a regex can be directly handed over to a backend implementation compatible with the subset without needing validation or translation.
+  Additional consideration of how the ``.`` character operates in relation to line breaks may be required for multiline text.
+  If the regex is applied to strings containing only the LINE FEED (U+000A) character and none of the other Unicode line break characters, most regex backend implementations are compatible with the defined behavior.
+  If the regex is applied to string data containing arbitrary combinations of line break characters and the right behavior cannot be achieved via environmental settings and regex options, implementations can consider a translation step where other line break characters are translated into LINE FEED in the text operated on.
```
rartino, line 4063:
```suggestion
  If the regex is applied to string data containing arbitrary combinations of Unicode line break characters and the right behavior cannot be achieved via environmental settings and regex options, implementations can consider a translation step where other line break characters are translated into LINE FEED in the text operated on.
```

optimade.rst
```diff
@@ -4002,3 +4002,73 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+
+  Compatibility with different regex implementations may change depending on the environment, language versions, and options and has to be verified by implementations.
```
rartino, line 4065:
```suggestion
  Compatibility with different regex implementations may change depending on the environment, implementation programming language versions, and options and has to be verified by implementations.
```

rartino (commented):
This implements some suggestions and tries to resolve some feedback from @sauliusg sent to me as pdf markup.

optimade.rst
```diff
@@ -4002,3 +4002,73 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+The subset is also intended to be compatible with, but even further restricted than, the subset recommended in the JSON Schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
+The compatibility with the JSON Schema standard is expressed as "intended" since there is some room for interpretation of the precise features included in the subset of regex features recommended in that standard.
+
+Hence, an OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/#sec-patterns>`__ with the additional restrictions described in the following.
+The regex is interpreted according to the processing rules that apply for an expression where only the Unicode variable is set to true of all variables set by the RegExp internal slot described by `ECMA-262, section 21.2.2.1 <https://262.ecma-international.org/11.0/#sec-notation>`__.
+Furthermore, it can only use the following tokens and features (this list is partially quoted from the JSON Schema standard):
```
rartino, line 4025:
```suggestion
An OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/#sec-patterns>`__ with additional restrictions described below which define a subset of the ECMA-262 format chosen to match features commonly available in different database backends.
The regex is interpreted according to the ECMA-262 processing rules that apply for an expression where only the Unicode variable is set to true of all variables set by the RegExp internal slot described by `ECMA-262, section 21.2.2.1 <https://262.ecma-international.org/11.0/#sec-notation>`__.

The subset includes only the following tokens and features:
```

optimade.rst
```diff
@@ -4002,3 +4002,73 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+  * SQLite supports regexes via libraries and thus can use a compatible format (e.g., PCRE2).
+  * XML Schema appears to use a compatible regex format, except it is implicitly anchored: i.e., the beginning-of-input ``^`` and end-of-input ``$`` anchors must be removed, and missing anchors replaced by ``.*``.
+  * POSIX Extended regexes (and their extended GNU implementations) are incompatible because ``\`` is not a special character in character classes.
+    POSIX Basic regexes also have further differences, e.g., the meaning of some escaped syntax characters is reversed.
```
rartino, line 4074:
```suggestion
  Compatibility notes:

  * The subset is intended to be compatible with, but even further restricted than, the subset recommended in the JSON Schema standard, see `JSON Schema: A Media Type for Describing JSON Documents 2020-12, section 6.4 <https://json-schema.org/draft/2020-12/json-schema-core#section-6.4>`__.
    The compatibility with the JSON Schema standard is expressed here as "intended" since there is some room for interpretation of the precise features included in the recommendation given in that standard.

  * The definition tolerates (with undefined behavior) regexes that use tokens and features beyond the defined subset.
    Hence, a regex can be directly handed over to a backend implementation compatible with the subset without needing validation or translation.
    
  * Additional consideration of how the ``.`` character operates in relation to line breaks may be required for multiline text.
    If the regex is applied to strings containing only the LINE FEED (U+000A) character and none of the other Unicode line break characters, most regex backend implementations are compatible with the defined behavior.
    If the regex is applied to string data containing arbitrary combinations of Unicode line break characters and the right behavior cannot be achieved via environmental settings and regex options, implementations can consider a translation step where other line break characters are translated into LINE FEED in the text operated on.

  * Compatibility with different regex implementations may change depending on the environment, implementation programming language versions, and options and has to be verified by implementations.
    However, as a general guide, we have used third-party sources, e.g., the `Regular Expression Engine Comparison Chart <https://gist.github.com/CMCDragonkai/6c933f4a7d713ef712145c5eb94a1816>`__ to collect the following information for compatibility when operating on text using LINE FEED as the line break character:

      * `ECMAScript (also known as javascript) <https://262.ecma-international.org/>`__ and version 1 and 2 of `PCRE <https://www.pcre.org/>`__ are meant to be compatible by design when used with appropriate options.

      * The following regex formats appear generally compatible when operating in Unicode mode: `Perl <https://perldoc.perl.org/perlre>`__, `Python <https://docs.python.org/3/library/re.html>`__, `Ruby <https://ruby-doc.org/3.2.2/Regexp.html>`__, `Rust <https://docs.rs/regex/latest/regex/>`__, `Java <https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html>`__, `.NET <https://learn.microsoft.com/en-us/dotnet/standard/base-types/details-of-regular-expression-behavior>`__, `MySQL 8 <https://dev.mysql.com/doc/refman/8.0/en/regexp.html>`__, `MongoDB <https://www.mongodb.com/docs/manual/reference/operator/query/regex/>`__, `Oracle <https://docs.oracle.com/cd/B13789_01/appdev.101/b10795/adfns_re.htm>`__, `IBM Db2 <https://www.ibm.com/docs/en/db2/11.5?topic=reference-regular-expressions>`__, `Elasticsearch <https://www.elastic.co/guide/en/elasticsearch/reference/current/regexp-syntax.html>`__, `DuckDB <https://duckdb.org/docs/sql/functions/patternmatching.html#regular-expressions>`__ (which uses the `re2 <https://github.com/google/re2/wiki/Syntax>`__ library).
      * SQLite supports regexes via libraries and thus can use a compatible format (e.g., PCRE2).
      * XML Schema appears to use a compatible regex format, except it is implicitly anchored: i.e., the beginning-of-input ``^`` and end-of-input ``$`` anchors must be removed, and missing anchors replaced by ``.*``.
      * POSIX Extended regexes (and their extended GNU implementations) are incompatible because ``\`` is not a special character in character classes.
        POSIX Basic regexes also have further differences, e.g., the meaning of some escaped syntax characters is reversed.
```

rartino (commented):
Restructuring to decrease bulk and improve order

optimade.rst
```diff
@@ -4002,3 +4002,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+
+- Individual Unicode characters matching themselves, as defined by the `JSON specification <https://json-schema.org/draft/2020-12/json-schema-core#RFC8259>`__.
+- The ``.`` character to match any one Unicode character except the line break characters LINE FEED (LF) (U+000A), CARRAGE RETURN (U+000D), LINE SEPARATOR (U+2028), PARAGRAPH SEPARATOR (U+2029) (see `ECMA-262 section 2.2.2.7 <https://262.ecma-international.org/14.0/?_gl=1*yqtzjq*_ga*MjEzNTE2ODEyNi4xNzA0NTQ1NTk5*_ga_TDCK4DWEPP*MTcwNzk5ODA1My43LjEuMTcwNzk5OTYxNC4wLjAuMA..#sec-compileatom>`__).
+- A literal escape of `one of the characters defined as syntax characters in the ECMA-262 standard <https://262.ecma-international.org/11.0/#prod-SyntaxCharacter>`__, i.e., the escape character (``\``) followed by one of the following characters ``^ $ \ . * + ? ( ) [ ] { } |`` to represent that literal character.
+  No other characters can be escaped.
+  (This rule prevents other escapes that are interpreted differently depending on regex flavor.)
+- Simple character classes (e.g., ``[abc]``), complemented character classes (e.g. [^abc]) and their ranged versions (e.g., ``[a-z]``, ``[^a-z]``) with the following constraints:
```
rartino, line 4027:
```suggestion
- Simple character classes (e.g., ``[abc]``), complemented character classes (e.g. [^abc]), and their ranged versions (e.g., ``[a-z]``, ``[^a-z]``) with the following constraints:
```

optimade.rst
```diff
@@ -4002,3 +4002,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+
+- Individual Unicode characters matching themselves, as defined by the `JSON specification <https://json-schema.org/draft/2020-12/json-schema-core#RFC8259>`__.
+- The ``.`` character to match any one Unicode character except the line break characters LINE FEED (LF) (U+000A), CARRAGE RETURN (U+000D), LINE SEPARATOR (U+2028), PARAGRAPH SEPARATOR (U+2029) (see `ECMA-262 section 2.2.2.7 <https://262.ecma-international.org/14.0/?_gl=1*yqtzjq*_ga*MjEzNTE2ODEyNi4xNzA0NTQ1NTk5*_ga_TDCK4DWEPP*MTcwNzk5ODA1My43LjEuMTcwNzk5OTYxNC4wLjAuMA..#sec-compileatom>`__).
+- A literal escape of `one of the characters defined as syntax characters in the ECMA-262 standard <https://262.ecma-international.org/11.0/#prod-SyntaxCharacter>`__, i.e., the escape character (``\``) followed by one of the following characters ``^ $ \ . * + ? ( ) [ ] { } |`` to represent that literal character.
+  No other characters can be escaped.
+  (This rule prevents other escapes that are interpreted differently depending on regex flavor.)
+- Simple character classes (e.g., ``[abc]``), complemented character classes (e.g. [^abc]), and their ranged versions (e.g., ``[a-z]``, ``[^a-z]``) with the following constraints:
```
rartino, line 4027:
```suggestion
- Simple character classes (e.g., ``[abc]``), complemented character classes (e.g. ``[^abc]``), and their ranged versions (e.g., ``[a-z]``, ``[^a-z]``) with the following constraints:
```

optimade.rst
```diff
@@ -4002,3 +4002,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+Furthermore, there is no support for escapes designating shorthand character classes as ``\`` and a letter or number, nor is there any way to represent a Unicode character by specifying a code point as a number, only via the Unicode character itself.
+(However, the regex can be embedded in a context that defines such escapes, e.g., in serialized JSON a string containing the character ``\u`` followed by four hexadecimal digits is deserialized into the corresponding Unicode character.)
+
+An OPTIMADE regex matches the string at any position unless it contains a leading beginning-of-input (``^``) or trailing end-of-input (``$``) anchor listed above, i.e., the anchors are not implicitly assumed.
+For example, the OPTIMADE regex "es" matches "expression".
+
+Regexes that utilize tokens and features beyond the designated subset are allowed to have an undefined behavior, i.e., they MAY match or not match *any* string or MAY produce an error.
+Implementations that do not produce errors in this situation are RECOMMENDED to generate warnings if possible.
+
```
rartino, line 4051:
```suggestion
Note that lazy quantifiers (``+?``, ``*?``, ``??``) are *not* included, nor are range quantifiers (``{x}``, ``{x,y}``, ``{x,}``).
Furthermore, there is no support for escapes designating shorthand character classes as ``\`` and a letter or number, nor is there any way to represent a Unicode character by specifying a code point as a number, only via the Unicode character itself.
(However, the regex can be embedded in a context that defines such escapes, e.g., in serialized JSON a string containing the character ``\u`` followed by four hexadecimal digits is deserialized into the corresponding Unicode character.)

An OPTIMADE regex matches the string at any position unless it contains a leading beginning-of-input (``^``) or trailing end-of-input (``$``) anchor listed above, i.e., the anchors are not implicitly assumed.
For example, the OPTIMADE regex "es" matches "expression".

Regexes that utilize tokens and features beyond the designated subset are allowed to have an undefined behavior, i.e., they MAY match or not match *any* string or MAY produce an error.
Implementations that do not produce errors in this situation are RECOMMENDED to generate warnings if possible.
```

optimade.rst
```diff
@@ -4002,3 +4002,71 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+Furthermore, there is no support for escapes designating shorthand character classes as ``\`` and a letter or number, nor is there any way to represent a Unicode character by specifying a code point as a number, only via the Unicode character itself.
+(However, the regex can be embedded in a context that defines such escapes, e.g., in serialized JSON a string containing the character ``\u`` followed by four hexadecimal digits is deserialized into the corresponding Unicode character.)
+
+An OPTIMADE regex matches the string at any position unless it contains a leading beginning-of-input (``^``) or trailing end-of-input (``$``) anchor listed above, i.e., the anchors are not implicitly assumed.
+For example, the OPTIMADE regex "es" matches "expression".
+
+Regexes that utilize tokens and features beyond the designated subset are allowed to have an undefined behavior, i.e., they MAY match or not match *any* string or MAY produce an error.
+Implementations that do not produce errors in this situation are RECOMMENDED to generate warnings if possible.
+  Compatibility notes:
```
rartino, line 4051:
```suggestion
Implementations that do not produce errors in this situation are RECOMMENDED to generate warnings if possible.

  Compatibility notes:
```

optimade.rst
```diff
@@ -4002,3 +4002,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+  * If the first character is ``^`` then the expression matches all characters *except* the ones specified by the class as defined by the characters that follows.
+  * The characters ``\ [ ]`` can only appear escaped with a preceding backslash, e.g. ``\\`` designates that the class includes a literal ``\`` character.
+    The other syntax characters may appear either escaped or unescaped to designate that the class includes them.
+    (This rule prevents other escapes inside classes that are not the same across regex flavors and expressions that, in some flavors, are interpreted as nested classes.)
+  * Except as specified above, all characters represent themselves literally (including syntax characters).
+  * Characters that represent themselves literally can only appear at most once.
+    (This rule prevents various kinds of extended character class syntax that differs between regex formats that assigns special meaning to duplicated characters, e.g., POSIX character classes, e.g., ``[:alpha:]``, equivalence classes, e.g., ``[=a=]``, set constructs, e.g. ``[A--B]``, ``[A&&B]``, etc.).
```
rartino, line 4036:
```suggestion
    (This rule prevents various kinds of extended character class syntax that differs between regex formats that assigns special meaning to duplicated characters such as POSIX character classes, e.g., ``[:alpha:]``, equivalence classes, e.g., ``[=a=]``, set constructs, e.g. ``[A--B]``, ``[A&&B]``, etc.).
```

sauliusg (dismissed):
After diffing commits fb2e2d3 and cba1227...

sauliusg (dismissed):

optimade.rst
```diff
@@ -4002,3 +4002,72 @@ An example of the sparse layout for multidimensional lists with three aggregated
...
+Likewise, if an OPTIMADE regex is embedded in a serialized data format (e.g., JSON), this section documents the format of the Unicode string resulting from the deserialization of that format.
+
+An OPTIMADE regex is a regular expression that adheres to `ECMA-262, section 21.2.1 <https://262.ecma-international.org/11.0/#sec-patterns>`__ with additional restrictions described below which define a subset of the ECMA-262 format chosen to match features commonly available in different database backends.
+The regex is interpreted according to the ECMA-262 processing rules that apply for an expression where only the Unicode variable is set to true of all variables set by the RegExp internal slot described by `ECMA-262, section 21.2.2.1 <https://262.ecma-international.org/11.0/#sec-notation>`__.
+
+The subset includes only the following tokens and features:
+
+- Individual Unicode characters matching themselves, as defined by the `JSON specification <https://json-schema.org/draft/2020-12/json-schema-core#RFC8259>`__.
```
ml-evs, line 4022:
```suggestion
- Individual Unicode characters matching themselves, as defined by the `JSON specification <https://www.rfc-editor.org/info/rfc8259>`__.
```

ml-evs: or

```suggestion
- Individual Unicode characters matching themselves, as defined by the JSON specification (:RFC:`8259`).
```

merkys (approved):

ml-evs (approved):

merkys (approved):

ml-evs (approved):

PR #491, opened 2024-01-12
Fix rst indentation and internal references

vaitkus:
This PR fixes a few formatting issues that cause warnings when converting the rst file to a pdf using pandoc. Not sure how and if these issues manifested on GitHub since currently I cannot force this file to render here (earlier version of this file also do not render).

Also, the `elements_ratios` field was misspelt as `element_ratios` in several places that deal with the metadata properties.

Finally, is there a simple way to make a field name to both link to the corresponding section in the document and be formatted in the same way as other fields (i.e. to combine ":field:`elements_ratios`" and "`elements_rations`_")?

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -508,7 +508,7 @@ The reason for this limitation is to avoid name collisions with metadata fields 
 Implementation of the :field:`meta` field is OPTIONAL.
 However, when an implementation supports the :field:`property_metadata` field, it SHOULD include metadata fields for all properties which have metadata and are present in the data part of the response.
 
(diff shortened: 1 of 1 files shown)


ml-evs:
Thanks for this @vaitkus, fixes look good (will review properly soon).

Just wanted to raise the issue that I don't seem to be able to preview RST anymore on GitHub... at all... does anyone else have this issue?

rartino (approved):
The corrections all seem correct to me.

ml-evs (approved):
Likewise, thanks for spotting these @vaitkus.

vaitkus:
@ml-evs Is it OK to merge this now or would it be better to wait until the RST rendering issue is resolved?

ml-evs:
I'll merge now, I think these are clear specification bugs and we can imagine what the rendered docs will look like. Not sure if the resolution will come anytime soon...

Issue #492, opened 2024-01-12
Specification not rendering on GitHub


ml-evs:
For some reason the spec can no longer be previewed on GitHub. I'm not sure if this an ephemeral problem, a GitHub bug or something wrong with our files (see https://github.com/orgs/community/discussions/86715).

ml-evs:
> Old versions like https://github.com/Materials-Consortia/OPTIMADE/blob/6f1e115094ec256c262df712a35c86f4d3e6d1c7/optimade.rst fail to render when they definitely used to (Firefox Developer Edition, no extensions)

_Originally posted by @ml-evs in https://github.com/Materials-Consortia/OPTIMADE/issues/493#issuecomment-1889488011_

rartino:
Clearly something changed on GitHub's side; even going back to the original rst file right after we converted the specification doesn't render.

ml-evs:
Feel free to pretend to be another effected project in my linked discussion :P I think this must have changed in the last hour or so...

rartino:
I think a fairly high level of complexity is needed to trigger this; here is a somewhat long rst file that still renders correctly: https://github.com/httk/rsttools/blob/main/docs/rstslide/presentation.rst

ml-evs:
I'll remove our custom roles and try to commit again to #493, see if it makes a difference

ml-evs:
After some fixes on GitHub's end, we are now *mostly* back to where we were. The only thing that no longer renders is the table of contents (and hopefully they are working on this). Given that we have HTML builds now anyway (tracked by #501), I will close this issue.

PR #493, opened 2024-01-12
Test RST rendering on GitHub

ml-evs:
See #492

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -51,56 +51,6 @@ OPTIMADE API specification v1.2.0~develop
      <anything>-fail : means this is a counter-example of something
                        that is meant to be on form <anything> but is not valid.
 
-.. role:: filter(code)
-   :language: filter
-
-.. role:: filter-fragment(literal)
-
-.. role:: filter-op(literal)
-
-.. role:: ere(literal)
-
-.. role:: pcre(literal)
-
-
-.. role:: entry(literal)
-
-.. role:: property(literal)
-
-.. role:: val(literal)
-
-.. role:: type(literal)
-
-.. role:: property-fail(literal)
-
-
-
-.. role:: endpoint(literal)
-
-.. role:: query-param(literal)
-
-.. role:: query-val(literal)
-
-.. role:: query-string(literal)
-
-.. role:: query-url(literal)
-
-
-.. role:: http-header(literal)
-
-.. role:: http-error(literal)
-
-
-.. role:: json(code)
-   :language: json
-
-.. role:: field(literal)
-
-.. role:: field-val(literal)
-
-.. role:: object(literal)
-
-
 .. sectnum::
 
 .. contents::
--- a/README.rst
+++ b/README.rst
@@ -1,0 +1,6 @@
+README
+------
+
+The Open Databases Integration for Materials Design (OPTIMADE) consortium aims to make materials databases interoperational by developing a common REST API.
+
+This repository contains the specification of the OPTIMADE API.


vaitkus:
This one does render properly. So do some more complex rst files on other repos, e.g. https://github.com/rstcheck/rstcheck/blob/main/README.rst.

ml-evs:
Hmmm yeah, trying to pin down which version stops rendering for us (it still seems like a recent change). If this is a rendering timeout issue, maybe this is a good excuse to cut down the complexity of the specification with @rartino's properties...

rartino:
Probably the same thing; rich diffs for PRs against optimade.rst now gives the error:

> Oh My
>
> Sorry... we’re unable to render the document prior to diffing.

ml-evs:
Old versions like https://github.com/Materials-Consortia/OPTIMADE/blob/6f1e115094ec256c262df712a35c86f4d3e6d1c7/optimade.rst fail to render when they definitely used to (Firefox Developer Edition, no extensions)

ml-evs:
Will close this to not pollute PRs, we can continue discussing at #492

ml-evs:
Okay, the first 312 lines of the spec render... https://github.com/Materials-Consortia/OPTIMADE/blob/3ba8299979001f0cfdd02bde55d241a9b2a3d9d3/optimade.rst

rartino:
> Okay, the first 312 lines of the spec render... https://github.com/Materials-Consortia/OPTIMADE/blob/3ba8299979001f0cfdd02bde55d241a9b2a3d9d3/optimade.rst

Maybe a sign that it is time to split it into separate files?...

ml-evs:
And there's no issue using custom roles/toc (in isolation, at least, could still be added complexity once the file is too large)

ml-evs:
> > Okay, the first 312 lines of the spec render... https://github.com/Materials-Consortia/OPTIMADE/blob/3ba8299979001f0cfdd02bde55d241a9b2a3d9d3/optimade.rst
> 
> Maybe a sign that it is time to split it into separate files?...

Sure, but for now we still won't have enough space to describe what's in those files!

rartino:
> Sure, but for now we still won't have enough space to describe what's in those files (assuming you mean properties at least!)

I mean chopping it up into chapters (and re-merge when rendering). It is a bit finky with the TOC, will require some non-trivial infrastructure work.

ml-evs:
> > Sure, but for now we still won't have enough space to describe what's in those files (assuming you mean properties at least!)
> 
> I mean chopping it up into chapters (and re-merge when rendering). It is a bit finky with the TOC, will require some non-trivial infrastructure work.

Sure, if each individual .rst can still be previewed fine. I still have https://ml-evs.science/OPTIMADE/optimade.html going, could also put some work into making this multi-versioned and deployed sensibly, as the "real" specification for users to view.

optimade.rst
```diff
@@ -913,3 +913,82 @@ An example of a full response:
 
   If the server is able to handle serialization in such a way that it can dictate the order of the top level object members in the response, it is RECOMMENDED to put the :field:`jsonapi` as the first top level member to simplify identification of the response.
 
+
```
ml-evs, line 916: As a yardstick, addition of this patch is what breaks rendering (as of today). I'll try to see if there's any weird feature we're using here that is causing it to lag

blokhin:
Since the spec becomes a standalone HTML document, not necessary connected to a GitHub-org, let's later add the "Community" section to the Appendix, so that the reader might get some idea about the Optimade people without googling.

PR #498, opened 2024-02-11
Enable `rstcheck` in CI and lint code blocks

ml-evs:
This PR follows up on #493 by following GitHub's advice of running the spec through `rstcheck`. We were already passing it, but in the process of adding it I think it makes sense to properly lint our code blocks as JSON data (which is possible in many cases without losing the structure of them). 

Also this PR:
- unifies the indent of all code blocks to start after the 3 chars indicated with the RST "`.. `". 
- fixes one occurrence of a non-standard field in a code block ("response_message" for the `meta` example)
- fixes indentation of one implementation note so that it refers to its proper field
- switches the partial data code blocks to JSONL format

Unfortunately its a bit tricky to see how this will render in the future in GitHub, so perhaps we need to hold off a little while for the official response.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -3907,54 +3907,54 @@ Examples
 Below follows an example of a dense response for a partial array data of integer values.
 The request returns the first three items and provides the next-marker link to continue fetching data:
 
-.. code:: json
+.. code:: jsonl
 
-    {"optimade-partial-data": {"format": "1.2.0"}, "layout": "dense", "returned_ranges": [{"start": 10, "stop": 20, "step": 2}]}
-    123
-    345
-    -12.6
-    ["PARTIAL-DATA-NEXT", ["https://example.db.org/value4"]]
+   {"optimade-partial-data": {"format": "1.2.0"}, "layout": "dense", "returned_ranges": [{"start": 10, "stop": 20, "step": 2}]}
+   123
+   345
+   -12.6
+   ["PARTIAL-DATA-NEXT", ["https://example.db.org/value4"]]
 
 Below follows an example of a dense response for a list property as a partial array of multidimensional array values.
 The item with index 10 in the original list is provided explicitly in the response and is the first one provided in the response since start=10.
 The item with index 12 in the list, the second data item provided since start=10 and step=2, is not included only referenced.
 The third provided item (index 14 in the original list) is only partially returned: it is a list of three items, the first and last are explicitly provided, the second one is only referenced.
 
-.. code:: json
+.. code:: jsonl
 
-    {"optimade-partial-data": {"format": "1.2.0"}, "layout": "dense", "returned_ranges": [{"start": 10, "stop": 20, "step": 2}]}
-    [[10,20,21], [30,40,50]]
-    ["PARTIAL-DATA-REF", ["https://example.db.org/value2"]]
-    [[11, 110], ["PARTIAL-DATA-REF", ["https://example.db.org/value3"]], [550, 333]]
-    ["PARTIAL-DATA-NEXT", ["https://example.db.org/value4"]]
+   {"optimade-partial-data": {"format": "1.2.0"}, "layout": "dense", "returned_ranges": [{"start": 10, "stop": 20, "step": 2}]}
+   [[10,20,21], [30,40,50]]
+   ["PARTIAL-DATA-REF", ["https://example.db.org/value2"]]
+   [[11, 110], ["PARTIAL-DATA-REF", ["https://example.db.org/value3"]], [550, 333]]
+   ["PARTIAL-DATA-NEXT", ["https://example.db.org/value4"]]
 
 Below follows an example of the sparse layout for multidimensional lists with three aggregated dimensions.
 The underlying property value can be taken to be sparse data in lists in four dimensions of 10000 x 10000 x 10000 x N, where the innermost list is a non-sparse list of arbitrary length of numbers.
 The only non-null items in the outer three dimensions are, say, [3,5,19], [30,15,9], and [42,54,17].
 The response below communicates the first item explicitly; the second one by deferring the innermost list using a reference-marker; and the third item is not included in this response, but deferred to another page via a next-marker.
 
-.. code:: json
+.. code:: jsonl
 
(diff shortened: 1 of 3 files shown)


optimade.rst
```diff
@@ -783,20 +773,20 @@ The response MAY also return resources related to the primary data in the field:
...
+
+       {
+         "links": {
+           "base_url": {
+             "href": "http://example.com/optimade",
+             "meta": {
+               "_exmpl_db_version": "3.2.1"
+             }
+           }
+           // ...
```
merkys, line 786: JSON "comments" are gone in all above examples, but not this one. Are they allowed or disallowed by `rstcheck`?

ml-evs: I think comments are allowed where they don't break the flow of the JSON doc, we were using them in lots of superfluous ways to indicate that there are more fields or more entries "available" so various lines had trailing commas that break the JSON format. I only removed comments that were flagged up by rstcheck. 

Probably we could revert the indentation changes (I'm not sure what effect it has in rst, from memory I think GitHub was always rendering them with no leading whitespace?) -- this would make the content changes clearer

rartino: Overall great idea to validate the spec, but I'm also puzzled by the changes to our comments in JSON.

I see no reason why comments would be allowed in a json or jsonlines code block, so if that passes `rstcheck` it may be a bug? However, the `jsonc` format that we've used where we have comments is meant to allow comments. Is `rstcheck` complaining about those comments? Why would it even be looking at the content of a code block specified to use a language it doesn't understand?

ml-evs: > Overall great idea to validate the spec, but I'm also puzzled by the changes to our comments in JSON.
> 
> I see no reason why comments would be allowed in a json or jsonlines code block, so if that passes `rstcheck` it may be a bug? However, the `jsonc` format that we've used where we have comments is meant to allow comments. Is `rstcheck` complaining about those comments? Why would it even be looking at the content of a code block specified to use a language it doesn't understand?
> 
> **Edit** and I agree with @merkys that in places where we use `// ...` to mean that we've cut out mandatory parts of the response, those comments are needed.

Fair enough, I wasn't entirely convinced these changes were the right idea either. I'll revert the jsonc -> JSON change, but yes, rstcheck doesn't check languages it doesn't understand (so essentially the JSON we have is not validated -- except by me changing it for this PR!)

optimade.rst
```diff
@@ -52,7 +52,6 @@ OPTIMADE API specification v1.2.0~develop
                        that is meant to be on form <anything> but is not valid.
 
 .. role:: filter(code)
-   :language: filter
```
ml-evs, line 55: For some reason this line caused rstcheck to fail, even when ignored.

optimade.rst
rartino: Maybe we seriously should look into extending rstcheck to check jsonc? It seems that would be genuinely useful for us.

ml-evs: [Here's the PR](https://github.com/rstcheck/rstcheck-core/pull/78/files) that added yaml support, really doesn't look too difficult actually. I'll open an issue with them and see what they say.

ml-evs: Just opened https://github.com/rstcheck/rstcheck-core/issues/92. Will see if they are happy to have it in the core or if there is another more modular approach.

rartino: Ok, but so if this has been pushed upstream (thanks @ml-evs!), and the specification is back to use jsonc, which for now is ignored by rstcheck, this comment can now be marked as resolved?

rartino:
@ml-evs This is marked as a draft, but is it not good to go now? The GitHub pre-commit checks pass and seem to include rstcheck.

(That said, in some future cleanup I would like to just float if it is possible to move these checks into 'make audit' instead, and then only have a single 'make audit' in the pre-commit hooks. That makes it a bit more straightforward to keep the pre-commit hooks aligned with the checks it is easy to do oneself when working on a clone, and it would also avoid us becoming unnecessarily dependent on git-specific features.)

optimade.rst
ml-evs: Sure; I opened a PR with them but not sure when it will be merged.

rartino: I suspect they will wait for the missing parts; documentation + changelog (and non-draft status) before anyone reviews. But in any case, we can still merge the rstcheck without that.

ml-evs:
In that case, feel free to review and we can merge here. Otherwise also happy to use `make audit` straight away. There might be some conflicts here with #499 -- perhaps this should be merged first as it introduces additional checks.

rartino (dismissed):

optimade.rst
```diff
@@ -639,7 +633,7 @@ Every response SHOULD contain the following fields, and MUST contain at least :f
   - **response\_message**: response string from the server.
   - **request\_delay**: a non-negative float giving time in seconds that the client is suggested to wait before issuing a subsequent request.
 
-  Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
+    Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__. Implementations are suggested to provide consistent handling of request overload through both mechanisms.
```
vaitkus, line 636:
```suggestion
    Implementation note: the functionality of this field overlaps to some degree with features provided by the HTTP error :http-error:`429 Too Many Requests` and the `Retry-After HTTP header <https://tools.ietf.org/html/rfc7231.html#section-7.1.3>`__.
    Implementations are suggested to provide consistent handling of request overload through both mechanisms.
```

vaitkus (approved):

vaitkus (approved):

rartino (approved):

optimade.rst
ml-evs: Haha, I was just wondering why, furiously clicking on the greyed out merge button... I think at least Antanas' change also held it up

Issue #501, opened 2024-03-19
Serve multiple rendered HTML specification versions from `gh-pages` branch


ml-evs:
Now #378 is merged, I would like to go through and manually update the gh pages branch to include HTML builds of our old spec versions (potentially with minor tweaks required that faithfully represent the old specs).

The idea would then be to set up the CI so that:

- pushes to develop trigger a build that replaces the `develop` version of the spec
- releases trigger a build that replaces a specific built tag of the spec
- this should probably be set up before we next release, to avoid needing to make any spurious changes after the fact.

I will probably nuke the `gh_pages` branch before doing this, then restrict push access to it.

The beauty of a divergent gh pages branch is that we could make any edits to the built versions (say we wanted to tweak the style sheet or template to include banner redirects to the latest version) we wanted directly to that, without needing PRs to develop.

ml-evs:
Just made a manual start on this on the `gh-pages` branch, so the development spec is now at https://www.optimade.org/OPTIMADE/specification/develop/

We can decide what the best canonical URLs are at some point. I'll try to make a PR to `develop` here that updates the workflow to automtically include any future releases or changes to develop (although maybe this is infrequent enough that we could juts keep develop up to date and manually add releases...) Will need to design some kind of splash page for gh-pages too

ml-evs:
We could also consider building and hosting PR previews.

ml-evs:
Hmm, I thought I had commented here with plans for the specification repo, but I guess it never hit send... I see @rartino has started the work already at https://github.com/Materials-Consortia/specification. For now @rartino, shall I just duplicate the gh pages in the this repo to there and then get the links set up correctly? I think going forward, we could either use an indirect workflow call to call the build in that repo, or simply build here and push to the spec repo with a bot account.

rartino:
Right; when I didn't see any answer or repo, I thought I'd do a quick experiment with a setup. The current specification repo works, in that it renders the OPTIMADE "default branch" specification to https://www.optimade.org/specification when the "Build OPTIMADE specification" gets triggered, which right now only is done manually by going here:

   https://github.com/Materials-Consortia/specification/actions/workflows/pages.yml

[...]

PR #503, opened 2024-03-22
Updates to HTML build in CI

ml-evs:
This PR fixes the CI so that gh-pages does not include all future changes.

It also fixes the CSS following some feedback.

(Addresses #501)

`develop` now gets deployed to `/specification/develop` from the root of the page (currently optimade.org/OPTIMADE).

Still missing:

- a list of deployed versions

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -35,10 +35,11 @@ jobs:
 
     - name: Commit to gh-pages
       run: |
-        git branch -D gh-pages || true
-        git checkout -b gh-pages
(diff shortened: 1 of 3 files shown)


merkys (dismissed):
I like it a lot that specific font names are no longer included! I did not attempt running the code in `.github/workflows/pages.yml` though.

sauliusg (approved):
This branch now works great on my machine.

merkys (approved):

rartino (approved):

PR #508, opened 2024-03-22
Use `fetch` instead of pull to avoid needing to rebase gh-pages branch

ml-evs:
I was too hasty with the previous build fixes. Hopefully this is the last one... I've also added `workflow_dispatch` here which should let us run the action from PR branches, primarily to test CI failures ahead of time (before merging into develop).

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -38,9 +38,9 @@ jobs:
       run: |
         git fetch origin gh-pages
         git checkout gh-pages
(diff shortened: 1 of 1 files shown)


vaitkus (approved):

merkys (approved):

rartino (approved):

ml-evs:
Ok this now working, see https://www.optimade.org/OPTIMADE/specification/develop/

PR #512, opened 2024-04-04
Add a development banner to generated development specification

ml-evs:
This is already deployed at http://specification.optimade.org/, but we need to make sure that new builds get the banner too.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -39,6 +39,8 @@ jobs:
         git fetch origin gh-pages
         git checkout gh-pages
         mkdir -p specification/develop
+        # patch dev specification to include banner
+        sed -i 's|</head>|</head><div class="banner-container"><div class="banner">This is a development version of the specification.</div></div>|g' optimade.html
         mv optimade.html specification/develop/index.html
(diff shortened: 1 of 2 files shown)


.github/workflows/pages.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         git fetch origin gh-pages
         git checkout gh-pages
         mkdir -p specification/develop
+        # patch dev specification to include banner
+        sed -i 's|</head>|</head><div class="banner-container"><div class="banner">This is a development version of the specification.</div></div>|g' optimade.html
```
vaitkus, line 43:
```suggestion
        sed -i 's|<body class="with-toc">|<body class="with-toc"><div class="banner-container"><div class="banner">This is a development version of the specification.</div></div>|g' optimade.html
```

vaitkus: `<div>...</div>` should appear inside `<body>` to be a valid HTML.

vaitkus (approved):

vaitkus (approved):

merkys (approved):

PR #514, opened 2024-04-04
Fixed standard symbol use in unitsystem definitions.

oke464:
Changed symbols used in unitsystem definitions to be consistent with the standard symbol defined in the referenced unit data schema.

### DIFF
--- a/schemas/src/defs/v1.2/constants/codata/2018/electromagnetic/bohrmagneton.yaml
+++ b/schemas/src/defs/v1.2/constants/codata/2018/electromagnetic/bohrmagneton.yaml
@@ -4,9 +4,9 @@ title: "Bohr magneton"
 symbol: "bohrmagneton"
 display-symbol: '\(\mu_B\)'
 description: |-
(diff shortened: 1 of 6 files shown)


ml-evs:
I think I'm a bit out of the loop on this one, I assume its correct but perhaps @rartino can confirm? (As I guess the schemas at schemas.optimade.org will also need to be regenerated?)

vaitkus:
Looks good to me as well.

@oke464, could you please also fix one minor unrelated issue in the `schemas/src/defs/v1.2/unitsystems/si/si_1985.yaml` file? The `description` field incorrectly states the release date as "in 1995" instead of "in 1985".

rartino:
Thanks @oke464! (This wasn't caught with the schemas and sanity-checks as it is an intentional feature that a unitsystem can be defined to include a unit but with a non-standard symbol; it just wasn't the intent to use that feature here.)

I took the liberty to also add the fix @vaitkus asked for and, while we are anyway doing an update, a couple of minor typographical adjustments I noticed since the big merge.

@ml-evs This indeed affected a couple of definitions/pages, in particular for unitsystems, under schemas.optimade.org. I've now updated them.

rartino (approved):

vaitkus (approved):

ml-evs (approved):

PR #515, opened 2024-04-07
Skip the 'build_html' job on repository forks

vaitkus:
The 'build_html' job fails on repository forks that do not have the `gh_pages` branch. This PR modifies the job so it is only run on the original `Materials-Consortia/OPTIMADE` repo.


rartino (dismissed):
This is better than the situation now (so approved). But, I wonder if it wouldn't be even better if possible to formulate it "if: a gh_pages branch exists" to make it possible for people to recover the main repo behavior (which can be useful also for forks) without having to add diverging commits to the repo.

rartino (approved):
Yes, lets merge this. But if we set up a separate repo for building and publishing the html version, this may end up not being used here anyway.

vaitkus:
Unfortunately, it seems that any non-zero status terminates the job so the job still fails on forks without this branch. I think 
rewriting to something like:
```
git ls-remote --exit-code --heads origin refs/heads/gh-pages > /dev/null && exit 0
```
should work, but I should probably test it first before submitting another PR.

vaitkus:
Ok, this approach does not seem to work in general. Exiting from the "Commit to gh-pages" job with a zero status results in the "Push changes" job still getting executed (and thus creating a "gh-pages" branch on the fork, but without the HTML pages). I will leave this alone unless the separate repo approach does not work out.

rartino:
Ah, the following steps would have to be updated as well to do nothing. So, should we revert this merge?

vaitkus:
We probably should since it is a bit misleading (comments do not match the actual behaviour). Is there an easy built-in way to do so in GitHub?

rartino:
The merge icon in this PR thread had a "Revert" button. I clicked that, and it created a revert PR.

PR #523, opened 2024-06-12
Introduce roles for files related to calculations

merkys:
During workshop discussions, @giovannipizzi suggested introducing a means to specify how files are related to calculations entries. This PR introduces `meta.role` field to specify whether an attached file is input or output relative to the calculation in question.

Furthermore, there was a suggestion to order the related output files in a way to make it easier for automated analysis software to parse the output files in order to identify the code and/or calculation type used. Since it is impossible to define the deterministic orders for all the used codes, the suggestion is to use an arbitrary order where output files most likely to contain the identifying information come first. This way analysis software would encounter the identifying output files earlier and stop before reading in all the output files.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -3533,12 +3533,12 @@ Calculations Entries
 --------------------
 
 The :entry:`calculations` entries have the properties described above in section `Properties Used by Multiple Entry Types`_.
-Related :entry:`files` entries specified as relationships as described in section `Relationships`_ list files used or produced by a calculation.
-They MAY be defined as either input or output files in the :field:`role` field inside the :field:`meta` dictionary of the JSON:API resource identifier object.
+Related :entry:`files` entries specified as relationships (as described in section `Relationships`_) list files used or produced by a calculation.
+They MAY be defined as either input or output files by the :field:`role` field inside the :field:`meta` dictionary of the JSON:API resource identifier object.
 The allowed values are :val:`input` for input files and :val:`output` for output files.
 Absence of :field:`role` or :val:`null` means that the role is undefined and SHOULD NOT be assumed.
-It is RECOMMENDED to order the output files in the list of calculation's relationships in a way to prioritize the files most likely to contain details describing the performed calculation.
-This way a piece of output parsing software would need to read in a small number of calculation output files to possibly identify the code and the type of calculation performed.
+It is RECOMMENDED to order the output files in the list of calculation's relationships in a way to prioritize the files most likely to contain details which could be used to automatically analyze the performed calculation.
+This way output parsing software would need to read in a small number of calculation output files to possibly identify the code, type and possibly parameters of calculation performed.
 
 References Entries
 ------------------
--- a/optimade.rst
+++ b/optimade.rst
@@ -3533,6 +3533,12 @@ Calculations Entries
 --------------------
 
 The :entry:`calculations` entries have the properties described above in section `Properties Used by Multiple Entry Types`_.
+Related :entry:`files` entries specified as relationships as described in section `Relationships`_ list files used or produced by a calculation.
+They MAY be defined as either input or output files in the :field:`role` field inside the :field:`meta` dictionary of the JSON:API resource identifier object.
+The allowed values are :val:`input` for input files and :val:`output` for output files.
+Absence of :field:`role` or :val:`null` means that the role is undefined and SHOULD NOT be assumed.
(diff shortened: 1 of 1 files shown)


blokhin (approved):

optimade.rst
```diff
@@ -3533,6 +3533,12 @@ Calculations Entries
 --------------------
 
 The :entry:`calculations` entries have the properties described above in section `Properties Used by Multiple Entry Types`_.
+Related :entry:`files` entries specified as relationships (as described in section `Relationships`_) list files used or produced by a calculation.
+They MAY be defined as either input or output files by the :field:`role` field inside the :field:`meta` dictionary of the JSON:API resource identifier object.
+The allowed values are :val:`input` for input files and :val:`output` for output files.
+Absence of :field:`role` or :val:`null` means that the role is undefined and SHOULD NOT be assumed.
```
vaitkus, line 3539:
```suggestion
If the :field:`role` field is not provided or the value is set to :val:`null`, a default role SHOULD NOT be assumed.
```

optimade.rst
```diff
@@ -3533,6 +3533,12 @@ Calculations Entries
...
+It is RECOMMENDED to order the output files in the list of calculation's relationships in a way to prioritize the files most likely to contain details which could be used to automatically analyze the performed calculation.
```
vaitkus, line 3540:
```suggestion
It is RECOMMENDED to order the output files in the list of calculation's relationships in a way that prioritizes the files most likely to contain details which could be used to automatically analyze the performed calculation.
```

optimade.rst
```diff
@@ -3533,6 +3533,12 @@ Calculations Entries
...
+This way output parsing software would need to read in a small number of calculation output files to possibly identify the code, type and possibly parameters of calculation performed.
```
vaitkus, line 3541:
```suggestion
Such an approach minimizes the number of calculation output files that need to be analyzed to possibly identify the code, type and parameters of the performed calculation.
```

merkys:
> My first reaction is: doesn't this belong in metadata rather than as data? In the sense `calculations` records the provenance of something, the "inputs", the "outputs" and the "process/workflow" seems the core data in such an entry?

My suggestion is to define the inputs/outputs division in metadata of relationships. Since we use them (relationships) to refer to other entries, then I think `meta` of relationships is the most appropriate place to store this metadata.

rartino:
JSON:API's meta of relationships is not something that exists in OPTIMADE outside of the JSON:API response format, so we have to define this "place" where you store this machine readable `role` in section 3.7. And, of we do that, why not encode that in the relationship name in JSON:API, instead of down inside the meta object?

From a "least surprise" perspective, it makes more sense to me that there is an "input" and an "output" relationship that contains the relationship linkage to inputs and outputs.

merkys:
We have discussed this proposal during the workshop and identified a drawback: there is no way to query on relationship metadata, nor now, nor after #524 gets merged in. A possible solution was put forward to use named relationships (now relationships are named after entry types), but this would possibly:
1. interfere with the common property-entrytype namespace
2. break the nice attribute which guarantees that relationships under the same name are of the same entry type

rartino:
I discussed with @merkys. He will add the below kind of construct as an example to #524 (which arguably follows from what is said there, and what already is in the specification in relation to "filtering on relationships"), which we think solves the remaining issue in the present PR:
```
structures?filter=references.target.role:references.target.doi HAS "32erk32rkj":"main"
```

merkys:
> I discussed with @merkys. He will add the below kind of construct as an example to #524 (which arguably follows from what is said there, and what already is in the specification in relation to "filtering on relationships"), which we think solves the remaining issue in the present PR:
> 
> ```
> structures?filter=references.target.role:references.target.doi HAS "32erk32rkj":"main"
> ```

Added in b8bb7937b10042168d1b090ba08863754652f790.

blokhin:
Would you like to standardize it further and to present a notion of the `master` output log file containing the most of the information? We did that in Nomad, MPDS, and Absolidix.

For example, for VASP that is `vasprun.xml` file, for FLEUR that is `out.xml` file, and for CRYSTAL that is simply the standard output, called by convention `*.o` or `*.cryst.out`.

blokhin:
The `master` has certain cultural references which we might be explicitly willing to support 💪

rartino (approved):
Adding more/stricter role definitions can be added later. I approve merging this as is.

blokhin (approved):

PR #524, opened 2024-06-12
Allow extended filtering on relationships

### LINKED ISSUE #437: Extended filtering on relationships
Currently we expected relationships filtering to be a two-step process:

> Note: formulating queries on relationships with entries that have specific property values is a multi-step process. For example, to find all structures with bibliographic references where one of the authors has the last name "Schmit" is performed by the following two steps:
>
>    - Query the references endpoint with a filter authors.lastname HAS "Schmit" and store the id values of the returned entries.
>    - Query the structures endpoint with a filter references.id HAS ANY <list-of-IDs>, where <list-of-IDs> are the IDs retrieved from the first query separated by commas.

In my implementation, I would like to support doing this in one-step, via e.g., 
  `/structures?filter=references.doi = "10.1234/12345"`.

This seems like a trivial extension to the specification (implementations MAY support relationship filtering via...). The only conflict would be the special `description` field we added for relationship filtering. I have not seen anyone using this, but we can maintain compatibility by reserving it as a keyword and never using `description` as an attribute name for any entry type (so that `references.description` is always unambiguously referring to the relationship description).

Am I missing some technical reason that we can't allow this syntax as optional, or was the issue that to be able to handle this robustly across different implementations we have to use the two-step process?

merkys:
This PR addresses #437 by allowing fully-fledged filtering on relationships. Prior to it, filtering was allowed, but only on `id` and `meta.description` of relationships. Now all properties of related entries should in principle be allowed to filter on, to arbitrary depth (OPTIONAL).

This introduces a slight backwards-incompatibility by abandoning special treatment for `description`. Thus queries which previously matched/mismatched, but were perfectly legal, now will return HTTP 400 for all standard entry types except `files` where `description` property is defined since v1.2.0. I am not sure how severe this incompatibility is - if we want to avoid it, we will have to mandate the special handling of `description` by re-routing it to `meta.description` of a relationship instead of entry's property.

Fixes #437.


rartino:
I'm confused. Even if we disregard that the `meta` subfield of the JSON:API relationship only makes sense in JSON:API, what in this PR explains why the meta subfield accesses the meta of the relationship, rather than the meta of the linked object?

optimade.rst
```diff
@@ -2000,30 +2000,34 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+Hence, the filter language acts as, for every entry type, there is a property with that name which contains a list of dictionaries with entry properties plus :property:`id`.
 For example: a client queries the :endpoint:`structures` endpoint with a filter that references :filter-fragment:`calculations.id`.
 For a specific structures entry, the nested property behaves as the list :filter-fragment:`["calc-id-43", "calc-id-96"]` and would then, e.g., match the filter :filter:`calculations.id HAS "calc-id-96"`.
 This means that the structures entry has a relationship with the calculations entry of that ID.
 
-    **Note**: formulating queries on relationships with entries that have specific property values is a multi-step process.
-    For example, to find all structures with bibliographic references where one of the authors has the last name "Schmidt" is performed by the following two steps:
+Support for queries on fields of arbitrary depth is OPTIONAL.
+For example, search for all structures with bibliographic references where one of the authors has the last name "Schmidt" could be performed with the following query:
```
ml-evs, line 2026: I think we actually need to revise this query example, since "name" is the required field for a `references` author and `lastname` is optional. How about a filter on DOI?

Or (in addition to this example) something more standardized, give me all references that correspond to structures of a given formula (or composition range)?

e.g., all literature references for Tantalum: `/v1/references?filter=structures.elements HAS "Ta"`

optimade.rst
```diff
@@ -2000,30 +2000,34 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
 
 Filtering on relationships
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 As described in the section `Relationships`_, it is possible for the API implementation to describe relationships between entries of the same, or different, entry types.
-The API implementation MAY support queries on relationships with an entry type :filter-fragment:`<entry type>` by using special nested property names:
+The API implementation MAY support queries on relationships with an entry type :filter-fragment:`<entry type>`:
 
 - :filter-fragment:`<entry type>.id` references a list of IDs of relationships with entries of the type :filter-fragment:`<entry type>`.
-- :filter-fragment:`<entry type>.description` references a correlated list of the human-readable descriptions of these relationships.
```
ml-evs, line 2013: See my main review comment, but I think we have to keep this still

ml-evs (commented):
I wholeheartedly support this, and indeed have been breaking the spec a bit myself to allow filtering on DOIs of references via this exact mechanism. This would be straightforward to enable in optimade-python-tools, as we currently hvae to *disable* filtering on fields other than ID in joins.

The only issue I see here is touched on by @rartino, previously we would allow filtering on the relationship description (which nobody ever used, and the filter was never implemented afaik), but somehow this is a breaking change that removes this functionality. Since we do not have a top-level description field of an entry, this could be kept in unambiguously, though slightly annoying for implementers, I think it makes sense (and we could never standardize a field called description for each entry in this case).

merkys:
> The only issue I see here is touched on by @rartino, previously we would allow filtering on the relationship description (which nobody ever used, and the filter was never implemented afaik), but somehow this is a breaking change that removes this functionality.

A somewhat drastic move would be to say that this design choice was a bug :sweat_smile: In a sense this would not be false, as `meta` for references object is JSON:API-specific, and we are trying to keep a generalized approach as much as possible.

> Since we do not have a top-level description field of an entry, this could be kept in unambiguously, though slightly annoying for implementers, I think it makes sense (and we could never standardize a field called description for each entry in this case).

`files` entry type has top-level `description` property since v1.2.0, which complicates things a bit.

optimade.rst
merkys: Query on all structures coming from the same DOI sounds good. One more suggestion of a bit more complicated query would be to filter for structures of certain anonymous formula related to publications appearing in year 2024.

optimade.rst
merkys: Yes, this has to be discussed. There are two options: retain full backwards compatibility by either encoding the special meaning of `description` or forbidding properties bearing such name, or scrap the special treatment of `description` altogether. This PR does the latter.

optimade.rst
merkys: I have added the said examples in 85c0a94bd339a3d03744a75d09f749d588171562, can you give them a look and mark as resolved if OK?

rartino:
I don't think filtering on the human readable description of a relationship is particularly crucial. However, the "role" of a relationship discussed in #523 is crucial to be able to filter on. (E.g. "does this calculation have an output structure that contains sodium").

merkys:
> However, the "role" of a relationship discussed in #523 is crucial to be able to filter on. (E.g. "does this calculation have an output structure that contains sodium").

Right, I agree that #523 needs further discussion. Nevertheless #523 should not block this PR.

rartino:
Since this PR asks to change how the dot operator works, I think we should try to peek forward enough so we do not paint ourselves into another corner.

Are there good reasons to not let the relationship still be the 'relationship object' (or strictly in OPTIMADE a list of relationship objects) and then let the target of the relationship object explicitly live under, e.g., 'target'?

To clarify, we could adopt a model of the structures relationship in an optimade calculation to work like: 
```
"structures": [
  {
    "description": "Initial structure for relaxation",
    "role": "input",
    "target": <here lives the structure entry>
},
{
    "description": "Final structure after relaxation",
    "role": "output",
    "target": <here lives the structure entry>
}
```

This would mean that a query for an output structure that contains sodium could be expressed as:
```
calculations?filter=structures.role:structures.target.elements HAS 'output':HAS "Na"
```
(As I type this up, I realize that we are actually missing the construct with the "inner" HAS for zip list constructs in our grammar, but that seems like an oversight that we should fix, given that we support e.g. the full set of fuzzy string operators in that position.)

merkys:
> Are there good reasons to not let the relationship still be the 'relationship object' (or strictly in OPTIMADE a list of relationship objects) and then let the target of the relationship object explicitly live under, e.g., 'target'?

I am not sure if I am getting this right, so bear with me. Do you mean retaining entry's relationships as they are (v1.2.0) and then including explicit copies of the related entries inside entry's `"attributes"`? If so, I think explicit inclusion might be redundant as JSON:API's `"included"` will cater for that already. Moreover, we may encounter arbitrary depth issues, which JSON:API avoids by only providing immediate relationships. Nevertheless such approach would indeed greatly simplify the specification of dot operator in filters.

ml-evs:
I have had a query related to this and think it would be worth discussing again at an upcoming meeting. For my implementations I am still breaking compliance by allowing filtering on `references.doi`...

rartino:
From the web meeting: how do one filter on `role=input` AND `target.doi = XYZ`? Some kind of adaption of the zip syntax, probably.

optimade.rst
```diff
@@ -2000,30 +2000,42 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+- :filter-fragment:`<entry type>.<property>` references a list of property :property:`<property>` values for related entries of type :filter-fragment:`<entry type>`.
```
ml-evs, line 2018:
```suggestion
- :filter-fragment:`<entry type>.target.<property>` references a list of property :property:`<property>` values for related entries (i.e., the :filter-fragment:`target` entry of the relationship) of type :filter-fragment:`<entry type>`.
```

optimade.rst
```diff
@@ -2000,30 +2000,42 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
 For example: a client queries the :endpoint:`structures` endpoint with a filter that references :filter-fragment:`calculations.id`.
 For a specific structures entry, the nested property behaves as the list :filter-fragment:`["calc-id-43", "calc-id-96"]` and would then, e.g., match the filter :filter:`calculations.id HAS "calc-id-96"`.
 This means that the structures entry has a relationship with the calculations entry of that ID.
 
-    **Note**: formulating queries on relationships with entries that have specific property values is a multi-step process.
-    For example, to find all structures with bibliographic references where one of the authors has the last name "Schmidt" is performed by the following two steps:
+Support for queries on fields of arbitrary depth is OPTIONAL.
+For example, search for all structures related to a publication having DOI 10.1234/1234 could be performed with the following query:
+
+    :query-url:`https://example.com/optimade/v1/structures?filter=references.doi=%2210.1234/1234%22`
```
ml-evs, line 2028:
```suggestion
    :query-url:`https://example.com/optimade/v1/structures?filter=references.target.doi=%2210.1234/1234%22`
```

optimade.rst
```diff
@@ -2000,30 +2000,42 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+
+Search for all literature references for structures with tantalum:
+
+    :query-url:`https://example.com/optimade/v1/references?filter=structures.elements+HAS+%22Ta%22`
```
ml-evs, line 2032:
```suggestion
    :query-url:`https://example.com/optimade/v1/references?filter=structures.target.elements+HAS+%22Ta%22`
```

optimade.rst
```diff
@@ -2000,30 +2000,42 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+
+Search for all structures of anonymous formula A2B from year 2024:
 
-    - Query the :endpoint:`references` endpoint with a filter :filter:`authors.lastname HAS "Schmidt"` and store the :filter-fragment:`id` values of the returned entries.
-    - Query the :endpoint:`structures` endpoint with a filter :filter-fragment:`references.id HAS ANY <list-of-IDs>`, where :filter-fragment:`<list-of-IDs>` are the IDs retrieved from the first query separated by commas.
+    :query-url:`https://example.com/optimade/v1/structures?filter=references.year=2023+AND+chemical_formula_anonymous=%22A2B%22`
```
ml-evs, line 2036:
```suggestion
    :query-url:`https://example.com/optimade/v1/structures?filter=references.target.year=2023+AND+chemical_formula_anonymous=%22A2B%22`
```

optimade.rst
```diff
@@ -2000,30 +2000,42 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
 - :filter-fragment:`identifier1` references a list of dictionaries that contain as an identifier :filter-fragment:`identifier2` and the filter matches for a flat list containing only the contents of :filter-fragment:`identifier2` for every dictionary in the list.
   E.g., if :filter-fragment:`identifier1` is the list :filter-fragment:`[{"identifier2":42, "identifier3":36}, {"identifier2":96, "identifier3":66}]`, then :filter-fragment:`identifier1.identifier2` is understood in the filter as the list :filter-fragment:`[42, 96]`.
 
+- :filter-fragment:`identifier1` references an entry type of a group of related entries each containing a property named after an identifier :filter-fragment:`identifier2` and the filter matches for a flat list containing only the values of :filter-fragment:`identifier2` properties for every related entry.
+  Support for such queries is OPTIONAL.
+  E.g., :filter-fragment:`references.doi` is understood in the filter as the list containing values of :property:`doi` for all related entries of type :entry:`references`.
+  This is explained in more detail in section `Filtering on relationships`_.
```
ml-evs, line 2006:
```suggestion
- :filter-fragment:`identifier1` references an entry type of a group of related entries each containing a property named after an identifier :filter-fragment:`identifier2` and the filter matches for a flat list containing only the values of :filter-fragment:`identifier2` properties for every related entry.
  In this case, :filter-fragment:`identifier2` is restricted to either :property:`id`, when matching directly on the ID of the related entry, or :property:`target`, which can be further nested to enable queries on the related entry's property.
  Support for such queries is OPTIONAL.
  E.g., :filter-fragment:`references.target.doi` is understood in the filter as the list containing values of :property:`doi` for all related entries of type :entry:`references`.
  This is explained in more detail in section `Filtering on relationships`_.
```

ml-evs (commented):
This quick review just works up the `target` that we discussed in the meeting. It feels a little cumbersome and feels strange to be introducing a filter-specific syntax inside a "field" name, but it achieves our goals in a backwards-compatible way and there seemed to be consensus (plus it is easy for servers to implement). I tried to come up with some other options (e.g., `relationships.structures.nsites = 20` or `included.structures.nsites=20`) that match other existing keywords, but they are also not satisfactory (and would be more of a breaking change).

optimade.rst
merkys: Re-added `description` in a40c0e00b4c3e3f63f0a7126704b8f276758a677. Now the PR retains backwards compatibility.

merkys:
I have updated the PR according to comments on Friday's web meeting. The PR now retains full backwards compatibility and allows for extended filtering on relationships through `target` field.

I did not implement `role` field yet which has as well been mentioned as I feel it deserves more discussion. If `role` is supposed to take values from some enumerator, its values has to be discussed. If `role` is supposed to be free-form, then `description` could be reused instead.

rartino (dismissed):
LGTM. But, the forward reference to that "roles" are defined in the section "Entry list" isn't ideal, given that we are set to remove that section and instead refer to the computer-readable property definitions part of the repository. I don't think it blocks merging for now, just that we need to also remember to update this when we remove that section.

rartino (approved):

optimade.rst
rartino: @ml-evs This PR is now quite urgently needed for our workshop work on Properties. Can you take a look @merkys reply?

rartino (dismissed):

optimade.rst
```diff
@@ -2002,30 +2007,55 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
 For example: a client queries the :endpoint:`structures` endpoint with a filter that references :filter-fragment:`calculations.id`.
 For a specific structures entry, the nested property behaves as the list :filter-fragment:`["calc-id-43", "calc-id-96"]` and would then, e.g., match the filter :filter:`calculations.id HAS "calc-id-96"`.
 This means that the structures entry has a relationship with the calculations entry of that ID.
 
-    **Note**: formulating queries on relationships with entries that have specific property values is a multi-step process.
-    For example, to find all structures with bibliographic references where one of the authors has the last name "Schmidt" is performed by the following two steps:
+Support for queries on fields of arbitrary depth is OPTIONAL.
+For example, search for all structures related to a publication having DOI 10.1234/1234 could be performed with the following query:
+
+    :query-url:`https://example.com/optimade/v1/structures?filter=references.target.doi=%2210.1234/1234%22`
```
ml-evs:
```suggestion
    :filter:`/structures?filter=references.target.doi="10.1234/1234"`
```

optimade.rst
```diff
@@ -2002,30 +2007,55 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+
+Search for all literature references for structures with tantalum:
+
+    :query-url:`https://example.com/optimade/v1/references?filter=structures.target.elements+HAS+%22Ta%22`
```
ml-evs:
```suggestion
    :filter:`/references?filter=structures.target.elements HAS "Ta"`
```

optimade.rst
```diff
@@ -2002,30 +2007,55 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+
+Search for all structures of anonymous formula A2B from year 2024:
+
+    :query-url:`https://example.com/optimade/v1/structures?filter=references.target.year=2024+AND+chemical_formula_anonymous=%22A2B%22`
```
ml-evs:
```suggestion
    :filter:`/structures?filter=references.target.year=2024 AND chemical_formula_anonymous="A2B"`
```

optimade.rst
```diff
@@ -2002,30 +2007,55 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+
+Note: the type of query discussed here corresponds to a "join"-type operation in a relational data model.
 
-    - Query the :endpoint:`references` endpoint with a filter :filter:`authors.lastname HAS "Schmidt"` and store the :filter-fragment:`id` values of the returned entries.
-    - Query the :endpoint:`structures` endpoint with a filter :filter-fragment:`references.id HAS ANY <list-of-IDs>`, where :filter-fragment:`<list-of-IDs>` are the IDs retrieved from the first query separated by commas.
+Search for all structures with primary citation from year 2024:
 
-    (Note: the type of query discussed here corresponds to a "join"-type operation in a relational data model.)
+    :query-url:`https://example.com/optimade/v1/structures?filter=references.role:references.target.year+HAS+%22_exmpl_primary%22:2024`
```
ml-evs:
```suggestion
    :filter:`/structures?filter=references.role:references.target.year HAS "_exmpl_primary":2024`
```

rartino (dismissed):

rartino (approved):

optimade.rst
```diff
@@ -2002,30 +2007,55 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+Hence, the filter language acts as, for every entry type, there is a property with that name which contains a list of dictionaries with keys :property:`id`, :property:`description`, :property:`role` and :property:`target`.
 For example: a client queries the :endpoint:`structures` endpoint with a filter that references :filter-fragment:`calculations.id`.
 For a specific structures entry, the nested property behaves as the list :filter-fragment:`["calc-id-43", "calc-id-96"]` and would then, e.g., match the filter :filter:`calculations.id HAS "calc-id-96"`.
 This means that the structures entry has a relationship with the calculations entry of that ID.
 
-    **Note**: formulating queries on relationships with entries that have specific property values is a multi-step process.
-    For example, to find all structures with bibliographic references where one of the authors has the last name "Schmidt" is performed by the following two steps:
+Support for queries on fields of arbitrary depth is OPTIONAL.
```
ml-evs, line 2351:
```suggestion
Support for queries on fields of arbitrary depth is OPTIONAL.

    **Note**: without this support, formulating queries on relationships with entries that have specific property values is a multi-step process.
        For example, to find all structures with bibliographic references where one of the authors has the last name "Schmidt" is performed by the following two steps:

        - Query the :endpoint:`references` endpoint with a filter :filter:`authors.lastname HAS "Schmidt"` and store the :filter-fragment:`id` values of the returned entries.
        - Query the :endpoint:`structures` endpoint with a filter :filter-fragment:`references.id HAS ANY <list-of-IDs>`, where :filter-fragment:`<list-of-IDs>` are the IDs retrieved from the first query separated by commas.
```

ml-evs (approved):
Thanks @merkys!

optimade.rst
```diff
@@ -2312,30 +2317,63 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
 - :filter-fragment:`identifier1` references a list of dictionaries that contain as an identifier :filter-fragment:`identifier2` and the filter matches for a flat list containing only the contents of :filter-fragment:`identifier2` for every dictionary in the list.
   E.g., if :filter-fragment:`identifier1` is the list :filter-fragment:`[{"identifier2":42, "identifier3":36}, {"identifier2":96, "identifier3":66}]`, then :filter-fragment:`identifier1.identifier2` is understood in the filter as the list :filter-fragment:`[42, 96]`.
 
+- :filter-fragment:`identifier1` references an entry type of a group of related entries each containing a property named after an identifier :filter-fragment:`identifier2` and the filter matches for a flat list containing only the values of :filter-fragment:`identifier2` properties for every related entry.
```
ndaelman-hu, line 2320: Okay, with the context above this is much more understandable. I think it can stay like this.
If I were to make it clearer, I would refer to the case above, as this is an extension:

 ```suggestion
- :filter-fragment:`identifier1` references an entry type, while :filter-fragment:`identifier2` denotes a property name present in each matched entry. The filter returns a flat list containing only the values of :filter-fragment:`identifier2` properties for every related entry. This is an extension to the previous case, across entries.
```

File: optimade.rst [hunk 14]
```diff
@@ -2312,30 +2317,63 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
 Filtering on relationships
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 As described in the section `Relationships`_, it is possible for the API implementation to describe relationships between entries of the same, or different, entry types.
-The API implementation MAY support queries on relationships with an entry type :filter-fragment:`<entry type>` by using special nested property names:
+The API implementation MAY support queries on relationships with an entry type :filter-fragment:`<entry type>`:
 
 - :filter-fragment:`<entry type>.id` references a list of IDs of relationships with entries of the type :filter-fragment:`<entry type>`.
 - :filter-fragment:`<entry type>.description` references a correlated list of the human-readable descriptions of these relationships.
+- :filter-fragment:`<entry type>.role` references a correlated list of roles between entries in relationships.
```
ndaelman-hu, line 2343: > correlated

Not sure what this means in this context.

optimade.rst
ndaelman-hu, line 2320: Note that it isn't clear from this paragraph:

1. whether the value corresponding to :filter-fragment:`identifier2` may be unset, i.e. be `Null`.
2. in what order the matched entry values should be returned.

If I understood correctly, a list is returned, rather than a dictionary mapping entry to corresponding property values.

optimade.rst
```diff
@@ -2312,30 +2317,63 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
+
+    ``/structures?filter=references.target.doi="10.1234/1234"``
+
+Search for all literature references for structures with tantalum:
+
+    ``/references?filter=structures.target.elements HAS "Ta"``
+
+Search for all structures of anonymous formula A2B from year 2024:
+
+    ``/structures?filter=references.target.year=2024 AND chemical_formula_anonymous="A2B"``
```
ndaelman-hu, line 2370: The examples and notes are very helpful here. But comparing with the notation above, i.e. :filter-fragment:`identifier1.identifier2`, it's unclear to which the `identifier`s map. This is mostly because the dot syntax is reserved to the specification and differs from the separator used in the queries.

I guess the identifiers in the first example are `structures` and `references.target.doi`, respectively. Applying the dot syntax verbatim would result in `references` and `target`, or `target` and `doi`.

I think that this is a relic from the syntax in cases 1 and 2, where no discrepancy arises. I would clarify this for case 3, though.

optimade.rst
```diff
@@ -2312,30 +2317,63 @@ A filter on a nested property name consisting of two identifiers :filter-fragmen
...
     For example, to find all structures with bibliographic references where one of the authors has the last name "Schmidt" is performed by the following two steps:
 
-    - Query the :endpoint:`references` endpoint with a filter :filter:`authors.lastname HAS "Schmidt"` and store the :filter-fragment:`id` values of the returned entries.
-    - Query the :endpoint:`structures` endpoint with a filter :filter-fragment:`references.id HAS ANY <list-of-IDs>`, where :filter-fragment:`<list-of-IDs>` are the IDs retrieved from the first query separated by commas.
+      - Query the :endpoint:`references` endpoint with a filter :filter:`authors.lastname HAS "Schmidt"` and store the :filter-fragment:`id` values of the returned entries.
+
+      - Query the :endpoint:`structures` endpoint with a filter :filter-fragment:`references.id HAS ANY <list-of-IDs>`, where :filter-fragment:`<list-of-IDs>` are the IDs retrieved from the first query separated by commas.
+
+For example, search for all structures related to a publication having DOI 10.1234/1234 could be performed with the following query:
```
ndaelman-hu, line 2360:
```suggestion
For example, search for all structures related to a publication (stored in a different entry type) having DOI 10.1234/1234 could be performed with the following query:
```

ndaelman-hu (approved):
I added some suggestions where the formulation could be clearer. Still, the specification is correct, just somewhat dense for fresh eyes. **This should not delay the merger of this PR.**

ml-evs:
> I added some suggestions where the formulation could be clearer. Still, the specification is correct, just somewhat dense for fresh eyes. **This should not delay the merger of this PR.**

Thanks @ndaelman-hu, I have to resolve your comments to be able to merge, but will open an issue tracking the clarifications you suggested.

optimade.rst
merkys: This sounds as if the response (to a filter) is supposed to return a flat list with values of said properties. This is not entirely true as filter returns all matching entries with all their properties.

optimade.rst
merkys: We use term "correlated list" throughout the specification when talking about parallel lists, i.e., lists of the same length where *i*-th positions are related.

optimade.rst
merkys: > Note that it isn't clear from this paragraph:
> 
> 1. whether the value corresponding to :filter-fragment:`identifier2` may be unset, i.e. be `Null`.

I need to check what the specification says about matching `NULL` values, but here their handling is supposed to be as elsewhere in the OPTIMADE specification.

> 2. in what order the matched entry values should be returned.

I am not sure what is the supposed default ordering, but I guess it is implementation-specific thing.

> If I understood correctly, a list is returned, rather than a dictionary mapping entry to corresponding property values.

No, this returns the same format as regular entry listing response would.

optimade.rst
merkys: Identifiers in this case map to dot-separated parts of nested property names (`references`, `target`, `doi`). `structures` is the entry type. I do not quite see how case 3 is different from cases 1 and 2, could you please elaborate?

optimade.rst
merkys: I would say "(described in a related `references` entry)".

ndaelman-hu:
```suggestion
For example, search for all structures related to a publication (described in a related `references` entry) having DOI 10.1234/1234 could be performed with the following query:
```

optimade.rst
ndaelman-hu: Okay, thank you for the clarification. That remark aside, what do you think of the overall suggestion?

```suggestion
- :filter-fragment:`identifier1` references an entry type, while :filter-fragment:`identifier2` denotes a property name present in each matched entry. The filter returns a flat list containing the matching entry. This specification is an extension to the previous case so it applies across entries.
```

optimade.rst
ndaelman-hu: Okay, this is helpful for my understanding of the specification. I did not entirely pick up on this from just this PR. More specifically:

1. I guess that there is no "pure" example of case 3, which led to my confusion. What you are saying is that `references.target.doi` -> `references` `target.doi` is case 3 and then `target.doi` -> `target` `doi` is case 1. It would help to add this stepwise breakdown to at least 1 example.
2. From the examples and discussions, I thought that `references` also were an entry type? Am I wrong here?

merkys: I might have misunderstood what you mean by "case". By them, do you mean the listed examples, or something else?

optimade.rst
merkys: Yes, this adheres better with the language of the two points above it, yet I would rephrase it like this:
> `identifier1` references a related entry type, while `identifier2` denotes a property present in each related entry.
> The filter matches for a flat list containing only the values of `identifier2` property for every related entry of type specified by `identifier1`.
> This specification is an extension to the previous point so it applies across entries.

ndaelman-hu:
```suggestion
- :filter-fragment:`identifier1` references a related entry type, while `identifier2` denotes a property present in each related entry.
The filter matches for a flat list containing only the values of `identifier2` property for every related entry of type specified by identifier1.
This specification is an extension to the previous point so it applies across entries.
```

optimade.rst
ndaelman-hu: No, sorry, I meant the bullet point specifications right under "Nested property names" (line 2309).
The one added in this PR marks the third case, though they aren't numbered in document.

optimade.rst
merkys: Nice. If we are in agreement over this, I will draft a PR with this text rephrased.

merkys: Um, just noticed I have to respect the fields listed below, will have to rephrase to pay attention to them, but I will do that later, not for this meeting.

merkys:
I have not forgotten to make the adjustments suggested by @ndaelman-hu. Please bear with me, I will make a PR tomorrow or early next week.

PR #529, opened 2024-06-14
Allow provider-specific data types

merkys:
This PR attempts to generalize #436 by allowing to define provider-specific data types. The following conditions for such data types apply:
* Their values have to be expressed as string literals both in responses and in filters;
* They have to be named using provider-specific prefixes;
* Properties can only be compared with values of the same type (as per specification up til now).

The latter condition might be too strict for some uses, but since casting is unclear at the moment, let us leave such extensions for future.

Merging this PR will allow to reassign #392 and #436 (SMILES) to `_cheminfo_` namespace where it can become a property with special set of rules for comparison.


optimade.rst
```diff
@@ -2070,6 +2072,10 @@ As the filter language syntax does not define a lexical token for timestamps, va
 In a comparison with a timestamp property, a string token represents a timestamp value that would result from parsing the string according to RFC 3339 Internet Date/Time Format.
 Interpretation failures MUST be reported with error :http-error:`400 Bad Request`.
 
+Database and definition providers MAY introduce custom types, representing them with string lexical tokens both in filters and responses.
+It is up to the providers to decide which comparison operators to support and how the comparisons should be performed.
+For example, if a provider intoduces a set-valued property :property:`_exmpl_set`, it may decide to define operator :val:`CONTAINS` so that :filter:`identifier CONTAINS set` is true if :val:`set` is a subset of a property.
```
vaitkus, line 2077:
```suggestion
For example, a provider that introduces a set-valued property :property:`_exmpl_set`, may decide to override the :val:`CONTAINS` operator so that :filter:`identifier CONTAINS set` is true when :val:`set` is a subset of a property.
```

optimade.rst
```diff
@@ -215,7 +215,7 @@ Hence, entry properties are described in this proposal using
 context-independent types that are assumed to have some form of
 representation in all contexts. They are as follows:
 
-- Basic types: **string**, **integer**, **float**, **boolean**, **timestamp**.
+- Basic types: **string**, **integer**, **float**, **boolean**, **timestamp**, database-provider-specific or definition-provider-specific data type.
```
vaitkus, line 218: The same `database-provider-specific or definition-provider-specific data type` phrase is repeated multiple times throughout the specification. I like that we are being really specific, but maybe we could defined a term somewhere (e.g. `custom data type`) and then refer to it troughout the text?

merkys: Fair point, I will check the term definition section to see how well it would fit there.

rartino: On another point: I think it seems odd to list provider-specific data types under "Basic types". To the extent that we do the separation into Basic vs. list/dictionary on the basis of "contains one thing" vs "many things", it also isn't clear to me why the defined types would have to be seen as the former. I suggest we move the segment about database-specific datatypes below the definitions of the datatypes that are explicitly defined by the standard.

merkys: My idea is to allow only those provider-specific data types that can be expressed as strings (e.g., symmetry operators, complex numbers etc). Timestamp type which OPTIMADE already has could fall under the same category of strings with internal semantics. That is why I lumped provider-specific data types together with them. But I agree that they could be split off to a separate segment.

merkys: > The same `database-provider-specific or definition-provider-specific data type` phrase is repeated multiple times throughout the specification. I like that we are being really specific, but maybe we could defined a term somewhere (e.g. `custom data type`) and then refer to it troughout the text?

What about calling them "namespace-specific data types"? The specification already has a section "Namespace Prefixes" which is later split into database-specific and definition provider-specific ones, thus I think "namespace" is a right term to use.

optimade.rst
```diff
@@ -215,11 +215,12 @@ Hence, entry properties are described in this proposal using
 context-independent types that are assumed to have some form of
 representation in all contexts. They are as follows:
 
-- Basic types: **string**, **integer**, **float**, **boolean**, **timestamp**.
+- Basic types: **string**, **integer**, **float**, **boolean**, **timestamp**
```
vaitkus, line 218:
```suggestion
- Basic types: **string**, **integer**, **float**, **boolean**, **timestamp**.
```

optimade.rst
merkys: I have replaced "database-provider-specific or definition-provider-specific" with "namespace-specific" in 3dd56bb2e500d933df6b2857078cb34ae17636ad.

optimade.rst
```diff
@@ -627,6 +628,7 @@ In the JSON response format, property types translate as follows:
 - **timestamp** uses a string representation of date and time as defined in `RFC 3339 Internet Date/Time Format <https://tools.ietf.org/html/rfc3339#section-5.6>`__.
 - **dictionary** is represented by the JSON object type.
 - **unknown** properties are represented by either omitting the property or by a JSON :field-val:`null` value.
+- namespace-specific data types use string representations.
```
vaitkus, line 631:
```suggestion
- Namespace-specific data types use string representations.
```

optimade.rst
```diff
@@ -1816,6 +1818,7 @@ The following tokens are used in the filter query component:
   (Note that at the end of the string value above the four final backslashes represent the two terminal backslashes in the value, and the final double quote is a terminator, it is not escaped.)
 
   String value tokens are also used to represent **timestamps** in form of the `RFC 3339 Internet Date/Time Format <https://tools.ietf.org/html/rfc3339#section-5.6>`__.
+  String value tokens as well are used to represent namespace-specific data types.
```
vaitkus, line 1821: Maybe this could be made into a small list, i.e.:

String value tokens are also used to represent:
- **timestamps** in form of the `RFC 3339 Internet Date/Time Format <https://tools.ietf.org/html/rfc3339#section-5.6>`__.
- Namespace-specific data types.

optimade.rst
```diff
@@ -2070,6 +2073,10 @@ As the filter language syntax does not define a lexical token for timestamps, va
 In a comparison with a timestamp property, a string token represents a timestamp value that would result from parsing the string according to RFC 3339 Internet Date/Time Format.
 Interpretation failures MUST be reported with error :http-error:`400 Bad Request`.
 
+Namespace providers MAY introduce custom types, representing them with string lexical tokens both in filters and responses.
```
vaitkus, line 2076:
```suggestion
Namespace providers MAY introduce namespace-specific data types, representing them with string lexical tokens both in filters and responses.
```

optimade.rst
```diff
@@ -220,6 +220,7 @@ representation in all contexts. They are as follows:
   A list can be empty, i.e., contain no items.
 - **dictionary**: an associative array of **keys** and **values**, where **keys** are pre-determined strings, i.e., for the same entry property, the **keys** remain the same among different entries whereas the **values** change.
   The **values** of a dictionary can be any basic type, list, dictionary, or unknown.
+- Namespace-specific data type.
```
vaitkus, line 223: Maybe it would make sense to add a link here to the part of the specification where "Namespace-specific" data types are defined, e.g. "Type handling and conversions in comparisons"?

blokhin:
I have a general question: how to understand, if a prefix denotes a provider-specific datatype, not a general-purpose common datatype. In other words, compare `_cheminfo_` and `_mpds_`, where is what?

blokhin:
apparently, by underscore as the first symbol, correct?

blokhin (dismissed):

optimade.rst
merkys: Link added, marking as resolved.

merkys:
> I have a general question: how to understand, if a prefix denotes a provider-specific datatype, not a general-purpose common datatype. In other words, compare `_cheminfo_` and `_mpds_`, where is what?

Provider-specific data type names are only going to appear in `x-optimade-type` in property definitions. There they will be identifiable by an initial underscore and namespacing prefix. General purpose data types should be elevated to the main specification, thus without `_<prefix>_`.

optimade.rst
```diff
@@ -2158,7 +2167,7 @@ A Property Definition MUST be composed according to the combination of the requi
 
 - :field:`x-optimade-type`: String.
   Specifies the OPTIMADE data type for this level of the defined property.
-  MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, or :val:`"dictionary"`.
+  MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, :val:`"dictionary"` or a string naming namespace-specific data type starting with provider-specific prefix.
```
vaitkus, line 2170:
```suggestion
  MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, :val:`"dictionary"` or the name of a namespace-specific data type that starts with an underscore followed by the provider-specific prefix.
```

vaitkus: Feel free to reword this in a different way than suggested,

vaitkus (approved):

vaitkus (approved):

blokhin (approved):

blokhin:
@merkys should we merge?

merkys:
> @merkys should we merge?

@blokhin I would like to get a review from @rartino as this change affects property definitions and might impact the way query filters are supposed to be handled by the backends.

ml-evs:
> Ok, I'll take a deeper look at this over the next couple of days.

Likewise, as I want to use this pretty much straight away!

ml-evs (commented):
Thanks for writing this up @merkys!

I only have one general comment; should we encourage the custom type to be described in the description of the fields that use it? Otherwise I don't see where this information would go (even if its only human-readable anyway). I've added two suggestions below along these lines.

optimade.rst
```diff
@@ -2158,7 +2167,7 @@ A Property Definition MUST be composed according to the combination of the requi
 
 - :field:`x-optimade-type`: String.
   Specifies the OPTIMADE data type for this level of the defined property.
-  MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, or :val:`"dictionary"`.
+  MUST be one of :val:`"string"`, :val:`"integer"`, :val:`"float"`, :val:`"boolean"`, :val:`"timestamp"`, :val:`"list"`, :val:`"dictionary"` or the name of a namespace-specific data type that starts with an underscore followed by the provider-specific prefix.
```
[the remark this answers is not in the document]
rartino: I guess in the end we will want this info to go into our "standards" definitions (see https://github.com/Materials-Consortia/OPTIMADE/tree/develop/schemas#standards-definitions ). For the OPTIMADE one see here: https://schemas.optimade.org/defs/v1.2/standards/optimade.html 

If forced to handle this today I would put that info into the description field there. Something along the lines of: 

> Description: The Magnetic materials standard node types and properties.
>
> Apart from the supplied property definitions, the Magnetic materials standard defines a namespace-specific data type for the magnetic susceptibility tensor "_mag_sustensor" that allows comparison using < and > to express filters of the magnetic susceptibility magnitude. In more detail, the following operators are valid: 
> ...

BUT, writing this up; even though the magnetic susceptibility tensor example is quite bogus, I realize we may be overly limiting deciding all custom datatypes must be represented as strings? Why not allow it to be expressed as any basic or container datatype? 

But, back to @ml-evs comment: to allow not repeating the definition a lot, maybe this is better?:

> When a namespace-specific data type is used, the human-readable property description should be used to describe its usage and format or to give a reference to where this information is available.

(That reference can then be to the "standards definition" if it is included there).

merkys:
Thanks for reviews, @ml-evs and @rartino!

Data type description in both human-readable text and Property Definitions is noncontroversial, thus I will address it later on.

Now on the extension beyond the string-based data types. Historically, I drafted this PR having SMILES in mind, thus strings were enough for me. Moreover, we already have strings with internal semantics (timestamps). Now I do not see why we cannot extend this to all basic+list+dict data types, but yes, this needs some more thought.

Some issues coming to my mind as I am writing this:

1. Intuition. A human reader is capable to perceive that a string has internal semantics. Just by looking at lists and dicts it is difficult to say whether they constitute a data type or not. And this risks becoming counterintuitive for number-based data types where, for example, `>` operator is overloaded to mean the opposite.

2. Accessing constituents. Should filters on constituents of list (items) and dict (keys and values) be supported, re-using compound identifier syntax as everywhere else in OPTIMADE?

rartino:
Just a note as I prepare for the meeting this afternoon: lets discuss the string-based vs. general custom data types at the meeting and decide whether it is fine with just string support for now.

@merkys Can you please look at @ml-evs edit suggestion above and act on it (either do it, reply, or close)

optimade.rst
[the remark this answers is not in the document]
merkys: I think this comment better fits to "Property Definition keys from JSON Schema" section, next to definition of `type`.

merkys:
During an online discussion on 2025-04-03 it has been decided to allow provider-specific data types to be based on any OPTIMADE data type, not just string. Thus the PR has to be adjusted to accommodate that.

optimade.rst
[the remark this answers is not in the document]
rartino: @ml-evs This is no longer relevant with the latest changes; I 'resolve' this now. If you still see issues, make a new comment.

optimade.rst
```diff
@@ -2160,7 +2166,8 @@ A Property Definition MUST be composed according to the combination of the requi
...
```
blokhin, line 2169: Should we add here `blob_raw_bytes` and `pointer_reference` 🤔

merkys: We may, but maybe as a separate PR? This one is made specifically for namespace provider-specific data types like SMILES in cheminformatics and the like.

rartino: @blokhin Please file an issue and we can discuss there. I don't think this is related to this PR? (I'll click 'resolve')

ml-evs (approved):
Thanks @merkys! I think this is in a good shape, even if there is no programmatic place to describe such types yet (e.g., if there are multiple properties that use the same custom type).

Issue #471, opened 2023-06-12
Deserializing a whole OPTIMADE API as a file


ml-evs:
DRAFT AS I NEED TO CLOSE MY LAPTOP

At the workshop, we discussed a file format specification that would allow an entire OPTIMADE API to be stored in a single file on disk. This would be useful for many applications (archival + restore, data transfer etc.). We (@unkcpz, @eimrek, @giovannipizzi) proposed a draft of this file format that is approximately the following:

- We use the JSON lines format: every line in the file is a valid JSON object, separated by newlines.
- First entry must be a dictionary with the key `x-optimade-api-version` with value e.g., `1.2.0`.
- Then comes any other metadata:
Then corresponding info endpoint data for the types in the file
Then each line contains an entry from the corresponding endpoints

optimade_example.jsonl:

{“x-optimade”: {...}}
{“meta”: {“api_version”: “1.1.0”}}\n
{“info”: {“properties”: [“_aiida_cell_volume”: {“...”}}}\n
{“info/structures”: {“properties”: [“_aiida_cell_volume”: {“type”: number}}}\n
{“info/references”: {...}}\n
{“data”: {“type”: “structures”, “id”: “1234”, “attributes”: {...}}\n
{“data”: {“type”: “structures”, “id”: “1234”, “attributes”: {...}}\n
{“data”: {“type”: “references”, “id”: “sfdas”, “attributes”: {...}}

-

ml-evs:
Just a note that we adopted a very similar format to this in https://github.com/materialscloud-org/optimade-maker, though the OP is slightly out of date. 

I can see the benefits of allowing both a single file and a directory structure (suggested above by @merkys). Somehow this could be capture in the top line of the root file, outlining whether the "client" should read the next lines of the file to find data, or look elsewhere in the filesystem.

PR #531, opened 2024-07-19
OPTIMADE JSON Lines specification appendix

ml-evs:
This PR begins the work on #471, by defining the conventions on top of JSON Lines for deserializing an entire OPTIMADE API into a file.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -4520,9 +4520,9 @@ The OPTIMADE JSON Lines format then extends these rules with the following conve
 - The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
 - The next block of lines provides the ``info`` endpoint responses.
   - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
-  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. These MUST appear in alphabetical order by entry type name.
-- The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_.
-  Again, these MUST appear in block alphabetical order by entry type name, but can appear in any order within those blocks.
+  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. 
+- The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_. 
+These entries can be provided in any order.
 - Finally, any custom extension endpoints (see `Custom Extension Endpoints`_), if present and desirable, MUST appear at the end of the file.
 
 This leaves the following overall file structure:
@@ -4551,6 +4551,6 @@ Example OPTIMADE JSON Lines File
   {"type": "info", "id": "/", "attributes": {"api_version": "1.2.0", "available_api_versions": ["1.2.0"], "formats": ["json"], "entry_types_by_format": {"json": ["references", "structures"]}, "license": "https://example.com/licenses/example_license.html"}, "homepage": "https://example.com", "name": "Example API", "provider": {"description": "A simple example provider", "name": "Example Provider"}}}
   {"type": "info", "id": "references", ...}
   {"type": "info", "id": "structures", ...}
-  {"type": "structures", "id": "1", "attributes": {...}}
   {"type": "references", "id": "2", "attributes": {...}}
+  {"type": "structures", "id": "1", "attributes": {...}}
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -4508,7 +4508,7 @@ The `JSON Lines <https://jsonlines.org/>`__ format enforces the following rules:
 - each line is a valid JSON object,
 - each line is separated by a newline character (``\n``), optionally ending the file with a newline.
 - each file must be UTF-8 encoded,
-- the recommended file extension is ``.jsonl``, with natural extensions to ``.jsonl.gz`` and ``jsonl.bz2`` for ``gzip`` and ``bzpi2`` compressed files, respectively.
+- the recommended file extension is ``.jsonl``, with natural extensions to ``.jsonl.gz`` and ``jsonl.bz2`` for ``gzip`` and ``bzip2`` compressed files, respectively.
 
 The OPTIMADE JSON Lines format then extends these rules with the following conventions:
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -4512,16 +4512,15 @@ The `JSON Lines <https://jsonlines.org/>`__ format enforces the following rules:
 
 The OPTIMADE JSON Lines format then extends these rules with the following conventions:
 
-- The first line of the file is a JSON object that contains metadata about the file. 
+- The first line of the file is a JSON object that contains metadata about the file.
   It MUST comprise of a dictionary with the key ``x-optimade``, under which the following key MUST be defined:
-  
+
   - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
 
 - The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
-- The next block of lines provides the ``info`` endpoint responses. 
+- The next block of lines provides the ``info`` endpoint responses.
   - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
-  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. 
-    These MUST appear in alphabetical order by entry type name.
+  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. These MUST appear in alphabetical order by entry type name.
 - The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_.
   Again, these MUST appear in block alphabetical order by entry type name, but can appear in any order within those blocks.
 - Finally, any custom extension endpoints (see `Custom Extension Endpoints`_), if present and desirable, MUST appear at the end of the file.
@@ -4536,7 +4535,7 @@ This leaves the following overall file structure:
   <entry info responses>
   <entries block ordered by entry type>
   <optional custom extension endpoints>
-  
+
 
 This JSONL format can also be used to share provider-specific properties.
 These should be consistent with any external definitions, and where appropriate, prefixes tied to the tools used to generate the file should be used.
--- a/optimade.rst
+++ b/optimade.rst
@@ -4592,7 +4592,7 @@ This leaves the following overall file structure:
 
 This JSONL format can also be used to share provider-specific or namespace-specific properties, following the normal rules defined in `Namespace Prefixes`_.
 Such static property definitions should be consistent with any externally defined prefixes, that is, JSONLines files should not use an existing prefix unless the desired definitions match exactly.
-Where appropriate, prefixes tied to the tools used to generate the file can be used, in cases where an external property definition is not provided (for example, a property that only applies to the current dataset).
+Where appropriate, a provider prefix can be chosen to represent the tool used to generate the file, in cases where an external property definition is not provided (for example, a property that only applies to the current dataset).
 
 It is RECOMMENDED that custom properties are defined in full within the JSONL file, or pointed to a specific versioned property definition.
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -4581,7 +4581,7 @@ This leaves the following overall file structure:
   <optional metadata>
   <base info response>
   <entry info responses>
-  <entries block ordered by entry type>
+  <unordered entries>
   <optional custom extension endpoints>
 
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -4566,11 +4566,16 @@ The OPTIMADE JSON Lines format then extends these rules with the following conve
   - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
 
 - The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
-- The next block of lines provides the ``info`` endpoint responses.
+
+- The next block of lines provides the ``info`` endpoint responses:
+
   - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
+
   - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_.
+
 - The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_.
   These entries can be provided in any order.
+
 - Finally, any custom extension endpoints (see `Custom Extension Endpoints`_), if present and desirable, MUST appear at the end of the file.
 
 This leaves the following overall file structure:
--- a/optimade.rst
+++ b/optimade.rst
@@ -4600,5 +4600,5 @@ Example OPTIMADE JSON Lines File
   {"type": "info", "id": "references", ...}
   {"type": "info", "id": "structures", ...}
   {"type": "references", "id": "2", "attributes": {...}}
-  {"type": "structures", "id": "1", "attributes": {...}}
+  {"type": "structures", "id": "1", "attributes": {...}, "relationships": {"references": {"data": [{"id": "2", "type": "references"}]}}}
 
--- a/optimade.rst
+++ b/optimade.rst
@@ -4520,9 +4520,9 @@ The OPTIMADE JSON Lines format then extends these rules with the following conve
 - The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
 - The next block of lines provides the ``info`` endpoint responses.
   - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
-  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. 
-- The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_. 
-These entries can be provided in any order.
+  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_.
+- The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_.
+  These entries can be provided in any order.
 - Finally, any custom extension endpoints (see `Custom Extension Endpoints`_), if present and desirable, MUST appear at the end of the file.
 
 This leaves the following overall file structure:
--- a/optimade.rst
+++ b/optimade.rst
@@ -4495,3 +4495,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to 
     * XML Schema appears to use a compatible regex format, except it is implicitly anchored: i.e., the beginning-of-input ``^`` and end-of-input ``$`` anchors must be removed, and missing anchors replaced by ``.*``.
     * POSIX Extended regexes (and their extended GNU implementations) are incompatible because ``\`` is not a special character in character classes.
       POSIX Basic regexes also have further differences, e.g., the meaning of some escaped syntax characters is reversed.
+
+
+The OPTIMADE JSON Lines Format for Database Exchange
+----------------------------------------------------
+
+There are many use cases for which it is beneficial to share all of the data served by an OPTIMADE API as a single file, for example, archival, transfer of entire databases and local-first clients.
+This appendix describes a lightweight standardization for doing this via the `JSON Lines <https://jsonlines.org/>`__ format, with some additional OPTIMADE-specific conventions.
+
+The `JSON Lines <https://jsonlines.org/>`__ format enforces the following rules:
+
+- each line is a valid JSON object,
+- each line is separated by a newline character (``\n``), optionally ending the file with a newline.
+- each file must be UTF-8 encoded,
+- the recommended file extension is ``.jsonl``, with natural extensions to ``.jsonl.gz`` and ``jsonl.bz2`` for ``gzip`` and ``bzpi2`` compressed files, respectively.
+
+The OPTIMADE JSON Lines format then extends these rules with the following conventions:
+
+- The first line of the file is a JSON object that contains metadata about the file. 
+  It MUST comprise of a dictionary with the key ``x-optimade``, under which the following key MUST be defined:
+  - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
+
+- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
+- The next block of lines provides the ``info`` endpoint responses. 
+  - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
+  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. 
+    These MUST appear in alphabetical order by entry type name.
+- The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_.
+  Again, these MUST appear in block alphabetical order by entry type name, but can appear in any order within those blocks.
+- Finally, any custom extension endpoints (see `Custom Extension Endpoints`_), if present and desirable, MUST appear at the end of the file.
+
+This leaves the following overall file structure:
+
+.. code :: txt
+
+  <header>
+  <optional metadata>
+  <base info response>
+  <entry info responses>
+  <entries block ordered by entry type>
+  <optional custom extension endpoints>
+  
+
+This JSONL format can also be used to share provider-specific properties.
+These should be consistent with any external definitions, and where appropriate, prefixes tied to the tools used to generate the file should be used.
+It is RECOMMENDED that custom properties are defined in full within the JSONL file, or pointed to a specific versioned property definition.
+
+Example OPTIMADE JSON Lines File
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code :: jsonc
+
+  {"x-optimade": {"api_version": "1.2.0"}}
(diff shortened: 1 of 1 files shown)


optimade.rst
```diff
@@ -4495,3 +4495,57 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+This leaves the following overall file structure:
+
+```
+<header>
+<optional metadata>
+<base info response>
+<entry info responses>
+<entries block ordered by entry type>
+<optional custom extension endpoints>
+```
```
ml-evs, line 4537:
```suggestion
.. code:: jsonc
    <header>
    <optional metadata>
    <base info response>
    <entry info responses>
    <entries block ordered by entry type>
    <optional custom extension endpoints>

```

optimade.rst
```diff
@@ -4495,3 +4495,57 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+```jsonlines
+{"x-optimade": {"api_version": "1.2.0"}}
+{"meta": {"time_stamp": "2024-07-19T11:47:10Z", "data_returned": 6, "provider": {"name": "Example JSONL", "description": "An example JSONL file.", "prefix": "_exmpl"}}}
+{"type": "info", "id": "/", "attributes": {"api_version": "1.2.0", "available_api_versions": ["1.2.0"], "formats": ["json"], "entry_types_by_format": {"json": ["references", "structures"]}, "license": "https://example.com/licenses/example_license.html"} "homepage": "https://example.com", "name": "Example API", "provider": {"description": "A simple example provider", "name": "Example Provider"}}}
+{"type": "info", "id": "references", ...}
+{"type": "info", "id": "structures", ...}
+{"type": "structures", "id": "1", "attributes": {...}
+```
```
ml-evs, line 4551:
```suggestion
.. code:jsonc
    {"x-optimade": {"api_version": "1.2.0"}}
    {"meta": {"time_stamp": "2024-07-19T11:47:10Z", "data_returned": 6, "provider": {"name": "Example JSONL", "description": "An example JSONL file.", "prefix": "_exmpl"}}}
    {"type": "info", "id": "/", "attributes": {"api_version": "1.2.0", "available_api_versions": ["1.2.0"], "formats": ["json"], "entry_types_by_format": {"json": ["references", "structures"]}, "license": "https://example.com/licenses/example_license.html"} "homepage": "https://example.com", "name": "Example API", "provider": {"description": "A simple example provider", "name": "Example Provider"}}}
    {"type": "info", "id": "references", ...}
    {"type": "info", "id": "structures", ...}
    {"type": "structures", "id": "1", "attributes": {...}

```

optimade.rst
```diff
@@ -4495,3 +4495,57 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+    <entries block ordered by entry type>
+    <optional custom extension endpoints>
+
+
+- Information about prefixes and provider fields
+
+Example OPTIMADE JSON Lines File
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:jsonc
```
ml-evs, line 4544:
```suggestion
.. code:: jsonc
```

optimade.rst
```diff
@@ -4495,3 +4495,57 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+  - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
+  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. 
+    These MUST appear in alphabetical order by entry type name.
+- The remaining lines of the file contain data entries themselves, described in `Entry Listng JSON Response Schema`_.
+  Again, these MUST appear in block alphabetical order by entry type name, but can appear in any order within those blocks.
+- Finally, any custom extension endpoints (see `Custom Extension Endpoints`_), if present and desirable, MUST appear at the end of the file.
+
+This leaves the following overall file structure:
+
+.. code:: jsonc
```
ml-evs, line 4530:
```suggestion
.. code
```

optimade.rst
```diff
@@ -4495,3 +4495,57 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+    <entries block ordered by entry type>
+    <optional custom extension endpoints>
+
+
+- Information about prefixes and provider fields
+
+Example OPTIMADE JSON Lines File
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: jsonc
```
ml-evs, line 4544:
```suggestion
.. code
```

optimade.rst
```diff
@@ -4495,3 +4495,57 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+
+This leaves the following overall file structure:
+
+.. code
+    <header>
+    <optional metadata>
+    <base info response>
+    <entry info responses>
+    <entries block ordered by entry type>
+    <optional custom extension endpoints>
```
ml-evs, line 4536:
```suggestion
.. code
  <header>
  <optional metadata>
  <base info response>
  <entry info responses>
  <entries block ordered by entry type>
  <optional custom extension endpoints>
```

optimade.rst
```diff
@@ -4495,3 +4495,57 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+Example OPTIMADE JSON Lines File
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code
+    {"x-optimade": {"api_version": "1.2.0"}}
+    {"meta": {"time_stamp": "2024-07-19T11:47:10Z", "data_returned": 6, "provider": {"name": "Example JSONL", "description": "An example JSONL file.", "prefix": "_exmpl"}}}
+    {"type": "info", "id": "/", "attributes": {"api_version": "1.2.0", "available_api_versions": ["1.2.0"], "formats": ["json"], "entry_types_by_format": {"json": ["references", "structures"]}, "license": "https://example.com/licenses/example_license.html"} "homepage": "https://example.com", "name": "Example API", "provider": {"description": "A simple example provider", "name": "Example Provider"}}}
+    {"type": "info", "id": "references", ...}
+    {"type": "info", "id": "structures", ...}
+    {"type": "structures", "id": "1", "attributes": {...}
```
ml-evs, line 4550:
```suggestion
.. code
  {"x-optimade": {"api_version": "1.2.0"}}
  {"meta": {"time_stamp": "2024-07-19T11:47:10Z", "data_returned": 6, "provider": {"name": "Example JSONL", "description": "An example JSONL file.", "prefix": "_exmpl"}}}
  {"type": "info", "id": "/", "attributes": {"api_version": "1.2.0", "available_api_versions": ["1.2.0"], "formats": ["json"], "entry_types_by_format": {"json": ["references", "structures"]}, "license": "https://example.com/licenses/example_license.html"} "homepage": "https://example.com", "name": "Example API", "provider": {"description": "A simple example provider", "name": "Example Provider"}}}
  {"type": "info", "id": "references", ...}
  {"type": "info", "id": "structures", ...}
  {"type": "structures", "id": "1", "attributes": {...}
```

optimade.rst
```diff
@@ -4495,3 +4495,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+The OPTIMADE JSON Lines format then extends these rules with the following conventions:
+
+- The first line of the file is a JSON object that contains metadata about the file.
+  It MUST comprise of a dictionary with the key ``x-optimade``, under which the following key MUST be defined:
+
+  - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
+
+- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
+- The next block of lines provides the ``info`` endpoint responses.
+  - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
```
eimrek, line 4522: > MUST

This breaks current optimade-maker jsonl files (well, just the li-ion-conductors one). Would a MAY, in principle, work here?

optimade.rst
```diff
@@ -4495,3 +4495,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_. These MUST appear in alphabetical order by entry type name.
```
eimrek, line 4523: > These MUST appear in alphabetical order by entry type name.

Fine for me, but is it needed?

optimade.rst
```diff
@@ -4495,3 +4495,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+
+.. code :: jsonc
+
+  {"x-optimade": {"api_version": "1.2.0"}}
+  {"meta": {"time_stamp": "2024-07-19T11:47:10Z", "data_returned": 6, "provider": {"name": "Example JSONL", "description": "An example JSONL file.", "prefix": "_exmpl"}}}
+  {"type": "info", "id": "/", "attributes": {"api_version": "1.2.0", "available_api_versions": ["1.2.0"], "formats": ["json"], "entry_types_by_format": {"json": ["references", "structures"]}, "license": "https://example.com/licenses/example_license.html"}, "homepage": "https://example.com", "name": "Example API", "provider": {"description": "A simple example provider", "name": "Example Provider"}}}
+  {"type": "info", "id": "references", ...}
+  {"type": "info", "id": "structures", ...}
+  {"type": "structures", "id": "1", "attributes": {...}}
+  {"type": "references", "id": "2", "attributes": {...}}
```
eimrek, line 4555: `references` block should be ahead of `structures`, as required above. But again, is the alphabetical ordering needed?

eimrek (commented):
Thanks @ml-evs, and sorry for the delays. Now had some time to take a look. I like the proposal and it can be used for optimade-maker as well. But I wrote some comments for consideration.

optimade.rst
```diff
@@ -4495,3 +4495,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+
+The OPTIMADE JSON Lines Format for Database Exchange
+----------------------------------------------------
+
+There are many use cases for which it is beneficial to share all of the data served by an OPTIMADE API as a single file, for example, archival, transfer of entire databases and local-first clients.
+This appendix describes a lightweight standardization for doing this via the `JSON Lines <https://jsonlines.org/>`__ format, with some additional OPTIMADE-specific conventions.
+
+The `JSON Lines <https://jsonlines.org/>`__ format enforces the following rules:
+
+- each line is a valid JSON object,
```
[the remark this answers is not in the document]
ml-evs: Hmm, not sure, to me value implies something stored under a key, whereas object is the smallest chunk of data that can be parsed entirely as JSON?

optimade.rst
ml-evs: This is quite important IMO, we don't provide any other way of identifying the file (and in fact we might want to even expand this with a version of the JSONL specification like ``{'x-optimade': {'api_version': 1.1, 'jsonl_version': 1.2}`` if we want the JSONL file to be able to store OPTIMADE data of any version. Would you be against simply adding it to the files where it is missing atm?

optimade.rst
ml-evs: Happy to relax this constraint, probably we'll never need to worry about an OPTIMADE JSONL file containing 100s of entry types (in which case you could use alphabetical order to short-circuit the search for a given entry type).

optimade.rst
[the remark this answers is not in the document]
eimrek: on https://jsonlines.org/ they mention "values" (as even e.g. just strings or integers are supported). But I agree that in the vast majority of cases, people will use objects.

optimade.rst
eimrek: Sorry, not sure if my comment was precise enough. I was commenting on the base info endpoint, which doesn't contain any crucial info as far as i understand, but I see how it can be useful to know what follows in the next lines.

optimade.rst
eimrek: ok, i see, but anyway, fine for me. good to keep consistent if we require the  entry blocks to be sorted, where this is more valuable, i think.

optimade.rst
```diff
@@ -4495,3 +4495,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+
+.. code :: jsonc
+
+  {"x-optimade": {"api_version": "1.2.0"}}
+  {"meta": {"time_stamp": "2024-07-19T11:47:10Z", "data_returned": 6, "provider": {"name": "Example JSONL", "description": "An example JSONL file.", "prefix": "_exmpl"}}}
+  {"type": "info", "id": "/", "attributes": {"api_version": "1.2.0", "available_api_versions": ["1.2.0"], "formats": ["json"], "entry_types_by_format": {"json": ["references", "structures"]}, "license": "https://example.com/licenses/example_license.html"}, "homepage": "https://example.com", "name": "Example API", "provider": {"description": "A simple example provider", "name": "Example Provider"}}}
+  {"type": "info", "id": "references", ...}
+  {"type": "info", "id": "structures", ...}
+  {"type": "structures", "id": "1", "attributes": {...}}
+  {"type": "references", "id": "2", "attributes": {...}}
```
eimrek, line 4555:
```suggestion
  {"type": "references", "id": "1", "attributes": {...}}
  {"type": "structures", "id": "2", "attributes": {...}}
```

switched em

optimade.rst
ml-evs: Revisiting this, I'm going to remove the constraint and allow entry types to be interleaved. I've found this useful when scraping another database and pulling included entries, for large databases this would require an additional final sorting step which is not ideal.

ml-evs:
CI is blocked by #536, but perhaps you could take another look when you get a chance @eimrek? I'll invite some other reviewers too.

rartino (approved):

optimade.rst
```diff
@@ -4543,3 +4543,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
     * XML Schema appears to use a compatible regex format, except it is implicitly anchored: i.e., the beginning-of-input ``^`` and end-of-input ``$`` anchors must be removed, and missing anchors replaced by ``.*``.
     * POSIX Extended regexes (and their extended GNU implementations) are incompatible because ``\`` is not a special character in character classes.
       POSIX Basic regexes also have further differences, e.g., the meaning of some escaped syntax characters is reversed.
+
+
+The OPTIMADE JSON Lines Format for Database Exchange
+----------------------------------------------------
+
+There are many use cases for which it is beneficial to share all of the data served by an OPTIMADE API as a single file, for example, archival, transfer of entire databases and local-first clients.
+This appendix describes a lightweight standardization for doing this via the `JSON Lines <https://jsonlines.org/>`__ format, with some additional OPTIMADE-specific conventions.
```
merkys, line 4552:
```suggestion
This appendix describes a lightweight standardization for doing this via the `JSON Lines <https://jsonlines.org/>`__ (JSONL) format, with some additional OPTIMADE-specific conventions.
```
The following text uses JSONL abbreviation, I suggest introducing it here.

optimade.rst
```diff
@@ -4543,3 +4543,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+  <header>
+  <optional metadata>
+  <base info response>
+  <entry info responses>
+  <entries block ordered by entry type>
+  <optional custom extension endpoints>
+
+
+This JSONL format can also be used to share provider-specific properties.
+These should be consistent with any external definitions, and where appropriate, prefixes tied to the tools used to generate the file should be used.
```
merkys, line 4589: I do not quite get the "prefixes tied to the tools used to generate the file should be used" part. Are there any examples for such cases?

merkys (commented):
Looks good to me either. The part on "prefixes tied to the tools used to generate the file" is unclear to me, otherwise happy to merge this.

optimade.rst
ml-evs: Hmm, yeah, this has lost its meaning a bit. The intention was that a tool like aiida or ASE could support OPTIMADE JSONL as an export format, and any "tool-specific definitions" like `_aiida_node_id` (or whatever) should be defined and use the tool prefix. I wouldn't be against removing this sentence as the format just follows the normal OPTIMADE rules about defining properties.

merkys: Right, thanks for the explanation. Since "MUST" is not used in this paragraph, I gather its main purpose is to recommend best practice. We may probably want that all the OPTIMADE JSONL files bear some sort of producer identification, but we probably do not want to standardize it at this point. But maybe we can reuse `meta.implementation` dictionary from the main specification?

Maybe it is worth stressing that OPTIMADE JSONL files have the same requirements ("MUST/SHOULD/MAY") as OPTIMADE responses, in terms of custom properties having prefixes and metadata descriptions. But maybe it is enough just to say that?

Also, what about sharing provider-specific entry types? If we mention provider-specific properties, should be allow entry types as well?

eimrek: Regarding non-optimade-spec properties, doesn't the spec say they must have either

- the provider specific prefix; or
- a defininition-provider-specific prefix?

I think we can just point to section 3.5 here, no?

optimade.rst
```diff
@@ -4543,3 +4543,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+- the recommended file extension is ``.jsonl``, with natural extensions to ``.jsonl.gz`` and ``jsonl.bz2`` for ``gzip`` and ``bzip2`` compressed files, respectively.
+
+The OPTIMADE JSON Lines format then extends these rules with the following conventions:
+
+- The first line of the file is a JSON object that contains metadata about the file.
+  It MUST be a dictionary with the key ``x-optimade``, under which the following key MUST be defined:
+
+  - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
+
+- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
```
eimrek, line 4568: Should we specify what meaning fields like `meta.data_returned` take in this context?

optimade.rst
```diff
@@ -4543,3 +4543,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+- The next block of lines provides the ``info`` endpoint responses.
+  - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
+  - The next lines MUST contain the entry info endpoint responses for the all entry types present later in the file, as described in `Entry Listing Info Endpoints`_.
+- The remaining lines of the file contain data entries themselves, described in `Entry Listing JSON Response Schema`_.
+  These entries can be provided in any order.
```
eimrek, line 4573: The order of the data entries is not clear to me. Below, we say `<entries block ordered by entry type>`, while here they can be in any order. I don't have strong feelings about this, but it should be clear. If we want to mandate block-order (e.g. all structures first, then all references, etc), i guess we should say something about both, 1) the order of the blocks (should they match the info endpoint order?); 2) the order of the entries within the blocks.

My own gut feeling is that perhaps the order of blocks is not important, but maybe it's good to have one type of entries all together in a block.

optimade.rst
```diff
@@ -4543,3 +4543,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+  <entry info responses>
+  <entries block ordered by entry type>
+  <optional custom extension endpoints>
+
+
+This JSONL format can also be used to share provider-specific properties.
+These should be consistent with any external definitions, and where appropriate, prefixes tied to the tools used to generate the file should be used.
+It is RECOMMENDED that custom properties are defined in full within the JSONL file, or pointed to a specific versioned property definition.
+
+Example OPTIMADE JSON Lines File
```
eimrek, line 4592:
```suggestion
Example file
```
Considering there is now two `OPTIMADE JSON Lines File` standards (other one being 9.4), maybe better to be more vague here (e.g. i already mixed these up when looking through the table of contents).

optimade.rst
```diff
@@ -4543,3 +4543,62 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+The OPTIMADE JSON Lines format then extends these rules with the following conventions:
+
+- The first line of the file is a JSON object that contains metadata about the file.
+  It MUST be a dictionary with the key ``x-optimade``, under which the following key MUST be defined:
+
+  - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
+
+- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
+- The next block of lines provides the ``info`` endpoint responses.
+  - First the base info response MUST be provided, following the description at `Base Info Endpoint`_.
```
eimrek, line 4570: this sublist is not rendering correctly to me

eimrek (commented):
Thanks a lot, Matthew, for working on this. Left some comments.

ml-evs:
> An additional question about custom prefixes, that is (still) not clear to me, is what should happen when `provider1` creates a JSONL file, and `provider2` subsequently serves this.
> 
> Should all `_provider1` prefixes be replaced by `_provider2`?
> 
> or would `_provider1` prefixes be retained and this would fall under [3.5.2](https://github.com/Materials-Consortia/OPTIMADE/blob/develop/optimade.rst?rgh-link-date=2025-02-05T11%3A19%3A16Z#352definition-provider-specific-namespace-prefixes)?

I think "serving" is an entirely different thing; `provider2` should decide whether to re-host the properties with their own prefix, or point to the `provider1` definition if it is available. It's a bit tricky to write this out (also somewhat addressed by d2c095a) -- somehow there are properties that are going to be namespace scoped to just the file in question as the file is the "provider". They can then be served with any prefix as long as they are consistent.

optimade.rst
ml-evs: fixed, thanks

optimade.rst
```diff
@@ -4543,3 +4543,69 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+  <optional metadata>
+  <base info response>
+  <entry info responses>
+  <unordered entries>
+  <optional custom extension endpoints>
+
+
+This JSONL format can also be used to share provider-specific or namespace-specific properties, following the normal rules defined in `Namespace Prefixes`_.
+Such static property definitions should be consistent with any externally defined prefixes, that is, JSONLines files should not use an existing prefix unless the desired definitions match exactly.
+Where appropriate, prefixes tied to the tools used to generate the file can be used, in cases where an external property definition is not provided (for example, a property that only applies to the current dataset).
```
eimrek, line 4595: @ml-evs thanks, this seems good. Although, would something like the suggestion below be more precise? Basically saying that there can only be one "custom" tool-based prefix and it should take "role" of the provider prefix in this case.

```suggestion
Where appropriate, a provider prefix can be chosen to represent the tool used to generate the file, in cases where an external property definition is not provided (for example, a property that only applies to the current dataset).
```

ml-evs: Yep, I like that, thanks!

optimade.rst
ml-evs:
```suggestion
- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule. For example, `meta.data_returned` can be omitted, or otherwise used to provide the total number of entries in the file.
```

Something like this?

eimrek: @ml-evs ok, yep, sounds good.

ml-evs:
Thanks for the comments @eimrek, @merkys and @rartino -- I think I've clarified them now so this should be ready to go again.

optimade.rst
```diff
@@ -4543,3 +4543,69 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+- the recommended file extension is ``.jsonl``, with natural extensions to ``.jsonl.gz`` and ``jsonl.bz2`` for ``gzip`` and ``bzip2`` compressed files, respectively.
+
+The OPTIMADE JSON Lines format then extends these rules with the following conventions:
+
+- The first line of the file is a JSON object that contains metadata about the file.
+  It MUST be a dictionary with the key ``x-optimade``, under which the following key MUST be defined:
+
+  - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
+
+- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule. For example, `meta.data_returned` can be omitted, or otherwise used to provide the total number of entries in the file.
```
eimrek, line 4568:
```suggestion
- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule. For example, ``meta.data_returned`` can be omitted, or otherwise used to provide the total number of entries in the file.
```

optimade.rst
```diff
@@ -4543,3 +4543,69 @@ Implementations that do not produce errors in this situation are RECOMMENDED to
...
+- the recommended file extension is ``.jsonl``, with natural extensions to ``.jsonl.gz`` and ``jsonl.bz2`` for ``gzip`` and ``bzip2`` compressed files, respectively.
+
+The OPTIMADE JSON Lines format then extends these rules with the following conventions:
+
+- The first line of the file is a JSON object that contains metadata about the file.
+  It MUST be a dictionary with the key ``x-optimade``, under which the following key MUST be defined:
+
+  - ``api_version``: The OPTIMADE API version used when generating the file, as described in the ``meta`` member in `JSON Response Schema: Common Fields`_.
+
+- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule. For example, ``meta.data_returned`` can be omitted, or otherwise used to provide the total number of entries in the file.
```
vaitkus, line 4568:
```suggestion
- The next line MAY contain a standard OPTIMADE ``meta`` object, following the same rules described in `JSON Response Schema: Common Fields`_, where every MUST and SHOULD rule can be reinterpreted as a MAY rule.
  For example, ``meta.data_returned`` can be omitted, or otherwise used to provide the total number of entries in the file.
```

vaitkus: Feel free to dismiss this comment if you would rather have this formatting change made after the merge.

vaitkus (dismissed):

optimade.rst
```diff
@@ -4592,11 +4592,11 @@ This leaves the following overall file structure:
...
 This JSONL format can also be used to share provider-specific or namespace-specific properties, following the normal rules defined in `Namespace Prefixes`_.
 Such static property definitions should be consistent with any externally defined prefixes, that is, JSONLines files should not use an existing prefix unless the desired definitions match exactly.
-Where appropriate, prefixes tied to the tools used to generate the file can be used, in cases where an external property definition is not provided (for example, a property that only applies to the current dataset).
+Where appropriate, a provider prefix can be chosen to represent the tool used to generate the file, in cases where an external property definition is not provided (for example, a property that only applies to the current dataset).
 
 It is RECOMMENDED that custom properties are defined in full within the JSONL file, or pointed to a specific versioned property definition.
 
-Example OPTIMADE JSON Lines File
+Example file
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
merkys, line 4600:
```suggestion
~~~~~~~~~~~~
```

merkys (dismissed):
I have no objections for this to be merged. I have myself encountered a need to store database subsets in OPTIMADE format, and this format seems to be just what I need.

vaitkus (approved):

PR #537, opened 2025-02-14
Update links

vaitkus:
Minor link cleanup.


merkys (approved):
We indeed have moved to JSON:API v1.1 everywhere. And `dx` part is no longer needed nor recommended part for DOI resolver URL.

ml-evs (approved):
Thanks!

PR #539, opened 2025-04-25
Fractional coordinates and spans

sauliusg:
Suggest specification of the fractional coordinate representation in a server response as discussed in the latest additions to the #416 issue.

Allow explicit designation of the atom site extent (was not possible previously).


.words.lst
```diff
@@ -1,12 +1,14 @@
...
 ABNF
 ACM
+APIs
 Aa
 Acta
 BaCa
 BadLuck
 Bb
 BibTeX
+CARRAGE
```
vaitkus, line 11: This word is a typo (CARRIAGE) that has already been corrected in the development branch. I would be better to exclude this word from the exception list to avoid future mistakes.

vaitkus (commented):
Overall looks good to me. I  suggested a few stylistic changes to not deviate from the rest of the specification too much and also added a few suggestions to further clarify a few of the concepts.

optimade.rst
```diff
@@ -3282,10 +3282,94 @@ cartesian\_site\_positions
...
+    If the implemenatation returns `cartesian_site_positions`_ without also returning the :property:`site_coordinate_span`, the coordinates MUST span the unit cell (i.e. the default value for the :property:`site_coordinate_span` MUST be :val:`"unit_cell"`), to maintain compatibility with the 1.1 OPTIMADE implementations.
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+  - The value of this property MUST be one of the following:
+
+    - :val:`"fundamental_domain"`: means that sites described in the response span a fundamental domain of a periodic system.
+      When a server indicates this span in the response, it MUST provide those and only those sites that enable reconstruction of the whole periodic system by applying symmetry operations from `space_group_symmetry_operations_xyz`_ property and then applying translations given by `lattice_vectors`_.
+      The fundamental domain does not need to be a connected space region.
+    - :val:`"asymmetric_unit"`: All sites are in a connected space region that is a fundamental domain, as per IUCr Online Dictionary of Crystallography definition.
```
ml-evs, line 3323: Great to see progress on this @sauliusg! My only initial query is whether the distinction between fundamental domain and asymmetric unit is a useful one -- I can't find a good reference for it in the IUCr Online Dictionary: https://dictionary.iucr.org/index.php?search=fundamental+domain

sauliusg: > Great to see progress on this @sauliusg! My only initial query is whether the distinction between fundamental domain and asymmetric unit is a useful one -- I can't find a good reference for it in the IUCr Online Dictionary: https://dictionary.iucr.org/index.php?search=fundamental+domain

"Fundamental domain" is a more general term, more often used in mathematics. A good intro is in Wikipedia [1]. I introduced it since, as Rickard pointed out, "asymmetric unit" needs to be a *connected* region of space, but in the COD we do not guarantee that in any sense. The I use "fundamental domain" to indicate that it can be also disconnected ([1]: "A fundamental domain always contains a free regular set U, an open set moved around by G into disjoint copies"); although I see now that this Wikipedia article is a bit crude. I'll look up a math textbook/handbook on that.

Ref.:

[1] https://en.wikipedia.org/wiki/Fundamental_domain

optimade.rst
```diff
@@ -3282,10 +3282,104 @@ cartesian\_site\_positions
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description** Indicates the extent of the material (crystal) described in the response.
+  In particular, properties `cartesian_site_positions`_ and `fractional_site_positions`_ MUST contain all sites *belonging* to the described extent, and only those sites.
+- **Type**: string
+- **Requirements/conventions**:
+
+  - **Support**: MUST be supported by all implementations if coordinates `fractional_site_positions`_ are returned.
+    It SHOULD be supported if coordinates in `cartesian_site_positions`_ are returned.
+    If the implementation returns `cartesian_site_positions`_ without also returning the :property:`site_coordinate_span`, the coordinates MUST span the unit cell (i.e. the default value for the :property:`site_coordinate_span` MUST be :val:`"unit_cell"`), to maintain compatibility with the 1.1 OPTIMADE implementations.
```
merkys, line 3316: I would say "implementations of OPTIMADE v1.2.0 and earlier", as v1.2.0 is the most recent release.

sauliusg: Included formulation of v1.2.0 or earlier into the text.

merkys:
I am trying to understand whether newly introduced properties are orthogonal to `"dimension_types"` property. In particular, when a supercell is returned, the system can still be periodic and non-periodic, right? If so, then there is no need to change anything in the definition of `"dimension_types"` and `nperiodic_dimensions`.

sauliusg:
> I am trying to understand whether newly introduced properties are orthogonal to `"dimension_types"` property. In particular, when a supercell is returned, the system can still be periodic and non-periodic, right? 

Correct.

>If so, then there is no need to change anything in the definition of `"dimension_types"` and `nperiodic_dimensions`.

Yes, that's how I understand it.

The proposed PR is a *conservative* addition to the existing site representation. Anything that was expressed using Cartesian coordinates can be (should be?) possible to express using fractional coordinates for the *same* sites. In addition, we can explicitly tell the client what extent of the structure is presented (i.e. what parts of the whole structure are *missing* ;)

optimade.rst
```diff
@@ -3282,10 +3282,94 @@ cartesian\_site\_positions
...
+  - Values of the :property:`site_span_extents` inner lists SHOULD be null for non-periodic dimensions.
+
+- **Examples**:
+
+  - :val:`[[0,0],[0,0],[0,0]]` indicates the response with the structure in the "first octant", i.e. with fractional coordinates in the range :val:`[0;1),[0;1),[0;1)`;
+
+  - :val:`[[-1,1],[-1,1],[-1,1]]` a 3x3 box around the origin;
+
+  - :val:`[[3,4],[2,3],[5,6]]` indicates a :val:`2x2x2` supercell with the origin shifted to the point :val:`(3,2,5)`;
+
```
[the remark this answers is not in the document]
vaitkus: Note no longer relevant in the updated text.

optimade.rst
```diff
@@ -3282,10 +3282,94 @@ cartesian\_site\_positions
...
+    Translations MUST be specified as multiples of the :property:`lattice_vectors` *a*, *b* and *c*; the first inner 2-element list corresponds to the minimum and maximum translations along *a*, the second – along *b*, and the third – along *c*.
+  - Values of the :property:`site_span_extents` inner lists SHOULD be null for non-periodic dimensions.
+
+- **Examples**:
+
+  - :val:`[[0,0],[0,0],[0,0]]` indicates the response with the structure in the "first octant", i.e. with fractional coordinates in the range :val:`[0;1),[0;1),[0;1)`;
+
+  - :val:`[[-1,1],[-1,1],[-1,1]]` a 3x3 box around the origin;
+
+  - :val:`[[3,4],[2,3],[5,6]]` indicates a :val:`2x2x2` supercell with the origin shifted to the point :val:`(3,2,5)`;
```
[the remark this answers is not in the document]
vaitkus: These examples no longer exist in the updated PR.

optimade.rst
```diff
@@ -3282,10 +3282,74 @@ cartesian\_site\_positions
...
+  - **Support**: MUST be supported by all implementations if coordinates `fractional_site_positions`_ are returned.
+    It SHOULD be supported if coordinates in `cartesian_site_positions`_ are returned.
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+  - The value of this property MUST be one of the following:
+
+    - :val:`"fundamental_domain"`: means that sites described in the response span a fundamental domain (Vinberg, 1994; European Mathematical Society, 2020) of a periodic system.
+      When a server indicates this span in the response, it MUST provide those and only those sites that enable reconstruction of the whole periodic system by applying symmetry operations from `space_group_symmetry_operations_xyz`_ property and then applying translations given by `lattice_vectors`_.
+      The fundamental domain does not need to be a connected space region.
+    - :val:`"asymmetric_unit"`: all sites are in a simply connected space region that is a fundamental domain, as per IUCr Online Dictionary of Crystallography definition (IUCr, 2017).
```
vaitkus, line 3323: It would be useful to explicitly mention with which of these spans the `space_group_symmetry_operations_xyz` or equivalent space group information becomes REQUIRED instead of RECOMMENDED. Alternatively, we might repeat a statement similar to the one given in `space_group_symmetry_operations_xyz`:

```
The property is REQUIRED if symmetry operations are
necessary to reconstruct the full model of the material
and no other symmetry information (e.g., the Hall symbol)
is provided that would allow the user to derive symmetry
operations unambiguously.
```

optimade.rst
```diff
@@ -3282,10 +3282,74 @@ cartesian\_site\_positions
...
+  A site is usually used to describe positions of atoms; what atoms can be encountered at a given site is conveyed by the :property:`species_at_sites` property, and the species themselves are described in the :property:`species` property.
+  Site coordinates MAY be given as `cartesian_site_positions`_, `fractional_site_positions`_, or both.
+  When symmetry operations given in `space_group_symmetry_operations_xyz`_ property are applied, they MUST be applied to coordinates given in the `fractional_site_positions`_ array.
+- **Type**: list of list of floats
+- **Requirements/Conventions**:
+
+  - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - It MUST be a list of length equal to the number of sites in the structure, where every element is a list of the three fractional coordinates of a site expressed as float values in the fractions of the unit cell vectors given by the `lattice_vectors`_ property.
```
vaitkus, line 3298: Support for the `lattice_vectors` property is RECOMMENDED (`SHOULD be supported`). It should be mentioned in the `lattice_vectors` definition, here or in both places that `lattice_vectors` property becomes MANDATORY when fractional coordinates are provided.

optimade.rst
```diff
@@ -3282,10 +3282,74 @@ cartesian\_site\_positions
...
+
+- **Description** Human-readable semi-formal characterization of the coordinate span when the `site_coordinate_span`_ property has value :val:`"other"`.
+- **Type**: string
+- **Example**: :val:`"Two fullerene molecules with a VdW contact."`
+
 nsites
 ~~~~~~
 
-- **Description**: An integer specifying the length of the :property:`cartesian_site_positions` property.
+- **Description**: An integer specifying the number of sites in a unit cell.
```
vaitkus, line 3352: Is a unit cell always properly defined here? I think OPTIMADE does not require to specify the `lattice_vectors` nor the space group information properties (although there are some requirements for space group information and fractional coordinates). For example, if we return Cartesian coordinates with the `molecular_entities` span and there is no unit cell or space group information, does the 'nsites' property then becomes `null`?

optimade.rst
```diff
@@ -3282,10 +3282,74 @@ cartesian\_site\_positions
...
+    - :val:`"unit_cell"`: a full unit cell of a periodic system (crystal).
+      For this span, the server MUST provide a set of sites in the response that can be used to reconstruct the whole periodic system (crystal) by simply applying translations from the `lattice_vectors`_ property to those sites.
+    - :val:`"molecular_unit_cell"`: same as :val:`"unit_cell"`, but in addition places atoms that are bound by covalent or coordination bonds at a bond distance from each other.
+    - :val:`"molecular_entities"`: sets of atoms that are bound by covalent or coordination bonds, as per IUPAC definition of a 'molecular entity'.
+      This set of sites MAY be larger than a fundamental domain.
+    - :val:`"other"`: any other collection of sites that does not fit the enumerated values above.
+    - :val:`null`: the span is not specified. In this case the best that the client can assume is that the sites span the whole unit cell (i.e. the `site_coordinate_span` is `"unit_cell"`). This is the assumed behavior of all main implentations before the `site_coordinate_span` definition was introduced.
+
```
vaitkus, line 3333: Maybe we should explicitly add a statement here about the default value being "`unit_cell`"? We could use the same phrasing that is used elsewhere in the specification, e.g.

If omitted or :val:`null`, the default value of :field:`site_coordinate_span` is :val:`unit_cell`.

sauliusg: This has been discussed with @rartino and we came to the wording as in the current text. We can not change defaults in the previous spec and we do not need defaults in the new spec (there should always be a `site_coordinate_span`).

optimade.rst
sauliusg: That's a good question. For aperiodic structures, 'nsites' should (probably) be the number of sites in the model. For slabs, the unit cell will be 2D, with the "natural" thickness of the slab defining the number of atoms in the third, non-periodic dimension and so on. Bu this is my understanding, we need to ask ML people what they want to search :)

optimade.rst
sauliusg: Technically, we can leave the `lattice_vectors` at SHOULD. The server then MAY return a response with fractional coordinates and without the lattice vectors, but it is up to the client what to do with this response – it will probably be useless, so why should anyone want to return it? I mean, we can leave to sort this out at a semantic/application level, without wanting to "police" the meaning of the responses.

Switching  `lattice_vectors` to MUST is undesired since some structures returned by OPTIMADE protocol are aperiodic, and making the requirement level depend on the values of other fields is against the current OPTIMADE design principle as I understood it from my conversation with Rickard last time (and I support the idea of property independence in the specification as well).

optimade.rst
rartino: From web meeting discussion: maybe adjust to:
```suggestion
- **Description**: An integer specifying the number of sites in a unit cell defined by the property :property:`lattice_vectors`.
```

optimade.rst
```diff
@@ -3282,10 +3282,74 @@ cartesian\_site\_positions
...
+      The fundamental domain does not need to be a connected space region.
+    - :val:`"asymmetric_unit"`: all sites are in a simply connected space region that is a fundamental domain, as per IUCr Online Dictionary of Crystallography definition (IUCr, 2017).
+    - :val:`"molecular_fundamental_domain"`: a fundamental domain where all atoms connected by covalent or donor-acceptor coordination bonds are adjacent to each other, placed at a bond distance.
+    - :val:`"molecular_asymmetric_unit"`: an asymmetric unit (a simply connected fundamental domain) where all atoms bound by covalent or donor-acceptor coordination bonds are adjacent to each other, placed at a bond distance.
+    - :val:`"unit_cell"`: a full unit cell of a periodic system (crystal).
+      For this span, the server MUST provide a set of sites in the response that can be used to reconstruct the whole periodic system (crystal) by simply applying translations from the `lattice_vectors`_ property to those sites.
```
rartino: From web meet discussion: should we somehow shortly express that the unit cell is allowed to be any repeating unit of the crystal?
```suggestion
    - :val:`"unit_cell"`: a full unit cell of a periodic system (crystal), i.e., any repeating unit chosen by the server defined by the property :property:`lattice_vectors`.
      For this span, the server MUST provide a set of sites in the response that can be used to reconstruct the whole periodic system (crystal) by simply applying translations from the `lattice_vectors`_ property to those sites.
```

optimade.rst
vaitkus: It was agreed to loosen the `space_group_symmetry_operations_xyz` requirement instead. This will be conducted in a separate PR. Closing this conversation to avoid blocking the merging.

optimade.rst
vaitkus: During yesterday's meeting this was discussed a bit further. It was agreed that the default value could be set to `unit_cell`.

optimade.rst
vaitkus: This suggestion was discussed during yesterday's meeting. It was agreed that the requirement level should not be changed from 'SHOULD' to 'MUST'.

optimade.rst
vaitkus: This issue was resolved by updating the definition to "<...> in unit cell defined by the field :property:`lattice_vectors`.".

optimade.rst
vaitkus: The reference text no longer belongs in the branch, so the conversation can be marked as resolved.

optimade.rst
vaitkus: @ml-evs are you happy with the response?

Looking to resolve this conversation as to not block the merging of the branch.

optimade.rst
sauliusg: Inserted the corresponding text.

optimade.rst
```diff
@@ -3282,10 +3282,76 @@ cartesian\_site\_positions
...
+    - :val:`"molecular_unit_cell"`: same as :val:`"unit_cell"`, but in addition places atoms that are bound by covalent or coordination bonds at a bond distance from each other.
+    - :val:`"molecular_entities"`: sets of atoms that are bound by covalent or coordination bonds, as per IUPAC definition of a 'molecular entity'.
+      This set of sites MAY be larger than a fundamental domain.
+    - :val:`"other"`: any other collection of sites that does not fit the enumerated values above.
+    - :val:`null`: if omitted or :val:`null`, the default value of :property:`site_coordinate_span` is :val:`unit_cell`. This is the assumed behavior of all main implementations before the :property:`site_coordinate_span` definition was introduced.
+
+- **Note**: In all cases it is RECOMMENDED that only the minimal set of the sites that is needed to reconstruct the whole material is provided. For example, for the 'unit_cell' span the server SHOULD NOT return sites that can be obtained from other returned sites through the translations given in :property:`lattice_vectors`; only a non-redundant set of sites SHOULD be provided.
```
vaitkus, line 3336: This statement seems to conflict slightly with the other parts of this definition. That is, it is stated here that the minimal set is RECOMMENDED, however, the description in line 3310 states that:
```
In particular, properties `cartesian_site_positions`_ and
`fractional_site_positions`_ MUST contain all sites *belonging*
to the described extent, and only those sites.
```
Similarly, the `fundamental_domain` definition includes as stricter note:
```
<...> those and only those sites <...>
```
If the idea is that this recommendation applies to all other definitions that do not have more restrictive requirement, I would make this clear:
`Unless specified otherwise, it is RECOMMENDED that only the minimal set...`

vaitkus (dismissed):
Looks good overall, although I would probably reword the one place I marked.

Also, there are a few layout issues, but I will create a separate PR for them, which can be merged either before or after merging this PR.

vaitkus (approved):

optimade.rst
```diff
@@ -3282,10 +3282,76 @@ cartesian\_site\_positions
 
   - :val:`[[0,0,0],[0,0,2]]` indicates a structure with two sites, one sitting at the origin and one along the (positive) *z*-axis, 2 Å away from the origin.
 
+fractional\_site\_positions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description**: fractional coordinates (positions) of each site in the structure.
+  A site is usually used to describe positions of atoms; what atoms can be encountered at a given site is conveyed by the :property:`species_at_sites` property, and the species themselves are described in the :property:`species` property.
+  Site coordinates MAY be given as `cartesian_site_positions`_, `fractional_site_positions`_, or both.
+  When symmetry operations given in `space_group_symmetry_operations_xyz`_ property are applied, they MUST be applied to coordinates given in the `fractional_site_positions`_ array.
```
```suggestion
  When symmetry operations given in the `space_group_symmetry_operations_xyz`_ property are applied, they MUST be applied to coordinates given in the `fractional_site_positions`_ array.
```

Or alternatively
```suggestion
  When symmetry operations given in `space_group_symmetry_operations_xyz`_ are applied, they MUST be applied to coordinates given in the `fractional_site_positions`_ array.
```

But also the MUST here is semantically a bit strange as it imposes a requirement on an action "MUST be applied", which isn't clear what it means in term of data or responses. So maybe also this:

```suggestion
  The symmetry operations given in `space_group_symmetry_operations_xyz`_ generates the equivalent sides when applied to the coordinates in the `fractional_site_positions`_ array.
```

optimade.rst
```diff
@@ -3282,10 +3282,76 @@ cartesian\_site\_positions
...
+- **Requirements/conventions**:
+
+  - **Support**: MUST be supported by all implementations if coordinates `fractional_site_positions`_ are returned.
+    It SHOULD be supported if coordinates in `cartesian_site_positions`_ are returned.
+  - **Query**: Support for queries on this property is OPTIONAL.
+
+  - The value of this property MUST be one of the following:
+
+    - :val:`"fundamental_domain"`: means that sites described in the response span a fundamental domain (Vinberg, 1994; European Mathematical Society, 2020) of a periodic system.
+      When a server indicates this span in the response, it MUST provide those sites that enable reconstruction of the whole periodic system by applying symmetry operations from `space_group_symmetry_operations_xyz`_ property and then applying translations given by `lattice_vectors`_.
```
rartino, line 3323: Here is a MUST that isn't a protocol-level-MUST and very difficult to think of in terms of validation. Could it be possible to instead rewrite it along the lines of this?

```suggestion
      A structure for which this span is set consists of (and only of) the periodic system given when applying the symmetry operations from `space_group_symmetry_operations_xyz`_ and then applying translations given by `lattice_vectors`_.
```

optimade.rst
```diff
@@ -3282,10 +3282,76 @@ cartesian\_site\_positions
...
+
+- **Examples**:
+
+  - :val:`[[0,0,0],[0,0,0.2]]` indicates a structure with two sites, one sitting at the origin and one along the third unit cell axis (*c*-axis), 1/5-th (0.2) of the vector *c* in the direction of the vector *c* from the origin.
+
+site\_coordinate\_span
+~~~~~~~~~~~~~~~~~~~~~~
+
+- **Description** Indicates the extent of the material (crystal) described in the response.
+  In particular, properties `cartesian_site_positions`_ and `fractional_site_positions`_ MUST contain all sites *belonging* to the described extent.
```
rartino, line 3312:
```suggestion
  In particular, the value of `site_coordinate_span` affects the precise meaning of `cartesian_site_positions`_ and `fractional_site_positions`_ as they are defined to contain all sites *belonging to the indicated extent.*
```

optimade.rst
sauliusg: The use of 'semantic "MUST"' is quite common in the current OPTIMADE specification (see issue https://github.com/Materials-Consortia/OPTIMADE/issues/545), so the usage here is similar to what was done up to now. If we decide to change this policy (and if there actually is a need to change), we should do the change systematically; this PR concerns different topic, and the two concerns should not be mixed in one PR or hold back the implementation of fractional coordinates.

optimade.rst
sauliusg: As in the conversation above:

>>The use of 'semantic "MUST"' is quite common in the current OPTIMADE specification (see issue https://github.com/Materials-Consortia/OPTIMADE/issues/545), so the usage here is similar to what was done up to now. If we decide to change this policy (and if there actually is a need to change), we should do the change systematically; this PR concerns different topic, and the two concerns should not be mixed in one PR or hold back the implementation of fractional coordinates.

PR #540, opened 2025-05-14
Change grammatica build source and target to Java 8

vaitkus:
Java 7 is no longer available on the more modern systems therefore the jar file needed to run the OPTIMADE grammar tests fails to build. Switching to Java 8 seems to resolve the issue even if it also results in multiple Java deprecation messages during the build.

### DIFF
--- a/tests/tools/grammatica-build.patch
+++ b/tests/tools/grammatica-build.patch
@@ -2,11 +2,11 @@
 <            source="1.4"
 <            target="1.5"
 ---
->            source="1.7"
->            target="1.7"
+>            source="1.8"
+>            target="1.8"
 116,117c116,117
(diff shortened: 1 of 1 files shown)


merkys (approved):
Looks good, although I have not tried this out myself.

Just FYI, Grammatica is available in Debian/Ubuntu package `libgrammatica-java` for some time now, thus I would suggest relying on that instead of building it from scratch.

vaitkus:
@merkys Thanks for the review. Personally, I am all for using packages from the standard repositories. However, I suggest that we merge this PR and then create a new one that deals with removing all  the Grammatica  download/build logic in case someone objects to that.

rartino (approved):

PR #542, opened 2025-06-05
Add a sentence to explain how identifier-only filters should work

### LINKED ISSUE #541: What does it mean to query for `filter=property`?
OPTIMADE filter language specification at v1.2.0 + 10d39fb4548c914f8cf42c96ebb2f5446454c6b0 allows queries containing only an identifier:
```
    Filter = [Spaces], Expression ;

    Expression = ExpressionClause, [ OR, Expression ] ;

    ExpressionClause = ExpressionPhrase, [ AND, ExpressionClause ] ;

    ExpressionPhrase = [ NOT ], ( Comparison | OpeningBrace, Expression, ClosingBrace ) ;

    Comparison = ConstantFirstComparison
               | PropertyFirstComparison ;
    (* Note: support for ConstantFirstComparison is OPTIONAL *)

    ConstantFirstComparison = ( OrderedConstant, ValueOpRhs
                              | UnorderedConstant, ValueEqRhs ) ;

    PropertyFirstComparison = Property, [ ValueOpRhs
                                        | KnownOpRhs
                                        | FuzzyStringOpRhs
                                        | SetOpRhs
                                        | SetZipOpRhs
                                        | LengthOpRhs ] ;
```
Here the last rule says that `Property` is a valid `PropertyFirstComparison`. However, I cannot find an explicit explanation how such comparisons should be handled at the backend. Most programming languages (C, Perl, Python) would interpret this as selection of non-falsey values (= exclusion of undefined, `false` boolean values, numbers equating to 0, empty strings etc). If OPTIMADE wants to follow the suit, I suggest writing this explicitly:

Filter `property` matches:
* if `property` is string: known AND nonempty string values
* if `property` is integer: known AND nonzero values
* if `property` is float: known AND nonzero values (can we compare float to 0?)
* if `property` is boolean: known AND true values
* if `property` is timestamp: known values
* if `property` is list: known AND nonempty lists
* if `property` is dictionary: known AND nonempty lists

Provider-specific data types will have to define such handling on their own.

merkys:
I noticed the behavior of filter `property` is undefined for non-boolean properties (see #541 for explanation). This PR fixes the issue by stating that for non-boolean properties such filter is equivalent to `property IS KNOWN`.

Fixes #541

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2035,6 +2035,8 @@ Filters that match when the property is known, or unknown, respectively can be c
     identifier IS KNOWN
     identifier IS UNKNOWN
 
+For properties having other than boolean values, :filter:`identifier IS KNOWN` MAY be written as :filter:`identifier`.
+
 Except for the above constructs, filters that use any form of comparison that involve properties of unknown values MUST NOT match.
 Hence, by definition, an :filter-fragment:`identifier` of value :filter-fragment:`null` never matches equality (:filter-op:`=`), inequality (:filter-op:`<`, :filter-op:`<=`, :filter-op:`>`, :filter-op:`>=`, :filter-op:`!=`) or other comparison operators besides :filter:`identifier IS UNKNOWN` and :filter:`NOT identifier IS KNOWN`.
 In particular, a filter that compares two properties that are both :val:`null` for equality or inequality does not match.


rartino (approved):

ml-evs (approved):
Thanks @merkys!

PR #546, opened 2025-07-09
Fix a few minor issues introduced by merging PR #539

vaitkus:


optimade.rst
```diff
@@ -3382,7 +3382,7 @@ species\_at\_sites
   - **Support**: SHOULD be supported by all implementations, i.e., SHOULD NOT be :val:`null`.
   - **Query**: Support for queries on this property is OPTIONAL.
     If supported, filters MAY support only a subset of comparison operators.
-  - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions`_ and/or `fractional_site_positions`_).
+  - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions`_ or `fractional_site_positions`_).
```
sauliusg: "or" is not what is meant, since it is ambiguous (inclusive vs exclusive or, and supposes that having just one list of the right length is enough, the other can have different length). Thus, the "and/or" was introduced deliberately and should stay. I agree that "and/or" is a laboratory jargon, but any more "linguistic" explanation would be longer and sometimes just obscures the main idea.

optimade.rst
```diff
@@ -3299,7 +3299,7 @@ fractional\_site\_positions
     If supported, filters MAY support only a subset of comparison operators.
   - It MUST be a list of length equal to the number of sites in the structure, where every element is a list of the three fractional coordinates of a site expressed as float values in the fractions of the unit cell vectors given by the `lattice_vectors`_ property.
   - An entry MAY have multiple sites at the same site position (for a relevant use of this, see e.g., the property `assemblies`_).
-  - **Note**: Since both `cartesian_site_positions`_ and the `fractional_site_positions`_ always describe the same sites, they MUST always have the same number of elements, equal to the number of elements in the `species_at_sites`_ array.
+  - **Note**: Since both the `cartesian_site_positions`_ property and the `fractional_site_positions`_ property always describe the same sites, they MUST always have the same number of elements, equal to the number of elements in the `species_at_sites`_ array.
```
sauliusg: I would be in favour of the previous shorter variant. Since names are quoted that are mentioned, not used, and thus grammatically can occupy position of a proper noun, don't they?

optimade.rst
```diff
@@ -3331,9 +3331,11 @@ site\_coordinate\_span
     - :val:`"molecular_entities"`: sets of atoms that are bound by covalent or coordination bonds, as per IUPAC definition of a 'molecular entity'.
       This set of sites MAY be larger than a fundamental domain.
     - :val:`"other"`: any other collection of sites that does not fit the enumerated values above.
-    - :val:`null`: if omitted or :val:`null`, the default value of :property:`site_coordinate_span` is :val:`unit_cell`. This is the assumed behavior of all main implementations before the :property:`site_coordinate_span` definition was introduced.
+    - :val:`null`: if omitted or :val:`null`, the default value of :property:`site_coordinate_span` is :val:`unit_cell`.
+      This is the assumed behavior of all main implementations before the :property:`site_coordinate_span` definition was introduced.
```
sauliusg, line 3342: Shouldn't this be one paragraph (i.e. no empty line in between)?

vaitkus: There is no empty line between these two sentences (lines 3334-3335). The sentences were refolded from a single line to two separate lines as is done throughout this document.

optimade.rst
vaitkus: The problem here was that one "the" was already present ("Since both `cartesian_site_positions` and the `fractional_site_positions` always <...>"). This is not grammatically correct.

Normally, you could omit the "the" altogether ("Both A and B") , however, in this case the quoted term is a plural, thus at the start of the sentence is ambiguous what "Both" refers to (i.e. at the start it could be read as "Both coordinates of Cartesian site positions" ...).

Do you have a different preferred rewording?

optimade.rst
vaitkus: My problem here is that the initial sentence of this definition (line 3377) uses only "or" in an almost identical sentence. So should both places be changed to "and/or"?

rartino:
@sauliusg I see that @vaitkus has answered all feedback. Could you take a look and indicate agreement or disagreement, so we can get the PR adjusted or merged?

optimade.rst
sauliusg: I think the text is OK as it is originally, but if the difference bothers anyone feel free to change to "and/or".

Just "or" is not enough IMHO since, in case the two lengths differ, this is an error and not a mandate to choose any of the two.

optimade.rst
sauliusg: Lets omit the second 'the'.

There is no ambiguity with the terms – once again, logically, the quoted term is singular and refers to the array as a whole, not to the distinct elements within that array.

In general – is there a risk that a potential implementer will not understand what is said or understand it incorrectly? If there is there is no such risk I would not change the text (here or anywhere else).

optimade.rst
vaitkus: Went with "and/or" in both places although I do find the phrasing more confusing.

optimade.rst
vaitkus: Reworded as suggested,

sauliusg (dismissed):

optimade.rst
```diff
@@ -3372,7 +3374,7 @@ nsites
 species\_at\_sites
 ~~~~~~~~~~~~~~~~~~
 
-- **Description**: Name of the species at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_ or `fractional_site_positions`_).
+- **Description**: Name of the species at each site (where values for sites are specified with the same order of the property `cartesian_site_positions`_ and/or `fractional_site_positions`_).
```
rartino, line 3384: How should one think about an edit of the description of a field that has already been released as a property definition, if the edit is only in the specification document and not in the corresponding property definition file? I think this requires editing the property definition file, and bumping its version number to 1.2.1.

optimade.rst
```diff
@@ -3299,7 +3299,7 @@ fractional\_site\_positions
     If supported, filters MAY support only a subset of comparison operators.
   - It MUST be a list of length equal to the number of sites in the structure, where every element is a list of the three fractional coordinates of a site expressed as float values in the fractions of the unit cell vectors given by the `lattice_vectors`_ property.
   - An entry MAY have multiple sites at the same site position (for a relevant use of this, see e.g., the property `assemblies`_).
-  - **Note**: Since both `cartesian_site_positions`_ and the `fractional_site_positions`_ always describe the same sites, they MUST always have the same number of elements, equal to the number of elements in the `species_at_sites`_ array.
+  - **Note**: Since both `cartesian_site_positions`_ and`fractional_site_positions`_ always describe the same sites, they MUST always have the same number of elements, equal to the number of elements in the `species_at_sites`_ array.
```
merkys:
```suggestion
  - **Note**: Since both `cartesian_site_positions`_ and `fractional_site_positions`_ always describe the same sites, they MUST always have the same number of elements, equal to the number of elements in the `species_at_sites`_ array.
```

rartino (approved):

Issue #550, opened 2025-09-17
Verify usage of the `sizes` field


ndaelman-hu:
There are spurious synonyms of the response field `sizes`, e.g. `lengths` throughout the documentation. Plan of action (in order):

1. Update the meta schema
2. Check that the wrong examples are captured.
3. Update the examples.
4. Update the schema (documentation).

rartino:
This is a duplicate of #549

PR #551, opened 2025-09-17
Fix wrong key in schemas (sizes vs lengths) and examples, harden metaschema

### LINKED ISSUE #549: Property definition bug
The [`lattice_vectors`](https://schemas.optimade.org/defs/v1.2/properties/optimade/structures/lattice_vectors) property definition uses `x-optimade-dimensions -> lengths`, but the specification and other properties uses `sizes`. This needs to be fixed, and a new patch version of that property definition released.

This also means we are missing an `additionalProperties: false` in `x-optimade-requirement` in the meta-schema source file `property_definition.yaml` that should have caught this in the meta-schema validation.

ndaelman-hu:
Closes #549 and #550

### DIFF
--- a/schemas/src/defs/v1.2/properties/optimade/structures/cartesian_site_positions.yaml
+++ b/schemas/src/defs/v1.2/properties/optimade/structures/cartesian_site_positions.yaml
@@ -5,7 +5,7 @@ x-optimade-type: "list"
 x-optimade-definition:
   label: "cartesian_site_positions_optimade_structures"
   kind: "property"
-  version: "1.2.0"
+  version: "1.2.1"
   format: "1.2"
   name: "cartesian_site_positions"
 x-optimade-unit-definitions:
--- a/schemas/src/defs/v1.2/properties/optimade/structures/lattice_vectors.yaml
+++ b/schemas/src/defs/v1.2/properties/optimade/structures/lattice_vectors.yaml
@@ -4,7 +4,7 @@ title: "lattice vectors"
 x-optimade-definition:
   label: "lattice_vectors_optimade_structures"
   kind: "property"
-  version: "1.2.0"
+  version: "1.2.1"
   format: "1.2"
   name: "lattice_vectors"
 x-optimade-type: "list"
--- a/schemas/src/defs/v1.2/properties/optimade/structures/lattice_vectors.yaml
+++ b/schemas/src/defs/v1.2/properties/optimade/structures/lattice_vectors.yaml
@@ -10,7 +10,7 @@ x-optimade-definition:
 x-optimade-type: "list"
 x-optimade-dimensions:
   names: ["dim_lattice", "dim_spatial"]
-  lengths: [3, 3]
+  sizes: [3, 3]
 x-optimade-unit-definitions:
   - $$inherit: "/v1.2/units/si/general/angstrom"
 type:
@@ -40,7 +40,7 @@ items:
   x-optimade-type: "list"
   x-optimade-dimensions:
     names: ["dim_spatial"]
-    lengths: [3]
+    sizes: [3]
   type:
     - "array"
   x-optimade-unit: "inapplicable"
--- a/schemas/src/meta/v1.2/optimade/property_definition.yaml
+++ b/schemas/src/meta/v1.2/optimade/property_definition.yaml
@@ -270,7 +270,7 @@ $defs:
           Specification of the dimensions of one or multi-dimensional data represented as multiple levels of lists.
           Each dimension is given a name and optionally a fixed size.
           It was decided that OPTIMADE Property Definitions cannot use the JSON Schema validating fields minItems and maxItems since that would require reprocessing the schema to handle requests asking for only parts of the data in the lists.
-          Instead, provide information on the length of a list for an implementation to validate. This is communicated in the `sizes` subfield of this field. At this time, `sizes` can only specify a fixed length requirement.)
+          Instead, provide information on the length of a list for an implementation to validate. This is communicated in the `sizes` subfield of this field. At this time, `sizes` can only specify a fixed length requirement.
         type: object
         additionalProperties: false
         patternProperties:
--- a/schemas/src/meta/v1.2/optimade/property_definition.yaml
+++ b/schemas/src/meta/v1.2/optimade/property_definition.yaml
@@ -267,6 +267,8 @@ $defs:
           It was decided that OPTIMADE Property Definitions cannot use the JSON Schema validating fields minItems and maxItems since that would require reprocessing the schema to handle requests asking for only parts of the data in the lists.
           Instead, list length information that an implementation can use to validate the length of lists is communicated in the `sizes` subfield of this field (however, at this time, it can only specify a fixed length requirement.)
         type: object
+        patternProperties:
+          "^x-(?!optimade-)": {}
         additionalProperties: false
         properties:
           names:
@@ -306,6 +308,7 @@ $defs:
       x-optimade-unit:
         type: string
       x-optimade-implementation:
+        additionalProperties: false
         patternProperties:
           "^x-(?!optimade-)": {}
         properties:
@@ -348,9 +351,9 @@ $defs:
         type: object
       x-optimade-requirements:
         additionalProperties: false
+        patternProperties:
+          "^x-(?!optimade-)": {}
         properties:
-          patternProperties:
-            "^x-(?!optimade-)": {}
           query-support:
             enum:
             - all mandatory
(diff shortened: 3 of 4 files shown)


ndaelman-hu:
At commit 01fb88b23f7a1637d6b08e053a6db0367203f408 running the grammar tests with the restriction (but without updating the examples) triggers the error message:

```
Validation of data failed. Error details:
- Schema validation failed:
  - Instance id: https://schemas.optimade.org/defs/v1.2/properties/optimade/structures/lattice_vectors
  - Schema: https://schemas.optimade.org/meta/v1.2/optimade/property_definition.json
  - Instance path: items/x-optimade-dimensions
  - Schema path: allOf/0/properties/items/allOf/0/properties/x-optimade-dimensions/additionalProperties
  - Error: Additional properties are not allowed ('lengths' was unexpected)
```

schemas/src/meta/v1.2/optimade/property_definition.yaml
```diff
@@ -265,10 +265,9 @@ $defs:
           Specification of the dimensions of one or multi-dimensional data represented as multiple levels of lists.
           Each dimension is given a name and optionally a fixed size.
           It was decided that OPTIMADE Property Definitions cannot use the JSON Schema validating fields minItems and maxItems since that would require reprocessing the schema to handle requests asking for only parts of the data in the lists.
-          Instead, list length information that an implementation can use to validate the length of lists is communicated in the `length` subfiled of this field (however, at this time, it can only specify a fixed length requirement.)
+          Instead, list length information that an implementation can use to validate the length of lists is communicated in the `sizes` subfield of this field (however, at this time, it can only specify a fixed length requirement.)
         type: object
```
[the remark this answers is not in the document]
ndaelman-hu: This was the suggestion to find any breaking changes

schemas/src/meta/v1.2/optimade/property_definition.yaml
```diff
@@ -265,10 +265,9 @@ $defs:
...
-        patternProperties:
-          "^x-(?!optimade-)": {}
+        additionalProperties: false
```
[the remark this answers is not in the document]
ndaelman-hu: @giovannipizzi what's your opinion on this alternative? Did you suggest `additionalProperties: false` just for testing, or to persist afterwards?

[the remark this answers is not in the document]
giovannipizzi: No, while we want to allow properties starting with `x-`, we still want to prevent properties starting with `x-optimade-` as these are reserved for us

giovannipizzi:
Thanks @ndaelman-hu - I pushed a commit fixing the issue that I had mentioned (the patternProperties needs to stay, so that all `x-*` properties are allowed, except for the `x-optimade-*` properties, and fixed a couple things more. I think it's ready to go, but I'd ask @rartino for a review as well

schemas/src/defs/v1.2/properties/optimade/structures/lattice_vectors.yaml
```diff
@@ -10,7 +10,7 @@ x-optimade-definition:
 x-optimade-type: "list"
```
rartino, line 10: I cannot do an edit request because how GitHub works with lines outside those the commit affects, but this PR needs to update the 'version' field in `x-optimade-definition` to `1.2.1`.

schemas/src/meta/v1.2/optimade/property_definition.yaml
```diff
@@ -265,10 +265,11 @@ $defs:
           Specification of the dimensions of one or multi-dimensional data represented as multiple levels of lists.
           Each dimension is given a name and optionally a fixed size.
           It was decided that OPTIMADE Property Definitions cannot use the JSON Schema validating fields minItems and maxItems since that would require reprocessing the schema to handle requests asking for only parts of the data in the lists.
-          Instead, list length information that an implementation can use to validate the length of lists is communicated in the `length` subfiled of this field (however, at this time, it can only specify a fixed length requirement.)
+          Instead, list length information that an implementation can use to validate the length of lists is communicated in the `sizes` subfield of this field (however, at this time, it can only specify a fixed length requirement.)
         type: object
         patternProperties:
           "^x-(?!optimade-)": {}
+        additionalProperties: false
```
rartino: I realize we don't have a "version" in the meta-schemas, and now that I see this update, I realize we probably really should, and perhaps under an `x-optimade-definition` stanza of `kind: meta`. One could argue it does not belong in this PR, but, since this is the first time we actually change a meta-schema, and in this case it is a *bugfix* and not an extension, IMO it does belong in this PR to indicate that. I.e., also addressing the bug that there is no place to indicate that we've patched the meta-schema.

rartino (changes requested):
A few adjustments IMO needed.

schemas/src/meta/v1.2/optimade/property_definition.yaml
rartino: Something like this:
```
x-optimade-definition:
  label: "property_definition_meta"
  kind: "meta"
  version: "1.2.1"
  format: "1.2"
  name: "property_definition"
```

ndaelman-hu: I applied your suggestion (which is understand to be the same as the comment above).
Pls let me know whether you agree, and if the minor edit to the description is fine.

schemas/src/defs/v1.2/properties/optimade/structures/lattice_vectors.yaml
rartino: You did the edit in `property_definition.yaml`. But this edit suggestion is in reference to `lattice_vectors.yaml` where you need to bump the previous  version number.

ndaelman-hu: I see, this is parallel to the comment below.

schemas/src/meta/v1.2/optimade/property_definition.yaml
ndaelman-hu: @rartino May this be resolved?

schemas/src/defs/v1.2/properties/optimade/structures/lattice_vectors.yaml
ndaelman-hu: Done. Given that this PR also modifies `schemas/src/defs/v1.2/properties/optimade/structures/cartesian_site_positions.yaml`, I bumped the version there too.

rartino (dismissed):
Thanks for the work on this. Looks good!

schemas/src/meta/v1.2/optimade/property_definition.yaml
```diff
@@ -260,13 +260,19 @@ $defs:
...
+        version: "1.2.1"
+        format: "1.2"
+        name: "property_definition"
       x-optimade-dimensions:
         description: |-
           Specification of the dimensions of one or multi-dimensional data represented as multiple levels of lists.
           Each dimension is given a name and optionally a fixed size.
           It was decided that OPTIMADE Property Definitions cannot use the JSON Schema validating fields minItems and maxItems since that would require reprocessing the schema to handle requests asking for only parts of the data in the lists.
-          Instead, list length information that an implementation can use to validate the length of lists is communicated in the `length` subfiled of this field (however, at this time, it can only specify a fixed length requirement.)
+          Instead, provide information on the length of a list for an implementation to validate. This is communicated in the `sizes` subfield of this field. At this time, `sizes` can only specify a fixed length requirement.)
```
giovannipizzi:
```suggestion
          Instead, provide information on the length of a list for an implementation to validate. This is communicated in the `sizes` subfield of this field. At this time, `sizes` can only specify a fixed length requirement.
```

giovannipizzi (approved):

giovannipizzi:
I left a comment for an unopened bracket, but otherwise approved

rartino (approved):

schemas/src/defs/v1.2/properties/optimade/structures/cartesian_site_positions.yaml
```diff
@@ -5,7 +5,7 @@ x-optimade-type: "list"
 x-optimade-definition:
   label: "cartesian_site_positions_optimade_structures"
   kind: "property"
-  version: "1.2.0"
+  version: "1.2.1"
```
merkys, line 8: Why definition versions have to be bumped?

merkys: @giovannipizzi explained this to me in person, LGTM.

merkys (approved):

PR #560, opened 2025-09-18
Allow provider-specific data types in property definitions

merkys:
After merging in #529 there is now a need to reflect the introduction of provider-specific data types in the property definitions. This PR does so by changing the definition of `x-optimade-type` and the related attributes.

Funnily, JSON Schema does not allow having both an enum and a pattern, thus I had to replace enum with a pattern. Not very elegant, but still readable.

### DIFF
--- a/schemas/src/defs/v1.3/entrytypes/optimade/structures.yaml
+++ b/schemas/src/defs/v1.3/entrytypes/optimade/structures.yaml
@@ -219,3 +219,11 @@ properties:
       sortable: false
       response-level: "may"
       query-support: "all mandatory"
+
+  optimization_type:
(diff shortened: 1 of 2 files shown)


rartino (approved):
👍

ml-evs (approved):

merkys:
This PR seems to have been done prior to merging in #551, thus fixes need to be propagated manually - please refrain for merging for the time being!

merkys:
I would be really happy if this PR could be re-approved. This change is needed to make the properties with provider-specific data types to pass the validation.

rartino (approved):
👍👍

PR #570, opened 2025-09-24
Introduce `wyckoff_positions` property

merkys:
This PR introduces Wyckoff positions (a.k.a. letters) for each site in a structure. Wyckoff positions were mentioned a couple times in 2025 OPTIMADE workshop; in the COD 1.5% of entries have them; a research project of mine would benefit from providers exposing known Wyckoff positions through OPTIMADE.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -3944,7 +3944,11 @@ wyckoff\_positions
   - **Query**: Support for queries on this property is OPTIONAL.
     If supported, filters MAY support only a subset of comparison operators.
   - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions`_ and/or `fractional_site_positions`_).
-  - If provided, MUST list a single letter (`a`-`z` or `α`) Wyckoff position for each site in the structure.
+  - If provided, MUST list a single letter (`a`-`z` or `α`) Wyckoff position for each site in the structure according to the International Tables for Crystallography vol. A (IUCr, 2016).
+
+- **Bibliographic References**:
+
+  IUCr (2016). International Tables for Crystallography vol. A. Space-group Symmetry, Ed. M. I. Aroyo, 6-th edition. Chichester, John Wiley & Sons.
 
 structure\_features
 ~~~~~~~~~~~~~~~~~~~
--- a/schemas/src/defs/v1.3/properties/optimade/structures/wyckoff_positions.yaml
+++ b/schemas/src/defs/v1.3/properties/optimade/structures/wyckoff_positions.yaml
@@ -19,7 +19,11 @@ description: |-
   **Requirements/Conventions**:
 
   - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions` and/or `fractional_site_positions`).
-  - If provided, MUST list a single letter (`a`-`z` or `α`) Wyckoff position for each site in the structure.
+  - If provided, MUST list a single letter (`a`-`z` or `α`) Wyckoff position for each site in the structure according to the International Tables for Crystallography vol. A (IUCr, 2016).
+
+  **Bibliographic References**:
+
+  - IUCr (2016). International Tables for Crystallography vol. A. Space-group Symmetry, Ed. M. I. Aroyo, 6-th edition. Chichester, John Wiley & Sons.
 x-optimade-unit: "inapplicable"
 items:
   $$inherit: "/v1.3/properties/optimade/common/wyckoff_position"


optimade.rst
```diff
@@ -3933,6 +3933,19 @@ assemblies
...
+
+- **Description**: A list of Wyckoff symbols of sites (where values for sites are specified with the same order of the property `cartesian_site_positions`_ and/or `fractional_site_positions`_).
+- **Type**: list of strings
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions`_ and/or `fractional_site_positions`_).
+  - If provided, MUST list a single letter (`a`-`z`) Wyckoff position for each site in the structure.
```
vaitkus: The IUCr coreCIF dictionary also allows α (\a) as one of the values, however, I am unsure how it should be interpreted.

vaitkus: On second though, they probably ran out of Latin letters and started using Greek ones.

sauliusg: I've never seen a Wyckoff position 'α'... There does not seem to be so may different W.p.'s, e.g. for s/group 230 that last letter is 'h'.

vaitkus: Group 47 (P m m m) uses α for some reason (ITC Volume A, 2006, page 263).

merkys: Then we should follow suite and add `α`. Will do so.

sauliusg: You are right, they ran out of Latin letters :) – the last special position W.l. is 'z', and the general position is 'α' :)

sauliusg: So we need to include 'α'. Not a big deal, JSON supports Unicode :)

rartino: It is not uncommon to instead use "A". Just to give the first google result supporting this: https://www.sciencedirect.com/science/article/pii/S2095927325009752

ndaelman-hu: `spglib` also falls back to uppercase "A". You can check this for yourself by running the little script attached. It should yield:

```python
Space group: Pmmm
Wyckoff letters: ['A', 'A', 'A', 'A', 'A', 'A', 'A', 'A']
```

[wyckoff_27.py](https://github.com/user-attachments/files/23334747/wyckoff_27.py)

optimade.rst
```diff
@@ -3933,6 +3933,23 @@ assemblies
...
+
+- **Description**: A list of Wyckoff symbols of sites (where values for sites are specified with the same order of the property `cartesian_site_positions`_ and/or `fractional_site_positions`_).
+- **Type**: list of strings
+- **Requirements/Conventions**:
+
+  - **Support**: OPTIONAL support in implementations, i.e., MAY be :val:`null`.
+  - **Query**: Support for queries on this property is OPTIONAL.
+    If supported, filters MAY support only a subset of comparison operators.
+  - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions`_ and/or `fractional_site_positions`_).
+  - If provided, MUST list a single letter (`a`-`z` or `α`) Wyckoff position for each site in the structure according to the International Tables for Crystallography vol. A (IUCr, 2016).
```
ndaelman-hu, line 3947: Given the discussion and considering the support of popular symmetry packages like spglib, maybe we should swap alpha for A?

```suggestion
  - If provided, MUST list a single letter (`a`-`z` or `A`) Wyckoff position for each site in the structure according to the International Tables for Crystallography vol. A (IUCr, 2016).
```

schemas/src/defs/v1.3/entrytypes/optimade/structures.yaml
```diff
@@ -212,6 +212,14 @@ properties:
       response-level: "may"
       query-support: "none"
 
+  wyckoff_positions:
+    $$inherit: "/v1.3/properties/optimade/structures/wyckoff_positions"
+    x-optimade-requirements:
+      support: "may"
+      sortable: false
+      response-level: "may"
+      query-support: "none"
```
ndaelman-hu, line 221: In NOMAD, I've actually been looking into supporting Wyckoff queries. Makes sense for people looking to break symmetry, e.g. generate a vacancy.

schemas/src/defs/v1.3/properties/optimade/common/wyckoff_position.yaml
```diff
@@ -0,0 +1,17 @@
...
+  version: "1.3.0"
+  format: "1.2"
+  name: "wyckoff_position"
+description: |-
+  The Wyckoff symbol for a site.
+x-optimade-unit: inapplicable
+type:
+  - "string"
+maxLength: 1
+enum: ["a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z", "α"]
```
ndaelman-hu, line 17: same alpha comment

schemas/src/defs/v1.3/properties/optimade/structures/wyckoff_positions.yaml
```diff
@@ -0,0 +1,29 @@
...
+type:
+  - "array"
+  - "null"
+description: |-
+  A list of Wyckoff symbols of sites (where values for sites are specified with the same order of the property `cartesian_site_positions` and/or `fractional_site_positions`).
+
+  **Requirements/Conventions**:
+
+  - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions` and/or `fractional_site_positions`).
+  - If provided, MUST list a single letter (`a`-`z` or `α`) Wyckoff position for each site in the structure according to the International Tables for Crystallography vol. A (IUCr, 2016).
```
ndaelman-hu, line 22: same alpha comment

schemas/src/defs/v1.3/properties/optimade/structures/wyckoff_positions.yaml
```diff
@@ -0,0 +1,29 @@
...
+  names: ["dim_sites"]
+type:
+  - "array"
+  - "null"
+description: |-
+  A list of Wyckoff symbols of sites (where values for sites are specified with the same order of the property `cartesian_site_positions` and/or `fractional_site_positions`).
+
+  **Requirements/Conventions**:
+
+  - MUST have length equal to the number of sites in the structure (first dimension of the list properties `cartesian_site_positions` and/or `fractional_site_positions`).
```
ndaelman-hu, line 21: side note: I would develop a schema field that allows for specifying this requirement in a more formal way

optimade.rst
vaitkus: I think that we more or less came to the conclusion to place the conversion load on the server rather than the client. That is, there might be multiple different conventions adopted by different software packages to represent this position therefore the best choice might be to use the convention given by the IUCr (`α`) and ask OPTIMADE servers to interconvert between the different notations (e.g. `A`).

ndaelman-hu: I was just going off a comment from @rartino during the release meeting. It's sensible, as most servers may simply fall back on `spglib` for their symmetry analysis.

schemas/src/defs/v1.3/entrytypes/optimade/structures.yaml
ndaelman-hu: From the release meeting: using unicode characters like `α` might generate issues with the regex.

rartino:
@merkys Something I didn't realize until now as I didn't carefully review this is that part of the Wyckoff property definition now ended up in a subdirectory to `.../properties/optimade` named `common`. This is the first time we do that, so it makes sense to pause a moment and ask how we want to do this going forward; and I'm not entirely sure if a subdirectory under optimade is the best idea (among other things I'm a bit worried how this looks when browsing the property defs), or if it would make sense to place it one level up side-by-side with `core` and `optimade` under some sensible name. I do understand the idea that it is more "optimade" than it is "core". Was this your thinking with this structure?

PR #574, opened 2025-11-04
Add calculations entries

rartino:
Apparently (this is on me) we missed adding property definitions for the calculations endpoint entirely. This PR fixes that bug for the v1.2 definitions, as well as add a v1.3 definition with the change we recently merged that clarifies how one is supposed to use relationships to describe inputs to and outputs from calculations.


schemas/src/defs/v1.2/entrytypes/optimade/calculations.yaml
```diff
@@ -0,0 +1,43 @@
+$$schema: "https://schemas.optimade.org/meta/v1.2/optimade/entrytype_definition"
+$id: "https://schemas.optimade.org/defs/v1.2/entrytypes/optimade/calculations"
+title: "calculations"
+description: "The calculations entry type describe calculations."
```
rartino:
```suggestion
description: "The calculations entry type describes calculations."
```

schemas/src/defs/v1.3/entrytypes/optimade/calculations.yaml
```diff
@@ -0,0 +1,44 @@
+$$schema: "https://schemas.optimade.org/meta/v1.2/optimade/entrytype_definition"
+$id: "https://schemas.optimade.org/defs/v1.3/entrytypes/optimade/calculations"
+title: "calculations"
+description: |-
+  The calculations entry type describe calculations. Related `files` entries specified as relationships list files used or produced by a calculation. They MAY be defined as either input or output files by the `role` field inside the `meta` dictionary of the JSON:API resource identifier object. The allowed values are `input` for input files and `output` for output files. If the `role` field is not provided or the value is set to `null`, a default role SHOULD NOT be assumed. It is RECOMMENDED to order the output files in the list of calculation's relationships in a way that prioritizes the files most likely to contain details which could be used to automatically analyze the performed calculation. Such an approach minimizes the number of calculation output files that need to be analyzed to possibly identify the code, type and parameters of the performed calculation.
```
rartino:
```suggestion
  The calculations entry type describes calculations. Related `files` entries specified as relationships list files used or produced by a calculation. They MAY be defined as either input or output files by the `role` field inside the `meta` dictionary of the JSON:API resource identifier object. The allowed values are `input` for input files and `output` for output files. If the `role` field is not provided or the value is set to `null`, a default role SHOULD NOT be assumed. It is RECOMMENDED to order the output files in the list of calculation's relationships in a way that prioritizes the files most likely to contain details which could be used to automatically analyze the performed calculation. Such an approach minimizes the number of calculation output files that need to be analyzed to possibly identify the code, type and parameters of the performed calculation.
```

PR #577, opened 2025-11-07
Add extra schema checker jobs to CI

### LINKED ISSUE #575: Validate machine-readable property definitions in CI of this repo
This should be done before the final v1.3 release, but not necessarily before the v1.3.0-rc.1 release candidate.

ml-evs:
Closes #575

I've also tweaked some of the rst roles in the spec to maybe allow us to scrape the spec for all fields (in a2cb422). It might be worth adding some special role to the headers that define fields to make this even easier, otherwise I'll concoct some grep incantation to try and do the same job without needing any changes to the spec.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -69,6 +69,8 @@ OPTIMADE API specification v1.3.0-rc.1
 
 .. role:: type(literal)
 
+.. role:: property-example(literal)
+
 .. role:: property-fail(literal)
 
 .. role:: endpoint(literal)
@@ -1321,7 +1323,7 @@ The API implementation MAY provide other entry types than the ones standardized 
 Such entry types MUST be prefixed by a database-provider-specific prefix (i.e., the resource objects' :property:`type` value should start with the database-provider-specific prefix, e.g., :property:`type` = :val:`_exmpl_workflows`).
 Each custom entry type SHOULD be served at a corresponding entry listing endpoint under the versioned or unversioned base URL that serves the API with the same name (i.e., equal to the resource objects' :property:`type` value, e.g., :endpoint:`/_exmpl_workflows`).
 It is RECOMMENDED to align with the OPTIMADE API specification practice of using a plural for entry resource types and entry type endpoints.
-Any custom entry listing endpoint MUST also be added to the :property:`available_endpoints` and :property:`entry_types_by_format` attributes of the `Base Info Endpoint`_.
+Any custom entry listing endpoint MUST also be added to the :field:`available_endpoints` and :field:`entry_types_by_format` attributes of the `Base Info Endpoint`_.
 
 For more on custom endpoints, see `Custom Extension Endpoints`_.
 
@@ -1870,9 +1872,9 @@ The links endpoint MUST be provided under the versioned or unversioned base URL 
 Link Types
 ~~~~~~~~~~
 
-Each link has a :property:`link_type` attribute that specifies the type of the linked relation.
+Each link has a :field:`link_type` attribute that specifies the type of the linked relation.
 
-The :property:`link_type` MUST be one of the following values:
+The :field:`link_type` MUST be one of the following values:
 
 - :field-val:`child`: a link to another OPTIMADE implementation that MUST be within the same provider.
   This allows the creation of a tree-like structure of databases by pointing to children sub-databases.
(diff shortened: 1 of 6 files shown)


tests/scripts/check_property_definitions.sh
```diff
@@ -0,0 +1,167 @@
...
+warnings=0
+
+echo "Checking property definitions against specification..."
+echo ""
+
+# Function to check if a property definition file exists
+check_property() {
+    local entry_type=$1
+    local property=$2
+    local version=${3:-v1.2}  # Default to v1.2
```
merkys, line 25: Instead of hard-coding, maybe we can extract version numbers (here and below for `v13_path`) from `optimade.rst` or `CHANGELOG.md`? Otherwise we may forget bumping them.

tests/scripts/check_property_definitions.sh
```diff
@@ -0,0 +1,167 @@
...
+
+    # Check in the main entry type directory first
+    local primary_path="schemas/src/defs/$version/properties/optimade/$entry_type/$property.yaml"
```
merkys, line 28: I think the proper way to check for the existence of a property is:
1. Take the entry types and their URLs from `schemas/src/defs/$version/standards/optimade.yaml`
2. Consult the entry type URLs (`schemas/src/defs/$DEFINITION_VERSION/entrytypes/optimade/$ENTRY_TYPE.yaml`) for properties

Important point here is that `$version` and `$DEFINITION_VERSION` might not be the same (see `schemas/src/defs/v1.3/standards/optimade.yaml` for example). I have recently implemented this for COD, maybe I could backport the code here. But maybe we want to do something really simple like `find schemas/src/defs/*/$property.yaml` instead?

tests/scripts/check_property_definitions.sh
```diff
@@ -0,0 +1,167 @@
...
+    if [[ "$entry_type" == "_common" ]]; then
+        properties=$(echo "$spec_fields" | jq -r '.["_common"][]')
+        for property in $properties; do
+            # Common properties should exist in core/ or be shared
+            if [[ -f "schemas/src/defs/v1.2/properties/core/$property.yaml" ]]; then
+                echo -e "  ${GREEN}✓${NC} $property (core)"
+            else
+                # Check if it exists in any entry type
+                found=false
+                for et in structures files references calculations; do
```
merkys: This again is hard-coded, will need updating for every new entry type.

optimade.rst
```diff
@@ -1321,7 +1323,7 @@ The API implementation MAY provide other entry types than the ones standardized
 Such entry types MUST be prefixed by a database-provider-specific prefix (i.e., the resource objects' :property:`type` value should start with the database-provider-specific prefix, e.g., :property:`type` = :val:`_exmpl_workflows`).
 Each custom entry type SHOULD be served at a corresponding entry listing endpoint under the versioned or unversioned base URL that serves the API with the same name (i.e., equal to the resource objects' :property:`type` value, e.g., :endpoint:`/_exmpl_workflows`).
 It is RECOMMENDED to align with the OPTIMADE API specification practice of using a plural for entry resource types and entry type endpoints.
-Any custom entry listing endpoint MUST also be added to the :property:`available_endpoints` and :property:`entry_types_by_format` attributes of the `Base Info Endpoint`_.
```
merkys, line 1324: I assume `:property:` -> `:field:` change is made to leave only per-entry properties marked with `:property:` in that section, right?

merkys (commented):
Thanks for developing this @ml-evs! This PR adds the much-needed automatic validation of the specification. However, I believe, there are certain cases when it might run into false-positives and false-negatives. Moreover, the code is quite tightly coupled with v1.2 and v1.3, thus it would need updating for each subsequent versions.

I think the `optimade-property-tools` might be reused for such tasks as entry type/property lookup inside the property definitions. If not, I have recently written some code for the COD implementation and could reuse it to expose simple executables for such tasks.

In any case, I am OK with merging the PR as it is, as it would help us in developing the v1.3.0 specification and schemas. However, I would keep a mental note to revisit the code later.

optimade.rst
ml-evs: That's the plan -- I think this matches the rst roles as described at the top of the spec too

tests/scripts/check_property_definitions.sh
ml-evs: Indeed, but its chicken-and-egg whether we add it here or wait for a folder name to be made with the relevant schemas. At least we could move it to the top of the file so its clear.

ml-evs: I've looked again and this is unnecessarily hardcoded, as it's already extracted from the spec further up. Might be a bit trickier to extract versions but I'll fix this at least.

rartino:
Workshop discussion: question if we want to do this since it adds a bit of extra work on specification writing, but the positive is that more formalism making sure things are correct are good.

PR #595, opened 2026-07-02
Add `runs` entry type and provenance relationships

giovannipizzi:
In this commit we introduce the following additions to the OPTIMADE specification:

- Entry Type Categories: classifies every   entry type as `data`,
  `execution`, or `metadata`. Standard types:
  `runs`/`calculations` → execution; `references` → metadata;
  `structures`/`trajectories`/`files` → data. An OPTIONAL `category`
  field is added to the Entry Listing Info Endpoint response.

- Reserved provenance relationship keys (in Entry Listing JSON Response Schema): `inputs`, `outputs`, `output_of`, `input_for`, `source_of`, `derived_from` are reserved and exempt from the grouping-by-entry-type rule. Category constraints, cardinality, and acyclicity rules are defined.

- Provenance Relation Objects (§ Provenance Relation Objects): the six provenance relationships resolve to reified edge objects (`input_rel`, `output_rel`, `derived_rel`) carrying a mandatory `label` and an optional `workflow_uri` (on `derived_rel`). The `target` key is reserved. Relationships can be filtered by `label`/`workflow_uri` via a standard OPTIMADE filters; fetching, pagination, and
  `include=target` conventions are   specified.

- Runs Entries: new `execution`-category entry type representing a specific workflow execution (could also be experimental). Properties: `workflow_uri` (REQUIRED, nullable), `inputs`/`outputs` relationships.

- Provenance relationships on data entries: `output_of` (to-one, exclusive creator rule), `input_for` (unbounded to-many), `source_of` (unbounded to-many, per-source label uniqueness), `derived_from` (unbounded to-many, label non-uniqueness on incoming side, optional `workflow_uri` on edge).

- Calculations classified as `execution` category (potentially to be deprecated?)

- Worked examples (in Sec. Provenance Examples) illustrating all new relationships.

Note: all these changes are backward compatible.

Please follow [the guidelines here](https://github.com/Materials-Consortia/OPTIMADE/wiki/How-to-contribute#submit-proposed-changes) to create a new pull request.
If your pull request requires discussion, make sure the OPTIMADE developer list is alerted (for more info see the link above.)

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -1474,46 +1474,54 @@ Example:
        // ...
      }
 
-The following additional provisions apply to the six provenance relationship keys :field:`inputs`, :field:`outputs`, :field:`output_of`, :field:`input_for`, :field:`source_of`, and :field:`derived_from` (see section `Relationships Used by Multiple Entry Types`_ for their definitions).
+The following additional provisions apply to the eight provenance relationship keys :field:`inputs`, :field:`input_for`, :field:`creates`, :field:`created_by`, :field:`returns`, :field:`returned_by`, :field:`source_of`, and :field:`derived_from` (see section `Relationships Used by Multiple Entry Types`_ for their definitions).
 
 **Reserved names.**
 The following names are RESERVED; an API implementation MUST NOT define an entry type with any of them:
-the six relationship keys :field:`inputs`, :field:`outputs`, :field:`output_of`, :field:`input_for`, :field:`source_of`, :field:`derived_from`;
-the three relation-object types :val:`input_rel`, :val:`output_rel`, :val:`derived_rel` (see section `Provenance Relation Objects`_);
+the eight relationship keys :field:`inputs`, :field:`input_for`, :field:`creates`, :field:`created_by`, :field:`returns`, :field:`returned_by`, :field:`source_of`, :field:`derived_from`;
+the four relation-object types :val:`input_relationships`, :val:`create_relationships`, :val:`return_relationships`, :val:`derived_relationships` (see section `Provenance Relation Objects`_);
 and the relationship key :field:`target` (see section `Provenance Relation Objects`_), used on relation objects.
 
 **Exemption from the grouping rule.**
-Where present, relationship objects under the six provenance keys are exempt from the grouping-by-entry-type rule above: the key need not equal the entry type name of any of their members.
-The members of these relationship objects are provenance relation objects (see section `Provenance Relation Objects`_) addressed by a semantic, direction-bearing key (e.g., :field:`output_of` vs. :field:`outputs`) rather than by target entry type.
+Where present, relationship objects under the eight provenance keys are exempt from the grouping-by-entry-type rule above: the key need not equal the entry type name of any of their members.
+The members of these relationship objects are provenance relation objects (see section `Provenance Relation Objects`_) addressed by a semantic, direction-bearing key (e.g., :field:`created_by` vs. :field:`creates`) rather than by target entry type.
 
 **Category constraints** (see section `Entry Type Categories`_):
 
-- :field:`inputs` and :field:`outputs` may be defined only on an ``execution`` entry; every referenced target MUST be a ``data`` entry.
-- :field:`output_of` and :field:`input_for` may be defined only on a ``data`` entry; every referenced target MUST be an ``execution`` entry.
+- :field:`inputs`, :field:`creates`, and :field:`returns` may be defined only on an ``execution`` entry; every referenced target MUST be a ``data`` entry.
+- :field:`input_for`, :field:`created_by`, and :field:`returned_by` may be defined only on a ``data`` entry; every referenced target MUST be an ``execution`` entry.
 - :field:`source_of` and :field:`derived_from` may be defined only on a ``data`` entry; every referenced target MUST be a ``data`` entry.
-- No ``metadata`` entry may define, or be the target of, any of the six provenance relationships.
+- No ``metadata`` entry may define, or be the target of, any of the eight provenance relationships.
 
 **Cardinality.**
-:field:`output_of` is to-one: at most one creating execution (see section `output\_of`_).
-:field:`inputs`, :field:`outputs`, :field:`input_for`, :field:`source_of`, and :field:`derived_from` are unbounded to-many relationships.
+:field:`created_by` is to-one: at most one creating execution (see section `created\_by`_).
+:field:`inputs`, :field:`input_for`, :field:`creates`, :field:`returns`, :field:`returned_by`, :field:`source_of`, and :field:`derived_from` are unbounded to-many relationships.
+
+**Graph structure.**
+The provenance graph has ``data`` and ``execution`` entries as its nodes and the provenance relation objects as its directed edges, each edge being directed from the entry declaring the outgoing relationship key to the entry named by the edge's :field:`target`.
+Every edge has one of the four types defined in section `Provenance Relation Objects`_.
+For a given ordered pair of nodes (a first and a second node), there MUST NOT be two edges of the same type both directed from the first to the second node.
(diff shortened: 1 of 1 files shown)


giovannipizzi:
## Open issues for discussion

The following points were flagged in the draft and might require explicit approval or discussion:

1. **`workflow_uri` nullability** (currently: REQUIRED-but-nullable): allows for providers to avoid creating an artificial URI just to respect the specs. -> DECISION: OK

2. **Filter/sort support on relation objects beyond `label` and `workflow_uri`**: the current text leaves broader query support unspecified (neither required nor prohibited). Should a minimum set of filter features be mandated for relation-object endpoints? -> DECISION: not decided, keep as is for now.

3. **`available_endpoints` and `entry_types_by_format`**: no wording change to Sec. 5.3/5.5.1 is needed (both already require implemented entry-listing endpoints to be reflected in the Base Info Endpoint). Implementers should add `"runs"` to `available_endpoints` and `entry_types_by_format.json`, and expose `/info/runs`. The three new relation types (`input_rel`/`output_rel`/`derived_rel`) are deliberately **not** added to `available_endpoints`. -> DECISION: OK

4. **`calculations` classified as `execution` — future deprecation**: the role of `calculations` is now unclear given the introduction of `runs`. Options: (a) deprecate in favour of `runs`; (b) give it non-overlapping semantics (e.g. its own `inputs`/`outputs`). This PR classifies it as `execution` and removes the old "future work" note. We should decide if we want to deprecate or remove in this PR directly. -> DECISION: deprecate

5. **`output_of` single-and-exclusive-creator rule**: a consequence is that two runs genuinely producing a byte-identical artifact MUST model them as two distinct data entries. The `source_of`/`derived_from` mechanism (unbounded) is the designated way to instead represent "returning" rather than "creating". Is this OK? -> DECISION: OK as is, but we will add a RETURN/CREATE separation of the output links (with different semantics).

6. **Promoting relation types to top-level entry types**: `input_rel`/`output_rel`/`derived_rel` are currently embedded-only (no `/v1/input_rel` endpoint). A future version MAY promote them to top-level entry points (in a backward-compatible way) if use cases for global cross-node queries emerge. For now, this does not seem needed. -> DECISION: Do NOT promote for now.

7. **Acyclicity is required but server-unenforceable**: the spec requires acyclicity but cannot mandate server-side verification. Clients MUST guard against cycles when traversing (this is written now as a note in the specs). -> DECISION: use SHOULD NOT for the requirement on the server (since it's not enforceable), but keep the discussion, and keep MUST NOT for the client recommendation.

8. **`meta` fields on the to-one `output_of` `related` response**: the `output_of` related endpoint returns a single relation object rather than a list. This PR treats it like a single-entry fetch (omits `data_returned`/`data_available`, shows only optional `more_data_available`). To be confirmed if this matches how the OPTIMADE spec usually populates `meta` fields on single-resource responses. -> DECISION: OK as is.

rartino:
I have started to work through this in tandem with a mock-implementation.

Quite a bit of this PR is centered around the new "categorization" concept now declared via `x-optimade-category`- which thus comes with a non-negligible complexity cost for the specification. Two questions:

- Why does this not rather live in the top `x-optimade-defintion` field as simply `category`? We have all the other declarative information about the entity being defined there.

- This might be me dimmed by a couple of weeks of vacation, but did we discuss these category-based provenance limitations during the workshop? Do we need them?

What is a use-case for an OPTIMADE server implementer and/or an OPTIMADE API user that is served by these category declarations, vs., them just not being there? Meaning: without them, a database is allowed to represent provenance relationships between any entry types for which it deems it makes sense to declare such relationships for.

ndaelman-hu:
*Disclaimer: I think @giovannipizzi knows this better, but I can give my view:*

@rartino To my understanding, `category` is a subcase of `kind`, not a peer. `kind` (`property`/`unit`/`prefix`/`constant`) already existed to say what sort of definition this is; `category` (`data`/`execution`/`metadata`) is new in this PR and only kicks in when `kind == "property"` at the outermost (entry) level — units, prefixes, and nested properties have no category.

That said, the semantics don't settle placement. Since `category` refines `kind`, and `x-optimade-definition` already holds optional keys (`version`, `resources`), the conditional-presence worry doesn't really apply — so I'd lean toward your suggestion: put it in `x-optimade-definition` next to `kind`, as `category`, rather than a separate top-level `x-optimade-category`. The real thing to pin down is whether it's normative (gates provenance) or just informative.

gmrigna:
relations

rartino:
We discussed this PR on the web meet today. Two issues that we would like to sort out in collaboration with, primarily, @giovannipizzi :

- What is the use-case for "category" as a concept in this PR. What benefits does it provide a client interrogating a provider via the OPTIMADE API? Or, what benefit does it provide a provider/API implementer?

- From the PR _"The eight provenance relationships do not link directly to the far-end entry.
Instead, each resolves to one or more **relation objects**: small resource objects that represent the edge itself."_ Is this necessary instead of straight JSON:API relationships? I understand that one can find some advantages in e.g. the filtering operations that can be done on these, but it seems very heavy handed to introduce a large number of JSON:API types, and makes the provenance graphs, and implementation, much more complex. I'd also argue that this goes a bit against the design of relationships in JSON:API.

I have an implementation that demonstrates how straight JSON:API relationships would look that I'm setting up publicly. I'll post another comment when it is accessible.

rartino:
@giovannipizzi @gmrigna 

I've implemented the way I envision OPTIMADE provenance can work under `_httk_`-provider specific keys in our altermagnets database running on httk₂.

Here is a run entry:

* https://altermagnets.anyterial.se/optimade/amdb/_httk_runs/anyt.am.runs-1-4

One can see a relationship to an input structure under the key: `_httk_has_input` with id: [`anyt.am.structures-1-95`](https://altermagnets.anyterial.se/optimade/amdb/structures/anyt.am.structures-1-95). It has the reverse relationship encoded in `_httk_is_output`.

One can see relationships to outputs under the key: `_httk_has_output`. These are:

* Some VASP output files: `vasprun.xml` [`anyt.am.files-1-10`](https://altermagnets.anyterial.se/optimade/amdb/files/anyt.am.files-1-10), `DOSCAR` [anyt.am.files-1-11](https://altermagnets.anyterial.se/optimade/amdb/files/anyt.am.files-1-11), a figure of the spin splitting of this altermagnet `Fe4O5-3D-splitting.png` [anyt.am.files-1-12](https://altermagnets.anyterial.se/optimade/amdb/files/anyt.am.files-1-12) (all OPTIMADE file entries, with the files themselves served via a URL pointing into the `extensions/files` endpoint).

* And a result record [`anyt.am.records-1-184"`](https://altermagnets.anyterial.se/optimade/amdb/_httk_records/anyt.am.records-1-184) with the single attribute `_httk_total_energy: -235.36149357`. It has the reverse relationship encoded as `_httk_is_output`.

Finally, looking at the structure again: [`anyt.am.structures-1-95`](https://altermagnets.anyterial.se/optimade/amdb/structures/anyt.am.structures-1-95), it has a `_httk_has_product` relationship to the output record: [`anyt.am.records-1-184"`](https://altermagnets.anyterial.se/optimade/amdb/_httk_records/anyt.am.records-1-184), and the output record has the reverse relationship in `_httk_product_of`.

To me this design seems to work out quite elegant, very closely in line with the spirit of JSON:API, and somewhat easily manually navigated. 

I can see the point of "what if some workflow has 10,000 outputs?", but I'm reluctant to the idea of making all representations of provenance far more complex to improve support for that case. Pagination of relationships is supported in JSON:API, and if "searchability" over inputs/outputs is important to someone, it would still be possible to add endpoints that support that without making the relationships themselves first-class JSON:API resource objects, and without complicating the relationships themselves. And aside all that: if one really generate 10,000 things from a workflow, why not simply encode it as the output being a collection instead of as 10,000 individual provenance outputs?

PR #597, opened 2026-07-03
Change the definition of `query-support: "equality only"` to encompass `IS KNOWN` and `IS UNKNOWN` operators.

### LINKED ISSUE #594: Does `query-support: "equality only"` imply support for `IS KNOWN`/`IS UNKNOWN`?
`query-support: "equality only"` is now formulated as:

> `equality only`: the defined property MUST be queryable using the OPTIMADE filter language equality and inequality operators. Other filter language features do not need to be available.

It does not say anything about `IS KNOWN`/`IS UNKNOWN`, but it seems to me that these operators are implied. Pinging @rartino for comment.

merkys:
Fixes #594 as if this was the intended behavior and left out as a bug of the specification.

### DIFF
--- a/optimade.rst
+++ b/optimade.rst
@@ -2601,7 +2601,8 @@ A Property Definition MUST be composed according to the combination of the requi
     The string MUST be one of the following:
 
     - :val:`all mandatory`: the defined property MUST be queryable using the OPTIMADE filter language with support for all mandatory filter features.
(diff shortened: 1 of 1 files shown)


rartino (approved):

sauliusg (approved):
Correct.

vaitkus (approved):

